PR #14, opened 2019-08-22
Parsing replicated license files

marccampbell:
This PR adds initial support to kots CLI for pulling and installing replicated apps using a local license file.


cmd/kots/cli/util.go
```diff
@@ -2,8 +2,17 @@ package cli
...
 	"os"
+	"strings"
 )
 
+func ExpandDir(path string) string {
+	if !strings.HasPrefix(path, "~") {
+		return path
+	}
+
+	return homeDir() + path[1:]
```
divolgin: `filepath.Join`... also, ouch.

marccampbell: 👍 I don't know of a built-in?

docs/apps/replicated/README.md
```diff
@@ -20,14 +20,14 @@ For disambiguation when there are multiple releases with the same name, use the
 kubectl kots pull replicated://app-slug#12
 ```
 
-To specify the channel to use (the license must be assigned to the requested channel):
+Some application channels require that they are specified on the URL:
```
divolgin: `on the URL` doesn't sound right to me

pkg/template/config_context.go
```diff
@@ -0,0 +1,110 @@
...
+
+			var built string
+			if builtValue != "" {
+				built = builtValue
+			} else {
+				built = builtDefault
+			}
+
+			if v, ok := templateContext[configItem.Name]; ok {
+				built = fmt.Sprintf("%v", v)
```
divolgin: I feel like this should be `%s` because that renders the result of `v.String()`

pkg/template/config_context.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		return false
+	}
+
+	return value != val
+}
+
+func (ctx ConfigCtx) getConfigOptionValue(itemName string) (string, error) {
+	if val, ok := ctx.ItemValues[itemName]; ok {
+		return fmt.Sprintf("%v", val), nil
```
divolgin: `%s` here too maybe

PR #31, opened 2019-08-31
Write bundle to admin console

marccampbell:


pkg/pull/archive.go
```diff
@@ -0,0 +1,97 @@
...
+	encoded := base64.StdEncoding.EncodeToString(archive)
+
+	// well.
+	//
+	// let's write this encoded value to a config map with a known label
+	// so that kotsadm will upload it and ingest it as an app
+	// it's really the only way we can get the archive
+	// but etcd and config maps are limited to 1 mb
+	// so let's split it across multiple, if it's larger than 1 mb
+	encodedParts, err := util.SplitStringOnLen(encoded, 756*1000)
```
laverya:
```suggestion
	// 768 * 1000 is sufficiently less than a megabyte to ensure we
	// won't have problems
	encodedParts, err := util.SplitStringOnLen(encoded, 768*1000)
```
moar comment!
Also using a round number (`768`) instead of some random number (`756`)
`(log(768)-log(3))/log(2)=8`

marccampbell: ha!

PR #41, opened 2019-09-06
Use the proxied, portforwarded address

marccampbell:


pkg/kotsadm/operator.go
```diff
@@ -92,6 +102,22 @@ func ensureOperatorRoleBinding(namespace string, clientset *kubernetes.Clientset
 	return nil
 }
 
+func ensuureOperatorServiceAccount(namespace string, clientset *kubernetes.Clientset) error {
```
divolgin: `ensuure`

PR #44, opened 2019-09-09
airgap upload

divolgin:


ffi/airgap.go
```diff
@@ -21,21 +23,90 @@ import (
...
+
+//export PullFromAirgap
+func PullFromAirgap(licenseData, airgapDir, downstream, outputFile, registryHost, registryNamesapce string) int {
+	workspace, err := ioutil.TempDir("", "kots-airgap")
 	if err != nil {
-		fmt.Printf("failed to download airgap archive: %s\n", err)
+		fmt.Printf("failed to create temp dir: %s\n", err)
 		return 1
 	}
+	// defer os.RemoveAll(workspace)
```
marccampbell: should this be put back in?

pkg/image/image.go
```diff
@@ -0,0 +1,77 @@
...
+	"path"
+	"path/filepath"
+
+	"github.com/pkg/errors"
+	"sigs.k8s.io/kustomize/v3/pkg/image"
+)
+
+func BuildRewriteList(rootDir string, host string, namespace string) ([]image.Image, error) {
+	images, err := findImages(rootDir, host, namespace, []string{})
+	return images, errors.Wrapf(err, "failed to find images")
```
marccampbell: why does this return an error always?

ffi/airgap.go
divolgin: yeah

pkg/image/image.go
divolgin: errors.Wrap will return nil when there is no error... it does feel a bit odd without the if

PR #48, opened 2019-09-12
pkg/upstream: Fix dropped errors

alrs:
Fixes a pair of swallowed error variables in `pkg/upstream`.

### DIFF
--- a/pkg/upstream/admin-console.go
+++ b/pkg/upstream/admin-console.go
@@ -56,6 +56,10 @@ func findPostgresPassword(renderDir string, files []os.FileInfo) (string, error)
 
 		decode := scheme.Codecs.UniversalDeserializer().Decode
 		obj, gvk, err := decode(data, nil, nil)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/upstream/admin-console.go
```diff
@@ -56,6 +56,10 @@ func findPostgresPassword(renderDir string, files []os.FileInfo) (string, error)
 
 		decode := scheme.Codecs.UniversalDeserializer().Decode
 		obj, gvk, err := decode(data, nil, nil)
+		if err != nil {
+			return "", errors.Wrap(err, "failed to decode")
```
divolgin: I think both of these should continue.  If we come across an invalid yaml file, we should keep looking.

marccampbell: agreed. there is no guarantee that we can decode all yaml using this. this PR would make it error if a custom resource was included that we didn't have the scheme loaded at compile time.

@divolgin will you update this to continue on these?

PR #50, opened 2019-09-13
add a command that returns version info

### LINKED ISSUE #28: There is no 'version' command
We pass version information as part of ldflags, but the package referenced does not exist.

https://github.com/replicatedhq/kots/blob/a0539db/deploy/.goreleaser.yml#L19-L22

laverya:
(old) example output:
```
{
    "version": "v0.9.1-3-g0dd52ca",
    "git": "0dd52ca",
    "buildTime": "2019-09-13T00:44:15Z",
    "go": {
        "version": "go1.12.9",
        "compiler": "gc",
        "os": "darwin",
        "arch": "amd64"
    },
    "runAt": "2019-09-13T00:44:18.601021Z"
}
```

new output:
`kots version` returns information like
```
Replicated Kots v0.9.3-4-ge30eba8

Version v0.9.3 is available for kots. To install updates, run
  $ curl https://kots.io/install | bash
```
If an error is encountered checking for updates, we display this
```
Unable to check for newer releases: %s
```

the version will be just `v0.10.0` for actual releases - the suffix is only present for non-tagged builds.

Fixes #28.


marccampbell:
@laverya Can we format the version output in plain text?

You are right, for tagged versions (actual releases) we should stick with the semver label only. The suffix is great on snapshot versions though. Most of the other information isn't too relevant for end users.

The thing we should be showing here is if the user is on the latest version. This should be pretty easy by comparing with the latest release tag from this repo.

I'd like to see:

```
$ kubectl kots version
Replicated Kots v0.10.0

Updates are available to kots. To install updates, run
  $ kubectl krew upgrade kots

```

(The bottom section is only printed when there are updates available)

laverya:
@marccampbell,
I think 'code formatted version' is pretty common within the k8s community - for instance, on my local:
```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.3", GitCommit:"2d3c76f9091b6bec110a5e63777c332469e0cba2", GitTreeState:"clean", BuildDate:"2019-08-19T12:36:28Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"12+", GitVersion:"v1.12.8-gke.10", GitCommit:"f53039cc1e5295eed20969a4f10fb6ad99461e37", GitTreeState:"clean", BuildDate:"2019-06-19T20:48:40Z", GoVersion:"go1.10.8b4", Compiler:"gc", Platform:"linux/amd64"}
$ kustomize version
Version: {KustomizeVersion:3.1.0 GitCommit:95f3303493fdea243ae83b767978092396169baf BuildDate:2019-07-26T19:21:45+01:00 GoOs:darwin GoArch:amd64}
$ go version
go version go1.12.9 darwin/amd64
```
Of those only go is really 'plaintext'.

Regarding checking for updates... That turns an offline command into an online one, which may not be desirable. It also doesn't really fit with my concept of what a 'version' command should do - tell the user what version they are running and provide information useful for filing bug reports. Shouldn't krew handle that anyways?

laverya:
Would changing the default to output on a single line (and perhaps also changing to use `%+v` formatting instead of json) and adding a `-q`/`--quiet` flag to output only the version be better?

marccampbell:
@laverya Understood and I appreciate that convention.

However, the JSON output doesn't achieve our primary goals with this, which is to clearly tell the user which version they are running. There's a lot of output here in the JSON format, most of which isn't needed by the end user. And why does someone check the version? Often it's to compare with the latest version to determine if they need to update or check out latest release notes.

Krew doesn't provide this functionality and doesn't have plans to. 

There are other popular OSS CLI tools (even written in Go) that perform similar functionality to what we want here. One example is `gcloud`. This version command shows non-JSON format and performs an update check for you.

We should make sure that if no internet connection is available, we fail the update check early and cleanly and just report the current version with a message that kots was unable to check for the latest version available.

laverya:
`kots version` now returns information like
```
Version v0.9.3-3-g4c40067 built at 2019-09-19T20:36:37Z
There is a newer version available, v0.9.3
```
Depending on the results of the update check, the second line may also be
```
This is the latest version
```
and for errors
```
Unable to check for newer releases: %s
```

`kots version --verbose` returns results similar to the previous output:
```
{
    "version": "v0.9.3-2-gd637ef7",
    "git": "d637ef7",
    "buildTime": "2019-09-19T20:32:13Z",
    "go": {
        "version": "go1.12.9",
        "compiler": "gc",
        "os": "darwin",
        "arch": "amd64"
    },
    "runAt": "2019-09-19T20:32:19.971591Z",
    "latestVersion": "v0.9.3"
}
```

PR #70, opened 2019-09-25
Base to helm

marccampbell:


marccampbell:
This adds a (hidden) CLI command and FFI function that performs a pull of a replicated app, and then uses the base as a new Helm chart.

PR #84, opened 2019-10-07
Don't mask errors when forwarding ports

divolgin:


pkg/k8sutil/portforward.go
```diff
@@ -120,28 +129,33 @@ func PortForward(kubeContext string, localPort int, remotePort int, namespace st
 		forwardedAdditionalPorts := map[AdditionalPort]chan struct{}{}
 
 		uri := fmt.Sprintf("http://localhost:%d/api/v1/kots/ports", localPort)
-		go func() error {
```
divolgin: this error is just getting thrown away right now

PR #129, opened 2019-11-06
kotsadm should use user 1001 or 101, not 0

laverya:


laverya:
minio and postgres are still a WIP

PR #133, opened 2019-11-08
Syncing license data

salahalsaleh1993:


pkg/pull/signature.go
```diff
@@ -90,7 +90,7 @@ func verify(message, signature, publicKeyPEM []byte) error {
 
 func getMessageFromLicense(license *kotsv1beta1.License) ([]byte, error) {
 	// JSON marshaller will sort map keys automatically.
-	fields := map[string]string{
+	fields := map[string]interface{}{
```
divolgin: I think changing this to interface opens this up to programming mistakes.  You can now add an object of any type to this map.

salahalsaleh1993: the license is signed with `licenseSequence` being an integer, that was the reason for changing this

ffi/main.go
```diff
@@ -133,6 +137,87 @@ func UpdateCheck(socket, fromArchivePath string) {
...
+
+	u, err := url.Parse(license.Spec.Endpoint)
+	if err != nil {
+		fmt.Printf("failed to parse endpoint from license: %s\n", err.Error())
+		return C.CString("")
+	}
+
+	hostname := u.Hostname()
+	if u.Port() != "" {
+		hostname = fmt.Sprintf("%s:%s", u.Hostname(), u.Port())
```
divolgin: `net.JoinHostPort` > `fmt.Sprintf`

pkg/pull/signature.go
salahalsaleh1993: i guess i should change signing the license to use `licenseSequence` as string

ffi/main.go
```diff
@@ -133,6 +137,87 @@ func UpdateCheck(socket, fromArchivePath string) {
...
+	}
+
+	url := fmt.Sprintf("%s://%s/release/%s/license", u.Scheme, hostname, license.Spec.AppSlug)
```
divolgin: actually, you can just assume license endpoint format is what should be used, so you don't need to generate `hostname`

PR #140, opened 2019-11-11
Support hidden entitlement fields

salahalsaleh1993:


kotskinds/apis/kots/v1beta1/license_types.go
```diff
@@ -94,6 +94,7 @@ type EntitlementField struct {
 	Title       string           `json:"title,omitempty"`
 	Description string           `json:"description,omitempty"`
 	Value       EntitlementValue `json:"value,omitempty"`
+	Hidden      bool             `json:"hidden,omitempty"`
```
marccampbell: I think this is better as `isHidden`.

salahalsaleh1993: okay, will update

PR #142, opened 2019-11-14
Backward compatable license sigining

salahalsaleh1993:


pkg/pull/signature.go
```diff
@@ -17,29 +17,49 @@ var (
...
+	LicenseData    string `json:"licenseData"`
+	InnerSignature string `json:"innerSignature"`
+}
+
 type KeySignature struct {
 	Signature   []byte `json:"signature"`
 	GlobalKeyId string `json:"globalKeyId"`
 }
 
 func VerifySignature(license *kotsv1beta1.License) error {
```
laverya: this should return the license data from the internal base64 string

kotskinds/apis/kots/v1beta1/license_types.go
```diff
@@ -99,7 +99,7 @@ type EntitlementField struct {
 
 // LicenseSpec defines the desired state of LicenseSpec
 type LicenseSpec struct {
-	Signature         []byte                      `json:"signature"`
+	Signature         string                      `json:"signature"`
```
divolgin: if you keep these as []byte, you won't have to do `base64.StdEncoding.DecodeString(license.Spec.Signature)`

pkg/pull/signature.go
```diff
@@ -48,16 +68,16 @@ func VerifySignature(license *kotsv1beta1.License) error {
...
 
-	licenseMessage, err := getMessageFromLicense(license)
+	licenseMessage, err := base64.StdEncoding.DecodeString(outerSignature.LicenseData)
 	if err != nil {
-		return errors.Wrap(err, "failed to convert license to message")
+		return errors.New("failed to decode license data from outer signature")
 	}
 
-	if err := verify(licenseMessage, signature.LicenseSignature, []byte(signature.PublicKey)); err != nil {
+	if err := verify(licenseMessage, innerSignature.LicenseSignature, []byte(innerSignature.PublicKey)); err != nil {
```
laverya: this verify can be run against the base64 encoded data itself
but really what matters is doing it the same way everywhere, I suppose
```suggestion
	if err := verify(outerSignature.LicenseData, innerSignature.LicenseSignature, []byte(innerSignature.PublicKey)); err != nil {
```

kotskinds/apis/kots/v1beta1/license_types.go
salahalsaleh1993: noted, will update

laverya: You'd still need a base64 decode of some form though?

pkg/pull/signature.go
salahalsaleh1993: we sign the data before encoding it, we can update both if that's the preferred approach

pkg/pull/signature.go
```diff
@@ -48,16 +68,16 @@ func VerifySignature(license *kotsv1beta1.License) error {
...
 		return errors.Wrap(err, "failed to verify license signature")
 	}
```
laverya: before returning here you should parse `licenseMessage` as a LicenseSpec
and then compare it to the other fields within the passed license's LicenseSpec (everything besides 'signature')

pkg/pull/signature.go
laverya: (or maybe overwrite it, since pass-by-reference?)

kotskinds/apis/kots/v1beta1/license_types.go
salahalsaleh1993: apparently not

pkg/pull/signature.go
laverya: If it's easier to sign before encoding, we can do that
(I don't have a preference here)

pkg/pull/signature.go
```diff
@@ -88,41 +101,48 @@ func verify(message, signature, publicKeyPEM []byte) error {
...
-		valStr, err := json.Marshal(val)
-		if err != nil {
-			return nil, errors.Wrapf(err, "failed to marshal entitlement value: %s", k)
-		}
-		fields[key] = string(valStr)
+func verifyLicenseData(outerLicense *kotsv1beta1.License, innerLicense *kotsv1beta1.License) error {
+	if outerLicense.Spec.AppSlug != innerLicense.Spec.AppSlug {
+		return errors.New("\"appSlug\" field has changed")
+	}
+	if outerLicense.Spec.Endpoint != innerLicense.Spec.Endpoint {
```
emosbaugh: @marccampbell mentioned that he may want to be able to change the endpoint. you should ask him about this

laverya: perhaps that should be done with an override param elsewhere?
And also have "can override endpoint" be a license param

pkg/pull/signature.go
```diff
@@ -17,51 +16,65 @@ var (
...
-	if err != nil {
-		return errors.Wrap(err, "failed to convert license to message")
+	if err := verify(outerSignature.LicenseData, innerSignature.LicenseSignature, []byte(innerSignature.PublicKey)); err != nil {
+		return nil, errors.Wrap(err, "failed to verify license signature")
 	}
 
-	if err := verify(licenseMessage, signature.LicenseSignature, []byte(signature.PublicKey)); err != nil {
-		return errors.Wrap(err, "failed to verify license signature")
+	verifiedLicense := &kotsv1beta1.License{}
+	if err := json.Unmarshal(outerSignature.LicenseData, verifiedLicense); err != nil {
```
laverya: I believe your current PR has `outerSignature.LicenseData` being base64'd here
One of the two will need to change (IMO `outerSignature.LicenseData` doesn't need base64, so you should remove it in vandoor)
Unless I'm misreading things

divolgin: Here it's defined as `[]byte` so it's automatically decoded.

salahalsaleh1993: Updated in license-api

pkg/pull/signature.go
marccampbell: @salahalsaleh1993 I agree with Ethan. The endpoint parameter should be configurable by the end customer here. It’s not part of the product yet, but some end customers could have proxies and want to redirect the endpoint to a proxy or other policy engine to pull. 

Let’s keep endpoint in the license spec, but it should not be validated during the signature.

salahalsaleh1993: cool, will update, thanks!

PR #145, opened 2019-11-14
support private registries with online installs

divolgin:


ffi/main.go
```diff
@@ -205,4 +206,29 @@ func init() {
...
+func parseConfigValuesFromFile(filename string) (*kotsv1beta1.ConfigValues, error) {
+	contents, err := ioutil.ReadFile(filename)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, nil
+		}
+		return nil, errors.Wrap(err, "failed to read config values file")
+	}
+
+	kotsscheme.AddToScheme(scheme.Scheme)
```
emosbaugh: you dont need this

ffi/online.go
```diff
@@ -79,9 +84,131 @@ func PullFromLicense(socket string, licenseData string, downstream string, outpu
...
+			return
+		}
+		licenseData, err := ioutil.ReadFile(expectedLicenseFile)
+		if err != nil {
+			fmt.Printf("failed to read license file: %s\n", err.Error())
+			ffiResult = NewFFIResult(-1).WithError(err)
+			return
+		}
+
+		kotsscheme.AddToScheme(scheme.Scheme)
```
emosbaugh: ditto

cmd/kots/cli/reset-password.go
```diff
@@ -109,7 +109,7 @@ func promptForNewPassword() (string, error) {
 	}
 
 	prompt := promptui.Prompt{
-		Label:     "Enter a new password to be used for the Admin Console:",
+		Label:     "1 Enter a new password to be used for the Admin Console:",
```
marccampbell: Remove the 1?

pkg/kotsadm/secrets.go
```diff
@@ -240,7 +240,7 @@ func promptForSharedPassword() (string, error) {
 	}
 
 	prompt := promptui.Prompt{
-		Label:     "Enter a new password to be used for the Admin Console:",
+		Label:     "2 Enter a new password to be used for the Admin Console:",
 		Templates: templates,
```
marccampbell: Remove this 2?

pkg/util/util.go
```diff
@@ -14,6 +15,14 @@ func IsURL(str string) bool {
 	return true
 }
 
+func IsLocalPath(str string) bool {
```
marccampbell: This concerns me. I think you are working around something wrong with IsURL which is ok. But I’d like to make this and IsURL a single function. And it can be called with ! When you need the other. So can you consolidate these two methods?

cmd/kots/cli/reset-password.go
divolgin: oops

pkg/util/util.go
divolgin: actually, I ended up not using this... removing

PR #165, opened 2019-11-27
Add license template context

salahalsaleh1993:


pkg/template/license_context.go
```diff
@@ -0,0 +1,27 @@
...
+}
+
+// FuncMap represents the available functions in the LicenseCtx.
+func (ctx LicenseCtx) FuncMap() template.FuncMap {
+	return template.FuncMap{
+		"LicenseFieldValue": ctx.licenseFieldValue,
+	}
+}
+
+func (ctx LicenseCtx) licenseFieldValue(name string) interface{} {
```
divolgin: this should return a string

PR #181, opened 2019-12-05
Load version label from airgap metadata if it's available

divolgin:


ffi/airgap.go
```diff
@@ -75,6 +75,7 @@ func PullFromAirgap(socket, licenseData, airgapDir, downstream, outputFile, regi
 			Downstreams:         []string{downstream},
 			LocalPath:           releaseDir,
 			LicenseFile:         licenseFile,
+			AirgapMetaFile:      filepath.Join(airgapDir, "airgap.yaml"),
```
marccampbell: @divolgin thinking about this and changes in the future... should we just walk the filetree of the archive, matching GVK instead?

pkg/upstream/fetch.go
```diff
@@ -43,7 +44,7 @@ func downloadUpstream(upstreamURI string, fetchOptions *FetchOptions) (*Upstream
 		return downloadHelm(u, fetchOptions.HelmRepoURI)
 	}
 	if u.Scheme == "replicated" {
-		return downloadReplicated(u, fetchOptions.LocalPath, fetchOptions.RootDir, fetchOptions.UseAppDir, fetchOptions.License, fetchOptions.ConfigValues, fetchOptions.CurrentCursor, fetchOptions.CurrentVersionLabel)
+		return downloadReplicated(u, fetchOptions.LocalPath, fetchOptions.RootDir, fetchOptions.UseAppDir, fetchOptions.License, fetchOptions.ConfigValues, fetchOptions.CurrentCursor, pickVersionLabel(fetchOptions))
```
marccampbell: I think we should update the tests on `downloadReplicated` (or create if necessary) to handle this new functionality. This seems like an area that's likely to break if we don't have a test around it.

divolgin: 👍

ffi/airgap.go
divolgin: @marccampbell this file is never in a subdirectory.... I think if we list files in the root of the bundle that should be sufficient.

PR #183, opened 2019-12-06
never write images to files when pulling, upload them immediately

laverya:
go.mod updates

also add a solution for docker hub images like 'nginx' that lack a domain

### DIFF
--- a/pkg/image/builder.go
+++ b/pkg/image/builder.go
@@ -62,9 +62,9 @@ func CopyImages(srcRegistry, destRegistry registry.RegistryOptions, appSlug stri
 				return err
 			}
 
-			newImagesSubset, err := cooyImagesBetweenRegistries(srcRegistry, destRegistry, appSlug, log, reportWriter, contents, savedImages)
+			newImagesSubset, err := copyImagesBetweenRegistries(srcRegistry, destRegistry, appSlug, log, reportWriter, contents, savedImages)
 			if err != nil {
-				return errors.Wrap(err, "failed to extract images")
+				return errors.Wrapf(err, "failed to copy images mentioned in %s", path)
 			}
 
 			newImages = append(newImages, newImagesSubset...)
@@ -165,7 +165,7 @@ func GetObjectsWithImages(upstreamDir string) ([]*k8sdoc.Doc, error) {
 	return objects, nil
 }
 
-func cooyImagesBetweenRegistries(srcRegistry, destRegistry registry.RegistryOptions, appSlug string, log *logger.Logger, reportWriter io.Writer, fileData []byte, savedImages map[string]bool) ([]kustomizeimage.Image, error) {
+func copyImagesBetweenRegistries(srcRegistry, destRegistry registry.RegistryOptions, appSlug string, log *logger.Logger, reportWriter io.Writer, fileData []byte, savedImages map[string]bool) ([]kustomizeimage.Image, error) {
 	newImages := []kustomizeimage.Image{}
 	err := listImagesInFile(fileData, func(images []string, doc *k8sdoc.Doc) error {
 		for _, image := range images {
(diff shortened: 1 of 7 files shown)


pkg/image/builder.go
```diff
@@ -60,19 +62,20 @@ func SaveImages(srcRegistry registry.RegistryOptions, appSlug string, log *logge
 				return err
 			}
 
-			err = saveImagesToFiles(srcRegistry, appSlug, log, imagesDir, reportWriter, contents, savedImages)
+			newImagesSubset, err := cooyImagesBetweenRegistries(srcRegistry, destRegistry, appSlug, log, reportWriter, contents, savedImages)
```
emosbaugh: cooy

pkg/image/builder.go
```diff
@@ -162,18 +165,23 @@ func GetObjectsWithImages(upstreamDir string) ([]*k8sdoc.Doc, error) {
...
 			}
 
-			log.ChildActionWithSpinner("Pulling image %s", image)
-			err := saveOneImage(srcRegistry, imagesDir, image, appSlug, reportWriter)
+			log.ChildActionWithSpinner("Transferring image %s", image)
+			newImage, err := copyOneImage(srcRegistry, destRegistry, image, appSlug, reportWriter)
 			if err != nil {
 				log.FinishChildSpinner()
-				return errors.Wrap(err, "failed to save image")
+				return errors.Wrap(err, "failed to transfer image")
```
emosbaugh: you should add the image name to this wrap

pkg/image/builder.go
```diff
@@ -60,19 +62,20 @@ func SaveImages(srcRegistry registry.RegistryOptions, appSlug string, log *logge
 				return err
 			}
 
-			err = saveImagesToFiles(srcRegistry, appSlug, log, imagesDir, reportWriter, contents, savedImages)
+			newImagesSubset, err := cooyImagesBetweenRegistries(srcRegistry, destRegistry, appSlug, log, reportWriter, contents, savedImages)
 			if err != nil {
 				return errors.Wrap(err, "failed to extract images")
```
emosbaugh: should add path name to this wrap

pkg/image/builder.go
```diff
@@ -209,41 +217,21 @@ func listImagesInFile(contents []byte, handler processImagesFunc) error {
...
-		return errors.Wrapf(err, "failed to parse local image name: %s", destStr)
+		return nil, errors.Wrap(err, "failed to create policy")
 	}
 
 	sourceCtx := &types.SystemContext{}
 
 	isPrivate, err := isPrivateImage(image)
 	if err != nil {
-		return errors.Wrapf(err, "failed to check if image is private: %s", image)
+		return nil, errors.Wrapf(err, "failed to check if image is private: %s", image)
```
emosbaugh: i see the name is here. but its a bit easier to just add it in the wrapping function and remove from everything in this function

pkg/image/image.go
```diff
@@ -43,3 +44,56 @@ func ImageInfoFromFile(registry registry.RegistryOptions, nameParts []string) (k
...
+}
+
+func buildImageAlts(destRegistry registry.RegistryOptions, image string) ([]kustomizeimage.Image, error) {
+	imgParts := strings.Split(image, "/")
+	if len(imgParts) == 1 {
+		imgParts = append([]string{"docker.io", "library"}, imgParts...)
+	}
+	imageInfo, err := ImageInfoFromFile(destRegistry, imgParts)
+	if err != nil {
+		return nil, errors.Wrapf(err, "info from %s", image)
```
emosbaugh: ditto

pkg/image/image.go
```diff
@@ -43,3 +44,56 @@ func ImageInfoFromFile(registry registry.RegistryOptions, nameParts []string) (k
...
+func DestImageName(registry registry.RegistryOptions, srcImage string) string {
+	imageParts := strings.Split(srcImage, "/")
+	lastPart := imageParts[len(imageParts)-1] // last part
+
+	image := fmt.Sprintf("%s/%s/%s", registry.Endpoint, registry.Namespace, lastPart)
+
+	return image
+}
+
+func buildImageAlts(destRegistry registry.RegistryOptions, image string) ([]kustomizeimage.Image, error) {
```
emosbaugh: could you add a unit test for this function? what if the image is `library/redis` or `replicated/kots`? does it work without the domain?

laverya: Can do
and I'll add some code to handle two-part images (since it currently handles 1 and 3+ part images)

laverya: that code ended up being rather wrong
resolved now, though

PR #184, opened 2019-12-06
Don't panic when ports cannot be forwarded

divolgin:


pkg/download/download.go
```diff
@@ -34,13 +34,23 @@ func Download(appSlug string, path string, downloadOptions DownloadOptions) erro
...
 		log.FinishSpinnerWithError()
 		return errors.Wrap(err, "failed to start port forwarding")
 	}
 	defer close(stopCh)
 
+	go func() {
+		select {
+		case err := <-errChan:
+			if err != nil {
+				log.Error(err)
```
divolgin: It's not really clear how this will terminate on error... Presumably `io.Copy` below will fail.

cmd/kots/cli/install.go
```diff
@@ -143,11 +143,22 @@ func InstallCmd() *cobra.Command {
 			}
 
 			if canPull {
-				stopCh, err := upload.StartPortForward(uploadOptions.Namespace, uploadOptions.Kubeconfig)
+				stopCh, errChan, err := upload.StartPortForward(uploadOptions.Namespace, uploadOptions.Kubeconfig)
```
emosbaugh: can we pass ctx into this function rather than returning stopCh?

cmd/kots/cli/install.go
```diff
@@ -160,12 +171,23 @@ func InstallCmd() *cobra.Command {
 				return errors.Wrap(err, "failed to wait for web")
 			}
 
-			stopCh, err := k8sutil.PortForward(v.GetString("kubeconfig"), 8800, 3000, namespace, podName, true)
+			stopCh, errChan, err := k8sutil.PortForward(v.GetString("kubeconfig"), 8800, 3000, namespace, podName, true)
```
emosbaugh: can we pass ctx into this function rather than returning stopCh?

cmd/kots/cli/install.go
```diff
@@ -160,12 +171,23 @@ func InstallCmd() *cobra.Command {
...
 			if err != nil {
 				return errors.Wrap(err, "failed to forward port")
 			}
 			defer close(stopCh)
```
emosbaugh: this would then be cancel() from context.WithCancel

cmd/kots/cli/install.go
```diff
@@ -160,12 +171,23 @@ func InstallCmd() *cobra.Command {
...
 
+			go func() {
+				select {
+				case err := <-errChan:
+					if err != nil {
+						log.Error(err)
+						os.Exit(-1)
+					}
+				case <-stopCh:
```
emosbaugh: and this would be <-ctx.Done()

pkg/k8sutil/portforward.go
```diff
@@ -86,26 +86,30 @@ func PortForward(kubeContext string, localPort int, remotePort int, namespace st
...
 	forwarder, err := portforward.New(dialer, []string{fmt.Sprintf("%d:%d", localPort, remotePort)}, stopChan, readyChan, out, errOut)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
+	errChan := make(chan error, 2) // 2 go routines are writing to this channel
+
 	go func() {
 		for range readyChan {
```
emosbaugh: this seems funny... why isnt it just?

```go
<-readyChan
```

pkg/upload/port-forward.go
```diff
@@ -9,19 +9,19 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 )
 
-func StartPortForward(namespace string, kubeconfig string) (chan struct{}, error) {
+func StartPortForward(namespace string, kubeconfig string) (chan struct{}, chan error, error) {
```
emosbaugh: ctx

cmd/kots/cli/install.go
```diff
@@ -143,11 +143,22 @@ func InstallCmd() *cobra.Command {
...
+						if err != nil {
+							log.Error(err)
+							os.Exit(-1)
+						}
+					case <-stopCh:
+					}
+				}()
+
 				if err := upload.Upload(rootDir, uploadOptions); err != nil {
 					return err
```
laverya: I think this needs a `close(stopCh)` since it isn't deferred

pkg/k8sutil/portforward.go
divolgin: this was written through trial and error... that channel is constantly written to, which eventually blocks, and it's nothing we care about

PR #187, opened 2019-12-09
Revert "Revert "never write images to files when pulling, upload them immediately""

laverya:
Reverts replicatedhq/kots#186, restores replicatedhq/kots#183

### DIFF
--- a/pkg/image/builder.go
+++ b/pkg/image/builder.go
@@ -24,6 +24,7 @@ import (
 	"github.com/replicatedhq/kots/pkg/k8sdoc"
 	"github.com/replicatedhq/kots/pkg/logger"
 	"gopkg.in/yaml.v2"
+	kustomizeimage "sigs.k8s.io/kustomize/v3/pkg/image"
 )
 
 var imagePolicy = []byte(`{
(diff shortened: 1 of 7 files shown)


pkg/image/builder.go
```diff
@@ -254,29 +243,85 @@ func saveOneImage(srcRegistry registry.RegistryOptions, imagesDir string, image
...
 		SignBy:                "",
 		ReportWriter:          reportWriter,
 		SourceCtx:             sourceCtx,
-		DestinationCtx:        nil,
+		DestinationCtx:        destCtx,
 		ForceManifestMIMEType: "",
 	})
 	if err != nil {
-		return errors.Wrap(err, "failed to copy image")
+		log.Error(errors.Wrap(err, "failed to copy image directly"))
```
divolgin: Suggestion for an improvement here:
If it is specifically that compatibility error, we should log it on info level and retry pull/push approach.
Otherwise we should log this on error level and bail.

laverya: Would an info level "failed to copy image directly, attempting fallback copy method" be good? I'm not confident in our ability to properly identify that compatibility error, and I don't think optimizing for failure cases is worthwhile in general

PR #199, opened 2019-12-12
Ensure airgap bundle belongs to the licensed app

divolgin:


pkg/pull/signature.go
```diff
@@ -13,8 +13,8 @@ import (
 )
 
 var (
-	ErrSignatureInvalid = errors.New("license signature is invalid")
-	ErrSignatureMissing = errors.New("license signature is missing")
+	ErrSignatureInvalid = errors.New("signature is invalid")
```
marccampbell: Is this sufficient to debug? Should we have license and bundle signature messages as separate strings to make diagnosing problems easier?

pkg/pull/signature.go
```diff
@@ -229,3 +229,29 @@ func getMessageFromLicense(license *kotsv1beta1.License) ([]byte, error) {
 
 	return message, err
 }
+
+func GetAppPublicKey(license *kotsv1beta1.License) ([]byte, error) {
+	if len(license.Spec.Signature) == 0 || len(license.Spec.Signature) == 1 {
```
marccampbell: why 1? i think a comment here explaining that might be useful

pkg/pull/signature.go
divolgin: yeah, stack trace string will be different

PR #201, opened 2019-12-16
Add support for replicated apps that contain one or more helm charts

marccampbell:
This PR adds support to include one or more Helm charts in a replicated app. Docs to follow.


kotskinds/apis/kots/v1beta1/helmchart_types.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	if m.valueType == "children" {
+		return m.children
+	}
+	if m.valueType == "childrenArray" {
+		return m.childrenArrays
+	}
+
+	panic("unknown value type")
```
divolgin: this function is called from `RenderValues`, which always returns nil error...  This should probably return an error and be handled there

divolgin: oh, and type info :)

pkg/base/replicated.go
```diff
@@ -57,8 +59,10 @@ func renderReplicated(u *upstream.Upstream, renderOptions *RenderOptions) (*Base
 	}
 
 	for _, upstreamFile := range u.Files {
-		rendered, err := builder.RenderTemplate(upstreamFile.Path, string(upstreamFile.Content))
+		content := fmt.Sprintf("%s", upstreamFile.Content)
```
divolgin: i think this doen't do anything

PR #202, opened 2019-12-17
Add namespaced serviceaccount to cluster role binding, if already exists

marccampbell:


marccampbell:
This fixes a bug that prevents kots from installing properly to a second namespace, even if the first was deleted because the cluster role was not updated.

PR #204, opened 2019-12-18
Upstream update

marccampbell:
This PR adds a new command:

```
kubectl kots upstream upgrade --namespace <namespace> <app-slug>
```

This command will do the same thing as clicking "Check for updates" in the admin console. If GitOps is enabled, it will result in a commit. If not, it will result in pending versions.

### DIFF
--- a/cmd/kots/cli/upstream-upgrade.go
+++ b/cmd/kots/cli/upstream-upgrade.go
@@ -75,6 +75,9 @@ func UpstreamUpgradeCmd() *cobra.Command {
 			if resp.StatusCode == 404 {
 				log.FinishSpinnerWithError()
 				return errors.New("The application was not found in the cluster in the specified namespace")
+			} else if resp.StatusCode != 200 {
+				log.FinishSpinnerWithError()
+				return errors.Errorf("Unexpected response from the API: %d", resp.StatusCode)
 			}
 
 			b, err := ioutil.ReadAll(resp.Body)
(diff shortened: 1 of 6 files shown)


cmd/kots/cli/upstream-upgrade.go
```diff
@@ -0,0 +1,138 @@
...
+			if err != nil {
+				log.FinishSpinnerWithError()
+				return errors.Wrap(err, "failed to get from kotsadm")
+			}
+			defer resp.Body.Close()
+
+			if resp.StatusCode == 404 {
+				log.FinishSpinnerWithError()
+				return errors.New("The application was not found in the cluster in the specified namespace")
+			}
```
divolgin: `else if resp.StatusCode != 200 { ...error`?

cmd/kots/cli/upstream-upgrade.go
```diff
@@ -0,0 +1,138 @@
...
+		return "", errors.Wrap(err, "failed to create kubernetes clientset")
+	}
+
+	pods, err := clientset.CoreV1().Pods(namespace).List(metav1.ListOptions{LabelSelector: "app=kotsadm-api"})
+	if err != nil {
+		return "", errors.Wrap(err, "failed to list pods")
+	}
+
+	for _, pod := range pods.Items {
+		if pod.Status.Phase == corev1.PodRunning {
```
divolgin: this could also check `pod.Status.ContainerStatuses`

PR #206, opened 2019-12-18
Add FFI function to render any template

marccampbell:
Given an app archive, this will find the config, values and license and render a single file via an FFI function.

Open questions: what should we do about password values? Will the cluster encryption key be processed because it runs in as an ffi in the api sandbox?

### DIFF
--- a/ffi/templates.go
+++ b/ffi/templates.go
@@ -1,0 +1,153 @@
+package main
+
+import "C"
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/mholt/archiver"
+	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


ffi/templates.go
```diff
@@ -0,0 +1,153 @@
...
+			decode := scheme.Codecs.UniversalDeserializer().Decode
+			obj, gvk, err := decode(content, nil, nil)
+			if err != nil {
+				return nil
+			}
+
+			if gvk.Group == "kots.io" && gvk.Version == "v1beta1" && gvk.Kind == "Config" {
+				config = obj.(*kotsv1beta1.Config)
+			}
+			if gvk.Group == "kots.io" && gvk.Version == "v1beta1" && gvk.Kind == "ConfigValues" {
```
divolgin: are you not a fan of elses? :)

marccampbell: interesting, tell me more about these `else` things!

PR #211, opened 2019-12-19
don't deploy velero.io/v1 Backup

areed:


pkg/kotsadm/kotsadm.go
```diff
@@ -16,6 +16,8 @@ import (
 
 var timeoutWaitingForKotsadm = time.Duration(time.Minute * 2)
 
+const KotsadmKey = "app.kubernetes.io/name"
```
marccampbell: What other label names could we use here, that is more "kots" like, less "kubernetes"?

areed: "kots.io/control-plane=1"

areed: decided on "kots.io/kotsadm=true" on code review call

PR #227, opened 2019-12-30
Split multi-doc yaml renders

marccampbell:
We had an option already present "SplitMultidocYAML" but it wasn't used.

This uses the same splitting technique implemented by ksplit, but in memory. Only affects files with multidocs in them, and them -1, -2 indexes them, instead of trying to come up with deterministically generated names

### DIFF
--- a/pkg/base/helm.go
+++ b/pkg/base/helm.go
@@ -108,7 +108,7 @@ func RenderHelm(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (*Base,
 
 		for idx, fileString := range fileStrings {
 			filename := strings.TrimSuffix(k, filepath.Ext(k))
-			filename = fmt.Sprintf("%s-%d%s", filename, idx, filepath.Ext(k))
+			filename = fmt.Sprintf("%s-%d%s", filename, idx+1, filepath.Ext(k))
 
 			baseFile := BaseFile{
 				Path:    filename,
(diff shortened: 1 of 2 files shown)


pkg/base/helm.go
```diff
@@ -83,12 +85,39 @@ func RenderHelm(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (*Base,
...
+			}
+
+			baseFiles = append(baseFiles, baseFile)
+			continue
 		}
 
-		baseFiles = append(baseFiles, baseFile)
+		for idx, fileString := range fileStrings {
+			filename := strings.TrimSuffix(k, filepath.Ext(k))
+			filename = fmt.Sprintf("%s-%d%s", filename, idx, filepath.Ext(k))
```
laverya: is 0-indexing intentional here?
I don't really care either way

pkg/base/helm.go
```diff
@@ -83,12 +85,39 @@ func RenderHelm(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (*Base,
...
+
+			baseFile := BaseFile{
+				Path:    filename,
+				Content: []byte(fileString),
+			}
+
+			baseFiles = append(baseFiles, baseFile)
+			continue
```
laverya: this `continue` is confusing IMO
since it's at the end of a for loop and is thus redundant

PR #235, opened 2020-01-02
Add support for optional resources

marccampbell:


pkg/base/base_test.go
```diff
@@ -122,16 +123,174 @@ connection:
...
+		},
+		{
+			name: "annotation of excluded false as string",
+			path: "config.yaml",
+			content: []byte(`apiVersion: kots.io/v1beta1
+kind: Config
+metadata:
+  name: sentry-enterprise
+  annotations:
+    kots.io/exclude: "false""
```
divolgin: something's up with quotes here

PR #237, opened 2020-01-03
Support airgap updates

divolgin:


pkg/cursor/cursor_test.go
```diff
@@ -0,0 +1,77 @@
...
+		},
+		{
+			name:         "comparable and after",
+			c1:           "100",
+			c2:           "10",
+			isAfter:      true,
+			isComparable: true,
+		},
+		{
+			name:    "not comparable",
```
laverya: this isn't testing the "not comparable" function though? Just whether C1 forms a cursor properly

PR #254, opened 2020-01-10
kots install with namespace on command line does nothing and exists without any errors

divolgin:


cmd/kots/cli/root.go
```diff
@@ -38,7 +37,6 @@ func RootCmd() *cobra.Command {
 
 func InitAndExecute() {
 	if err := RootCmd().Execute(); err != nil {
-		fmt.Println(err)
```
divolgin: This is double  logging.

cmd/kots/cli/install.go
```diff
@@ -64,11 +64,7 @@ func InstallCmd() *cobra.Command {
 				if len(errs) > 0 {
 					return errors.New(errs[0])
 				}
-
-				return nil
```
laverya: welp
yeah, that'll break things

PR #258, opened 2020-01-12
refactor to move kotsadm DeployOptions and UpgradeOptions types into subpackage

laverya:


pkg/k8sutil/config.go
```diff
@@ -6,10 +6,10 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/client-go/kubernetes"
 
-	"github.com/replicatedhq/kots/pkg/kotsadm"
+	"github.com/replicatedhq/kots/pkg/kotsadm/types"
```
marccampbell: import as kotsadmtypes?

PR #259, opened 2020-01-14
Handle helm values with a dot

marccampbell:
This PR updates the helm render utility to handle values with a `.`.

We pass these in to the render function the same way that the `--set` CLI flag does in `helm render`.  The problem this addresses is when there's a values.yaml that contains something like this:

```yaml
ingress:
  annotations:
    kubernetes.io/ingress.class: nginx
```

Before this PR, this would be rendered as:

```yaml
  annotations:
    kubernetes: "map[io/ingress:map[class:nginx]]"
    kubernetes.io/ingress.class: "nginx"
```

This is because helm splits on the `.` character. By adding escaping in, this now renders as:

```yaml
  annotations:
    kubernetes.io/ingress.class: "nginx"
```

### DIFF
--- a/kotskinds/apis/kots/v1beta1/helmchart_types.go
+++ b/kotskinds/apis/kots/v1beta1/helmchart_types.go
@@ -178,7 +178,7 @@ func renderOneLevelValues(values map[string]MappedChartValue, parent []string) (
 					}
 
 					for _, childKey := range childKeys {
-						key := fmt.Sprintf("%s[%d].%s", k, i, childKey)
+						key := fmt.Sprintf("%s[%d].%s", escapeIfNeeded(k), i, escapeIfNeeded(childKey))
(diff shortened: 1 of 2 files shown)


kotskinds/apis/kots/v1beta1/helmchart_types.go
```diff
@@ -194,14 +194,22 @@ func renderOneLevelValues(values map[string]MappedChartValue, parent []string) (
...
 
 	return keys, nil
 }
 
+func escapeIfNeeded(in string) string {
+	if !strings.Contains(in, ".") {
+		return in
+	}
+
+	return fmt.Sprintf(`%s`, strings.ReplaceAll(in, ".", `\.`))
```
emosbaugh: Why sprintf?

PR #280, opened 2020-01-24
Get latest license on app install

salahalsaleh1993:


pkg/license/license.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/pkg/errors"
+	kotsv1beta1 "github.com/replicatedhq/kots/kotskinds/apis/kots/v1beta1"
+	"github.com/replicatedhq/kots/pkg/version"
+	"io/ioutil"
+	"k8s.io/client-go/kubernetes/scheme"
+	"net/http"
+)
+
+func GetLatestLicense(license *kotsv1beta1.License) (*kotsv1beta1.License, error) {
+	url := fmt.Sprintf("%s/release/%s/license", license.Spec.Endpoint, license.Spec.AppSlug)
```
marccampbell: Why is exchanging a license a method in the /release/app-slug/ route? This feels like license syncing shouldn’t be a part of the app routes to me,

salahalsaleh1993: you're right, should change that to be something like `/license/:appSlug`

PR #284, opened 2020-01-28
include auth string in calls to kotsadm

laverya:


laverya:
role change is also made in kurl in https://github.com/replicatedhq/kURL/pull/197

pkg/util/util.go
```diff
@@ -60,3 +66,42 @@ func IntPointer(x int) *int64 {
 	xout = int64(x)
 	return &xout
 }
+
+var passwordLetters = []rune("0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
+
+// generates a [0-9a-zA-Z] password of the specified length
+func GenPassword(length int) string {
```
areed: The signature kind of suggests that longer passwords are stronger but the generator is determined by a single int64 seed. crypto/rand seems like a better fit, or at least re-seed every ~12 reads.

pkg/auth/auth.go
```diff
@@ -0,0 +1,78 @@
...
+	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+const KotsadmAuthstringSecretName = "kotsadm-authstring"
+const KotsadmAuthstringSecretKey = "kotsadm-authstring"
+
+var authSlugCache string
+
+// ThisIsATest disables secret fetch/creation in the auth package
```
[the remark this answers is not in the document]
laverya: whoops - I'll fix that comment
Really it's that setting the cache string in a test prevents this code from attempting to create a secret

pkg/util/util.go
laverya: I'm not sure reseeding would be more secure, but I'll look at moving to crypto/rand

laverya: moved to crypto/rand now, kept math/rand as fallback

PR #296, opened 2020-01-31
add scope test instrumentation

fshot:


pkg/image/image_test.go
```diff
@@ -78,6 +81,8 @@ func Test_ImageNameFromNameParts(t *testing.T) {
 }
 
 func TestDestRef(t *testing.T) {
+	test := scopeagent.StartTest(t)
+	defer test.End()
```
laverya: looks like this one got missed

fshot: Thanks, fixed.

PR #314, opened 2020-02-06
Scope instrumentation

fshot:


pkg/base/base_test.go
```diff
@@ -120,10 +123,11 @@ spec:
 			assert.Equal(t, test.expected, string(b.Content))
 		})
 	}
-
 }
 
 func Test_ShouldBeIncludedInBaseKustomization(t *testing.T) {
+	test := scopeagent.StartTest(t)
+	defer test.End()
```
laverya: this is also instrumented within the test case loop

pkg/midstream/lists_test.go
```diff
@@ -4,9 +4,12 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"go.undefinedlabs.com/scopeagent"
 )
 
 func Test_findNewStrings(t *testing.T) {
+	scopetest := scopeagent.StartTest(t)
+	defer scopetest.End()
```
laverya: move to inside the test case loop?

pkg/base/base_test.go
fshot: thanks, fixed

pkg/midstream/lists_test.go
fshot: good catch, thanks

PR #317, opened 2020-02-06
fix config yaml rendering with long template functions

laverya:
use laverya/yaml.v3 to marshal yaml with unlimited line length
also add MarshalYAML interface for the BoolOrString type (so that it renders the same as it did previously)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,15 +1,13 @@
 package config
 
 import (
-	"bytes"
-
 	"github.com/pkg/errors"
 	kotsv1beta1 "github.com/replicatedhq/kots/kotskinds/apis/kots/v1beta1"
 	"github.com/replicatedhq/kots/kotskinds/multitype"
 	"github.com/replicatedhq/kots/pkg/base"
 	"github.com/replicatedhq/kots/pkg/logger"
 	"github.com/replicatedhq/kots/pkg/template"
-	"k8s.io/apimachinery/pkg/runtime/serializer/json"
+	"github.com/replicatedhq/kots/pkg/util"
 	"k8s.io/client-go/kubernetes/scheme"
 )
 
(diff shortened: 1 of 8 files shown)


pkg/config/config.go
```diff
@@ -45,31 +43,21 @@ func TemplateConfig(log *logger.Logger, configSpecData string, configValuesData
...
+	rendered, err := builder.RenderTemplate("config", string(configDocWithData))
 	if err != nil {
 		return "", errors.Wrap(err, "failed to render config template")
 	}
 
 	return rendered, nil
 }
 
-func marshalConfig(config *kotsv1beta1.Config) (string, error) {
-	s := json.NewYAMLSerializer(json.DefaultMetaFactory, scheme.Scheme, scheme.Scheme)
```
emosbaugh: why is or isnt this necessary?

laverya: it's replaced with the `util.MarshalIndent` function which then calls the yaml render function directly

emosbaugh: but that doesnt use the kubernetes scheme. what is the difference?

laverya: The kubernetes yaml marshaller isn't special
It just renders json and converts it to yaml (with gopkg.in/yaml.v2, even)
The util function also renders with that package (well, a newer version of it) but sets the maximum line length to unlimited to avoid wrapping

marccampbell: Having 2 different marshallers will be difficult over time. What options do we have to eventually get this back to the built-in?  I understand that it's not special, but now our code is because this one file is handled using a different marshal/unmarhshal library. For simplicity and maintainability moving forward, it would be good to have a path to getting this back to the same library as everything else.

fshot: Perhaps we could add additional tests that do side-by-side renderings of a range of config files, to ensure that the alternate 'unlimited-width' render path produces exactly the same output as the k8s marshaller.  Once the goyaml project merges in Andrew's PR, and that fix gets included with k8s, that would be our path to using a single library everywhere, but until then, the tests would give us some confidence that the behavior hasn't diverged in the meantime.

laverya: side-by-side rendering added in #320

PR #322, opened 2020-02-08
add basic testing of the 'NewConfigContext' function

laverya:


pkg/template/config_context_test.go
```diff
@@ -0,0 +1,183 @@
...
+			want: &ConfigCtx{
+				ItemValues: map[string]ItemValue{
+					"abcItem": {
+						Value: "replacedAbcItemValue",
+					},
+				},
+			},
+		},
+		{
+			name: "chained configOptionValue failure", // this isn't really the desired behavior, test just demonstrates it
```
fshot: Maybe leave a little more explanation about the bug that this test with unexpected results is trying to illustrate, and/or what the expected results are meant to be...

PR #325, opened 2020-02-10
Generate openapischema for kotskinds

marccampbell:
Adds a new `make -C kotskinds schemas` command that will use controller-gen to generate CRD specs, and then build a small go binary that uses the apiextensions/v1beta1 clientset to parse these and export the schema as a json object using the json unmarshaller.

This can then be copied into projects that use the [Kubeval](https://github.com/instrumenta/kubeval/) validation tool to compare a given manifest to the spec.


go.mod
```diff
@@ -58,6 +58,7 @@ require (
 	gopkg.in/alecthomas/kingpin.v3-unstable v3.0.0-20180810215634-df19058c872c // indirect
 	gopkg.in/yaml.v2 v2.2.8
 	k8s.io/api v0.17.2
+	k8s.io/apiextensions-apiserver v0.0.0-20190918161926-8f644eb6e783
```
laverya: why not https://github.com/kubernetes/apiextensions-apiserver/releases/tag/v0.17.2 ?
```suggestion
	k8s.io/apiextensions-apiserver v0.17.2
```

MikaelSmith:
Do you have any notes for using this with kubeval?

PR #326, opened 2020-02-11
Add when field support for config groups

salahalsaleh:


marccampbell:
@salahalsaleh1993 thanks for the PR, i think this has deeper implication that we need to consider.

when a field is required, but the group is excluded because of when: false, will the config still be usable? if a group is excluded but the config item has a previous value, what will or should we use in config template rendering?

PR #327, opened 2020-02-12
`kots admin-console allow-namespace` command

marccampbell:
This adds a new subcommand to kots that's useful for existing clusters. It will modify the role and role binding that we use to allow progression to support multiple namespaces. 

Also, this adds the velero rbac policies in.


AmberAlston:
@marccampbell This PR is a year+, can we close it or do you believe it still needs to be done?

PR #332, opened 2020-02-13
Don't allow upgrade command on kurl clusters

divolgin:


pkg/kotsadm/main.go
```diff
@@ -168,6 +166,40 @@ func Deploy(deployOptions types.DeployOptions) error {
...
+		err := errors.New("The namespace cannot be found or accessed")
+		return err
+	}
+
+	if upgradeOptions.ForceUpgradeKurl {
+		return nil
+	}
+
+	// don't upgrade kurl clusters.  kurl only installs into default namespace.
+	if upgradeOptions.Namespace != "" && upgradeOptions.Namespace != "default" {
```
marccampbell: @divolgin can we also check the the existence of the kurl.sh node label?  Require both the node label and the default namespace?

pkg/kotsadm/main.go
```diff
@@ -168,6 +166,40 @@ func Deploy(deployOptions types.DeployOptions) error {
...
+		return nil
+	}
+
+	kotsAPI, err := clientset.AppsV1().Deployments(upgradeOptions.Namespace).Get("kotsadm-api", metav1.GetOptions{})
```
marccampbell: I think we can ignore from here down in the function. Why should we check that env var?

cmd/kots/cli/admin-console-upgrade.go
```diff
@@ -51,9 +52,11 @@ func AdminConsoleUpgradeCmd() *cobra.Command {
 		},
 	}
 
+	cmd.Flags().Bool("force-upgrade-kurl", false, "set to force upgrade even if this is a kurl cluster")
```
marccampbell: If this is enabled, let's print an message on the screen that we are proceeding with something that's not recommended

pkg/kotsadm/main.go
divolgin: @marccampbell the node that the pods are on? or just all nodes?

pkg/kotsadm/main.go
divolgin: @marccampbell because kotsamd-api can exist and not be installed with kurl?  isn't the only way to test that is to test the `ENABLE_KURL` variable?

pkg/kotsadm/main.go
marccampbell: all nodes

PR #338, opened 2020-02-19
Show better error message when dest dir exists

divolgin:


cmd/kots/cli/download.go
```diff
@@ -23,7 +23,7 @@ func DownloadCmd() *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			v := viper.GetViper()
 
-			if len(args) == 0 {
```
emosbaugh: @divolgin how about using cobra for this? https://godoc.org/github.com/spf13/cobra#ExactArgs

PR #339, opened 2020-02-19
Don't remove version/cursor info from installation on kots pull from local files

divolgin:


pkg/pull/pull.go
```diff
@@ -604,3 +622,50 @@ func licenseIsExpired(license *kotsv1beta1.License) (bool, error) {
 	}
 	return partsed.Before(time.Now()), nil
 }
+
+func findConfig(localPath string) (*kotsv1beta1.Config, *kotsv1beta1.ConfigValues, *kotsv1beta1.License, *kotsv1beta1.Installation, error) {
```
divolgin: This is copied from the ffi package.  Didn't feel the need to refactor since that whole package is going away.

PR #341, opened 2020-02-20
Support for additional images

marccampbell:
This supports moving the RewriteImages function out of TS and into kotsadm.


pkg/template/config_context.go
```diff
@@ -144,6 +145,10 @@ func (ctx ConfigCtx) configOptionNotEquals(name string, value string) bool {
 	return value != val
 }
 
+func (ctx ConfigCtx) localRegistryAddress() string {
+	return ""
```
divolgin: TODO?

PR #346, opened 2020-02-22
allow chaining template functions within the config page

laverya:
this allows 'chaining' template functions on the config page - for instance, generating a certificate based on a hostname provided on the settings page, and making that certificate available via `repl{{ ConfigOption ...`


pkg/template/depgraph.go
```diff
@@ -0,0 +1,117 @@
...
+		d.AddDep(parent, dep)
+		return dep
+	}
+
+	return template.FuncMap{
+		"ConfigOption":          addDepFunc,
+		"ConfigOptionIndex":     addDepFunc,
+		"ConfigOptionData":      addDepFunc,
+		"ConfigOptionEquals":    addDepFunc,
+		"ConfigOptionNotEquals": addDepFunc,
```
divolgin: There are registry templates here too now.

laverya: yeah, but they don't add dependencies within the config page

PR #357, opened 2020-02-26
Better retry logic

marccampbell:


pkg/image/builder.go
```diff
@@ -471,6 +472,37 @@ func CopyFromFileToRegistry(path string, name string, tag string, digest string,
...
+
+			start := time.Now()
+			for {
+				// TODO should this support http registry also?
+				response, err := quickClient.Get(fmt.Sprintf("https://%s/v2/", registryEndpoint))
+				if err == nil && response.StatusCode < 500 {
+					break
+				}
+				if time.Now().Sub(start) > time.Duration(time.Second*5) {
+					return errors.Wrap(err, "unable to connect to registry")
```
laverya: err might be nil here
if statuscode was > 500

PR #368, opened 2020-02-28
allow using the namespace flag with the reset-password command

laverya:
basically, `kubectl kots reset-password mynamespace`, `kubectl kots reset-password -n mynamespace` and `kubectl kots reset-password --namespace mynamespace` should all work.

### DIFF
--- a/cmd/kots/cli/reset-password.go
+++ b/cmd/kots/cli/reset-password.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"fmt"
 	"os"
 
 	"github.com/manifoldco/promptui"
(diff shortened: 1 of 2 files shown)


moonlight16:
lgtm  

Also could we remove the -n from the Global flags section?  or is this used with other kots commands?

jacohn@jeremy-prod:~$ kubectl kots reset-password -n default
Change the password on the Admin Console

```
Usage:
  kots reset-password [namespace] [flags]

Flags:
  -h, --help   help for reset-password

Global Flags:
...
  -n, --namespace string               If present, the namespace scope for this CLI request
```

laverya:
That global flag is used with other kots functions too (and is the flag that this PR starts using)

Why do you want to remove it?

moonlight16:
> Why do you want to remove it?

Nvrmind.  i just realized this patch is allowing both.

PR #372, opened 2020-03-03
Additional namespaces flag

marccampbell:


pkg/kotsadm/namespaces.go
```diff
@@ -0,0 +1,66 @@
...
+		return nil // no error here
+	}
+
+	if gvk.Group != "kots.io" || gvk.Version != "v1beta1" || gvk.Kind != "Application" {
+		return nil
+	}
+
+	application := obj.(*kotsv1beta1.Application)
+
+	if application.Spec.AdditionalNamespaces == nil {
```
divolgin: don't need this.  loop works with nil arrays

pkg/kotsadm/namespaces.go
```diff
@@ -0,0 +1,66 @@
...
+		return nil
+	}
+
+	for _, additionalNamespace := range application.Spec.AdditionalNamespaces {
+		// We support "*" for additional namespaces to handle pullsecret propagation
```
divolgin: "We `don't` support", right?

marccampbell: We do, but we don't create that.  It (will be) handled in the kotsadm code to handle imagePulLSecret propagation

divolgin: Does this then break on something like this:
```
 - custom_ns
 - worker*
```

marccampbell: unsupported.  only "*" and/or specific (exact) namespaces

PR #373, opened 2020-03-04
improve configitem resolution failure messaging

laverya:
Now error messages include the offending configitem, and why:
```
no config options exist with 0 dependencies - "alpha" depends on "bravo"; "bravo" depends on "alpha", "charlie"
```
or
```
no config options exist with 0 dependencies - "bravo" depends on "doesnotexist"
```

Previously, the error was always the same:
```
No nodes exist with 0 dependencies
```

### DIFF
--- a/pkg/template/depgraph.go
+++ b/pkg/template/depgraph.go
@@ -4,9 +4,9 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"strings"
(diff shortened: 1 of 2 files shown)


laverya:
unfortunately, the error messages themselves are nondeterministic due to ordering... (map traversal order in golang is unspecified)

PR #385, opened 2020-03-06
use quotedbool type in config item when field

laverya:


kotskinds/multitype/quotedbool.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	*b = QuotedBool(s)
+	return nil
+}
+
+// UnmarshalJSON implements the yaml.Unmarshaller interface.
+func (b *QuotedBool) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	boolTry := false
+	stringTry := ""
+	err := unmarshal(&boolTry)
```
emosbaugh: why dont you just use this logic in the function above as well?

laverya: Because this is relying on a function passed in as a param, and UnmarshalJSON has a different signature?

PR #388, opened 2020-03-07
ADD support for kURL context template functions 

mzaneri:


pkg/template/kurl_context.go
```diff
@@ -0,0 +1,197 @@
...
+	retrieved := GetKurlValues(installerName, nameSpace)
+
+	if retrieved != nil {
+		ctx.AddValuesToKurlContext(retrieved)
+	}
+
+	return ctx
+}
+
+func (ctx KurlCtx) AddValuesToKurlContext(retrieved *kurlv1beta1.Installer) {
```
mzaneri: This is the generic way the addons and their fields are parsed out using reflection. Basically, it iterates from all addons, then through all their fields, and adds them to a map[string]interface{}. All that is need to be done when changes in kURL are made is to go get the kURL release

go.mod
```diff
@@ -44,6 +46,7 @@ require (
 	github.com/otiai10/copy v1.0.2
 	github.com/phayes/freeport v0.0.0-20180830031419-95f893ade6f2
 	github.com/pkg/errors v0.9.1
+	github.com/replicatedhq/kurl/kurlkinds v0.0.0-20200227234931-c098073f445c
```
mzaneri: this should probably be done off of tags I think?

pkg/template/kurl_context.go
```diff
@@ -0,0 +1,197 @@
...
+	case string:
+		return t
+	case bool:
+		return strconv.FormatBool(t)
+	default:
+		fmt.Printf("The yamlPath '%s' corresponds to value '%v' of type '%T'. The KurlOption function supports only string, integer, and boolean values\n", yamlPath, result, result)
+		return ""
+	}
+}
+
```
mzaneri: Leaving in this debug function as I think it may have utility to vendors

go.mod
divolgin: i agree, we do tag every release https://github.com/replicatedhq/kURL/releases

pkg/template/kurl_context.go
```diff
@@ -0,0 +1,197 @@
...
+		"KurlInt":    ctx.kurlInt,
+		"KurlBool":   ctx.kurlBool,
+		"KurlOption": ctx.kurlOption,
+		"KurlAll":    ctx.kurlAll,
+	}
+}
+
+func (ctx KurlCtx) kurlBool(yamlPath string) bool {
+	if len(ctx.KurlValues) == 0 {
+		fmt.Printf("No Installer CRD has been found, is this a kURL cluster?\n")
```
divolgin: I think this is too verbose.  Same app can be installed on kurl and non-kurl clusters, so this is going to be just noise almost always.

pkg/template/kurl_context.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	result, ok := ctx.KurlValues[yamlPath]
+	if !ok {
+		fmt.Printf("There is no value found at the yamlPath '%s'\n", yamlPath)
+		return ""
+	}
+
+	switch t := interface{}(result).(type) {
+	case int:
```
divolgin: don't really need this.   this works for anything that can be "stringified": `fmt.Sprintf("%v", strOrBoolOrInt)`

pkg/template/kurl_context.go
mzaneri: changed

PR #389, opened 2020-03-07
combine all builder init into one function

laverya:
the caller should not be responsible for adding the static context, license context and config context


pkg/template/builder.go
```diff
@@ -18,6 +20,23 @@ type Builder struct {
...
+	b := Builder{}
+	configCtx, err := b.newConfigContext(configGroups, existingValues, localRegistry, cipher, license)
+	if err != nil {
+		return Builder{}, nil, errors.Wrap(err, "create config context")
+	}
+
+	b.Ctx = []Ctx{
+		StaticCtx{},
+		LicenseCtx{License: license},
+		NewKurlContext("base", "default"), // can be hardcoded because kurl always deploys to the default namespace
```
divolgin: ❤️

mzaneri: default is fine being hardcoded, not sure if 'base' (the name of that installer CRD object) is? dunno if we ever expect that to change

PR #406, opened 2020-03-17
ADD distribution template function

mzaneri:


pkg/template/static_context.go
```diff
@@ -420,3 +423,41 @@ func (ctx StaticCtx) isKurl() bool {
 
 	return configMap != nil
 }
+
+func (ctx StaticCtx) distribution() string {
```
laverya: would it be better to have this return an array of strings? If something can be both "AWS" and "OpenShift" for instance

mzaneri: per marc, Openshift is the proper return here

PR #433, opened 2020-03-31
include kurl functions in config context rendering

laverya:


laverya:
It's a builder that has everything but the config, though 🤷‍♂

PR #441, opened 2020-04-02
tlsMap should not be a global

laverya:
I'm not sure if this will have side effects - will this make the calls to TLSMap/TLSKey order-dependent? Were they actually order-dependent before?


AmberAlston:
@laverya @divolgin should this PR still be open?

laverya:
honestly this should almost certainly be closed - there's been a lot of change to that code in the last year.

PR #446, opened 2020-04-07
Check kind before annotations on backup

areed:


laverya:
Should we move _all_ of the kots kinds before the annotation check?

areed:
> Should we move _all_ of the kots kinds before the annotation check?

That broke some of the unit tests.

PR #454, opened 2020-04-21
Fix install upload endpoint

areed:


areed:
This flag still needs specs so I can't say the changed endpoint in this PR is the "correct" endpoint.

PR #459, opened 2020-04-23
Change `kots install --license-file` to be the same as `kots install`, but automate license installation

### LINKED ISSUE #450: kots CLI: kots install with --license-file hangs after "Creating base" with no feedback
```
 k kots install my-kots-app --license-file ~/Desktop/path/to/license.yaml
```

Output gets to here and then just hangs for minutes 

```
Enter the namespace to deploy to: my-kots-app

  • Pulling upstream ✓  
  • Creating base ✓  
```

I'm guessing this is starting datastores, apis, etc but we lose all the regular `kubectl kots install` output.

**Update** after several minutes I get this:

```
  • Creating base ✓  
  • Creating midstream ✓  
  • Creating downstream "this-cluster" ✓  
  • Deploying Admin Console
    • Creating namespace ✓  
    • Waiting for datastore to be ready ✓  
Enter a new password to be used for the Admin Console: 
```

marccampbell:
This fixes #450 by changing this functionality completely. Before this PR, `kots install .. --license-file` used the license file and pulled locally.

With this PR, the behavior of kots install works the same, with or without a license file. If a license file is passed, it's written to a secret with a label and kotsadm will pick it up during start (replicatedhq/kotsadm#1376)

### DIFF
--- a/pkg/kotsadm/license.go
+++ b/pkg/kotsadm/license.go
@@ -1,0 +1,67 @@
+package kotsadm
+
+import (
+	"bytes"
+
+	"github.com/pkg/errors"
+	"github.com/replicatedhq/kots/pkg/kotsadm/types"
+	corev1 "k8s.io/api/core/v1"
+	kuberneteserrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/serializer/json"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
+)
+
+func getLicenseSecretYAML(deployOptions *types.DeployOptions) (map[string][]byte, error) {
+	docs := map[string][]byte{}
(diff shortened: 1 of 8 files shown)


pkg/kotsadm/license.go
```diff
@@ -0,0 +1,67 @@
...
+	return docs, nil
+}
+
+func ensureLicenseSecret(deployOptions *types.DeployOptions, clientset *kubernetes.Clientset) error {
+	existingSecret, err := getLicenseSecret(deployOptions.Namespace, clientset)
+	if err != nil {
+		return errors.Wrap(err, "failed to check for existing license secret")
+	}
+
+	if existingSecret == nil {
```
divolgin, line 40: Should this always replace the secret?  What if the user tries this a second time (for whatever reason) with a different license.

marccampbell: @divolgin I think this should remove the secret. If the user puts a different license in (and restarts the kotsadm pod), it will create a secret again and import the app, and then delete that secret.  Unless I'm not understanding the concern?

PR #466, opened 2020-04-23
Upgrade sprig

robbiet480:
Need this to get access to the new `htpasswd` function added at https://github.com/Masterminds/sprig/pull/225. Also ran a `go mod tidy`.


laverya:
Thanks for the bump!

Issue #457, opened 2020-04-21
KOTS install should support brining an external object store


dexhorthy:
Right now we install minio or (in the case of a kURL cluster) plug into an embedded object store -- it should be possible during kots install to pass a flag to bring an external object store instance. Unclear how we configure auth for this, and I wonder if we'd like to make this configurable in kotsadm or not.

dexhorthy:
@marccampbell do you have thoughts on design here?

PR #474, opened 2020-04-28
adding proposal for external object store

dexhorthy:
Assigning @marccampbell but would also love to hear from @divolgin and @emosbaugh since y'all have done a lot of work with custom S3 endpoints -- what other parameters do you think we might need to consider?

### DIFF
--- a/docs/proposals/object-store.md
+++ b/docs/proposals/object-store.md
@@ -1,0 +1,161 @@
+Adding Capability for BYO Object Store
+================
+
+This document outlines the current architecture of how Object Stores are used by [kots](https://github.com/replicatedhq/kots) and [kotsadm](https://github.com/replicatedhq/kotsadm), and a proposal for modifying how it is configured to allow end users of `kotsadm` to bring their own object store rather than using an embedded minio instance.
+
+### Goals
+
+- For `kots install` and `kots pull`, allow an end user to bring their own object store rather than using an embedded minio instance that requires the cluster to include a storage class.
+- Move closer to unifying the YAML used across `kots install` and the [kURL addon](https://kurl.sh/docs/add-ons/kotsadm) (or, at the very least, avoid creating any more differences between the two methods)
+
+### Architecture Background
+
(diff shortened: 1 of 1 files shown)


docs/proposals/object-store.md
```diff
@@ -0,0 +1,161 @@
...
+
+In every case, five environment variables are configured for `kotsadm` and `kotsadm-api`:
+
+- `S3_ACCESS_KEY_ID`
+- `S3_SECRET_ACCESS_KEY`
+- `S3_ENDPOINT`
+- `S3_BUCKET_NAME`
+- `S3_BUCKET_ENDPOINT` 
+
+`S3_BUCKET_ENDPOINT` appears to be a boolean, configured whether to use a custom endpoint. Seems to always be `true`, I believe this controls whether to override the default S3 endpoint in our S3 client.
```
marccampbell, line 36: My understanding of this param is that it controls whether the client (our s3 internal client) uses path style or dns style bucket naming.  (https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story)

I believe that minio doesn't currently support bucket dns names in K8s, so we rely on this legacy pattern of addressing buckets

dexhorthy: ahhhh okay gotcha -- do you think we need an extra flag for this then?

dexhorthy:
@marccampbell agree that reducing scope by skipping the secret migration is probably the right path, I hesitated to put the migration stuff in here at all. I'll take another pass at a version of this where we leave the secret unmigrated. 


> Is it possible to use a kURL install with an external object store?

It should also be possible with kURL, but I don't consider this in-scope for this proposal. I'll note that explicitly. Do you think that's acceptable?

> Passing those params feels like it could be verbose. What about reading AWS Creds from the env (using the standard methods), by opting in using an flag?
> Can we support "read the creds from a secret in the cluster"
>How about "use instance role"?

I'm in favor of supporting alternate methods of providing credentials, including reading from environment/instance roles/secrets. Do you think it makes sense to implement with the flags listed here, and then from there talk about different methods? Are you also concerned with the length of the flag names themselves?

marccampbell:
I think this is good. My only remaining thought is that the 2 flag names `minio,external` might be misinterpreted. It's possible, using this proposal, to BYO minio and someone should use `external` to configure this. Along the same lines, `minio` pushes this implementation detail (which internal object store we use) to something that's more difficult to change, if we decide to.

How would you feel about `internal,external`. Or something like this? Simply changing the `minio` option name to a generic term to tell KOTS "whatever kots wants to provision to handle this" (which today, is minio)

dexhorthy:
works for me

dexhorthy:
(Relates to #457 )

PR #481, opened 2020-05-01
Fail on invalid yaml

emosbaugh:
This pr will render yaml parse errors as part of the kots.io/v1beta1 Installation yaml document

Including the following yaml file in a release would render no files. There would be no messaging to the user and the installation would proceed.

```yaml
not: kubernetes
---
{{
```

With the addition of this commit the user will see errors in the Installation yaml document.

bad.yaml

Screen Shot 2020-05-19 at 12 44 38 PM

### DIFF
--- a/kotsadm/web/src/queries/AppsQueries.js
+++ b/kotsadm/web/src/queries/AppsQueries.js
@@ -27,6 +27,10 @@ export const listAppsRaw = `
           createdOn
           sequence
           deployedAt
+          yamlErrors {
+            path
+            error
+          }
         }
         lastUpdateCheckAt
         downstreams {
@@ -113,6 +117,10 @@ export const getKotsAppRaw = `
         sequence
         releaseNotes
         deployedAt
+        yamlErrors {
+          path
+          error
+        }
       }
       lastUpdateCheckAt
       bundleCommand
--- a/kotskinds/apis/kots/v1beta1/installation_types.go
+++ b/kotskinds/apis/kots/v1beta1/installation_types.go
@@ -22,17 +22,23 @@ import (
 
 // InstallationSpec defines the desired state of InstallationSpec
 type InstallationSpec struct {
-	UpdateCursor  string              `json:"updateCursor,omitempty"`
-	ChannelName   string              `json:"channelName,omitempty"`
-	VersionLabel  string              `json:"versionLabel,omitempty"`
-	ReleaseNotes  string              `json:"releaseNotes,omitempty"`
-	EncryptionKey string              `json:"encryptionKey,omitempty"`
-	KnownImages   []InstallationImage `json:"knownImages,omitempty"`
+	UpdateCursor     string              `json:"updateCursor,omitempty"`
+	ChannelName      string              `json:"channelName,omitempty"`
+	VersionLabel     string              `json:"versionLabel,omitempty"`
+	ReleaseNotes     string              `json:"releaseNotes,omitempty"`
+	EncryptionKey    string              `json:"encryptionKey,omitempty"`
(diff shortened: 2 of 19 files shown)


marccampbell:
@emosbaugh I'm trying to fully understand the implications of this:

We'd silently ignore all files that don't have a `.yml` or `.yaml` extension. But kots would fail if a file did have a valid yaml extension but wasn't unmarshallable?

How does KOTS fail in the second scenario? Would you be able to include an example of the user experience here?

emosbaugh:
@marccampbell description updated

marccampbell:
I think a more compatible approach here could be to change this to use the old behavior, store the unmarshallable files, allow the deployment to continue, and show the warnings in the Admin Console / CLI?

Something like

```
kots pull ...
...
Completed with warnings. The following files were not able to be parsed and were not included in the application base:

file1.yaml
file2.yaml
```

emosbaugh:
@marccampbell I will have to return warnings from the base package along with errors for functions `base.RenderUpstream`, `base.RenderHelm` and `base.Base.WriteBase`. The return signature would look something like `(Warnings, error)`. That is the only way i can achieve the style of warnings you are asking for. This will affect the public function signatures. Is that a direction you would be ok with? If so I will write up a proposal.

pkg/base/base.go
```diff
@@ -92,10 +93,16 @@ func (f BaseFile) ShouldBeIncludedInBaseKustomization(excludeKotsKinds bool) (bo
 	o := OverlySimpleGVK{}
 
 	if err := yaml.Unmarshal(f.Content, &o); err != nil {
+		// check if this is a yaml file
+		if ext := filepath.Ext(f.Path); ext == ".yaml" || ext == ".yml" {
+			fmt.Printf("File %s is failed to unmarshal yaml: %v\n", f.Path, err)
```
emosbaugh, line 98: return warning

pkg/base/base.go
```diff
@@ -92,10 +93,16 @@ func (f BaseFile) ShouldBeIncludedInBaseKustomization(excludeKotsKinds bool) (bo
...
+		}
 		return false, nil
 	}
 
+	// check if this is a kubernetes document
 	if o.APIVersion == "" || o.Kind == "" {
+		fmt.Printf("File %s is not a kubernetes document\n", f.Path)
```
emosbaugh, line 105: return warning

kotskinds/apis/kots/v1beta1/installation_types.go
```diff
@@ -22,19 +22,25 @@ import (
...
-	ReleaseNotes  string              `json:"releaseNotes,omitempty"`
-	EncryptionKey string              `json:"encryptionKey,omitempty"`
-	KnownImages   []InstallationImage `json:"knownImages,omitempty"`
+	UpdateCursor     string              `json:"updateCursor,omitempty"`
+	ChannelName      string              `json:"channelName,omitempty"`
+	VersionLabel     string              `json:"versionLabel,omitempty"`
+	ReleaseNotes     string              `json:"releaseNotes,omitempty"`
+	EncryptionKey    string              `json:"encryptionKey,omitempty"`
+	KnownImages      []InstallationImage `json:"knownImages,omitempty"`
+	UnparseableFiles []UnparseableFile   `json:"unparseableFiles,omitempty"`
```
emosbaugh, line 31: can you think of a better property name?

kotsadm/web/src/queries/AppsQueries.js
```diff
@@ -113,6 +117,10 @@ export const getKotsAppRaw = `
         sequence
         releaseNotes
         deployedAt
+        yamlErrors {
```
emosbaugh, line 120: i need some ui work done

PR #492, opened 2020-05-04
Initial design of dashboard links respecting the browser address

marccampbell:


divolgin:
Will this support templated links set via config options, in which case it will be different from localhost and from the address used to access the dashboard?

marccampbell:
@divolgin what do you think is needed here WRT templates?

I imagine that optional buttons would be a nice to have (in scope/out of scope?).

But the current spec refers to a service name and an external and internal port. What are you thinking about?

divolgin:
@marccampbell I thought we template those, but apparently we don't.  So it's not a concern.

marccampbell:
@divolgin is there any additional detail needed here or is this good to go?

PR #493, opened 2020-05-04
Proposal for OCI/Docker image repository instead of S3 Object Store

marccampbell:


design/registry-instead-of-s3.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Detailed Design
+
+New CLI flags will be added to `kots install` and `kots pull` to provide a local registry to be used.
+When these flags are not provided, KOTS will install Docker Distribution statefulset and service using a ClusterIP.
+This will be an insecure registry.
+
+
+Application manifests are currently packaged as a `.tar.gz` file.
+KOTS will continue to do this, but use the [ORAS Go Module](https://github.com/deislabs/oras#oras-go-module) to package these into an image.
```
divolgin, line 34: How about adding each file as a layer?  This opens possibility for server side deduplication.

divolgin:
There is no mention of kURL in this.  Would it have to provide a different package/install script based on the chosen version of kotsadm?

design/registry-instead-of-s3.md
marccampbell: Hmm, that feels like a small optimization for a lot of complexity. A big application could have 1000s of files. With most YAML documenting being small, would there be a big win in deduplication?

An interesting alternative is to consider 1 image, where we just add each tgz to it, sequenced. So we could pull, add the sequence as a layer, and then push.

design/registry-instead-of-s3.md
```diff
@@ -0,0 +1,50 @@
+# Optionally replace object store requirement with local registry
+
+KOTS requires an object store in order to store persistent versions of application version artifacts.
+KOTS also supports optionally providing a local OCI/Docker image repository to store application container images.
+This proposal removes the object store requirement in favor of the image repository.
+
+## Goals
+
+We could remove the requirement to have an object store (currently minio or rook) by using an image repository.
```
laverya, line 9: will the image repository not also require an object store?

laverya: and it seems I misunderstood
externally hosted registry

design/registry-instead-of-s3.md
divolgin: yeah, ok.  good point.  images do have a limit on the number of layers.

Issue #476, opened 2020-04-28
Multiple resources cannot be included with same gvks and name in different namespaces


emosbaugh:
This loop does not account for namespace. As a result the first resource gets added to the resources and the second gets added to the patchStrategicMerge in the base/kustomization.yaml. As a result only one of the two gets created.

https://github.com/replicatedhq/kots/blob/46fe2b9a9215c9636d078a5dc4e61451a7a0d2bf/pkg/base/write.go#L132-L148

To reproduce:
1) create two resources that are identical. One with an alternate namespace provided in the metadata.namespace property. the other without a name space
2) deploy the app
3) one resource will be created. the other will not

```
➜  replicated kubectl -n defaultns get secret my-secret
Error from server (NotFound): secrets "my-secret" not found
➜  replicated kubectl -n otherns get secret my-secret
NAME                      TYPE                DATA   AGE
my-secret   Opaque          3      24m
```

PR #494, opened 2020-05-04
fix detection of independent resources in different namespaces

laverya:
if resources have different namespaces, they aren't the same namespace

fixes https://github.com/replicatedhq/kots/issues/476

### DIFF
--- a/pkg/base/base.go
+++ b/pkg/base/base.go
@@ -2,8 +2,8 @@ package base
 
 import (
 	"bytes"
-	"crypto/sha256"
 	"fmt"
+	"os"
 	"strconv"
 
 	"github.com/pkg/errors"
@@ -35,6 +35,7 @@ type OverlySimpleGVK struct {
 
 type OverlySimpleMetadata struct {
 	Name        string                 `yaml:"name"`
+	Namespace   string                 `yaml:"namespace"`
 	Annotations map[string]interface{} `json:"annotations"`
 }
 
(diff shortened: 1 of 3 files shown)


pkg/base/base.go
```diff
@@ -43,16 +44,19 @@ func init() {
...
-func GetGVKWithNameHash(content []byte) []byte {
+func GetGVKWithNameAndNs(content []byte) string {
 	o := OverlySimpleGVK{}
 
 	if err := yaml.Unmarshal(content, &o); err != nil {
-		return nil
+		return ""
+	}
+
+	namespace := os.Getenv("POD_NAMESPACE")
```
emosbaugh, line 54: base.Base has a Namespace property. I dont think this will work with helm charts

laverya: isn't this after the helm template code, though? So that'd already be written into the yaml

emosbaugh: https://github.com/replicatedhq/kots/blob/master/pkg/base/write.go#L103-L105

pkg/base/base.go
```diff
@@ -43,16 +43,19 @@ func init() {
...
-func GetGVKWithNameHash(content []byte) []byte {
+func GetGVKWithNameAndNs(content []byte, baseNS string) string {
 	o := OverlySimpleGVK{}
 
 	if err := yaml.Unmarshal(content, &o); err != nil {
-		return nil
+		return ""
+	}
+
+	namespace := baseNS
```
emosbaugh, line 53: so its possible that both baseNS and o.Metadata.Namespace are empty here but that is ok

laverya: pretty much my thought
kustomize will match based on the empty ns in that case anyways, so adding some other value would potentially cause errors even if the resources end up colliding during deployment

PR #497, opened 2020-05-05
Proposal: scheduled update checks

salahalsaleh:


design/scheduled-update-check.md
```diff
@@ -0,0 +1,66 @@
...
+## Background
+
+Many customers assume that scheduled update checks are a given and are surprised when they do not receive 
+updates automatically.
+
+## High-Level Design
+
+1. This proposal is only for online installations in KOTS Admin console.
+2. The default period of time (X) is configurable.
+3. The ability to disable this feature completely.
```
marccampbell, line 24: :+1: this will be a nice to have.  By default, it's enabled?

design/scheduled-update-check.md
```diff
@@ -0,0 +1,66 @@
...
+The "message" key will hold a value as to why the status is "failed".
+
+By default, when the KOTS Admin Console's api starts:
+
+1. The api will get the configuration values from the database.
+2. If the feature is enabled, start a goroutine service "loop". 
+3. The service sleeps for X period of time.
+4. Use the same logic from the update check request to check for updates for each application.
+5. The update check request logic already creates new versions automatically if there are updates available.
+6. If the api fails to start the service, a "failed" status along with the reason will be saved in
```
marccampbell, line 52: I'm not sure we need a UI element showing that a background process failed. Many clusters will have whitelists to access the public internet, and failure will be expected. Because the default it on, it might fail sometimes. 

How about not showing this status on this page, but building a custom troubleshoot analyzer that's built into kotsadm to show a warning tile if this is failing?

design/scheduled-update-check.md
```diff
@@ -0,0 +1,66 @@
...
+On each trigger, the process will get all available applications and run update checks for each app.
+The update check will automatically create new versions for each update available in an app.
+
+## Detailed Design
+
+3 new columns (`update_checker_enabled` & `update_checker_interval` & `update_checker_status`) will be 
+added to the `app` table in the postgress database using a migration.
+
+* The `update_checker_enabled` column is a bool and defaults to `true`. 
+* The `update_checker_interval` column is an integer and defaults to `360` minutes.
```
marccampbell, line 36: I think a Go duration would be more compatible (we use this for snapshots).  So `6h` would be the default?

design/scheduled-update-check.md
```diff
@@ -0,0 +1,66 @@
...
+2. If the feature is enabled, start a goroutine service "loop". 
+3. The service sleeps for X period of time.
+4. Use the same logic from the update check request to check for updates for each application.
+5. The update check request logic already creates new versions automatically if there are updates available.
+6. If the api fails to start the service, a "failed" status along with the reason will be saved in
+the `update_checker_status` column in the database and displayed in the Admin Console's dashboard page.
+
+To configure these options:
+
+1. There will be an additional card in the Admin Console's dashboard page.
```
marccampbell, line 57: An entire tile? What do you think about putting this in a modal for now, accessible via a link in the Version History tile that we have (and on the version history page)?

design/scheduled-update-check.md
salahalsaleh: yes, good point, i'll update

design/scheduled-update-check.md
salahalsaleh: yes, i'll update to mention that here as well.

design/scheduled-update-check.md
salahalsaleh: good point, i agree

design/scheduled-update-check.md
salahalsaleh: good idea

design/scheduled-update-check.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Non Goals
+
+- Deployment of the new updates
+- Scheduled Update Checks for Airgap Installations
+
+## Background
+
+Many customers assume that scheduled update checks are a given and are surprised when they do not receive 
+updates automatically.
```
marccampbell, line 18: nit pick,  there should not be a \n in the doc here, don't force the line breaks.

design/scheduled-update-check.md
```diff
@@ -0,0 +1,64 @@
+# Scheduled Update Checks for Online Installs
+
+KOTS (Admin Console) does not automatically check for new updates. This proposal adds the 
+ability for KOTS (Admin Console) to check for updates every X period of time for online installations.
```
divolgin, line 4: This be a cron job style schedule.  The check should happen at specified time and not drift, which will happen if it's a time interval.

design/scheduled-update-check.md
```diff
@@ -0,0 +1,64 @@
...
+2. The default period of time (X) is configurable.
+3. The ability to disable this feature completely. It will be enabled by default.
+
+The update check will happen in a background process/thread which is triggered every X period of time.
+On each trigger, the process will get all available applications and run update checks for each app.
+The update check will automatically create new versions for each update available in an app.
+
+## Detailed Design
+
+3 new columns (`update_checker_enabled` & `update_checker_interval` & `update_checker_status`) will be
```
divolgin, line 32: How about instead of adding new column, we use config maps for this?

design/scheduled-update-check.md
```diff
@@ -0,0 +1,64 @@
...
+The update check will automatically create new versions for each update available in an app.
+
+## Detailed Design
+
+3 new columns (`update_checker_enabled` & `update_checker_interval` & `update_checker_status`) will be 
+added to the `app` table in the postgress database using a migration.
+
+* The `update_checker_enabled` column is a bool and defaults to `true`. 
+* The `update_checker_interval` column is a text and defaults to `6h`.
+* The `update_checker_status` column is a text and defaults to `NULL`
```
divolgin, line 37: There is already a mechanism for tracking update status.  Scheduled update checks should also be visible on the UI and the user should not be able to click the `Check for updates` button while a scheduled check is running.  This is already built, so let's use it.

design/scheduled-update-check.md
marccampbell: @divolgin this makes sense, but we are worried about "configmap explosion".  We are working on multiple projects that need config right now, and if each need a new config map, that feels a little bad. Ideally we'd have a global config map to use, here, but we don't. 

Agreed that this shouldn't be in the database, but I'm also not opposed to it for now. Would you be ok with this as is?

design/scheduled-update-check.md
marccampbell: :+1: this is a great suggestion.

design/scheduled-update-check.md
divolgin: Yep, db is ok.  We have some of this built already (specifically status and progress), which I mentioned in other comments.

design/scheduled-update-check.md
salahalsaleh: This is the status of the checker, not the update itself. To check if it is running or not, and why.

design/scheduled-update-check.md
salahalsaleh: Right, my initial thoughts were to save those values in a configmap as well. But then i changed my mind due to the reasons marc mentioned above. Also https://github.com/replicatedhq/kots/pull/497#discussion_r420445806

design/scheduled-update-check.md
divolgin: I don't think that's needed.  There is no way to view or use this info.  And you can only run 1 update check at a time, manual and scheduled.

salahalsaleh: There is a way to view this info, it is mentioned here https://github.com/replicatedhq/kots/pull/497/files#diff-ea548131f69005447776a7f865c3c0b2R54

salahalsaleh: How would a customer know if automatic update checks are even running then? the value in the database can indicate it's enabled, but an error can occur and the customer will never know.

divolgin: That's already built.  If you click "check for updates" and there is an error, it will be shown on the UI.  There should be no difference in manual and scheduled update checks other than how they are triggered.

design/scheduled-update-check.md
salahalsaleh: How would this work if the interval is configurable? Would it be better to use a cron job interval instead?

design/scheduled-update-check.md
salahalsaleh: I don't get how that's already built. There is a new thing introduced here "Update Checker". If the checker is not running, there will be _no update_ to fail with an error in the first place.

design/scheduled-update-check.md
divolgin: There's a go package that converts cron spec to time.   I don't think an actual kubernetes CronJob should be used here.

PR #505, opened 2020-05-06
GitOps cluster tracers design proposal

salahalsaleh:


design/gitops-cluster-tracers.md
```diff
@@ -0,0 +1,40 @@
...
+
+When (re)writing the midstream in KOTS, the following will be injected into the kustomization.yaml file:
+
+```
+commonAnnotations:
+  kots.io/app-slug: <app-slug>
+  kots.io/app-sequence: <app-sequence>
+```
+
+the `<app-slug>` will be obtained from the license
```
marccampbell, line 36: I think this should be the kotsadm app slug, which may differ from the license app slug, in the event that there are two instances of the app?

design/gitops-cluster-tracers.md
```diff
@@ -0,0 +1,40 @@
...
+which is currently not possible.
+
+## High-Level Design
+
+Add annotations for both the application slug and sequence to all of the application pods via Kustomize.
+Those annotations will be applied to the downstream manifests before commiting the new changes.
+
+## Detailed Design
+
+When (re)writing the midstream in KOTS, the following will be injected into the kustomization.yaml file:
```
marccampbell, line 28: midtream/kustomization.yaml or downstreams/this-cluster/kustomization.yaml?

design/gitops-cluster-tracers.md
salahalsaleh: nice catch 👍

design/gitops-cluster-tracers.md
salahalsaleh: I was discussing this with @divolgin, and he suggested the midstream. I believe it makes more sense to have it in the midstream, since the annotation will hold the app version and not the app downstream version?

salahalsaleh: since we currently always have 1 downstream, i think it won’t matter, but if we decide to support multiple downstreams in the future, it is preferable to have those annotations for all the downstreams

design/gitops-cluster-tracers.md
```diff
@@ -0,0 +1,35 @@
...
+
+```
+commonAnnotations:
+  kots.io/app-slug: <app-slug>
+  kots.io/app-sequence: <app-sequence>
+```
+
+both the `<app-slug>` & `<app-sequence>` will be added to the pull/rewrite options before calling kots.pull or kots.rewrite from the KOTS Admin Console.
```
marccampbell, line 33: I assume these will be optional annotations? How will this work on the first release, IIRC the app-slug is not yet created when we create this archive.

design/gitops-cluster-tracers.md
```diff
@@ -0,0 +1,35 @@
...
+
```
marccampbell, line 34: Which overlay will the annotations be applied to?

salahalsaleh: the midstream overlay

salahalsaleh: https://github.com/replicatedhq/kots/pull/505/files#diff-509fb7aee9e1aee6a2acfbe20b2603abR25

design/gitops-cluster-tracers.md
salahalsaleh: the app will be available by then, because we create the app before we create the archive. the sequence will be 0 and the app-slug will also be available

PR #509, opened 2020-05-08
Gitops cluster tracers

salahalsaleh:


pkg/midstream/write.go
```diff
@@ -58,6 +61,14 @@ func (m *Midstream) WriteMidstream(options WriteOptions) error {
 		return errors.Wrap(err, "failed to write patches")
 	}
 
+	if options.AppSlug != "" {
+		if m.Kustomization.CommonAnnotations == nil {
+			m.Kustomization.CommonAnnotations = make(map[string]string)
+		}
+		m.Kustomization.CommonAnnotations["kots.io/app-slug"] = options.AppSlug
```
divolgin, line 68: Any concerns that this might cause deployments to fail when annotations are added to objects that cannot be modified?

salahalsaleh: tested and confirmed updated without issues.

PR #515, opened 2020-05-08
Kotsadm gitops cluster tracers

salahalsaleh:


kotsadm/pkg/version/version.go
```diff
@@ -14,6 +14,21 @@ import (
 	"github.com/replicatedhq/kots/kotsadm/pkg/version/types"
 )
 
+// GetNextAppSequence determines next available sequence for this app
+// we shouldn't assume that a.CurrentSequence is accurate. Returns 0 if currentSequence is nil
```
divolgin, line 18: If CurrentSequence  is not accurate, this function should fail with an error.

I think this function should be reverted.  The `createVersion` should be unchanged and use the current logic because it makes sense there.  Elsewhere, just use `app.CurrentSequence + 1` to determine next sequence.

salahalsaleh: I wanted to do that to make sure the new created version and the app sequence we send to as an annotation are always the same

PR #517, opened 2020-05-08
Add '--config-values' option for 'kots pull'

### LINKED ISSUE #514: Testing workflow with different 'userdata' inputs
I'm trying to use `kots pull` as part of a unit testing workflow where I render templates from some config, pipe through kustomize, then validate with conftest:
```
kubectl kots pull <slug> --local-path <dir> --rootdir output --exclude-admin-console --exclude-kots-kinds=false --license-file license.yaml
kubectl kustomize output/<slug>/overlays/midstream | conftest test -
```

`<dir>` is a directory that has my application manifests. For now I have some example config at `<dir>/userdata`. However, that interferes with linting `<dir>` and makes it hard to provide different inputs to test.

I'd like to be able to specify an alternate path for `userdata` so I can easily provide different test inputs.

A step further would be to have a KOTS CLI command that takes the manifest and userdata as arguments and produces the kustomized output on stdout to pipe to conftest.

MikaelSmith:
Adds the `--config-values` option - recentry introduced in `kots install` - to `kots pull` to facilitate a similar workflow for validating rendered manifests.

Fixes #514.


MikaelSmith:
Anything blocking this from being merged?

PR #523, opened 2020-05-12
Using an OCI compliant image registry instead of S3 for storage

### LINKED ISSUE #520: KOTS should use an image registry instead of S3-compatible object store
Instead of requiring an S3 object store, it's common for Kubernetes clusters to have access to an image registry (DockerHub, Quay, ECR, GCR, Harbor, etc). This is already provisioned and is compatible with all environments. Currently, by requiring an object store, we have to distribute Minio / Rook with KOTS, in addition to the Docker Distribution kURL add-on. Finally, since Docker registries are not environment specific, this change would make it easier to support the "bring your own" storage backend.

The design for this is at https://github.com/replicatedhq/kots/blob/c1ada4b1910cc1bf9a918673bdde459e70f30d41/design/registry-instead-of-s3.md

marccampbell:
This is an experimental feature to fix #520. It also adds docker/distribution to the local dev env (leaving minio also).  This is not enabled in dev by default, so no dev env migration is necessary.
This will not impact deployed versions because this feature is not enabled by default.

- [x] Add an OCI compliant image registry to the dev stack
- [x] Update the go code to push image versions to the registry, when configured
- [x] Update the go code to pull image versions from the registry, when configured
- [x] Move the typescript code that reads from s3 (view files page) into go so we don't need to port that
- [x] Upload support bundles to the OCI registry
- [x] Download support bundles from the OCI registry
- [ ] Provide validation during startup that the registry supports ORA
- [ ] Path for migrating from S3 to OCI
- [x] Document the registries we support, put an alpha label on this one (https://github.com/replicatedhq/kots.io/pull/206)

Some quality of life todos before merging:

- [x] Create some functions and shared config in that archive.go file. Building OCI refs is fragile.
- [ ] Update the e2e tests to include an installation on registry

Some new things:

- [ ] Injected analyzers don't contain messages


kotsadm/kustomize/overlays/dev/deployment.yaml
```diff
@@ -42,3 +42,7 @@ spec:
               value: http://replicated-app:3000
             - name: API_ADVERTISE_ENDPOINT
               value: http://127.0.0.1:30065
+            # - name: STORAGE_BASEURI
+            #   value: "docker://kotsadm-storage-registry:5000"
+            # - name: STORAGE_BASEURI_PLAINHTTP
+            #   value: "true"
```
laverya, line 48: should these stay commented out?

kotsadm/pkg/apiserver/server.go
```diff
@@ -78,6 +80,7 @@ func Start() {
 	r.Path("/api/v1/upload").Methods("PUT").HandlerFunc(handlers.UploadExistingApp)
 	r.Path("/api/v1/download").Methods("GET").HandlerFunc(handlers.DownloadApp)
 	r.Path("/api/v1/app/{appSlug}/sequence/{sequence}/renderedcontents").Methods("OPTIONS", "GET").HandlerFunc(handlers.GetAppRenderedContents)
+	r.Path("/api/v1/app/{appSlug}/sequence/{sequence}/contents").Methods("OPTIONS", "GET").HandlerFunc(handlers.GetAppContents)
```
laverya, line 83: is this an entirely new route?

kotsadm/pkg/handlers/contents.go
```diff
@@ -0,0 +1,82 @@
...
+		archiveFiles[strings.TrimPrefix(path, archivePath)] = contents
+		return nil
+	})
+	if err != nil {
+		logger.Error(err)
+		w.WriteHeader(500)
+		return
+	}
+
+	JSON(w, 200, map[string]interface{}{"files": archiveFiles})
```
laverya, line 81: could we make this into a better defined type?
at least map[string]map[string][]byte

kotsadm/kustomize/overlays/dev/deployment.yaml
marccampbell: I was torn on this one. Enabling this makes the environment use the new OCI storage. But there's no migration, so if these are present, it will break envs. 

Any thoughts on how you'd like to see this rolled out as an alpha feature to the skaffold environment?

kotsadm/pkg/apiserver/server.go
marccampbell: Yes.  It used to be a GQL route. This is used on the ViewFiles page to render the application contents. I didn't want to put the OCI get/put code in Typescript, so I ported this endpoint to Go /Rest.

kotsadm/pkg/handlers/troubleshoot.go
```diff
@@ -13,15 +16,219 @@ import (
...
+	if r.Method == "OPTIONS" {
+		w.WriteHeader(200)
+		return
+	}
+
+	getSupportBundleFilesResponse := GetSupportBundleFilesResponse{
+		Success: false,
+	}
+
+	sess, err := session.Parse(r.Header.Get("Authorization"))
```
laverya, line 54: I really wish we could do this in a middleware like with gin

marccampbell: Agreed.  We should write a design proposal up for this change

kotsadm/pkg/handlers/troubleshoot.go
```diff
@@ -13,15 +16,219 @@ import (
...
+		return
+	}
+	defer f.Close()
+
+	w.Header().Set("Content-Type", "application/gzip")
+	w.Header().Set("Content-Disposition", `attachment; filename="supportbundle.tar.gz"`)
+	io.Copy(w, f)
+}
+
+func UploadTroubleshootBundle(w http.ResponseWriter, r *http.Request) {
```
laverya, line 131: should we call this "UploadSupportBundle", like the other functions here?

kotsadm/pkg/supportbundle/supportbundle.go
```diff
@@ -2,38 +2,20 @@ package supportbundle
...
-	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/mholt/archiver"
 	"github.com/pkg/errors"
 	"github.com/replicatedhq/kots/kotsadm/pkg/persistence"
-	kotss3 "github.com/replicatedhq/kots/kotsadm/pkg/s3"
 	"github.com/replicatedhq/kots/kotsadm/pkg/supportbundle/types"
 	"github.com/segmentio/ksuid"
 )
 
-func SetBundleAnalysis(id string, insights []byte) error {
```
laverya, line 19: was this never called?

marccampbell: moved to analysis.go.  this file was big, splitting it up since i added a lot more to it

laverya: 🤦 the downsides of marking files as 'read' as you go is that you can't cmd+f them

PR #524, opened 2020-05-13
Kotsadm scheduled updates checks

salahalsaleh:
Screenshots:

Dashboard configure icon:
Screen Shot 2020-05-12 at 5 08 54 PM

Dashboard tooltip:
Screen Shot 2020-05-12 at 5 09 02 PM

Dashboard modal:
Image 2020-05-12 at 5 09 38 PM

App version history configure icon:
Screen Shot 2020-05-12 at 5 10 18 PM

App version history tooltip:
Screen Shot 2020-05-12 at 5 10 25 PM

App version history modal:
Image 2020-05-12 at 5 10 39 PM


kotsadm/web/src/components/apps/Dashboard.jsx
```diff
@@ -631,6 +713,39 @@ class Dashboard extends Component {
             </div>
           </Modal>
         }
+
```
GraysonNull, line 716: Wondering if we can move this to a `UpdateCheckerModal {...props} />` component since it's being copied multiple places. Could probably even move most if not all of the methods associated with it to the component to clean up repeated code. Also make it easier to update and maintain.

kotsadm/pkg/updatechecker/updatechecker.go
```diff
@@ -0,0 +1,210 @@
...
+// if disabled: stop the current running cron job (if exists)
+// returns a "not supported" error for airgap applications
+func Configure(appID string) error {
+	a, err := app.Get(appID)
+	if err != nil {
+		return errors.Wrap(err, "failed to get app")
+	}
+
+	if a.IsAirgap {
+		return errors.New("airgap scheduled update checks are not supported")
```
divolgin, line 61: this will cause an error to be logged.  this is not an error.

kotsadm/pkg/updatechecker/updatechecker.go
```diff
@@ -0,0 +1,210 @@
...
+	if err != nil {
+		return errors.Wrap(err, "failed to list installed apps")
+	}
+
+	for _, a := range appsList {
+		if a.IsAirgap {
+			continue
+		}
+		if err := Configure(a.ID); err != nil {
+			return errors.Wrapf(err, "failed to configure app %s", a.Slug)
```
divolgin, line 42: this should continue the loop on error and configure the remaining apps

kotsadm/pkg/updatechecker/updatechecker.go
```diff
@@ -0,0 +1,210 @@
...
+	"github.com/replicatedhq/kots/kotsadm/pkg/task"
+	"github.com/replicatedhq/kots/kotsadm/pkg/upstream"
+	"github.com/replicatedhq/kots/kotsadm/pkg/version"
+	kotspull "github.com/replicatedhq/kots/pkg/pull"
+	cron "github.com/robfig/cron/v3"
+	"go.uber.org/zap"
+)
+
+// jobs maps app ids to their cron jobs
+var jobs map[string]*cron.Cron
```
divolgin, line 24: `var jobs = map[string]*cron.Cron`

kotsadm/pkg/updatechecker/updatechecker.go
```diff
@@ -0,0 +1,210 @@
...
+	if a.UpdateCheckerSpec == "@never" || a.UpdateCheckerSpec == "" {
+		Stop(a.ID)
+		return nil
+	}
+
+	if jobs == nil {
+		jobs = make(map[string]*cron.Cron)
+	}
+
+	job, ok := jobs[a.ID]
```
divolgin, line 79: I think there is a bug here somewhere.  If this job was started and never stopped, the function will call `Start` on it twice in a row.

kotsadm/web/src/components/modals/UpdateCheckerModal.jsx
```diff
@@ -0,0 +1,113 @@
...
+        "Authorization": Utilities.getToken(),
+        "Content-Type": "application/json",
+      },
+      method: "PUT",
+      body: JSON.stringify({
+        updateCheckerSpec: updateCheckerSpec,
+      })
+    })
+      .then(async (res) => {
+        if (!res.ok) {
```
divolgin, line 34: The handler always returns a JSON object on error, with the message in `error` key.  This does not use that message.  This will display a JSON string to the user.

kotsadm/pkg/updatechecker/updatechecker.go
salahalsaleh: will or should?

kotsadm/pkg/updatechecker/updatechecker.go
salahalsaleh: which is fine

Image 2020-05-14 at 9 08 18 AM

PR #526, opened 2020-05-13
Don't append null port to URL

divolgin:


kotsadm/web/src/components/apps/DashboardCard.jsx
```diff
@@ -48,20 +48,13 @@ export default class DashboardCard extends React.Component {
     const parsedUrl = url.parse(uri);
 
     let port;
-    if (parsedUrl.port === "80") {
+    if (!parsedUrl.port) {
```
jgruica, line 51: per documentation 

> If the port is `80`, it will not be included in the link.
should this be?
`  if (!parsedUrl.port) {
    port = "";
  } else {
    if (parsedUrl.port === "80") {
      port = "";
    } else {
      port = port = ":" + parsedUrl.port;
    }
  }`


and
>  If the path is `/`, there will not be a duplicate `/` on the link.
`  if (parsedUrl.path === "/") {	
    path = "";	
  } else {	
    path = "/"	
  }`

PR #527, opened 2020-05-13
Helm V3

emosbaugh:


kotsadm/go.mod
```diff
@@ -59,6 +59,7 @@ require (
 )
 
 replace (
+	github.com/replicatedhq/kots => ../
```
emosbaugh, line 62: This is needed for the e2e tests. Feedback please

PR #535, opened 2020-05-14
Don't leave orphaned pods when deleting jobs

### LINKED ISSUE #534: Jobs leave orphaned pods
Create a release with a job that has a hardcoded and and a delete policy hook

```
apiVersion: batch/v1
kind: Job
metadata:
  name: myjob
  annotations:
    "kots.io/hook-delete-policy": "hook-succeeded, hook-failed"
```

When a new version is shipped, the Job object is deleted successfully, but the pod is left on the system

divolgin:
Fixes #534


divolgin:
Fixes https://github.com/replicatedhq/kotsadm/issues/919

PR #547, opened 2020-05-15
Early apply namespaces in addition to CRDs

marccampbell:
This makes it possible for a KOTS application include a namespace definition and manifest that will be deployed into that namespace, while supporting a single "apply" release.

The operator, which handles the application phase, will extract namespaces from the manifest, apply them before applying the rest of the documents.

There's a 5 second hard coded sleep after CRD and namespace that already exists. The Kubernetes API server will have this much time to actually apply the namespace or the deploy will still fail.

A better solution could be to use an informer to allow for the namespace creation to happen in > 5 seconds. This has not been a problem yet, but this is something to watch for.


divolgin:
We'll need to add this to `kind: List` when that's supported

PR #555, opened 2020-05-16
Encrypt plain text values on install

### LINKED ISSUE #485: kots install with plainText --config-values does not encrypt the values
Running a fully automated install command in KOTS 1.15.0 such as :

```shell
kubectl kots install app --license-file ./license.yaml --config-values ./config-values.yaml
```

Will not encrypt any plain text passwords that are supplied in config values. This prohibits a completely automated install if the config screen has any password types.

marccampbell:
Fixes #485


AmberAlston:
@marccampbell should this PR still be open?

PR #558, opened 2020-05-18
default support bundle spec includes velero and restic logs

divolgin:


kotsadm/pkg/handlers/troubleshoot.go
```diff
@@ -538,6 +539,29 @@ func makeKurlCollectors() []*v1beta1.Collect {
...
+	collectors := []*v1beta1.Collect{}
+	for _, selector := range selectors {
+		collectors = append(collectors, &v1beta1.Collect{
+			Logs: &v1beta1.Logs{
+				CollectorMeta: v1beta1.CollectorMeta{
+					CollectorName: "velero",
+				},
+				Name:      "velero",
+				Selector:  []string{selector},
+				Namespace: "velero",
```
marccampbell, line 557: is there a function in the snapshots package that finds this?

divolgin: good idea. updated.

PR #563, opened 2020-05-19
Delete objects in different namespaces

### LINKED ISSUE #564: Resources are not removed with a namespace specified in the yaml when upgrading
From https://github.com/replicatedhq/kotsadm/issues/1408

Steps to reproduce:
1. create a release, specify a namespace other than the target namespace in the yaml for a deployment
2. install that release
3. create a new release and remove that deployment
4. install the new release, the deployment is not removed

additionalNamespaces does not resolve the issue

diffAndRemovePreviousManifests does not call docsByNamespace
https://github.com/replicatedhq/kotsadm/blob/master/operator/pkg/client/deploy.go#L105

emosbaugh:
Fixes #564

### DIFF
--- a/kotsadm/operator/pkg/client/deploy.go
+++ b/kotsadm/operator/pkg/client/deploy.go
@@ -15,6 +15,7 @@ import (
 	batchv1beta1 "k8s.io/api/batch/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	kuberneteserrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
 	k8sschema "k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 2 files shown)


kotsadm/operator/pkg/client/deploy.go
```diff
@@ -44,18 +47,23 @@ func (c *Client) diffAndRemovePreviousManifests(applicationManifests Application
 		return errors.Wrap(err, "failed to decode manifests")
 	}
 
+	targetNamespace := c.TargetNamespace
+	if applicationManifests.Namespace != "." {
```
laverya, line 51: != `"."`, or != `""`?

laverya: I see that that's what it was before, but it feels odd

emosbaugh: ya, idk. but its correct

kotsadm/operator/pkg/client/gvkn.go
```diff
@@ -20,20 +19,24 @@ type OverlySimpleGVKWithName struct {
 }
 
 type OverlySimpleMetadata struct {
-	Name      string `yaml:"name"`
-	Namespace string `yaml:"namespace"`
+	Name        string                 `yaml:"name"`
+	Namespace   string                 `yaml:"namespace"`
+	Annotations map[string]interface{} `json:"annotations"`
```
laverya, line 24: why'd this change? It doesn't seem to be used anywhere in this PR

PR #567, opened 2020-05-20
support bundle redactions api

laverya:
relates to https://github.com/replicatedhq/kots/pull/565 and https://github.com/replicatedhq/troubleshoot/pull/198


kotsadm/pkg/supportbundle/redactions.go
```diff
@@ -0,0 +1,51 @@
...
+	q := `select redact_report from supportbundle where id = $1`
+
+	var redactString sql.NullString
+	row := db.QueryRow(q, bundleID)
+	err := row.Scan(&redactString)
+	if err != nil {
+		return redact.RedactionList{}, errors.Wrap(err, "select redact_report")
+	}
+
+	if !redactString.Valid {
```
divolgin, line 24: Better check `redactString.String == ""` because that column allows empty string values

PR #568, opened 2020-05-20
Proxy public images

divolgin:


design/proxy_public_images.md
```diff
@@ -0,0 +1,53 @@
...
+- All images will be pulled through proxy.replicated.com or registry.replicated.com.
+- None of the other registries need to be accessible through firewalls.
+
+## Non Goals
+
+- There will be no magic anonymous access to registries via proxy.replicated.com:
+    - Vendors will still need to connect to external registry with valid credentials even if all images are public.
+    - `imagePullSecret` will be required in order to pull public images.
+- Public images will not retain original public names on the private instance.  All image names will begin with `proxy.replicated.com`.
+- This change will not apply to airgap builds and installs.  If instance has access to pubic registry, missing image from airgap bundle will still be pulled from the internet.
```
dexhorthy, line 16: I think this make sense, and I understand. I have a (perhaps tangential) question -- what do you expect to happen in the case that

1. `proxyPublicImages` set to true
2. kURL airgap bundle is brought onto an online node
3. kURL airgap bundle is installed with `cat install.sh | sudo bash -s airgap`
4. license is uploaded, online mode selected

In this case, public images would still get pulled from proxy.replicated.com, right?

Does this count as an "airgap install"? Or is "airgap install" only in reference to how the license is configured on initial setup?

divolgin: Yes, images will be proxied.  This still counts as an online install.  None of this is about kURL.

dexhorthy: cool. thank you!

design/proxy_public_images.md
```diff
@@ -0,0 +1,53 @@
...
+
+The call to `IsPrivateImage` function will skipped when `ProxyPublicImages` is set to `true`.  This will require modifying the following functions:
+- `localImageName`
+- `GetPrivateImages`
+- `copyOneImage`
+- `localImageName` (the template function used with `additionalImages` key)
+
+## Alternatives Considered
+
+An alternative to this method is to push images to private repos.  These can be hosted by any registry, including registry.replicated.com.  The downside of this method is the the requirement that vendors modify their CI process or make currently public repos private.
```
dexhorthy, line 48:
```suggestion
An alternative to this method is to push images to private repos.  These can be hosted by any registry, including registry.replicated.com.  The downside of this method is the requirement that vendors modify their CI process or make currently public repos private.
```

design/proxy_public_images.md
```diff
@@ -0,0 +1,53 @@
...
+- This change will not apply to airgap builds and installs.  If instance has access to pubic registry, missing image from airgap bundle will still be pulled from the internet.
+
+## Background
+
+An end customer may desire to keep access to public internet minimal from a KOTS managed cluster.  Delivering public images to the cluster is the only part that requires access to domains other than replicated.com.
+
+## High-Level Design
+
+1. This will be an opt-in feature, enabled by including the new `proxyPublicImages` flag in the `kots.io/Application` spec.
+1. A valid link to the external registry will need to be created by the vendor.
```
dexhorthy, line 25: Does this mean the vendor would need to create an external registry for e.g. `quay.io` if they are pulling public images from quay? Even if they have no private images in quay?

divolgin: Yes.

PR #569, opened 2020-05-20
proposal to add an API to edit/list/etc individual redactors

laverya:
proposal with images: https://github.com/replicatedhq/kots/blob/laverya/individual-redact-spec-api/design/redactors-api.md


design/redactors-api.md
```diff
@@ -0,0 +1,110 @@
...
+Storing metadata about individual redactors requires additional changes.
+Currently, all redactors are stored as a yaml document at the `kotsadm-redact` key within the `kotsadm-redact` configmap.
+Now, redactors will be stored individually at the `<name>` key within the `kotsadm-redact` configmap.
+Further, instead of storing raw yaml there, an object containing metadata and the redactor will be stored.
+
+```go
+type RedactorMetadata struct {
+	Metadata RedactorList `json:"metadata"`
+
+	Redactor v1beta1.Redact `json:"redactor"`
```
laverya, line 104: I should change this to redact
instead of redactor
for consistency

design/redactors-api.md
```diff
@@ -0,0 +1,125 @@
...
+## Background
+
+Custom redactors are being moved to a more informative UI, planned to look something like this:
+![](./images/redactors-api-spec-list.png)
+API routes for individual redactors would greatly simplify implementation and move logic from frontend to backend.
+
+
+## High-Level Design
+
+Five new API routes will be added - `/api/v1/redact/list` to list redactor names, GET/POST/DELETE `/api/v1/redact/individual/{name}` to get/set/delete them by name, and GET/POST `/api/v1/redact/metadata/{name}` to get/update metadata for a redactor.
```
marccampbell, line 25: I think all of the redact route should be under /api/v1/troubleshoot/redact/...

divolgin, line 25: I think `name` should be `slug`

laverya: We already have a get/set route under redact
That can of course be moved, though

laverya: Hmm
Yeah, should be doable

marccampbell: lets move them, unless redactors are going to be used outside of troubleshoot somehow?

design/redactors-api.md
```diff
@@ -0,0 +1,128 @@
...
+	Updated     time.Time `json:"updatedAt"`
+	Enabled     bool      `json:"enabled"`
+	Description string    `json:"description"`
+}
+```
+
+`post` to `/api/v1/redact/metadata/{slug}` should be of the following form:
+```go
+type PostRedactorMetadata struct {
+	Name        bool   `json:"name"`
```
laverya, line 93: this is actually string 🤦

design/redactors-api.md
laverya: I'll make another proposal to move those routes then
But I'm not sure it fits with this proposal

laverya: added that move to this proposal

laverya: undoing the api/v1/troubleshoot move

PR #572, opened 2020-05-21
github actions for kotsadm

divolgin:


.github/workflows/kotsadm.yaml
```diff
@@ -0,0 +1,565 @@
...
+  
+    - name: Download go_api artifact
+      if: ${{ steps.get_tag.outputs.GIT_TAG != '' }}
+      uses: actions/download-artifact@v2
+      with:
+        name: go_api
+        path: ./kotsadm/bin
+
+    - name: Download web artifact
+      if: ${{ steps.get_tag.outputs.GIT_TAG != '' }}
```
laverya, line 356: is there a way to put the 'if' a level higher?

divolgin: no, because github has no global level tag variable.  it has to be generated as a step inside each job

divolgin: job level condition can be controlled with `on.push.tags`, but that applies to the entire workflow and then we'd have to have separate workflows for tagged and untagged steps, and there would be a lot of duplication.

PR #574, opened 2020-05-21
Yaml errors for all app versions

emosbaugh:


kotsadm/api/src/kots_app/kots_app_store.ts
```diff
@@ -748,6 +753,14 @@ export class KotsAppStore {
         commitUrl: row.git_commit_url || "",
         gitDeployable: row.git_deployable
       };
+      if (row.kots_installation_spec) {
+        try {
+          const installationSpec = yaml.safeLoad(row.kots_installation_spec).spec as InstallationSpec;
+          versionItem.yamlErrors = installationSpec.yamlErrors;
+        } catch (err) {
+          console.log(`Failed to unmarshal installation spec yaml for sequence ${versionItem.sequence}`, err);
```
GraysonNull, line 761: Is this supposed to throw an error rather than just console log it?

emosbaugh:
its not critical. i didnt want to fail anything. just catch the error and
log.

PR #577, opened 2020-05-22
redact spec api

laverya:
implements https://github.com/replicatedhq/kots/pull/569

### DIFF
--- a/kotsadm/pkg/handlers/redact.go
+++ b/kotsadm/pkg/handlers/redact.go
@@ -5,9 +5,11 @@ import (
 	"io/ioutil"
 	"net/http"
 
+	"github.com/gorilla/mux"
 	"github.com/replicatedhq/kots/kotsadm/pkg/logger"
 	"github.com/replicatedhq/kots/kotsadm/pkg/redact"
 	"github.com/replicatedhq/kots/kotsadm/pkg/session"
+	"github.com/replicatedhq/kots/pkg/util"
 )
 
 type UpdateRedactRequest struct {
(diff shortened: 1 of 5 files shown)


kotsadm/pkg/handlers/redact.go
```diff
@@ -159,3 +187,271 @@ func GetRedact(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		logger.Error(err)
+		getMetadataResponse.Error = "failed to parse authorization header"
+		JSON(w, 401, getMetadataResponse)
+		return
+	}
+
+	// we don't currently have roles, all valid tokens are valid sessions
+	if sess == nil || sess.ID == "" {
+		getMetadataResponse.Error = "failed to parse authorization header"
```
divolgin, line 214: maybe a different message here, because this one is already used

laverya: sounds like a good idea

kotsadm/pkg/redact/redact.go
```diff
@@ -132,10 +173,278 @@ func SetRedactSpec(spec string) (string, error) {
...
+	}
+
+	newConfigMap, err := clientset.CoreV1().ConfigMaps(os.Getenv("POD_NAMESPACE")).Update(configMap)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to update configmap")
+	}
+	return newConfigMap, nil
+}
+
+func getSlug(name string) string {
```
divolgin, line 357: I don't think this function is 1-to-1.  why is this needed?

laverya: I thought there were restrictions on the keys a configmap could have in k8s
if there aren't, I'd love to drop slugs entirely

PR #587, opened 2020-05-26
Support proxying public images

divolgin:


pkg/template/config_context_test.go
```diff
@@ -425,6 +451,12 @@ func Test_localImageName(t *testing.T) {
 			image:    "redis:latest",
 			expected: "redis:latest",
 		},
+		{
+			name:     "do not rewrite public image",
```
laverya, line 455: this test name seems opposite of the behavior

```suggestion
			name:     "rewrite public image when ProxyPublicImages: true",
```

divolgin: thanks.  copy/paste.

PR #605, opened 2020-05-28
Kotsadm works with HTTP_PROXY

areed:
Includes template functions for proxy and no_proxy.

There was a docker library that that didn't have an option to configure the http client so the original plan of using `PROXY_ADDRESS` env var and opting-in for non-cluster requests didn't work. Instead https://github.com/replicatedhq/kURL/pull/486 expands the NO_PROXY var to include all in-cluster hostnames.

### DIFF
--- a/kotsadm/api/src/snapshots/resolvers/veleroClient.ts
+++ b/kotsadm/api/src/snapshots/resolvers/veleroClient.ts
@@ -66,7 +66,7 @@ export class VeleroClient {
     if (!cluster) {
       throw new Error("No cluster available from kubeconfig");
     }
-    this.server = cluster.server;
+    this.server = "https://kubernetes.default"
   }
 
   // tslint:disable-next-line cyclomatic-complexity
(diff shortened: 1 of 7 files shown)


kotsadm/api/src/snapshots/resolvers/veleroClient.ts
```diff
@@ -66,7 +66,7 @@ export class VeleroClient {
     if (!cluster) {
       throw new Error("No cluster available from kubeconfig");
     }
-    this.server = cluster.server;
+    this.server = "https://kubernetes.default"
```
areed, line 69: This was the cluster IP of the kubernetes service. No proxy includes the whole service CIDR but this node library does not handle cidrs.

pkg/docker/registry/auth.go
```diff
@@ -22,6 +22,7 @@ var (
 			TLSClientConfig: &tls.Config{
 				InsecureSkipVerify: true,
 			},
+			Proxy: http.ProxyFromEnvironment,
```
areed, line 25: GraphQL calls this by FFI when testing registry credentials.

PR #625, opened 2020-06-01
Snapshot in progress check inside ping request

jgruica:


kotsadm/web/src/Root.jsx
```diff
@@ -186,17 +187,25 @@ class Root extends Component {
   }
 
   ping = async (tries = 0) => {
-    await fetch(`${window.env.API_ENDPOINT}/ping`, {
-    }).then(() => {
-      this.setState({ connectionTerminated: false });
+    let url;
+    if (window.location.pathname.split("/").length === 3 && window.location.pathname[2] !== "config") {
+      const appSlug = window.location.pathname.split("/").pop();
```
divolgin, line 192: does this have to be parsed from URL?  Is there no app slug available in some property?

jgruica: i don't think that's necessary. we have to use the appslug only when on "Application Dashboard Component". That's the only component with length 3

divolgin: This is what makes me concerned.  This is very fragile.  Isn't there anywhere else we can get app slug from?  There are a lot of queries using app slug in the UI.

jgruica: This is root component without any props. I can get it from the listAllApps query but I still have to compare what's in URL to find the current app slug.

kotsadm/pkg/handlers/ping.go
```diff
@@ -2,10 +2,60 @@ package handlers
...
+			logger.Error(err)
+			pingResponse.Error = "failed to list backups"
+			return
+		}
+	
+		for _, backup := range backups {
+			if backup.Status == "InProgress" {
+				snapshotInProgressApps := make([]string, 0, 0)
+				snapshotInProgressApps = append(snapshotInProgressApps, currentApp.Slug)
+				pingResponse.SnapshotInProgressApps = snapshotInProgressApps
```
divolgin, line 56: I think you want this here
```
pingResponse.SnapshotInProgressApps = append(pingResponse.SnapshotInProgressApps, currentApp.Slug)
```

PR #627, opened 2020-06-02
failure to get metadata from staging is nonfatal

laverya:


pkg/upstream/replicated.go
```diff
@@ -738,13 +739,14 @@ func GetApplicationMetadata(upstream *url.URL) ([]byte, error) {
...
 		if err != nil {
-			return nil, errors.Wrap(err, "failed to get metadata from staging.replicated.app")
+			otherMetadata = nil
 		}
 
 		metadata = otherMetadata
 	}
 
 	if metadata == nil {
+		log.Info("unable to find metadata for upstream %q", upstream.String())
```
marccampbell, line 749: Should we show this message for applications that do not contain metadata?

laverya: why shouldn't we?

marccampbell: The documentation states that the kots.io/v1beta1 Kind=application (which is the metadata) is an optional resource. We shouldn't log a message about it being missing unless we change our position and make it required or recommended.

if i delivered an application without metadata, i'd want to remove this log line, as it's expected behavior

PR #657, opened 2020-06-16
Don't overwrite existing config values

divolgin:


laverya:
Before we merge this...

I want a “this is what the behavior is going to be” doc
with “this is how you chain functions” and “this is how you make something run once and never again” and “this is how you have something run every time”

Because I'm pretty sure this changes behavior

PR #661, opened 2020-06-17
Moving backup to the API pod and adding a restore container

divolgin:


kotsadm/deploy/Dockerfile
```diff
@@ -1,8 +1,13 @@
 FROM debian:stretch-slim
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
-    curl ca-certificates git \
-  && rm -rf /var/lib/apt/lists/*
+    curl ca-certificates git gnupg2 \
+  && for i in 1 2 3 4 5 6 7 8; do mkdir -p "/usr/share/man/man$i"; done \
```
divolgin, line 5: stretch deletes these directories and they are needed to install some packages

PR #666, opened 2020-06-18
more snapshots

divolgin:


kotsadm/kustomize/overlays/dev/deployment.yaml
```diff
@@ -42,7 +54,9 @@ spec:
               value: http://replicated-app:3000
             - name: API_ADVERTISE_ENDPOINT
               value: http://127.0.0.1:30065
-            # - name: STORAGE_BASEURI
```
marccampbell, line 45: this is handy for testing the ORAs code...

divolgin: oops

Issue #629, opened 2020-06-02
License expiration isn't checked until after fetching application


MikaelSmith:
For large applications - especially in airgap environments - uploading the app bundle can take several minutes. If you provided an expired license, you don't get feedback that it's expired until after the application is uploaded. Then once you get a valid license, you have to upload the application again.

Ideally feedback that the license is expired would happen immediately after uploading it.

marccampbell:
@salahalsaleh will you take a look at this for 1.17?

PR #681, opened 2020-06-23
check license expiration on license upload

salahalsaleh:
Addresses #629 

#### If the license is airgap supported:
The license will not be synced before checking for expiration date, and uploading the license will fail with `License is expired` if the uploaded license is expired.

#### If the license is NOT airgap supported:
The license WILL be synced before checking for expiration date, and if the upstream license is not expired, installation will not be interrupted.

### DIFF
--- a/kotsadm/pkg/handlers/license.go
+++ b/kotsadm/pkg/handlers/license.go
@@ -187,7 +187,7 @@ func UploadNewLicense(w http.ResponseWriter, r *http.Request) {
 	verifiedLicense, err := kotspull.VerifySignature(unverifiedLicense)
 	if err != nil {
 		uploadLicenseResponse.Error = "License signature is not valid"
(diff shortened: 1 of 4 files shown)


kotsadm/pkg/handlers/license.go
```diff
@@ -190,6 +191,32 @@ func UploadNewLicense(w http.ResponseWriter, r *http.Request) {
...
+	expired, err := kotspull.LicenseIsExpired(verifiedLicense)
+	if err != nil {
+		logger.Error(err)
+		uploadLicenseResponse.Error = err.Error()
+		JSON(w, 500, uploadLicenseResponse)
+		return
+	}
+	if expired {
+		uploadLicenseResponse.Error = "License is expired"
+		JSON(w, 200, uploadLicenseResponse)
```
divolgin, line 216: 200 is not the right response here probably

PR #695, opened 2020-06-25
Add NodeCount function

### LINKED ISSUE #696: Template function that returns the number of nodes in a cluster
I'd like a template function that returns the number of nodes in a cluster so I can provide customized defaults for how to deploy applications.

MikaelSmith:
Adds a NodeCount template function that returns the number of nodes currently in the target cluster.

Resolves #696.


MikaelSmith:
This will also require docs additions, but I'm not sure where those live.

marccampbell:
Thanks! We can get this into the docs and included in 1.17.0.

PR #697, opened 2020-06-25
Alpha support for generating yaml only

marccampbell:
With this change:

```
kubectl kots install --yaml --offline --kotsadm-tag 1.16.1 appname > kotsadm.yaml
```

will generate the admin console yaml for kots 1.16.1


marccampbell:
Closing this one in favor of the more kots install to existing airgap cluster workflow. This can be revisited in the future, but at this time, it's not planned.

PR #702, opened 2020-06-26
remove gitops tracers

salahalsaleh:


pkg/midstream/write.go
```diff
@@ -102,20 +93,18 @@ func (m *Midstream) mergeKustomization(existing *kustomizetypes.Kustomization) {
 	newResources := findNewStrings(m.Kustomization.Resources, existing.Resources)
 	m.Kustomization.Resources = append(existing.Resources, newResources...)
 
-	m.Kustomization.CommonAnnotations = mergeMaps(m.Kustomization.CommonAnnotations, existing.CommonAnnotations)
```
divolgin, line 105: this is needed

pkg/midstream/write.go
```diff
@@ -102,20 +93,18 @@ func (m *Midstream) mergeKustomization(existing *kustomizetypes.Kustomization) {
...
 }
 
 func mergeMaps(new map[string]string, existing map[string]string) map[string]string {
 	merged := existing
-	if merged == nil {
-		merged = make(map[string]string)
-	}
-	for key, value := range existing {
-		merged[key] = value
+	if new != nil {
```
divolgin, line 101: why this change?  it is safe to use nil map in a loop

pkg/midstream/write.go
salahalsaleh: hmmm, before gitops tracers were added, this statement never existed. why is it needed now?

pkg/midstream/write.go
salahalsaleh: oh, i did not know that, will update 👍

PR #710, opened 2020-06-29
Airgap existing cluster proposal

marccampbell:


design/airgap-existing-cluster.md
```diff
@@ -0,0 +1,110 @@
+# Support for installing KOTS Admin Console in Airgapped Clusters
+
+## Goals
+
+- Install kotsadm to an existing cluster where the cluster does not have internet access
+- Install kotsadm to an existing cluster from a workstation where the workstation does not have internet access at the time of installation
+- Support different credentials for push / pull (workstation / cluster) to the local registry
+- Support temporary/dynamic and persistent/static credentials for pushing kotsadm images to the registry
+- Support only persistent/static credentials for pulling kotsadm images to the registry
+- Support different credentials for pushing vs pulling images to the local registry
```
dexhorthy, line 10: This seems redundant with 


```
- Support different credentials for push / pull (workstation / cluster) to the local registry
```

is there a difference between these?

design/airgap-existing-cluster.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Non Goals
+
+- Create a single installer for kotsadm + application
+- Create a single download for kotsadm, kots plugin, application and license
+- Automated installs of kotsadm to existing clusters in airgap mode
```
dexhorthy, line 16: Could we clarify "Automated" Here?

```suggestion
- Headless/Automated installs of kotsadm to existing clusters in airgap mode
```

design/airgap-existing-cluster.md
```diff
@@ -0,0 +1,110 @@
...
+- Support private (authenticated) local hosting of kotsadm images
```
dexhorthy, line 17: I'm not sure what we mean by "local hosting" but this needs to include the ability to store the kotsadm images in my private registry

design/airgap-existing-cluster.md
```diff
@@ -0,0 +1,110 @@
...
+
+From the public workstation:
+1. The user downloads the kots plugin suitable for Airgap Workstation.
+1. The user downloads the kotsadm archive containing kots container images
+1. The user downloads the airgap bundle of the application
+1. The user copies the kots plugin, the kotsadm archive and the application airgap bundle onto a medium (eg a USB stick) that can be transfered into the airgapped network.
+
+From the airgap workstation:
+1. The user will install the kots plugin 
+1. The user will copy the container images in the kotsadm archive to the local registry using the kots kubectl plugin
```
dexhorthy, line 37:
```suggestion
1. The user will run a `kubectl kots` command to copy the container images in the kotsadm archive to the local registry, providing registry details for pushing
```

design/airgap-existing-cluster.md
```diff
@@ -0,0 +1,110 @@
...
+1. The user will install kotsadm to the cluster using the kots plugin via `kots install` passing local registry information which will be used to generate the kotsadm yaml and image pull secrets (if neccessary)
+1. The user will install thte application and license using the existing airgap installation process
```
dexhorthy, line 39:
```suggestion
1. The user will install the application and license using the existing airgap installation process 
```

design/airgap-existing-cluster.md
```diff
@@ -0,0 +1,110 @@
+# Support for installing KOTS Admin Console in Airgapped Clusters
+
+## Goals
+
+- Install kotsadm to an existing cluster where the cluster does not have internet access
+- Install kotsadm to an existing cluster from a workstation where the workstation does not have internet access at the time of installation
+- Support different credentials for push / pull (workstation / cluster) to the local registry
+- Support temporary/dynamic and persistent/static credentials for pushing kotsadm images to the registry
```
dexhorthy, line 8: I know it would be great to support a wide variety of credential methods, but I think if we wanted to we could cut a little scope here and have any registry push situation require passing a username and password via environment or CLI flags. Support for other fanciers auth methods could come later.

design/airgap-existing-cluster.md
```diff
@@ -0,0 +1,110 @@
...
+1. The user will install the kots plugin 
+1. The user will copy the container images in the kotsadm archive to the local registry using the kots kubectl plugin
+1. The user will install kotsadm to the cluster using the kots plugin via `kots install` passing local registry information which will be used to generate the kotsadm yaml and image pull secrets (if neccessary)
+1. The user will install thte application and license using the existing airgap installation process 
+
+## Detailed Design
+
+The CI process, on tag, will create the kotsadm archive and include it in the GitHub Release.
+The filename will be `kots_airgap_<version>.tar.gz` and this file contains .tar file exports of all kotsadm images.
+The files SHOULD be stored in the archive in the same format as application airgap files are stored.
```
dexhorthy, line 45: wish we had that format documented somewhere

design/airgap-existing-cluster.md
```diff
@@ -0,0 +1,110 @@
...
+Add a `kots admin-console push-images` to the kots cli to retag and push all kotsadm images to an authenticated registry that is accessible from the cluster, using the desired push credentials.
+This new command is used to copy all of the KOTS Admin Console images from one location to another.
+
+Example:
+
+```
+kubectl kots admin-console push-images ./kotsadm-1.16.1.tar.gz registry.somebigbank.com/my-app \
+  --registry-username push-user \
+  --registry-password abcdef
+```
```
dexhorthy, line 56: lgtm

design/airgap-existing-cluster.md
```diff
@@ -0,0 +1,110 @@
...
+
+`--registry-username` and `--registry-password` are both optional.
+If not present, attempt to authenticate using the already authenticated local docker credentials.
```
dexhorthy, line 59: As noted above, I would vote this goal be a `Stretch` as well, as docker credential providers can get weird to integrate from what I understand, they involve forking out to other binaries sometimes

design/airgap-existing-cluster.md
```diff
@@ -0,0 +1,110 @@
...
+For example:
+
+```
+kubectl kots install --kotsadm-registry registry.somebigbank.com/my-app \
+  --registry-username pull-user \
+  --registry-password fedcba \
+  app-name
+```
+
+If a `registry-username` and `registry-password` are provided, an `imagePullSecret` will be created and deployed with the registry credentials and added to all kotsadm pods.
```
dexhorthy, line 81: I think we also need to consider the case where no registry-username/password are provided, but the user wants to direct kots to add an already-existing imagePullSecret to all workloads

design/airgap-existing-cluster.md
```diff
@@ -0,0 +1,110 @@
...
+Stretch: Additionally, a new `--yaml` flag will be added to have `kots` only generate the yaml (not apply) to support gitops workflows.
+
+## Limitations
+
+- No support for custom branding in airgap installations to existing clusters
+- No support for `requireMinimalRBACPrivileges` set to `true` (all installs will be cluster admin, which matches all use cases of KOTS install today)
+
+## Assumptions
+
+- The workstation will have kubectl access to the cluster
```
dexhorthy, line 96: should these specify which workstation?

design/airgap-existing-cluster.md
markpundsack: Yeah, there's a difference and this part is necessary, although we can make it clearer. Pushing is a one-time action, so you can use temporary credentials to do it, but pulling can happen any time later, as Kubernetes needs to pull in images, so the credentials need to be long-lived. But even with that, it's possible to design a system such that both credentials need to be the same because we only accept the credentials once. We consciously designed it such that you could specify different credentials.

design/airgap-existing-cluster.md
markpundsack: How about "Unattended"?

design/airgap-existing-cluster.md
markpundsack: I think that was a leftover from an earlier iteration.
```suggestion
- Support private (authenticated) local hosting of kotsadm images
```

design/airgap-existing-cluster.md
markpundsack: Good point. I'm fine with making it a stretch.
```suggestion
Stretch: If not present, attempt to authenticate using the already authenticated local docker credentials.
```

design/airgap-existing-cluster.md
markpundsack: @dexhorthy Hmm... how likely is that? Can we start without it? We briefly discussed using advanced secret management like short-lived Vault credentials but decided it was out of scope. If the existing imagePullSecret is a simple username and password, then having to replicate it isn't horrible at least, although it's not DRY and may miss regular rotating if there's some existing mechanism to rotate creds. I think we can cut that for now.

design/airgap-existing-cluster.md
markpundsack:
```suggestion
- The airgap workstation will have kubectl access to the cluster
```

design/airgap-existing-cluster.md
```diff
@@ -0,0 +1,110 @@
...
+- The workstation must have push access to an image registry that the cluster has pull access
```
markpundsack, line 97:
```suggestion
- The airgap workstation must have push access to an image registry that the cluster has pull access 
```

design/airgap-existing-cluster.md
markpundsack: Although if we don't do that stretch, then what's the behavior without it? Are the credentials required then?

design/airgap-existing-cluster.md
```diff
@@ -0,0 +1,110 @@
...
+1. The user will install thte application and license using the existing airgap installation process 
+
+## Detailed Design
+
+The CI process, on tag, will create the kotsadm archive and include it in the GitHub Release.
+The filename will be `kots_airgap_<version>.tar.gz` and this file contains .tar file exports of all kotsadm images.
+The files SHOULD be stored in the archive in the same format as application airgap files are stored.
+
+Add a `kots admin-console push-images` to the kots cli to retag and push all kotsadm images to an authenticated registry that is accessible from the cluster, using the desired push credentials.
+This new command is used to copy all of the KOTS Admin Console images from one location to another.
```
markpundsack, line 48:
```suggestion
* The CI process, on tag, will create the kotsadm archive and include it in the GitHub Release.
* The filename will be `kots_airgap_<version>.tar.gz` and this file contains .tar file exports of all kotsadm images.
* The files SHOULD be stored in the archive in the same format as application airgap files are stored.

* Add a `kots admin-console push-images` to the kots cli to retag and push all kotsadm images to an authenticated registry that is accessible from the cluster, using the desired push credentials.
* This new command is used to copy all of the KOTS Admin Console images from one location to another.
```

design/airgap-existing-cluster.md
markpundsack:
```suggestion
- Unattended, automatic installs of kotsadm to existing clusters in airgap mode
```

design/airgap-existing-cluster.md
```diff
@@ -0,0 +1,110 @@
...
+
+Example:
+
+```
+kubectl kots admin-console push-images ./kotsadm-1.16.1.tar.gz registry.somebigbank.com/my-app \
+  --registry-username push-user \
+  --registry-password abcdef
+```
+
+`--registry-username` and `--registry-password` are both optional.
```
markpundsack, line 58:
```suggestion
Stretch: `--registry-username` and `--registry-password` are both optional.
```

design/airgap-existing-cluster.md
markpundsack: To be clear, we're not really doing anything too special to support this other than asking for the push credentials separately from the pull credentials, and only using the push credentials once. If it required more work, then we could see about cutting it from scope, but as is, it was a natural part of the design anyway, so no additional work, yet still beneficial.

design/airgap-existing-cluster.md
```diff
@@ -0,0 +1,109 @@
...
+The files SHOULD be stored in the archive in the same format as application airgap files are stored.
+
+Add a `kots admin-console push-images` to the kots cli to retag and push all kotsadm images to an authenticated registry that is accessible from the cluster, using the desired push credentials.
+This new command is used to copy all of the KOTS Admin Console images from one location to another.
+
+Example:
+
+```
+kubectl kots admin-console push-images ./kotsadm-1.16.1.tar.gz registry.somebigbank.com/my-app \
+  --registry-username push-user \
```
divolgin, line 53: I propose we replace `registry-user` and `registry-password` with a `registry-secret-name`.   The secret will contain a `.dockerconfigjson` key and the json will contain a key for every registry used by the process (the push registry and the pull registry).  This is the standard k8s way to define registry secrets and this gets us automation at no extra cost.

design/airgap-existing-cluster.md
divolgin: When credentials are not provided, we should assume the registry is anonymous.  If it is not, the command will fail with an appropriate error message.

design/airgap-existing-cluster.md
```diff
@@ -0,0 +1,109 @@
+# Support for installing KOTS Admin Console in Airgapped Clusters
+
+## Goals
+
+- Install kotsadm to an existing cluster where the cluster does not have internet access
+- Install kotsadm to an existing cluster from a workstation where the workstation does not have internet access at the time of installation
+- Support different credentials for push / pull (workstation / cluster) to the local registry
+- Support temporary/dynamic and persistent/static credentials for pushing kotsadm images to the registry
+- Support only persistent/static credentials for pulling kotsadm images to the registry
```
divolgin, line 9: Are these credentials really needed?  All kotsadm images are public, but also they will be downloaded as a github asset according to this doc.

design/airgap-existing-cluster.md
```diff
@@ -0,0 +1,109 @@
...
+It's not currently possible to install KOTS Admin Console into an existing cluster that doesn't have outbound internet access.
+
+## High-Level Design
+
+Airgap Workstation: the computer with kubectl access to the airgap cluster
+Public Workstation: the computer with access to the internet
+
+From the public workstation:
+1. The user downloads the kots plugin suitable for Airgap Workstation.
+1. The user downloads the kotsadm archive containing kots container images
```
divolgin, line 30: I'm not clear on this.  Does the user download the archive from github or does the "kots install" command download everything for the user?

design/airgap-existing-cluster.md
```diff
@@ -0,0 +1,109 @@
...
+1. The user downloads the airgap bundle of the application
+1. The user copies the kots plugin, the kotsadm archive and the application airgap bundle onto a medium (eg a USB stick) that can be transfered into the airgapped network.
+
+From the airgap workstation:
+1. The user will install the kots plugin
```
divolgin, line 35: Is this a manual `cp` command or do we provide a script to do this?

design/airgap-existing-cluster.md
dexhorthy: should this say "pushing" to the registry? I think that's how I had read it

markpundsack: The problem is just that the enterprise isn't going to have a registry that is open, so we expect they'll need a password to fetch the images, even if the images are otherwise public. But also just realized this is phrased wrong. The password isn't necessary to pull images **to** the registry, but **from** the local registry.

markpundsack:
```suggestion
- Support only persistent/static credentials for pulling kotsadm images from the airgapped registry
```

design/airgap-existing-cluster.md
divolgin: i think this refers to image format.  if that's the case, the answer is, `docker-archive`, not `OCI`.

design/airgap-existing-cluster.md
markpundsack: To capture: We talked and agreed to have an install script like kurl airgapped install.

design/airgap-existing-cluster.md
markpundsack: @divolgin said on a call that we already have logic for handling trying the registry anonymously and then trying `.docker/config.json` creds if needed, so sounds like this might be easy to do. Again, it can still be a stretch as it's not strictly necessary.

design/airgap-existing-cluster.md
markpundsack: Talked with @divolgin and we can skip this for now. If we hit the stretch goal above of trying to pull anonymously and if rejected then try with `~/.docker/config.json` credentials, then someone could `docker login` before hand and this would work without leaking credentials into bash history.

design/airgap-existing-cluster.md
markpundsack: We were vague there. :) Worst case, the vendor would send an email to the admin with a link to download. They might add a redirector or cname but it's a straight download. It is **not** downloaded via `kots install`.

It's conceivable to combine downloading the kots plugin and kotsadm archive into the same download tarball, and that would probably be a better experience, but it's not a requirement. Heck, it would be better to download all three things (those two plus the application airgapped bundle) in a single download. But we're trying to limit scope here to a minimal iteration with a passable workflow. We could probably enhance that later.

marccampbell:
Merging this because it's being worked on.

PR #735, opened 2020-07-03
Build kotsadm airgap bundle

divolgin:


.github/workflows/kotsadm.yaml
```diff
@@ -232,7 +232,7 @@ jobs:
         SCOPE_DSN_PUBLIC: ""
       run: SCOPE_DSN=$SCOPE_DSN_PUBLIC make -C kotsadm/kurl_proxy test build
 
-    - name: Upload operator artifact
+    - name: Upload kurl_proxy artifact
```
divolgin, line 235: just a typo

.github/workflows/kotsadm.yaml
```diff
@@ -302,12 +302,6 @@ jobs:
         name: typescript_api_build
         path: ./kotsadm/api/build
 
-    - name: Upload Go API bin artifact
```
divolgin, line 305: this not needed. this is not an artifact that exists

divolgin:
@marccampbell docker push now runs in a container with skopeo, and `azure/docker-login` creates the config on the host in a location not available in that container.

PR #754, opened 2020-07-07
change/fix logging for registry auth token parse failures

laverya:


AmberAlston:
@laverya @divolgin should this PR still be open?

emosbaugh:
@laverya i see this function is never called from the cli yet its using the cli logger. it seems like this may be worth putting a little more effort and merging although i'm not the one to make review and make that call.

laverya:
I'll take another look at things here

PR #780, opened 2020-07-10
404 when app download not found

laverya:


kotsadm/pkg/handlers/download.go
```diff
@@ -24,7 +24,7 @@ func DownloadApp(w http.ResponseWriter, r *http.Request) {
 	a, err := app.GetFromSlug(r.URL.Query().Get("slug"))
 	if err != nil {
 		logger.Error(err)
-		w.WriteHeader(500)
```
divolgin, line 27: how about a `if err == sql.ErrNoRows ` check

laverya: yeah, with `errors.Is`
makes sense

PR #790, opened 2020-07-13
metadata pull failure is not fatal (needed for airgap installs)

divolgin:


cmd/kots/cli/install.go
```diff
@@ -76,7 +76,7 @@ func InstallCmd() *cobra.Command {
 
 			applicationMetadata, err := pull.PullApplicationMetadata(upstream)
 			if err != nil {
-				return errors.Wrap(err, "failed to pull app metadata")
+				log.Info("Failed to pull app metadata: %v", err)
```
marccampbell, line 79: How about making this less negative?

`Unable to pull application metadata. This can be ignored, but custom branding will not be available in the Admin Console until a license is installed`.

PR #804, opened 2020-07-14
Round backup size on snapshot list

jgruica:


AmberAlston:
@jgruica should this PR still be open?

PR #807, opened 2020-07-15
ability to rerun preflight checks

salahalsaleh:
### During app installation, if preflight checks fail:
Image 2020-07-15 at 12 20 20 PM

### Re-run is always visible when viewing preflight results from the Version History page:
Image 2020-07-15 at 12 19 28 PM

### DIFF
--- a/kotsadm/pkg/apiserver/server.go
+++ b/kotsadm/pkg/apiserver/server.go
@@ -93,7 +93,7 @@ func Start() {
 	// Implemented handlers
 	r.Path("/api/v1/license/platform").Methods("OPTIONS", "POST").HandlerFunc(handlers.ExchangePlatformLicense)
 	r.Path("/api/v1/app/{appSlug}/sequence/{sequence}/preflight/ignore-rbac").Methods("OPTIONS", "POST").HandlerFunc(handlers.IgnorePreflightRBACErrors)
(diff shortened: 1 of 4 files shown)


kotsadm/pkg/apiserver/server.go
```diff
@@ -93,7 +93,7 @@ func Start() {
 	// Implemented handlers
 	r.Path("/api/v1/license/platform").Methods("OPTIONS", "POST").HandlerFunc(handlers.ExchangePlatformLicense)
 	r.Path("/api/v1/app/{appSlug}/sequence/{sequence}/preflight/ignore-rbac").Methods("OPTIONS", "POST").HandlerFunc(handlers.IgnorePreflightRBACErrors)
-	r.Path("/api/v1/app/{appSlug}/preflight/run").Methods("OPTIONS", "POST").HandlerFunc(handlers.StartPreflightChecks)
+	r.Path("/api/v1/app/{appSlug}/sequence/{sequence}/preflight/run").Methods("OPTIONS", "POST").HandlerFunc(handlers.StartPreflightChecks)
```
salahalsaleh, line 96: @divolgin calls to this route were removed in this pr, so it should be safe to modify https://github.com/replicatedhq/kots/pull/713/files

The commit which added this route: https://github.com/replicatedhq/kots/commit/c9e1f635

PR #814, opened 2020-07-17
add a route to only update a redactor's enabled status

laverya:
the existing route requires you to update the description, redactor yaml, etc at the same time

Usage is to post `{"enabled":true}` or `{"enabled":false}` to `/api/v1/redact/enabled/{slug}` to enable/disable a redactor. It returns the full redactor yaml+metadata struct

### DIFF
--- a/kotsadm/pkg/redact/redact.go
+++ b/kotsadm/pkg/redact/redact.go
@@ -44,7 +44,7 @@ type RedactorMetadata struct {
 func GetRedactSpec() (string, string, error) {
 	configMap, errstr, err := getConfigmap()
 	if err != nil || configMap == nil {
-		return "", errstr, err
+		return "", errstr, errors.Wrap(err, "get redactors configmap")
 	}
 
 	return getRedactSpec(configMap)
@@ -66,7 +66,7 @@ func getRedactSpec(configMap *v1.ConfigMap) (string, string, error) {
 func GetRedact() (*v1beta1.Redactor, error) {
 	configmap, _, err := getConfigmap()
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "get redactors configmap")
 	}
 	if configmap == nil {
 		return nil, nil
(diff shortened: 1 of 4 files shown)


kotsadm/pkg/redact/redact.go
```diff
@@ -186,6 +186,52 @@ func SetRedactYaml(slug, description string, enabled, newRedact bool, yamlBytes
 	return redactorEntry, nil
 }
 
+// sets whether an individual redactor is enabled
+func SetRedactEnabled(slug string, enabled bool) (*RedactorMetadata, error) {
+	configMap, _, err := getConfigmap()
+	if err != nil {
+		return nil, err
```
salahalsaleh, line 193: errors.wrap?

kotsadm/pkg/redact/redact.go
```diff
@@ -186,6 +186,52 @@ func SetRedactYaml(slug, description string, enabled, newRedact bool, yamlBytes
...
+	}
+
+	newData, redactorEntry, err := setRedactEnabled(slug, enabled, time.Now(), configMap.Data)
+	if err != nil {
+		return nil, err
```
salahalsaleh, line 198: dito

laverya: this one shouldn't be errors.wrap, since it's an internal function only called in one place
the functions are separated for testing purposes

kotsadm/pkg/redact/redact.go
laverya: getConfigmap returns wrapped errors already: https://github.com/replicatedhq/kots/blob/cf3809d6209d06d12b2fa2a21db5e4fb2f5072cc/kotsadm/pkg/redact/redact.go#L327

Adding another layer doesn't seem useful here

laverya: that being said, we're inconsistent in whether this function has errors wrapped
I'll just add it everywhere

salahalsaleh: Okay, thanks

PR #815, opened 2020-07-17
Conditional status informers design proposal

salahalsaleh:


design/conditional-status-informers.md
```diff
@@ -0,0 +1,40 @@
...
+Many of them have requested the ability to have conditional status informers so that the status informer can be excluded when selecting the external db option.
+
+## High-Level Design
+
+In the kots application spec, modify the `statusInformers` array type to include an optional `exclude` field with each entry.
+Then, config template functions (e.g. ConfigOptionEquals) can be used in the `exclude` field value to indicate when/if to exclude the entry (informer) from the list.
+
+## Detailed Design
+
+First, modify the `statusInformers` array in the kots application spec to be an array of interfaces instead of strings to support both the old and the new format (explanation of new format below).
```
divolgin, line 27: Is this Go or TypeScript?

design/conditional-status-informers.md
```diff
@@ -0,0 +1,40 @@
...
+
+In the new format, each entry can be either a string (same as old format), or an object that consists of two fields, a string `resource` field, and a boolstring `exclude` field.
+The `resource` field is required, but the `exclude` field is optional.
+
+Second, when creating or updating an app version:
+
+1- Parse the kots application spec and read the `statusInformers` array.
+2- Check the type for each entry in the array, if not a string, parse in the new format.
```
divolgin, line 35: Can you link to the part of the code that will be doing this?

design/conditional-status-informers.md
```diff
@@ -0,0 +1,36 @@
...
+Add the ability to render/template the `statusInformers` array in the kots application spec (`kind: Application`).
+This way, config template functions (e.g. ConfigOptionEquals) can be used to decide whether or not an informer should be included/applied.
+
+## Detailed Design
+
+## Notes:
+- Getting the list of status informers happens during the app deploy loop in TypeScript.
+- The list of informers is then sent through a socket to the operator to be applied.
+
+1- To add the ability to exclude specific status informers, the `statusInformers` array will be rendered using the `RenderFile` function from kots ffi before it's sent to the operator (the `RenderFile` function can take any string as input to render).
```
markpundsack, line 31: Is *only* the `statusInformers` array going to be rendered, or will the whole file be rendered?

design/conditional-status-informers.md
```diff
@@ -0,0 +1,36 @@
...
+2- This way, conditional template functions can be used as an entry in the `statusInformers` array in the kots application spec.
+3- The template function will then resolve to either a valid status informer entry, or not (an empty string `""` for example).
+4- Invalid status informers (entries) will be excluded from the array by the operator.
```
markpundsack, line 34: Do we currently silently ignore invalid status informers? This seems risky. If someone typos something, we shouldn't silently ignore it. This seems like the intent is to provide an alternate way to ignore informers for missing objects, but since we provide a way to explicitly handle that, I'd rather than also have an implicit way to handle it, that could backfire.

design/conditional-status-informers.md
markpundsack: Rendering the whole file would make sense from a user mental-model perspective, but probably has complications, like some content from the file is used prior to having config available. e.g. if the application icon is conditional on config, we wouldn't have that config when first rendering it.

design/conditional-status-informers.md
```diff
@@ -0,0 +1,36 @@
...
+
+Note: the pattern of rendering a kotskinds file on the fly already exists and used for preflights where the preflight spec is rendered before loading and executing preflights. For reference: https://github.com/replicatedhq/kots/blob/7f621f7e/kotsadm/pkg/preflight/preflight.go#L59
```
markpundsack, line 36: Are there any special concerns for airgap mode?

design/conditional-status-informers.md
salahalsaleh: Yes, in this proposal, only the `statusInformers` array is going to be rendered. I've discussed this from a technical perspective with @divolgin, and this seems to be the preferable approach because:

1- There are some fields that we shouldn't support rendering/templating for in the application spec (similar to what you pointed above. @divolgin might have a better idea of what those fields are).
2- `kotskinds` objects are usually only rendered on the fly (except for the config) and the rendered version is not stored in the database or s3 (because `kotskinds` are not part of the deployable manifests). This way, the spec can be rendered when needed without modifying the original data.

@divolgin can give his input on this one too.

design/conditional-status-informers.md
salahalsaleh: Hmmmm, not that I can think of, no.

design/conditional-status-informers.md
salahalsaleh: Currently, if a status informer fails to parse (has invalid format), we only log the error. In some cases, this might lead to the app status being shown as `Missing`, but I think we need to improve the UX for this.

design/conditional-status-informers.md
markpundsack: Thanks, that makes sense.

design/conditional-status-informers.md
markpundsack: Given that only this portion of this file is rendered, I can't think of anything either.

design/conditional-status-informers.md
markpundsack: Yeah, some improvements on the UX would be good there. Any suggestions?

salahalsaleh: If a status informer is not running, the application status will not be `Ready`, in this case (or maybe in all cases), we can show a list of the running status informers along with the ones that are not running. This can help the user detect typos in the status informers. Also, we can update the linter to look for those types of mistakes, where it can detect if a status informer is pointing to a non-existing object (even when using template functions).

divolgin: @salahalsaleh what is the difference between an invalid informer and an empty string informer?

Also, can you link to code where these operations will happen?  As far as I know there is no way to render anything in TS.   Are we adding new FFI functions for this?

salahalsaleh: @divolgin An invalid informer will typically have a typo, but an empty string informer will be the result of templating and be excluded.

We do have an existing ffi render function in TS and it's used in multiple places: https://github.com/replicatedhq/kots/blob/ca6a40ba/kotsadm/api/src/kots_app/kots_ffi.ts#L202

PR #819, opened 2020-07-20
Hide yamlErrors for now

emosbaugh:


kotsadm/web/src/components/apps/AppVersionHistory.jsx
```diff
@@ -140,7 +140,8 @@ class AppVersionHistory extends Component {
     const { app } = this.props;
     const downstream = app.downstreams[0];
     const diffSummary = this.getVersionDiffSummary(version);
-    const hasYamlErrors = this.hasYamlErrors(downstream, version);
+    let hasYamlErrors = this.hasYamlErrors(downstream, version);
+    hasYamlErrors = false;
```
GraysonNull, line 144: Is the intent to have this always return false?

emosbaugh: yes

Issue #827, opened 2020-07-22
kots upstream upgrade ignores the deploy flag if no updates are downloaded.


divolgin:
1. Ship some updates and ensure they show up in Version history (but not deployed).
2. Run the upgrade command with `--deploy` flag.  No version will be deployed.

```
kubectl kots upstream upgrade <app slug> -n <namespace> --deploy
  • Checking for application updates ✓
  • There are no application updates available, ensuring latest is marked as deployed
  • To access the Admin Console, run kubectl kots admin-console --namespace <namespace>
```

PR #831, opened 2020-07-22
Fix: upstream update should deploy

marccampbell:
Fixes: #827

### DIFF
--- a/kotsadm/pkg/updatechecker/updatechecker.go
+++ b/kotsadm/pkg/updatechecker/updatechecker.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/replicatedhq/kots/kotsadm/pkg/app"
+	"github.com/replicatedhq/kots/kotsadm/pkg/downstream"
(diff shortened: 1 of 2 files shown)


kotsadm/pkg/updatechecker/updatechecker.go
```diff
@@ -205,6 +206,32 @@ func CheckForUpdates(appID string, deploy bool) (int64, error) {
...
+
+		// ensure that the latest version is deployed
+		allVersions, err := version.GetVersions(a.ID)
+		if err != nil {
+			return 0, errors.Wrap(err, "failed to list app versions")
+		}
+
+		// get the first version, the array must contain versions at this point
+		// this function can't run without an app
+		latestVersion := allVersions[len(allVersions)-1]
```
divolgin, line 221: judging by the implementation for `GetVersions`, this could be an empty array

marccampbell: i'll add a check here.  my comment was more about "many many other places above me in the call stack would have failed, and it's impossible to reach this point".  but fair, i'll add

PR #832, opened 2020-07-22
YAML File Errors Proposal

emosbaugh:


design/yaml_file_errors.md
```diff
@@ -0,0 +1,49 @@
...
+    path: deployment.yaml
+```
+
+An example file tree:
+
+```
+> upstream
+> base
+> overlays
+v errors
```
emosbaugh, line 47: is there a better name for this?

design/yaml_file_errors.md
```diff
@@ -0,0 +1,49 @@
...
+spec:
+  channelName: Stable
+  knownImages:
+  - image: docker.io/sentry-enterprise/app:1.0
+    isPrivate: true
+  releaseNotes: |
+    Release 1.0.3
+  updateCursor: "3"
+  versionLabel: 1.0.3
+  yamlErrors:
```
emosbaugh, line 36: it would be more clear in my opinion if we dropped a file into the errors dir that has yaml errors in it. how do we ensure the file name does not collide with a vendor yaml file?

design/yaml_file_errors.md
markpundsack: `skipped`?

Or maybe we need to be verbose, since it's a totally unexpected directory. `skipped files due to YAML errors`? I'm only half kidding.

design/yaml_file_errors.md
markpundsack: I'm not sure I understand this comment. Can you rephrase?

design/yaml_file_errors.md
```diff
@@ -0,0 +1,49 @@
...
+
+- Surface the rendered YAML content to the user when unmarshalling fails
+
+## Non Goals
+
+- Attempt to automate fixing any common errors such as unescaped quotes or line breaks
+
+## Design
+
+Add a new directory at the root of the file tree `errors` that will include the raw content of all rendered output for yaml files that failed unmarshalling.
```
markpundsack, line 18: As a stretch, it would be nice if when someone viewed the error'd file, we could display the error message(s) in context. Like at least a note at the header of the page saying `line 26: did not find expected key`, but maybe even highlight the line(s) for them, even if they didn't come there from the version history with a deep link to the line.

design/yaml_file_errors.md
markpundsack: Do you mean to include a meta file with all of the errors listed? Like `errors.yaml`? And then your concern is that `errors.yaml` could collide with a vendor file?

If so, I'd say a better UX would be to show errors in context when viewing each file.

design/yaml_file_errors.md
emosbaugh: renderErrors? ... or render_errors or render-errors or rendererrors

design/yaml_file_errors.md
emosbaugh: yes. i agree that we could make lots of ux improvements. but we also have to take into account kots pull. in this case only the file tree exists.

i suppose i could just add a file _errors.yaml which should not collide. does this file need to have a gvk? or am i overcomplicating things and should i just leave it in the installation?

design/yaml_file_errors.md
emosbaugh: the idea would be that with the additions to the api (file tree and structured kots.io/v1beta1.Installation) the ui team should be able to achieve this or whatever is designed.

design/yaml_file_errors.md
markpundsack: Hmm... OK. I hadn't thought of `kots pull`. Makes sense. I'll defer to your judgement on gvk.

design/yaml_file_errors.md
emosbaugh: i ended up using skippedFiles

design/yaml_file_errors.md
emosbaugh: not gonna add a gvk

design/yaml_file_errors.md
markpundsack: I'd prefer to be explicit here, and include that stretch goal for the UI in the design. If, for whatever reason, the API doesn't provide enough information for the frontend to display the in-context messages, we should consider changes. If we're lucky and there are no changes, then great.

Issue #518, opened 2020-05-08
TLS certs and keys do not match if generated on the config page


laverya:
The `TLSCert` and `TLSKey` template functions do not seem to coordinate properly when used on the config page.

For example,
```
apiVersion: kots.io/v1beta1
kind: Config
metadata:
  name: app-slug
spec:
  groups:
    - name: certs
      title: generated certs
      items:
          - name: platform_host
            title: platform host
            type: text
          - name: upload_keys
            title: Upload SSL keys
            type: bool
            default: false
          - name: generated_tls_key
            readonly: true
            type: textarea
            value: repl{{ TLSKey "self_signed_cert" }}
          - name: generated_tls_cert
            readonly: true
            type: textarea
            value: repl{{ TLSCert "self_signed_cert" (ConfigOption "platform_host") nil nil 365 }}
          - name: uploaded_tls_key
            title: TLS Key (SSL)
            type: textarea
            required: true
            when: '{{repl (ConfigOptionEquals "upload_keys" "1")}}'
          - name: uploaded_tls_cert
            title: TLS Cert (SSL)
            type: textarea
            required: true
            when: '{{repl (ConfigOptionEquals "upload_keys" "1")}}'
          - name: tls_key
            hidden: true
            readonly: true
            type: textarea
            value: '{{repl if (ConfigOptionEquals "upload_keys" "1")}}repl{{ConfigOption "uploaded_tls_key"}}{{repl else}}repl{{ConfigOption "generated_tls_key"}}{{repl end}}'
          - name: tls_cert
            hidden: true
            readonly: true
            type: textarea
            value: '{{repl if (ConfigOptionEquals "upload_keys" "1")}}repl{{ConfigOption "uploaded_tls_cert"}}{{repl else}}repl{{ConfigOption "generated_tls_cert"}}{{repl end}}'
```
the above config should produce a valid cert and key at `tls_cert` and `tls_key`.

However, this can be shown not to be the case with the following configmap:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: example-config
data:
    TLS_CERT: '{{repl ConfigOption "tls_cert" | Base64Encode }}'
    TLS_KEY: '{{repl ConfigOption "tls_key" | Base64Encode }}'
    LOCAL_TLS_KEY: '{{repl TLSKey "self_signed_cert" | Base64Encode }}'
```
While all will be populated, LOCAL_TLS_KEY does not match TLS_KEY, and neither match TLS_CERT.

[...]

PR #833, opened 2020-07-22
add TLSCert and TLSKey to configoption dep resolution

laverya:


laverya:
should resolve #518

emosbaugh:
Will need to add more to rep resolution after merging https://github.com/replicatedhq/kots/pull/845

PR #845, opened 2020-07-24
Generate CA and certificates from CA

### LINKED ISSUE #844: Need to create certificates signed by a generated CA
This is primarily a requirement to work around https://github.com/openssl/openssl/issues/1418. All current OpenSSL releases won't verify certificates when provided a self-signed certificate. It will only accept a self-signed certificate if the verifying certificate includes the KeyCertSign bit in KeyUsage. The simplest way to work around this is first generate a CA cert/key pair, then use that to generate a certificate. The application that needs to trust the generated certificate can verify against the generated CA certificate.

This can be approximately accomplished with
```
repl{{$ca := genCA "ca" 365}}repl{{$cert := genSignedCert "host" nil nil 365 $ca}}repl{{cat ($ca.Cert | b64enc) ($cert.Cert | b64enc) ($cert.Key | b64enc)}}
```
and then using split and indexing to access individual entries. However using this in a Config reliably is again difficult because you either regenerate it every time the config is evaluated (producing a different certificate), or generate it once and never change it (so the certs can't depend on other config inputs).

I'd like to have an equivalent of TLSCert/TLSKey that caches the generated certs and can access them again by name. This provides sufficient longevity to make them useful as an easy onboard tool.

MikaelSmith:
Adds `TLSCACert`, `TLSCertFromCA`, and `TLSKeyFromCA` to generate a named CA cert/key pair and access the cert, as well as generate cert/key pairs from that CA as needed (based on the combination of CA name, cert name, and common name).

Useful when you need a separate CA certificate, such as working around https://github.com/openssl/openssl/issues/1418.

Fixes #844.

### DIFF
--- a/pkg/template/static_context.go
+++ b/pkg/template/static_context.go
@@ -395,18 +395,9 @@ func (ctx StaticCtx) tlsKey(certName string, args ...interface{}) string {
 	return p.Key
 }
 
-func (ctx StaticCtx) tlsCaCert(caName string, args ...interface{}) string {
+func (ctx StaticCtx) tlsCaCert(caName string, daysValid int) string {
 	cap, ok := caMap[caName]
 	if !ok {
-		if len(args) != 1 {
-			return ""
-		}
-
-		daysValid, ok := args[0].(int)
-		if !ok {
-			return ""
-		}
-
 		cap = genCa(caName, daysValid)
 		caMap[caName] = cap
 	}
@@ -414,7 +405,7 @@ func (ctx StaticCtx) tlsCaCert(caName string, args ...interface{}) string {
 	return cap.Cert
 }
 
-func (ctx StaticCtx) tlsCertFromCa(caName, certName, cn string, ips []interface{}, alternateDNS []interface{}, daysValid int) string {
+func (ctx StaticCtx) tlsCertFromCa(caName, certName, cn string, ips, alternateDNS []interface{}, daysValid int) string {
 	key := fmt.Sprintf("%s:%s:%s", caName, certName, cn)
 	if p, ok := tlsMap[key]; ok {
 		return p.Cert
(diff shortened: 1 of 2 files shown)


pkg/template/static_context.go
```diff
@@ -390,14 +395,99 @@ func (ctx StaticCtx) tlsKey(certName string, args ...interface{}) string {
 	return p.Key
 }
 
+func (ctx StaticCtx) tlsCaCert(caName string, args ...interface{}) string {
```
emosbaugh, line 398: any reason not to make the args explicit here? i believe the reason tlsKey uses arbitrary args was to preserve backwards compatibility

https://github.com/replicatedhq/kots/commit/38898458f0cbc5bc2b3b785b5a7b2520a9a7f7dd#diff-b82706ecb35ffdea1e9f78aa454d2ec8R348

pkg/template/static_context.go
```diff
@@ -390,14 +395,99 @@ func (ctx StaticCtx) tlsKey(certName string, args ...interface{}) string {
...
+	if p, ok := tlsMap[key]; ok {
+		return p.Cert
+	}
+
+	p := genSignedCert(caName, cn, ips, alternateDNS, daysValid)
+	tlsMap[key] = p
+	return p.Cert
+}
+
+func (ctx StaticCtx) tlsKeyFromCa(caName, certName, cn string, args ...interface{}) string {
```
emosbaugh, line 428: same here. can you make this take explicit args if no other reason than following the convention from tlsKey? thanks

laverya:
I think that if we merge this, I'll definitely need to add it to https://github.com/replicatedhq/kots/pull/833 (which won't be hard to do, since the framework's already there)

pkg/template/static_context.go
MikaelSmith: Ah, wasn't sure why that choice was made. It seemed reasonable to allow calling TLSCertFromCA then call TLSCACert with just the CA identifier to get the cert that had already been generated.

pkg/template/static_context.go
MikaelSmith: I'm not sure I understand, this is basically following the convention from tlsKey (except dropping the legacy reference by only certName).

MikaelSmith:
Pushed up changes to always require all parameters.

emosbaugh:
Looks good. Thanks for the contribution!

PR #848, opened 2020-07-24
Get app dashboard

salahalsaleh:


kotsadm/pkg/handlers/dashboard.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	parentSequence, err := downstream.GetParentSequence(a.ID, clusterID)
+	if err != nil {
+		logger.Error(err)
+		w.WriteHeader(500)
+		return
+	}
+
+	metrics, err := version.GetMetricCharts(a.ID, parentSequence)
```
divolgin, line 60: i think this handler should not fail if we can't get metrics of prometheus address.

PR #854, opened 2020-07-24
Viewing invalid yaml files

jgruica:
Screen Shot 2020-07-27 at 3 36 55 PM
Screen Shot 2020-07-27 at 3 37 22 PM

### DIFF
--- a/kotsadm/web/src/components/apps/AppVersionHistory.jsx
+++ b/kotsadm/web/src/components/apps/AppVersionHistory.jsx
@@ -61,7 +61,8 @@ class AppVersionHistory extends Component {
     showUpdateCheckerModal: false,
     displayShowDetailsModal: false,
     yamlErrorDetails: [],
-    deployView: false
+    deployView: false,
+    versionSequence: ""
   }
 
   componentDidMount() {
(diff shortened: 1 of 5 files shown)


kotsadm/web/src/components/apps/AppVersionHistory.jsx
```diff
@@ -149,43 +150,67 @@ class AppVersionHistory extends Component {
...
+                      secondSequence: version.parentSequence
+                    });
+                  }
+                }}
+              >
+                <span className="files">{diffSummary.filesChanged} files changed </span>
+                <span className="lines-added">+{diffSummary.linesAdded} </span>
+                <span className="lines-removed">-{diffSummary.linesRemoved}</span>
+              </div>
+              <div className="flex flex1 alignItems--center u-marginLeft--10">
```
salahalsaleh, line 171: i feel like the `yamlErrorsDetails` can be used here instead of duplicating the whole div

kotsadm/web/src/components/apps/AppVersionHistory.jsx
```diff
@@ -61,7 +61,8 @@ class AppVersionHistory extends Component {
     showUpdateCheckerModal: false,
     displayShowDetailsModal: false,
     yamlErrorDetails: [],
-    deployView: false
+    deployView: false,
+    versionSequence: ""
```
salahalsaleh, line 65: hmmm, i feel like there can be a better name for this. this does not describe what this variable is used for IMO.

kotsadm/web/src/components/apps/AppVersionHistory.jsx
salahalsaleh, line 171: i feel like the yamlErrorsDetails can be used here instead of duplicating the whole div

PR #859, opened 2020-07-27
move listSupportBundles query to go

salahalsaleh:


kotsadm/pkg/handlers/troubleshoot.go
```diff
@@ -99,6 +119,81 @@ func GetSupportBundleFiles(w http.ResponseWriter, r *http.Request) {
...
+	sess, err := session.Parse(r.Header.Get("Authorization"))
+	if err != nil {
+		logger.Error(err)
+		w.WriteHeader(401)
+		return
+	}
+
+	// we don't currently have roles, all valid tokens are valid sessions
+	if sess == nil || sess.ID == "" {
+		logger.Error(err)
```
divolgin, line 140: wrong log.  err is nil

kotsadm/pkg/handlers/troubleshoot.go
```diff
@@ -99,6 +119,81 @@ func GetSupportBundleFiles(w http.ResponseWriter, r *http.Request) {
...
+		w.WriteHeader(500)
+		return
+	}
+
+	responseSupportBundles := []ResponseSupportBundle{}
+	for _, bundle := range supportBundles {
+		licenseType, err := supportbundle.GetLicenseType(bundle.ID)
+		if err != nil {
+			logger.Error(errors.Wrapf(err, "failed to get license type for bundle %s", bundle.Slug))
```
divolgin, line 165: missing return

kotsadm/pkg/handlers/troubleshoot.go
```diff
@@ -99,6 +119,81 @@ func GetSupportBundleFiles(w http.ResponseWriter, r *http.Request) {
...
+		}
+
+		analysis, err := supportbundle.GetBundleAnalysis(bundle.ID)
+		if err != nil {
+			logger.Error(errors.Wrapf(err, "failed to get analysis for bundle %s", bundle.Slug))
+		}
```
divolgin, line 171: missing return

kotsadm/pkg/handlers/troubleshoot.go
salahalsaleh: hmmm, i don't think failing to get the license type should fail the whole request or not include the support bundle in the list

kotsadm/pkg/handlers/troubleshoot.go
salahalsaleh: dito

PR #875, opened 2020-07-28
move collectoSupportBundle query to go

salahalsaleh:


kotsadm/web/src/components/troubleshoot/GenerateSupportBundle.jsx
```diff
@@ -275,11 +284,4 @@ class GenerateSupportBundle extends React.Component {
   }
 }
 
-export default withRouter(compose(
-  withApollo,
```
jgruica, line 279: is this imported on top? (withApollo,compose,graphql) we should delete it because it's unused

salahalsaleh: yep, good catch

PR #880, opened 2020-07-29
Use proxy info from environment in kotsadm API

### LINKED ISSUE #770: Ability to install kots/adm in a network that requires HTTP_PROXY and friends
Hi team -- I'd like to check on what kind of support we have for

1. running `kots install` from behind a proxy
2. running kotsadm in an existing cluster from behind a proxy
3. running other installation methods like embbeded-cluster via https://kurl.sh from behind a proxy

- In what scenarios, if any, is this supported? Existing Cluster? Embedded Cluster via https://kurl.sh?
- What is the level of support for this today? Alpha? Beta? GA
- If it is supported, then perhaps this becomes a documentation issue? What do I need to do to leverage this?

divolgin:
closes #770

### DIFF
--- a/cmd/kots/cli/install.go
+++ b/cmd/kots/cli/install.go
@@ -124,6 +124,9 @@ func InstallCmd() *cobra.Command {
 				IncludeMinio:              v.GetBool("deploy-minio"),
 				IncludeDockerDistribution: v.GetBool("deploy-dockerdistribution"),
 				Timeout:                   time.Minute * 2,
+				HTTPProxyEnvValue:         v.GetString("http-proxy"),
+				HTTPSProxyEnvValue:        v.GetString("https-proxy"),
+				NoProxyEnvValue:           v.GetString("no-proxy"),
 
 				KotsadmOptions: kotsadmtypes.KotsadmOptions{
 					OverrideVersion:   v.GetString("kotsadm-tag"),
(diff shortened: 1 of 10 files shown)


marccampbell:
@divolgin the implementation here looks good, but should this be opt-in via a CLI flag?

cmd/kots/cli/install.go
```diff
@@ -222,6 +240,10 @@ func InstallCmd() *cobra.Command {
 	cmd.Flags().String("config-values", "", "path to a manifest containing config values (must be apiVersion: kots.io/v1beta1, kind: ConfigValues)")
 	cmd.Flags().Bool("port-forward", true, "set to false to disable automatic port forward")
 	cmd.Flags().String("wait-duration", "2m", "timeout out to be used while waiting for individual components to be ready.  must be in Go duration format (eg: 10s, 2m)")
+	cmd.Flags().String("http-proxy", "", "sets HTTP_PROXY environemnt variable in kotsadm containers")
```
marccampbell, line 243: (applies to all 4):

- change spelling of environment.
- replaced `kotsadm containers` with "all KOTS Admin Console components"

cmd/kots/cli/pull.go
```diff
@@ -86,6 +104,10 @@ func PullCmd() *cobra.Command {
 	cmd.Flags().Bool("exclude-kots-kinds", true, "set to true to exclude rendering kots custom objects to the base directory")
 	cmd.Flags().Bool("exclude-admin-console", false, "set to true to exclude the admin console (replicated apps only)")
 	cmd.Flags().String("shared-password", "", "shared password to use when deploying the admin console")
+	cmd.Flags().String("http-proxy", "", "sets HTTP_PROXY environemnt variable in kotsadm containers")
```
marccampbell, line 107: Same comment as above on help text

pkg/kotsadm/proxy.go
```diff
@@ -0,0 +1,34 @@
...
+			Name:  "HTTP_PROXY",
+			Value: deployOptions.HTTPProxyEnvValue,
+		},
+		{
+			Name:  "HTTPS_PROXY",
+			Value: deployOptions.HTTPSProxyEnvValue,
+		},
+	}
+
+	kotsadmNoProxy := "kotsadm-postgres,kotsadm-minio,kotsadm-api-node"
```
marccampbell, line 20: Do we need this on the kotsadm service also?

divolgin: @marccampbell only operator connects to that and it has no proxy defined, so it's always a direct connection.

PR #886, opened 2020-07-30
Degraded status during updates

emosbaugh:
This will now report status degraded when replicas are not yet updated (when a deployment is rolling replicas or a rollout fails).

### DIFF
--- a/kotsadm/operator/pkg/appstate/deployments.go
+++ b/kotsadm/operator/pkg/appstate/deployments.go
@@ -120,8 +120,11 @@ func calculateDeploymentState(r *appsv1.Deployment) types.State {
 	if desiredReplicas == 0 {
 		// TODO: what to do here?
 	}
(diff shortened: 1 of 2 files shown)


emosbaugh:
We need another state here, e.g. "upgrading" rather than "degraded"

AmberAlston:
@emosbaugh should this PR still be open?

emosbaugh:
I'm not sure. It's been a while since i submitted that PR and I'm pretty far removed from the day-to-day of the kots project. Is it still an open issue?

AmberAlston:
Showing "degraded" status during updates isn't the right approach. We'd need to introduce a new app status that is more appropriate to what is actually happening (vs unavailable, degraded, ready). This PR is very old at this point, and not the right approach without more work. We should close this and have a fresh PR when the problem statement is properly tackled.

PR #895, opened 2020-07-31
move getKotsDownstreamOutput query to go

salahalsaleh:


kotsadm/pkg/downstream/downstream.go
```diff
@@ -134,3 +136,84 @@ func SetIgnorePreflightPermissionErrors(appID string, sequence int64) error {
...
+	var status sql.NullString
+	var statusInfo sql.NullString
+	var dryrunStdout sql.NullString
+	var dryrunStderr sql.NullString
+	var applyStdout sql.NullString
+	var applyStderr sql.NullString
+
+	if err := row.Scan(&status, &statusInfo, &dryrunStdout, &dryrunStderr, &applyStdout, &applyStderr); err != nil {
+		if err == sql.ErrNoRows {
+			return &types.DownstreamOutput{
```
divolgin, line 170: `return &types.DownstreamOutput{}` is enough here.

salahalsaleh: ah, you're right, trying to mimic the logic in TS is sometimes a trap :)

PR #905, opened 2020-08-04
Handle showing errors for bad kots release blocking diff creation for newer releases

GraysonNull:


kotsadm/web/src/components/apps/AppVersionHistory.jsx
```diff
@@ -1194,6 +1215,27 @@ class AppVersionHistory extends Component {
           </div>
         </Modal>
 
+        {this.state.showDiffErrModal &&
```
salahalsaleh, line 1218: why not have this condition in `isOpen` instead? that way, the modal does not have to re-render.

PR #909, opened 2020-08-05
fetch versions downloaded by automatic updates

salahalsaleh:


marccampbell:
@salahalsaleh we should only refetch the full version list if there were changes. This query is a expensive and I think we should try to use the cache more.

PR #913, opened 2020-08-06
Generate new certs to be used for joining nodes

divolgin:


kotsadm/pkg/k8s/cluster.go
```diff
@@ -42,22 +44,26 @@ func GenerateBootstrapToken(client kubernetes.Interface, ttl time.Duration) (str
...
+
+	config, err := kubeadmconfig.DefaultedInitConfiguration(&kubeadmapiv1beta2.InitConfiguration{}, &kubeadmapiv1beta2.ClusterConfiguration{})
+	if err != nil {
+		return "", errors.Wrap(err, "failed to get kotsadm config")
 	}
 
 	key, err := copycerts.CreateCertificateKey()
 	if err != nil {
-		return "", err
+		return "", errors.Wrap(err, "failed to genertae key")
```
marccampbell, line 60: spelling

kotsadm/pkg/kurl/join_cert.go
```diff
@@ -0,0 +1,199 @@
...
+		status, err := client.CoreV1().Pods(pod.Namespace).Get(ctx, pod.Name, metav1.GetOptions{})
+		if err != nil {
+			return "", errors.Wrap(err, "failed to get pod")
+		}
+		if status.Status.Phase == corev1.PodRunning ||
+			status.Status.Phase == corev1.PodFailed ||
+			status.Status.Phase == corev1.PodSucceeded {
+			break
+		}
+		time.Sleep(time.Second * 1)
```
marccampbell, line 49: this could hang forever, wdyt about adding a timeout in here?

kotsadm/pkg/kurl/join_cert.go
```diff
@@ -0,0 +1,199 @@
...
+
+	return key, nil
+}
+
+func getCertGenLogs(ctx context.Context, client kubernetes.Interface, pod *corev1.Pod) ([]byte, error) {
+	// maxLines := int64(10000)
+	podLogOpts := corev1.PodLogOptions{
+		Follow:    false,
+		Container: pod.Spec.Containers[0].Name,
+		// TailLines:
```
marccampbell, line 70: want to delete?

kotsadm/pkg/kurl/join_cert.go
```diff
@@ -0,0 +1,199 @@
...
+	_, err = io.Copy(buf, podLogs)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to copy log")
+	}
+
+	return buf.Bytes(), nil
+}
+
+func parseCertGenOutput(logs []byte) (string, error) {
+	// Output looks like this:
```
marccampbell, line 91: yikes, there's no JSON output or less brittle?

kotsadm/pkg/kurl/join_cert.go
```diff
@@ -0,0 +1,199 @@
...
+						HostPath: &corev1.HostPathVolumeSource{
+							Path: "/usr/bin/kubeadm",
+							Type: &binVolumeType,
+						},
+					},
+				},
+			},
+			Containers: []corev1.Container{
+				{
+					Image:           existingDeployment.Spec.Template.Spec.Containers[apiContainerIndex].Image,
```
marccampbell, line 178: :thinking: the effort to get the container image is appreciated but also seems hard and potentially fragile. 

I am assuming this is needed for airgap installs? Can't we just look for the kurl config map / image pull secret with a label or some other methods to look it up and construct it? Maybe not, this isn't available in the downward api either.

No objections here, but it feels tricky and brittle.

kotsadm/pkg/kurl/configmap.go
```diff
@@ -69,7 +69,7 @@ func UpdateConfigMap(client kubernetes.Interface, generateBootstrapToken, upload
 	if uploadCerts {
 		certsDuration := time.Hour * 2
 		certsExpiration := time.Now().Add(certsDuration)
-		key, err := k8s.UploadCertsWithNewKey(client)
+		key, err := createCertAndKey(context.TODO(), client, "default")
```
emosbaugh, line 72: should you use `os.Getenv("POD_NAMESPACE")` here?

kotsadm/pkg/kurl/join_cert.go
emosbaugh: better yet use the context with timeout

```go
if err := ctx.Err(); err != nil {
  return err
}
```

kotsadm/pkg/kurl/join_cert.go
emosbaugh: it seems like the other option would be to write this in go and import kotsadm lib?

kotsadm/pkg/kurl/join_cert.go
```diff
@@ -0,0 +1,199 @@
...
+			Kind:       "Pod",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: existingDeployment.Namespace,
+			Labels:    existingDeployment.Labels,
+		},
+		Spec: corev1.PodSpec{
+			SecurityContext: &securityContext,
+			NodeSelector: map[string]string{
```
emosbaugh, line 151: i think you should add toleration here as well for the future

```
      tolerations:
      - key: node-role.kubernetes.io/master
        effect: NoSchedule
```

kotsadm/pkg/kurl/join_cert.go
```diff
@@ -0,0 +1,199 @@
...
+					},
+				},
+			},
+			Containers: []corev1.Container{
+				{
+					Image:           existingDeployment.Spec.Template.Spec.Containers[apiContainerIndex].Image,
+					ImagePullPolicy: corev1.PullNever,
+					Name:            "join-cert-gen",
+					Command:         []string{"/usr/bin/kubeadm"},
+					Args:            []string{"init", "phase", "upload-certs", "--upload-certs"},
```
emosbaugh, line 182: i guess `--config` not required here?

kotsadm/pkg/kurl/join_cert.go
divolgin: @marccampbell @emosbaugh I kept this in its own commit intentionally.  We should figure out how to make this work, maybe in 1.19: https://github.com/replicatedhq/kots/commit/6ba2ca6a#diff-d813b91502ac37baa1004104af667747R16

kotsadm/pkg/kurl/join_cert.go
divolgin: nope.  just works.  there must be some defaults that we need to copy to make the Go version of this work.

kotsadm/pkg/kurl/configmap.go
```diff
@@ -69,10 +70,15 @@ func UpdateConfigMap(client kubernetes.Interface, generateBootstrapToken, upload
 	if uploadCerts {
 		certsDuration := time.Hour * 2
 		certsExpiration := time.Now().Add(certsDuration)
-		key, err := createCertAndKey(context.TODO(), client, "default")
+
+		ctx, cancel := context.WithTimeout(context.TODO(), 30*time.Second)
```
emosbaugh, line 74: I wonder if you want to give it a little more time?

PR #923, opened 2020-08-08
Convert secrets to sealed secrets

marccampbell:
This adds an experimental feature to integrated the bitnami sealedsecrets controller into kots.

With this change, if a secret is deployed to the same namespace as kots, and the secret looks like

```
apiVersion: v1
data:
  cert.pem: <base64 encoded certificate?
kind: Secret
metadata:
  labels:
    kots.io/buildphase: secret
    kots.io/secrettype: sealedsecrets

  name: kots-sealed-secret
  namespace: default
type: Opaque

```

All secrets in the base will be replaced with encrypted sealed secrets.

### DIFF
--- a/kotsadm/pkg/secrets/sealed_secrets.go
+++ b/kotsadm/pkg/secrets/sealed_secrets.go
@@ -48,7 +48,10 @@ func replaceSecretsWithSealedSecrets(archivePath string, config map[string][]byt
 			return nil
 		}
 
-		secret := decoded.(*v1.Secret)
+		secret, ok := decoded.(*v1.Secret)
+		if !ok {
+			return nil
+		}
 
 		// sealed secrets require a namespace
 		if secret.Namespace == "" {
--- a/kotsadm/pkg/secrets/sealed_secrets.go
+++ b/kotsadm/pkg/secrets/sealed_secrets.go
@@ -1,0 +1,80 @@
+package secrets
+
+import (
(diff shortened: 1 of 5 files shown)


kotsadm/pkg/secrets/sealed_secrets.go
```diff
@@ -0,0 +1,80 @@
...
+		if err != nil {
+			return errors.Wrap(err, "failed to read file")
+		}
+
+		decoded, _, err := decode(contents, nil, nil)
+		if err != nil {
+			return nil
+		}
+
+		secret := decoded.(*v1.Secret)
```
divolgin, line 51: i'm gonna ok this lack of error checking :)

kotsadm/pkg/secrets/secrets.go
```diff
@@ -0,0 +1,93 @@
...
+		}
+
+		contents, err := ioutil.ReadFile(path)
+		if err != nil {
+			return err
+		}
+
+		_, gvk, err := decode(contents, nil, nil)
+		if err != nil {
+			return nil
```
emosbaugh, line 77: a comment like this may be good here

https://github.com/replicatedhq/kots/blob/86d1ab88b92a46841cee25b83ea128fc033f39d4/kotsadm/pkg/kotsutil/kots.go#L303
```
return nil // not an error because the file might not be yaml
```

PR #924, opened 2020-08-09
Refactor PG and S3 code to be swappable with other implementations

marccampbell:
This change moves all of the S3 and Postgresql implementation into a new `S3PGStore`, that implements a new `KOTStore` interface. This interface wraps all persistence in the Admin Console, and nothing else will write to S3 and Postgres, or read from S3 and Postgres.

The goal of this is to build out the `OCIStore` that exists as an alternate interface to the `S3PGStore`. KOTS will be transitioning to a more flexible backend storage option, and the implementation details of where to store app archives, support bundles, and everything else will be handled by the store.

This store is rather large at this point. The goals with this PR was to refactor the store to the interface and stub the not implemented OCIStore. But there's more work that can be designed to simplify this store in the future.


salahalsaleh:
@marccampbell yes, I highly expect that the `listApps` query will be moved to go in 1.19.

marccampbell:
hold for 1.19

kotsadm/api/src/healthz/healthz_store.ts
```diff
@@ -21,7 +21,14 @@ export class HealthzStore {
     const query = `select count(1)`;
     await this.pool.query(query);
 
-    const storageReady = await bucketExists(this.params, this.params.shipOutputBucket);
+    let storageReady = false;
+
+    if (this.params.storageBaseURI.startsWith("docker://")) {
+      storageReady = true;  // TODO
```
divolgin, line 27: console.log something here?  this is bound to be forgotten.

kotsadm/kustomize/overlays/dev/kustomization.yaml
```diff
@@ -1,6 +1,6 @@
 bases:
   - ../../base
-  - ./minio
+  # - ./minio
```
divolgin, line 3: i don't see where registry is added to replace minio in dev

marccampbell: 8 lines down

kotsadm/pkg/store/ocistore/supportbundle_store.go
```diff
@@ -1,44 +1,54 @@
...
-	kotss3 "github.com/replicatedhq/kots/kotsadm/pkg/s3"
+	supportbundletypes "github.com/replicatedhq/kots/kotsadm/pkg/supportbundle/types"
+	troubleshootredact "github.com/replicatedhq/troubleshoot/pkg/redact"
 	"go.uber.org/zap"
 )
 
-func refFromBundleID(bundleID string, baseURI string) string {
-	baseURI = strings.TrimSuffix(baseURI, "/")
+func (s OCIStore) ListSupportBundles(appID string) ([]*supportbundletypes.SupportBundle, error) {
+	return nil, errors.New("not implemented")
```
emosbaugh, line 24: im curious to know how this will be implemented

PR #928, opened 2020-08-10
move listApps query to go

salahalsaleh:


kotsadm/web/src/Root.jsx
```diff
@@ -140,20 +138,29 @@ class Root extends Component {
...
+      }
+      const response = await res.json();
+      const apps = response.apps;
+      this.setState({
+        appsList: apps,
+        rootDidInitialWatchFetch: true
+      });
+      return apps;
+    } catch(err) {
       throw error;
```
GraysonNull, line 162: Won't `error` be undefined here?

kotsadm/web/src/components/gitops/GitOpsDeploymentManager.jsx
```diff
@@ -83,9 +84,33 @@ class GitOpsDeploymentManager extends React.Component {
...
+      }
+      const response = await res.json();
+      const apps = response.apps;
+      this.setState({
+        appsList: apps,
+      });
+      return apps;
+    } catch(err) {
+      console.log(err);
+      throw error;
```
GraysonNull, line 108: Same here, I think `error` will be undefined here

PR #929, opened 2020-08-10
Error handling updates

jgruica:
Screen Shot 2020-08-10 at 10 57 50 AM
Screen Shot 2020-08-10 at 1 12 05 PM
Screen Shot 2020-08-10 at 1 47 08 PM

### DIFF
--- a/kotsadm/web/src/components/apps/AppDetailPage.jsx
+++ b/kotsadm/web/src/components/apps/AppDetailPage.jsx
@@ -140,14 +140,14 @@ class AppDetailPage extends Component {
   }
 
   getApp = async (slug = this.props.match.params.slug) => {
-    if (!this.props.match.params.slug) {
+    if (!slug) {
       return;
     }
 
     try {
       this.setState({ loadingApp: true, gettingAppErrMsg: "", displayErrorModal: false });
 
-      const res = await fetch(`${window.env.API_ENDPOINT}/apps/app/${this.props.match.params.slug}`, {
+      const res = await fetch(`${window.env.API_ENDPOINT}/apps/app/${slug}`, {
         headers: {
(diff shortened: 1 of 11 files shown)


kotsadm/web/src/components/apps/AppDetailPage.jsx
```diff
@@ -129,14 +140,14 @@ class AppDetailPage extends Component {
   }
 
   getApp = async (slug = this.props.match.params.slug) => {
```
GraysonNull, line 142: This seems odd, especially since we're just checking for `this.props.match.params.slug` everywhere else in this function? does a slog get passed to this at all?

jgruica: it is used: https://github.com/replicatedhq/kots/blob/90426f19b1405d3cf6dec4430d6c96b9e2aae6b5/kotsadm/web/src/components/apps/AppDetailPage.jsx#L196

which is used in did mount and did update:
https://github.com/replicatedhq/kots/blob/90426f19b1405d3cf6dec4430d6c96b9e2aae6b5/kotsadm/web/src/components/apps/AppDetailPage.jsx#L205

GraysonNull: So I still am not really clear why were setting it to this.props.match.slug as a default value. Do we call it somewhere without passing a slug so we just set it to what're that param is?

jgruica: correct, there are few places where we're calling it without slug.

GraysonNull: I see. Ok sounds good

PR #938, opened 2020-08-11
Helm yaml errors

emosbaugh:


austinchambers:
# To Repro: 
1. Ingest the Redis HA official helm chart into a new KOTS vendor app
2. Attempt to deploy the application

**Expected:** Deploy succeeds. Helm charts with multi-resources and one of them is empty deploys without problem. 
**Actual:** Deploy fails. Admin Console incorrectly reports the chart as invalid YAML. Upon inspection, it appears to be because there are multi-documents, and https://github.com/helm/charts/blob/master/stable/redis-ha/templates/redis-ha-announce-service.yaml#L6 one of them has only whitespace between `---` lines.

PR #939, opened 2020-08-12
move snapshotConfig query to Go

GraysonNull:


kotsadm/pkg/handlers/snapshots.go
```diff
@@ -406,3 +417,104 @@ func GetGlobalSnapshotSettings(w http.ResponseWriter, r *http.Request) {
...
+
+	JSON(w, 200, getSnapshotConfigResponse)
+}
+
+func parseTTL(s string) (*snapshottypes.ParsedTTL, error) {
+	parsedTTLResponse := &snapshottypes.ParsedTTL{}
+
+	ttlMatch, err := regexp.Compile(`^\d+(s|m|h)$`)
+	if err != nil {
+		fmt.Printf("Invalid snapshot TTl %v", s)
```
salahalsaleh, line 476: missing return

salahalsaleh: i would replace with `Println` or add a `\n` at the end

kotsadm/pkg/handlers/snapshots.go
```diff
@@ -406,3 +417,104 @@ func GetGlobalSnapshotSettings(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	unit := ttlMatch.FindStringSubmatch(s)[1]
```
salahalsaleh, line 479: length check

kotsadm/pkg/handlers/snapshots.go
```diff
@@ -406,3 +417,104 @@ func GetGlobalSnapshotSettings(w http.ResponseWriter, r *http.Request) {
...
+	quantity := strings.Split(ttlMatch.FindStringSubmatch(s)[0], unit)
+	quantityInt, err := strconv.ParseInt(quantity[0], 10, 64)
+	if err != nil {
+		fmt.Printf("failed to parseInt quanitity")
```
salahalsaleh, line 483: missing return

kotsadm/pkg/handlers/snapshots.go
```diff
@@ -406,3 +417,104 @@ func GetGlobalSnapshotSettings(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	switch unit {
```
salahalsaleh, line 486: missing `break`s

kotsadm/pkg/handlers/snapshots.go
```diff
@@ -406,3 +417,104 @@ func GetGlobalSnapshotSettings(w http.ResponseWriter, r *http.Request) {
...
+			break
+		}
+		if quantityInt/24 >= 1 && quantityInt%24 == 0 {
+			parsedTTLResponse.Quantity = quantityInt / 24
+			parsedTTLResponse.Unit = "days"
+			break
+		}
+		parsedTTLResponse.Quantity = quantityInt
+		parsedTTLResponse.Unit = "hours"
+	default:
```
salahalsaleh, line 516: how about returning an error on `default` because it did not match any `unit`

kotsadm/web/src/components/apps/AppSnapshotSchedule.jsx
```diff
@@ -118,17 +121,43 @@ class AppSnapshotSchedule extends Component {
...
-    if (this.props.snapshotConfig.snapshotConfig && this.props.snapshotConfig.snapshotConfig !== lastProps.snapshotConfig.snapshotConfig) {
+  getSnapshotConfig = async () => {
+    this.setState({ loadingConfig: true, gettingConfigErrMsg: "", displayErrorModal: false });
+    await fetch(`${window.env.API_ENDPOINT}/app/${this.props.app.slug}/snapshot/config`, {
+      method: "GET",
+      headers: {
+        "Authorization": Utilities.getToken(),
+        "Content-Type": "application/json",
+      }
+    }).then(async (res) => {
```
salahalsaleh, line 132: i would recommend using the `async/await` in a `try/catch` syntax. it gives the choice of `awaiting` to the caller of the function. (this function is called without await, so in this case, you can also just get rid of `async/await` here, i would recommend the first approach though).

kotsadm/web/src/components/apps/AppSnapshotSchedule.jsx
```diff
@@ -159,19 +188,26 @@ class AppSnapshotSchedule extends Component {
...
+    const { hasValidCron, updatingSchedule, updateConfirm, loadingConfig } = this.state;
     const selectedRetentionUnit = RETENTION_UNITS.find((ru) => {
       return ru.value === this.state.selectedRetentionUnit?.value;
     });
     const selectedSchedule = SCHEDULES.find((schedule) => {
       return schedule.value === this.state.selectedSchedule?.value;
     });
+
+    if (loadingConfig) {
+      <div className="flex-column flex1 alignItems--center justifyContent--center">
```
salahalsaleh, line 200: missing `return`

kotsadm/web/src/components/apps/AppSnapshotSchedule.jsx
```diff
@@ -118,17 +121,43 @@ class AppSnapshotSchedule extends Component {
...
     }
   }
 
   componentDidMount = () => {
-    this.getReadableCronExpression();
-    if (this.props.snapshotConfig.snapshotConfig) {
+    if (this.state.snapshotConfig) {
       this.setFields();
+    } else {
+      this.getSnapshotConfig()
```
salahalsaleh, line 158: writing in go made you forget a few `;` 🙂

kotsadm/web/src/components/apps/AppSnapshotSchedule.jsx
```diff
@@ -118,17 +121,43 @@ class AppSnapshotSchedule extends Component {
...
+    await fetch(`${window.env.API_ENDPOINT}/app/${this.props.app.slug}/snapshot/config`, {
+      method: "GET",
+      headers: {
+        "Authorization": Utilities.getToken(),
+        "Content-Type": "application/json",
+      }
+    }).then(async (res) => {
+      if (!res.ok) {
+        this.setState({ loadingConfig: false, gettingConfigErrMsg: `Unable to get snapshot config: Unexpected status code: ${res.status}`, displayErrorModal: true });
+        return
```
salahalsaleh, line 135: missing `;`

kotsadm/web/src/components/apps/AppSnapshotSchedule.jsx
```diff
@@ -118,17 +121,43 @@ class AppSnapshotSchedule extends Component {
...
+      }
+      const body = await res.json();
+      this.setState({
+        snapshotConfig: body,
+        loadingConfig: false
+      })
```
salahalsaleh, line 141: missing `;`

kotsadm/web/src/components/apps/AppSnapshotSchedule.jsx
GraysonNull: haha ah yeah the switching back and forth definitely has me forgetting those!

kotsadm/pkg/handlers/snapshots.go
```diff
@@ -406,3 +416,110 @@ func GetGlobalSnapshotSettings(w http.ResponseWriter, r *http.Request) {
...
+		return nil, errors.Wrap(nil, "failed to parseInt quanitity")
```
salahalsaleh, line 522:
```suggestion
		return nil, errors.Wrap(nil, "unsupported unit type")
```

kotsadm/pkg/handlers/snapshots.go
```diff
@@ -406,3 +416,110 @@ func GetGlobalSnapshotSettings(w http.ResponseWriter, r *http.Request) {
...
+			parsedTTLResponse.Unit = "weeks"
+			break
+		}
+		if quantityInt/24 >= 1 && quantityInt%24 == 0 {
+			parsedTTLResponse.Quantity = quantityInt / 24
+			parsedTTLResponse.Unit = "days"
+			break
+		}
+		parsedTTLResponse.Quantity = quantityInt
+		parsedTTLResponse.Unit = "hours"
```
salahalsaleh, line 520: missing `break` here

PR #941, opened 2020-08-12
move testGitOpsConnection mutation to go

salahalsaleh:


kotsadm/pkg/handlers/gitops.go
```diff
@@ -105,6 +110,129 @@ func DisableAppGitOps(w http.ResponseWriter, r *http.Request) {
...
+		logger.Error(err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	go func(appID, clusterID, appSlug, appName, downstreamName string) {
+		currentVersion, err := downstream.GetCurrentVersion(appID, clusterID)
+		if err != nil {
+			logger.Error(err)
+			w.WriteHeader(http.StatusInternalServerError)
```
divolgin, line 183: you can't write back to the client in a go routine.   when this runs, the HTTP request has already been completed

salahalsaleh: oh, good catch, i had this outside of the go routine, then i wrapped it and forgot about this

kotsadm/pkg/handlers/gitops.go
```diff
@@ -105,6 +110,129 @@ func DisableAppGitOps(w http.ResponseWriter, r *http.Request) {
...
+			defer os.RemoveAll(pendingVersionArchive)
+
+			_, err = gitops.CreateGitOpsCommit(downstreamGitOps, appSlug, appName, int(pendingVersion.ParentSequence), pendingVersionArchive, downstreamName)
+			if err != nil {
+				logger.Error(err)
+				w.WriteHeader(http.StatusInternalServerError)
+				return
+			}
+		}
+	}(a.ID, d.ClusterID, a.Slug, a.Name, d.Name)
```
divolgin, line 231: it's legal to use `a` inside the go routine.  you don't need to pass these in as arguments.  your call on this one.

kotsadm/pkg/handlers/gitops.go
```diff
@@ -105,6 +110,129 @@ func DisableAppGitOps(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	if err := gitops.SetGitOpsError(a.ID, d.ClusterID, ""); err != nil {
+		logger.Error(err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	go func(appID, clusterID, appSlug, appName, downstreamName string) {
+		currentVersion, err := downstream.GetCurrentVersion(appID, clusterID)
```
divolgin, line 180: This should work like checking for updates works or how setting registry credentials works.  even if we can't show a status update, a second request to this handler should fail if the first one is still running

PR #944, opened 2020-08-13
Helm hooks support

emosbaugh:


pkg/base/base.go
```diff
@@ -102,129 +77,142 @@ func (e ParseError) Error() string {
...
 	}
 	// In addition to kotskinds, we exclude the application crd for now
-	if o.APIVersion == "app.k8s.io/v1beta1" {
+	if gvk.Group == "app.k8s.io" {
 		return true
 	}
 	return false
 }
 
+func isExcludedByAnnotation(annotations map[string]string) (bool, error) {
```
emosbaugh, line 174: this may break backwards compatibility for apps with annotations with bool values. i should probably test and fix this

emosbaugh:
Closing this to be revisited in the future.

PR #951, opened 2020-08-14
Read-only registry proposal

divolgin:


design/read-only-registry.md
```diff
@@ -0,0 +1,76 @@
...
+Online install will download updates, but will not push images to the private registry as long as the read-only flag is set.
+
+Aigapped installs will accept airgap bundle, but will not push images to the private registry as long as the read-only flag is set.
+
+Stretch: Replicated could provide imageless airgap bundles.
+
+### Registry Settings page
+
+1. Registry settings page will have a checkbox named Read only.
+1. In aigapped environments this is always checked and disabled and all other fields can be modified.
```
laverya, line 53: Why disable this in airgap? Do you mean inn kurl? If this is kurl, wouldn't it always be unchecked?

divolgin: We can't push images in airgapped mode with current UI design. These settings are disabled right now.

laverya: I'm confused. Doesn't uploading an airgap package currently result in kotsadm pushing images to the configured registry?

divolgin: @laverya this is about the Registry Settings page

laverya: @divolgin is the intent to disallow changing from RO to RW (or RW to RO) on airgap, but still allow either to be set depending on what was chosen during initial setup?

divolgin: yeah...   this is actually wrong.  it should be "always disabled" but not "always checked"

design/read-only-registry.md
```diff
@@ -0,0 +1,79 @@
...
+
+Aigapped installs will accept airgap bundle, but will not push images to the private registry as long as the read-only flag is set.
+
+Stretch: Replicated could provide imageless airgap bundles.
+
+### Registry Settings page
+
+1. Registry settings page will have a checkbox named Read only.
+1. In aigapped environments this is always disabled
+   - if read-only set, all fields can be modiefied
```
laverya, line 55:
```suggestion
   - if read-only set, all fields can be modified
```

design/read-only-registry.md
```diff
@@ -0,0 +1,79 @@
...
+Online install will download updates, but will not push images to the private registry as long as the read-only flag is set.
+
+Aigapped installs will accept airgap bundle, but will not push images to the private registry as long as the read-only flag is set.
+
+Stretch: Replicated could provide imageless airgap bundles.
+
+### Registry Settings page
+
+1. Registry settings page will have a checkbox named Read only.
+1. In aigapped environments this is always disabled
```
laverya, line 54:
```suggestion
1. In aigapped environments this checkbox is always disabled (you can't change from readonly to read-write on airgap, or the other way around)
```

design/read-only-registry.md
```diff
@@ -0,0 +1,79 @@
...
+
+### Install
+
+A new argument will be added to `kots install` command to flag the registry as a read-only registry.
+
+```
+kubectl kots install --kotsadm-registry my.private.registry/myapp \
+  --registry-username pulluser \
+  --registry-password pullpassword \
+  --registry-is-readonly \
```
markpundsack, line 30: The semantics feel a bit off to me. This isn't **just** about having the registry be read-only, but also then skipping pushing any images to the registry, and requiring some out-of-band way to side-load images into the registry. I don't have a suggestion for improvement, but I can imagine some folks getting confused by this.

design/read-only-registry.md
```diff
@@ -0,0 +1,79 @@
...
+```
+
+When the above command is used, the following will happen:
+
+1. All kotsadm pods will be rewritten to pull images from the private registry. This is existing functionality.
+1. Provided registry credentials will be saved in `.dockerconfigjson` secret in the app's namespace.
+1. During license upload in Admin Console UI, registry settings will be populated with these values.
+1. After the application is installed, Registry Settings page will be populated with these settings.
+
+Note: this changes current functionality where registry information specified in the command line arguments is not saved and is not applied to the application pods.
```
markpundsack, line 41: Is anything else necessary to handle more edge cases, like what if the images haven't been pushed yet, or worse, some images have been pushed but not all of them? Usually when updating, we have reasonable assurance that the image pulls will work, so I wouldn't be surprised if our robustness will take a hit under this new flow. Do we need to do a preflight check to see if all the images actually exist before trying to use them?

markpundsack: Going further, maybe we even need visual indication of whether all images exist or not. Or maybe we need a UI to list the images for someone to fetch out-of-band. Hopefully we don't need to go that far, but otherwise, how do people know what images to fetch?

design/read-only-registry.md
```diff
@@ -0,0 +1,174 @@
+# Support for Read-Only Registries
+
+## Goals
+
+- Configure kotsadm to use application images from a private registry
```
markpundsack, line 5:
```suggestion
- Configure kotsadm to use application images from a private registry which kotsadm only has read-only access to (images must be installed or copied to the private registry outside of KOTS)
```

design/read-only-registry.md
```diff
@@ -0,0 +1,174 @@
...
+
+## Non Goals
+
+- Delivering images to the read-only registry.
+- Creating imageless airgap bundles.
+- Addressing existing inconsistencies and bugs in multi-app support.
+
+## Background
+
+Some production enviroments restrict image registry use to read-only mode.  This is an environment wide restriction.  Images are side-loaded into the registry from a trusted internal source.  Right now kotsadm will always push images to the private registry, making it impossible to use in such an environment.
```
markpundsack, line 19:
```suggestion
Some production environments restrict image registry use to read-only mode.  This is an environment wide restriction.  Images are side-loaded into the registry from a trusted internal source.  Right now kotsadm will always push images to the private registry, making it impossible to use in such an environment.
```

design/read-only-registry.md
```diff
@@ -0,0 +1,174 @@
...
+- Support using read-only registry during initial installation
+- Support using read-only registry on application updates
+- Support using read-only registry when settings are modified on the Registry Settings page in Admin Console
+- Support read-only registries in airgap and on-line modes
+
+## Non Goals
+
+- Delivering images to the read-only registry.
+- Creating imageless airgap bundles.
+- Addressing existing inconsistencies and bugs in multi-app support.
```
markpundsack, line 15: This is a bit of a strange one. I can't guess why this is included. Can you elaborate?

divolgin: There are some issues with registry settings in multi-app setups.  They are tied to application but also used globally for Admin Console itself.

design/read-only-registry.md
```diff
@@ -0,0 +1,174 @@
...
+  --registry-username <username> \
+  --registry-password <password> \
+  --disable-image-push
+```
+
+When the above command is used, the following will happen:
+
+1. All kotsadm pods will be rewritten to pull images from the private registry. This is existing functionality.
+1. Provided registry credentials will be saved in `kotsadm-private-registry` secret in the app's namespace. This is existing functionality.
+1. The `kots admin-console push-images` command can be ommitted.
```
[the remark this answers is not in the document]
divolgin: If the push-images command is used, and it actually works, it won't have any side-effects.  In a hardened environment (with truly read-only access), this command should fail, which will also have no other side-effects.

design/read-only-registry.md
```diff
@@ -0,0 +1,174 @@
...
+1. During license upload in Admin Console UI, registry settings will be populated with these values.
+1. After the application is installed, Registry Settings page will be populated with these values.
+
+During application installation:
+- If airgap option is selected, images will **not** be pushed to the private registry specified in the UI.
+- If online option is selected, updating Registry Settings page will **not** result in images being pushed.
```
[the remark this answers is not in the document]
divolgin: During the install, registry settings are only used with airgap bundle.

design/read-only-registry.md
```diff
@@ -0,0 +1,174 @@
...
+
+#### Fully airgapped mode
+
+In this case, both Admin Console and application are installed in airgapped mode:
```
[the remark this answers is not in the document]
divolgin: No, just wanted to cover this case explicitly.  It's a separate path in the code.  This and fully unattended would be the same though. The command in the example below is the minimal set of flags that would be used, and more can be added for fully unattended.

design/read-only-registry.md
```diff
@@ -0,0 +1,174 @@
...
+
+When the above command is used, CLI will skip pushing images and proceed with the rest of the instalaltion.
+User will be redirect to the Config screen.
+Other command line arguments can be used for headless and/or unattended install experience without any additional changes.
+
+### Updates
+
+#### Online Admin Console
+
+Upgrading Admin Console installed in online mode does not change.
```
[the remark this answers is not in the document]
divolgin: The basic `kots install --kotsadm-registry...` command does not push any images.  I guess the command changes form `kots install <app>` to `kots install --kotsadm-registry...`.  The `--disable-image-push` would be used during installation and isn't needed here.

design/read-only-registry.md
```diff
@@ -0,0 +1,174 @@
...
+
+#### Airgapped Admin Console
+
+Upgrading Admin Console installed in airgapped mode has to be done with the `--disable-image-push` flag specified on the command line:
```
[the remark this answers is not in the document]
divolgin: Right now registry credentials can be omitted in kURL installs only.  We'd need to move registry settings out of postgres since this command does not interact with the API.  I think this should be out of scope.  The flag could potentially be in a config map, which would not require the API.

design/read-only-registry.md
```diff
@@ -0,0 +1,174 @@
...
+Upgrading airgapped application using Admin Console CLI does not change from user's perspective.
+Admin Console will still create a new application version and rewrite all images to be pulled from the private registry.
+
+```
+kubectl kots upstream upgrade app-slug \
+  --airgap-bundle /path/to/app.airgap \
+  --kotsadm-namespace myapp \
+  --kotsadm-registry my.private.registry \
+  --registry-username <username> \
+  --registry-password <password> \
```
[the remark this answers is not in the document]
divolgin: Yes

design/read-only-registry.md
```diff
@@ -0,0 +1,174 @@
...
+Internally the functionality changes as follows:
+- Images will not be pushed to private registry.
+
+### Registry Settings page
+
+Registry settings page will have a checkbox named `Disable Image Push`.
+
+#### Online application installs
+
+In online installs this checkbox is can be toggled freely.
```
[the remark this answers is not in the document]
divolgin: If only the checkbox changes, technically we don't need to create a new version and deploy it.  So we could allow simply toggling it without any side effects.

design/read-only-registry.md
```diff
@@ -0,0 +1,174 @@
...
+  --registry-password <password> \
+  --disable-image-push
+  ```
+
+Internally the functionality changes as follows:
+- Images will not be pushed to private registry.
+
+### Registry Settings page
+
+Registry settings page will have a checkbox named `Disable Image Push`.
```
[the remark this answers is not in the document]
divolgin: @markpundsack there is not, and now that you mention it, maybe this should be a text label in this iteration.

design/read-only-registry.md
```diff
@@ -0,0 +1,174 @@
...
+#### Online application installs
+
+In online installs this checkbox is can be toggled freely.
+
+- When the checkbox is unchecked, saving registry works without any changes.
+- When checkbox is checked, saving registry settings will not push images.
+
+#### Airgapped installs
+
+In airgapped environments, changing registry host and namespace is not supported today.
```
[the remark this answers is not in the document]
divolgin: @markpundsack host and namespace are disabled.

design/read-only-registry.md
```diff
@@ -0,0 +1,174 @@
...
+Therefore this checkbox will reflect the value of the `--disable-image-push` flag used to install latest app version.
+
+### Preflight checks
+
+A new preflight check will be added to validate that images are present in their respective registries.
```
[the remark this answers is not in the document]
divolgin: @markpundsack I that we will can create a special preflight check.  Vendor would have to choose to add it.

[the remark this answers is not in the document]
markpundsack: Hmm... since the vendor isn't really part of the decision to use read-only registry or not, it doesn't make sense to rely on the vendor choosing to add the preflight check. Seems to me like we should inject that when necessary based on end-user enterprise choices.

design/read-only-registry.md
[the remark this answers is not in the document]
markpundsack: Makes sense, but let's make sure the UX works. It's an anti-pattern to have one checkbox in a form be immediate, whereas the others need to be submitted. This goes even further in that some changes need to be deployed after hitting submit and others won't, and a user won't be able to tell which ahead of time.

In other worse, this should probably be a separate form.

design/read-only-registry.md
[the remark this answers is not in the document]
markpundsack: Not sure I understand. Sure, the command you listed doesn't push any images, but that's not the command I'd normally run in online mode so at least the change would be to run that command instead. But also, I'd expect not to need to repeat the `kotsadm-registry` and other registry settings.

PR #952, opened 2020-08-15
pkg/docker/registry: replaced deprecated AWS session.New

alrs:
From the docs:
> Deprecated: Use NewSession functions to create sessions instead. NewSession has the same functionality as New except an error can be returned when the func is called instead of waiting to receive an error until a request is made.

This uses `NewSession`, and changes the signature of `getECRService` to return the `err`.

### DIFF
--- a/pkg/docker/registry/ecr.go
+++ b/pkg/docker/registry/ecr.go
@@ -45,7 +45,10 @@ func GetECRBasicAuthToken(ecrEndpoint, username, password string) (string, error
 		return "", errors.Wrap(err, "failed to parse ECR endpoint")
 	}
 
-	ecrService := getECRService(username, password, zone)
+	ecrService, err := getECRService(username, password, zone)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to get ECR service")
+	}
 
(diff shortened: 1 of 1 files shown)


pkg/docker/registry/ecr.go
```diff
@@ -63,10 +66,11 @@ func GetECRBasicAuthToken(ecrEndpoint, username, password string) (string, error
 	return *ecrToken.AuthorizationData[0].AuthorizationToken, nil
 }
 
-func getECRService(accessKeyID, secretAccessKey, zone string) *ecr.ECR {
+func getECRService(accessKeyID, secretAccessKey, zone string) (*ecr.ECR, error) {
 	awsConfig := &aws.Config{Region: aws.String(zone)}
 	awsConfig.Credentials = credentials.NewStaticCredentials(accessKeyID, secretAccessKey, "")
-	return ecr.New(session.New(awsConfig))
+	sess, err := session.NewSession(awsConfig)
```
emosbaugh, line 72: Handle the error here

emosbaugh (changes requested):
This looks good. One comment and can you please sign the CLA? Thanks!

PR #957, opened 2020-08-17
Move deploy socket to go

salahalsaleh:


kotsadm/pkg/app/app.go
```diff
@@ -150,6 +151,31 @@ func ListInstalled() ([]*App, error) {
...
+
+	apps := []*App{}
+	for rows.Next() {
+		var appID string
+		if err := rows.Scan(&appID); err != nil {
+			return nil, errors.Wrap(err, "failed to scan")
+		}
+		app, err := Get(appID)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to get app")
```
emosbaugh, line 171: wanna add the app id to the message?

kotsadm/pkg/socketservice/socketservice.go
```diff
@@ -0,0 +1,543 @@
...
+		socketMtx.Lock()
+		defer socketMtx.Unlock()
+
+		d, err := downstream.GetFromDeployToken(c.RequestURL().Query().Get("token"))
+		if err != nil {
+			logger.Error(errors.Wrap(err, "failed to get cluster id from deploy token"))
+			return
+		}
+
+		logger.Info(fmt.Sprintf("cluster %s connected", d.ClusterID))
```
emosbaugh, line 81: maybe capitalize and add a little more info in the message as its at info level. connected to what?

kotsadm/pkg/socketservice/socketservice.go
```diff
@@ -0,0 +1,543 @@
...
+	Informers []string `json:"informers"`
+}
+
+type SupportBundleArgs struct {
+	URI string `json:"uri"`
+}
+
+var server *socket.Server
+var clusterSocketHistory = []ClusterSocket{}
+var socketMtx sync.Mutex
```
emosbaugh, line 66: shouldnt this be a class with member vars?

kotsadm/pkg/socketservice/socketservice.go
```diff
@@ -0,0 +1,543 @@
...
+
+func deployLoop() {
+	for _, clusterSocket := range clusterSocketHistory {
+		apps, err := app.ListInstalledForDownstream(clusterSocket.ClusterID)
+		if err != nil {
+			logger.Error(errors.Wrap(err, "failed to list installed apps for downstream"))
+			continue
+		}
+
+		for _, a := range apps {
```
emosbaugh, line 130: could you make everything inside here a function and then you could return rather than continue and get rid of a lot of logger.Error statements?

kotsadm/pkg/socketservice/socketservice.go
```diff
@@ -0,0 +1,543 @@
...
+			}
+
+			d, err := downstream.Get(clusterSocket.ClusterID)
+			if err != nil {
+				logger.Error(errors.Wrap(err, "failed to get downstream"))
+				continue
+			}
+
+			var deployError error
+			defer func() {
```
emosbaugh, line 157: this defer function will not get called until the deploy loop ends. does it ever? also perhaps this warrants deploy() being its own function?

kotsadm/pkg/socketservice/socketservice.go
```diff
@@ -0,0 +1,543 @@
...
+		}
+
+		for _, a := range apps {
+			if a.RestoreInProgressName != "" {
+				continue
+			}
+
+			deployedVersion, err := downstream.GetCurrentVersion(a.ID, clusterSocket.ClusterID)
+			if err != nil {
+				logger.Error(errors.Wrap(err, "failed to get current downstream version"))
```
emosbaugh, line 137: if you wrap this in a function it would be nice to add appid and cluster id to the message outside the function call

kotsadm/pkg/socketservice/socketservice.go
```diff
@@ -0,0 +1,543 @@
...
+			// get previous manifests (if any)
+			base64EncodedPreviousManifests := ""
+			previouslyDeployedSequence, err := downstream.GetPreviouslyDeployedSequence(a.ID, clusterSocket.ClusterID)
+			if err != nil {
+				err = errors.Wrap(err, "failed to get previously deployed sequence")
+				logger.Error(err)
+				deployError = err
+				continue
+			}
+			if previouslyDeployedSequence != -1 {
```
emosbaugh, line 224: man functions with returns would be nice to avoid all this nesting

kotsadm/pkg/socketservice/socketservice.go
```diff
@@ -0,0 +1,543 @@
...
+
+			c, err := server.GetChannel(clusterSocket.SocketID)
+			if err != nil {
+				logger.Error(errors.Wrap(err, "failed to get socket channel from server"))
+				continue
+			}
+			c.Emit("deploy", deployArgs)
+			clusterSocket.LastDeployedSequences[a.ID] = deployedVersion.ParentSequence
+
+			// deploy status informers
```
emosbaugh, line 287: this could be its own function

kotsadm/pkg/socketservice/socketservice.go
```diff
@@ -0,0 +1,543 @@
...
+func supportBundleLoop() {
+	for _, clusterSocket := range clusterSocketHistory {
+		pendingSupportBundles, err := supportbundle.ListPendingForCluster(clusterSocket.ClusterID)
+		if err != nil {
+			logger.Error(errors.Wrap(err, "failed to list pending support bundles"))
+			continue
+		}
+
+		for _, pendingSupportBundle := range pendingSupportBundles {
+			a, err := app.Get(pendingSupportBundle.AppID)
```
emosbaugh, line 344: another case where this should be a function

kotsadm/pkg/socketservice/socketservice.go
```diff
@@ -0,0 +1,543 @@
...
+	restore, err := snapshot.GetRestore(a.RestoreInProgressName)
+	if err != nil {
+		return errors.Wrap(err, "failed to get restore")
+	}
+
+	if restore == nil {
+		return startVeleroRestore(a.RestoreInProgressName)
+	}
+
+	return checkRestoreComplete(a, restore)
```
emosbaugh, line 425: maybe wrap this error

kotsadm/pkg/socket/README.md
```diff
@@ -0,0 +1,8 @@
+# Socket
+
+We started using https://github.com/graarh/golang-socketio
```
laverya, line 3: Should we include a license header in the copied files? https://github.com/graarh/golang-socketio/blob/master/LICENSE.MIT

laverya: Also... Is this the library that operator is using? Would it be best to use the same lib on both sides there?

laverya: actually, is there just a way that we could share `kotsadm/pkg/socket/` and `kotsadm/operator/pkg/socket/`?

salahalsaleh: i think that's a good idea

kotsadm/pkg/socketservice/socketservice.go
salahalsaleh: good catch

salahalsaleh: the deployLoop function does end, but it gets called every 1 second. but i think the location of this should be changed, not sure where to yet

salahalsaleh: moving the deploy to it's own function is a good suggestion that would solve this 👍

kotsadm/pkg/socket/README.md
salahalsaleh: it is the same library that the operator uses, yeah.

kotsadm/operator/kustomize/base/deployment.yaml
```diff
@@ -25,7 +25,7 @@ spec:
               memory: 500Mi
           env:
             - name: KOTSADM_API_ENDPOINT
-              value: http://kotsadm-api-node:3000
+              value: http://kotsadm:3000
```
divolgin, line 28: can this be changed before we move everything?  shouldn't there be 2 endpoints now?

salahalsaleh: the operator does not talk to the old api directly anymore

kotsadm/pkg/socket/README.md
divolgin: I'll leave this duplicated. 
a) it's easier to remove duplicate code than to unravel dependencies
b) the entire operator package is going away soon
c) it's the quickest path to merging this PR

kotsadm/pkg/socketservice/socketservice.go
```diff
@@ -0,0 +1,542 @@
...
+		// this version is already the currently deployed version
+		return nil
+	}
+
+	d, err := downstream.Get(clusterSocket.ClusterID)
+	if err != nil {
+		return errors.Wrap(err, "failed to get downstream")
+	}
+
+	var deployError error
```
emosbaugh, line 169: id love to see another function start here :) but i will approve

kotsadm/pkg/socketservice/socketservice.go
```diff
@@ -0,0 +1,542 @@
...
+			if err != nil {
+				if ee, ok := err.(*exec.ExitError); ok {
+					err = fmt.Errorf("kustomize stderr: %q", string(ee.Stderr))
+				}
+				deployError = errors.Wrap(err, "failed to run kustomize for previously deployed app version")
+				return deployError
+			}
+			base64EncodedPreviousManifests = base64.StdEncoding.EncodeToString(previousRenderedManifests)
+		}
+	}
```
emosbaugh, line 256: and end here i guess?

PR #962, opened 2020-08-18
handle error when listing resources on restore

salahalsaleh:


kotsadm/operator/pkg/client/deploy.go
```diff
@@ -361,7 +361,11 @@ func (c *Client) clearNamespace(slug string, namespace string) (bool, error) {
 			continue
 		}
 		// there may be other resources that can't be listed besides what's in the skip set so ignore error
```
laverya, line 363: any reason this comment is no longer relevant?

PR #964, opened 2020-08-18
Migrate getKotsPreflightResult to go api

emosbaugh:


kotsadm/web/src/components/PreflightResultPage.jsx
```diff
@@ -155,7 +164,10 @@ class PreflightResultPage extends Component {
   }
 
   renderErrors = (errors) => {
-    this.props.data?.startPolling(2000);
```
emosbaugh, line 158: i dont know why this starts polling again

kotsadm/pkg/downstream/downstream.go
```diff
@@ -431,6 +480,29 @@ func versionFromRow(appID string, row scannable) (*types.DownstreamVersion, erro
...
+	var preflightResult sql.NullString
+	var preflightResultCreatedAt sql.NullTime
+
+	if err := row.Scan(
+		&preflightResult,
+		&preflightResultCreatedAt,
+		&r.AppSlug,
+		&r.ClusterSlug,
+	); err != nil {
+		return nil, errors.Wrap(err, "failed to scan")
```
divolgin, line 495: Is it possible that this returns NoRows, which would cause an error to be rendered in the UI?

emosbaugh: But i think in this case we want an error. Probably better to have a 404 and a message...

Screen Shot 2020-08-18 at 4 29 45 PM

PR #965, opened 2020-08-18
Move drainNode to Go

areed:


kotsadm/pkg/kurl/drain.go
```diff
@@ -0,0 +1,119 @@
...
+
+	return nil
+}
+
+func shouldDrain(pod *corev1.Pod) bool {
+	if pod == nil {
+		return false
+	}
+
+	// completed pods always ok to drain
```
laverya, line 69: if completed pods are always ok to drain, should this return true?

kotsadm/pkg/kurl/drain_test.go
```diff
@@ -0,0 +1,75 @@
...
+		pod    *corev1.Pod
+		expect bool
+	}{
+		{
+			name:   "should drain by default",
+			pod:    &corev1.Pod{},
+			expect: true,
+		},
+		{
+			name: "succeeded pod should not drain",
```
laverya, line 23: succeeded pod SHOULD drain, in the old code:
```
  // completed pods always ok to drain	
  if (isFinished(pod)) {	
    return true;	
  }
```

PR #966, opened 2020-08-19
Add all auth providers

emosbaugh:


laverya:
the relevant code: https://github.com/kubernetes/client-go/blob/eb8ca25/plugin/pkg/client/auth/plugins.go#L20-L24

PR #969, opened 2020-08-20
live render config in golang

laverya:


kotsadm/pkg/handlers/config.go
```diff
@@ -27,7 +28,9 @@ import (
 	versiontypes "github.com/replicatedhq/kots/kotsadm/pkg/version/types"
 	kotsv1beta1 "github.com/replicatedhq/kots/kotskinds/apis/kots/v1beta1"
 	"github.com/replicatedhq/kots/kotskinds/multitype"
+	config2 "github.com/replicatedhq/kots/pkg/config"
```
salahalsaleh, line 31: how about naming this `kotsconfig` and the other one `kotsadmconfig`? i feel like it would be clearer

laverya: sure
this is what happens when I rely on IDE autoimports

kotsadm/pkg/apiserver/server.go
```diff
@@ -130,6 +130,7 @@ func Start() {
 	r.Path("/api/v1/app/{appSlug}/registry").Methods("OPTIONS", "GET").HandlerFunc(handlers.GetAppRegistry)
 	r.Path("/api/v1/app/{appSlug}/registry/validate").Methods("OPTIONS", "POST").HandlerFunc(handlers.ValidateAppRegistry)
 	r.Path("/api/v1/app/{appSlug}/config").Methods("OPTIONS", "PUT").HandlerFunc(handlers.UpdateAppConfig)
+	r.Path("/api/v1/app/{appSlug}/liveConfig").Methods("OPTIONS", "PUT").HandlerFunc(handlers.LiveAppConfig)
```
salahalsaleh, line 133: if you wanna keep the pattern alive :)
```suggestion
	r.Path("/api/v1/app/{appSlug}/liveconfig").Methods("OPTIONS", "PUT").HandlerFunc(handlers.LiveAppConfig)
```

salahalsaleh: would it make more sense to use `GET` here? since this doesn't update anything

laverya: also probably true

kotsadm/pkg/handlers/config.go
```diff
@@ -141,6 +155,103 @@ func UpdateAppConfig(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	liveAppConfigResponse := LiveAppConfigResponse{
+		Success: false,
+	}
+
+	liveAppConfigRequest := LiveAppConfigRequest{}
+	if err := json.NewDecoder(r.Body).Decode(&liveAppConfigRequest); err != nil {
+		logger.Error(err)
+		liveAppConfigResponse.Error = "failed to decode request body"
```
divolgin, line 174: If these messages are not including the original error intentionally, there should be a comment about it.  Because someone will change it.

kotsadm/pkg/handlers/config.go
```diff
@@ -141,6 +155,103 @@ func UpdateAppConfig(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	renderedConfig, err := config2.TemplateConfigObjects(kotsKinds.Config, configValues, appLicense, template.LocalRegistry{})
+	if err != nil {
+		liveAppConfigResponse.Error = "failed to render templates"
+		JSON(w, 500, liveAppConfigResponse)
+		return
+	}
+
+	JSON(w, 200, LiveAppConfigResponse{Success: true, ConfigGroups: renderedConfig.Spec.Groups})
```
divolgin, line 252: would be cool to use `http.Status...` constants in handlers

pkg/config/config.go
```diff
@@ -19,6 +19,44 @@ func TemplateConfig(log *logger.Logger, configSpecData string, configValuesData
 	return templateConfig(log, configSpecData, configValuesData, licenseData, localRegistry, MarshalConfig)
 }
 
+func TemplateConfigObjects(configSpec *kotsv1beta1.Config, configValues map[string]template.ItemValue, license *kotsv1beta1.License, localRegistry template.LocalRegistry) (*kotsv1beta1.Config, error) {
+	templatedString, err := templateConfigObjects(configSpec, configValues, license, localRegistry, MarshalConfig)
+	if err != nil {
+		return nil, errors.Wrap(err, "call templateConfig")
```
divolgin, line 25: `"failed to template config"`

kotsadm/pkg/handlers/config.go
laverya: This is the pattern through the entire file, though
these errors are supposed to be pretty

kotsadm/pkg/apiserver/server.go
laverya: "why not use GET":
`TypeError: Failed to execute 'fetch' on 'Window': Request with GET/HEAD method cannot have body.`

PR #973, opened 2020-08-20
Do not re-deploy first version of the app when preflights succeed

emosbaugh:


kotsadm/pkg/store/s3pg/app_store.go
```diff
@@ -74,8 +74,9 @@ func (s S3PGStore) ListInstalledApps() ([]*apptypes.App, error) {
 }
 
 func (s S3PGStore) GetApp(id string) (*apptypes.App, error) {
-	logger.Debug("getting app from id",
-		zap.String("id", id))
+	// too noisy
```
salahalsaleh, line 77: finally, thank you!

PR #982, opened 2020-08-21
move getGitOpsRepo query to golang

laverya:


kotsadm/web/src/components/gitops/GitOpsDeploymentManager.jsx
```diff
@@ -72,21 +71,10 @@ class GitOpsDeploymentManager extends React.Component {
...
-      if (enabled) {
-        const selectedService = find(SERVICES, service => service.value === provider);
-        this.setState({
-          selectedService: selectedService ? selectedService : this.state.selectedService,
-          hostname: hostname || ""
-        });
-      }
-    }
-  }
+  componentDidUpdate(lastProps) {}
```
salahalsaleh, line 77: no need for this

kotsadm/web/src/components/gitops/GitOpsDeploymentManager.jsx
```diff
@@ -113,21 +101,57 @@ class GitOpsDeploymentManager extends React.Component {
...
+  getGitops = async () => {
+    try {
+      const res = await fetch(`${window.env.API_ENDPOINT}/gitops/get`, {
+        headers: {
+          "Authorization": Utilities.getToken(),
+          "Content-Type": "application/json",
+        },
+        method: "GET",
+      });
+      if (!res.ok) {
```
salahalsaleh, line 130: should log out on `401` status, example: https://github.com/replicatedhq/kots/blob/5a6baa27c54e10529b88f73c860f127f64ff8ee8/kotsadm/web/src/Root.jsx#L151

kotsadm/web/src/components/gitops/GitOpsDeploymentManager.jsx
```diff
@@ -113,21 +101,57 @@ class GitOpsDeploymentManager extends React.Component {
     }
   }
 
+  getGitopsState = async () => {
+    const freshGitops = await this.getGitops()
+
+    if (freshGitops.enabled) {
```
salahalsaleh, line 107: since it can `undefined`
```suggestion
    if (freshGitops?.enabled) {
```

kotsadm/web/src/components/gitops/GitOpsDeploymentManager.jsx
```diff
@@ -168,13 +192,16 @@ class GitOpsDeploymentManager extends React.Component {
     const gitOpsInput = this.getGitOpsInput(provider, repoUri, branch, path, format, action, hostname);
 
     try {
-      const getGitOpsRepo = this.props.getGitOpsRepoQuery?.getGitOpsRepo;
-      if (getGitOpsRepo?.enabled && this.providerChanged()) {
-        const success = await this.resetGitOps();
-        if (!success) {
-          return false;
+      const getGitOpsRepo = await this.getGitops();
```
salahalsaleh, line 195: should use the `gitops` in `this.state` instead of requesting it again

salahalsaleh: actually, you can just check for `if (this.providerChanged())` immediately here

kotsadm/web/src/components/gitops/GitOpsDeploymentManager.jsx
```diff
@@ -261,12 +288,12 @@ class GitOpsDeploymentManager extends React.Component {
       return;
     }
 
-    const getGitOpsRepo = this.props.getGitOpsRepoQuery?.getGitOpsRepo;
-    if (!getGitOpsRepo) {
+    const freshGitops = await this.getGitops();
```
salahalsaleh, line 291: same here, should use `gitops` from `this.state` instead

kotsadm/web/src/components/gitops/GitOpsDeploymentManager.jsx
```diff
@@ -113,21 +101,57 @@ class GitOpsDeploymentManager extends React.Component {
...
+    if (freshGitops.enabled) {
+      const selectedService = find(SERVICES, service => service.value === freshGitops.provider);
+      this.setState({
+        selectedService: selectedService ? selectedService : this.state.selectedService,
+        hostname: freshGitops.hostname || "",
+        gitops: freshGitops
+      })
+    } else {
+      this.setState({
+        gitops: freshGitops
```
salahalsaleh, line 116: should add initial value like `gitops: {}` to the state here https://github.com/replicatedhq/kots/pull/982/files#diff-47d43402b26a6d7686c467ce611e0197R62

kotsadm/web/src/components/gitops/GitOpsDeploymentManager.jsx
```diff
@@ -528,20 +555,19 @@ class GitOpsDeploymentManager extends React.Component {
...
       );
     }
 
-    const gitopsRepo = getGitOpsRepoQuery.getGitOpsRepo;
+    const gitopsRepo = this.state.gitops;
     const activeStep = find(STEPS, { step: this.state.step });
     return (
       <div className="GitOpsDeploymentManager--wrapper flex-column flex1">
-        {gitopsRepo.enabled && this.state.step !== "action" ?
+        {gitopsRepo && gitopsRepo.enabled && this.state.step !== "action" ?
```
salahalsaleh, line 570: just a suggestion, you can do this:
```suggestion
        {gitopsRepo?.enabled && this.state.step !== "action" ?
```

kotsadm/web/src/components/gitops/GitOpsDeploymentManager.jsx
```diff
@@ -113,21 +104,61 @@ class GitOpsDeploymentManager extends React.Component {
     }
   }
 
+  getGitopsState = async () => {
+    const freshGitops = await this.getGitops()
```
salahalsaleh, line 108: `getGitops` throws an error on `catch` but it's not caught here. you can wrap this whole function in `try/catch` or move the `setState` functionality to `getGitops` function (which i recommend)

kotsadm/web/src/components/gitops/GitOpsDeploymentManager.jsx
```diff
@@ -192,13 +199,10 @@ class GitOpsDeploymentManager extends React.Component {
...
 
     try {
-      const getGitOpsRepo = await this.getGitops();
-      if (getGitOpsRepo?.enabled) {
-        if (this.providerChanged()) {
-          const success = await this.resetGitOps();
-          if (!success) {
-            return false;
-          }
+      if (this.providerChanged()) {
```
salahalsaleh, line 202: sorry, i didn't notice this before, but this should check for `this.state.gitops?.enabled` as well.

kotsadm/web/src/components/gitops/GitOpsDeploymentManager.jsx
salahalsaleh: i feel like merging `getGitopsState` with `getGitops` would prevent future mistakes of calling the wrong function

laverya: yeah, initially there was a reason to have those independent
but not really anymore

kotsadm/pkg/gitops/gitops.go
```diff
@@ -43,6 +43,13 @@ type GitOpsConfig struct {
 	IsConnected bool   `json:"isConnected"`
 }
 
+type GlobalGitopsConfig struct {
```
salahalsaleh, line 46: Would recommend using GitOps instead of Gitops to keep the pattern going (noticed this in some other places as well in the pr)

kotsadm/web/src/components/gitops/GitOpsDeploymentManager.jsx
```diff
@@ -113,21 +104,59 @@ class GitOpsDeploymentManager extends React.Component {
...
+          hostname: freshGitops.hostname || "",
+          gitops: freshGitops
+        })
+      } else {
+        this.setState({
+          gitops: freshGitops
+        })
+      }
+
+      return freshGitops
```
salahalsaleh, line 140: No need to return anything here

kotsadm/web/src/components/gitops/GitOpsDeploymentManager.jsx
```diff
@@ -113,21 +104,59 @@ class GitOpsDeploymentManager extends React.Component {
...
+    } catch(err) {
+      console.log(err);
+      throw err;
```
salahalsaleh, line 143: probably shouldn’t throw error on catch here

kotsadm/web/src/components/gitops/GitOpsDeploymentManager.jsx
```diff
@@ -261,12 +290,11 @@ class GitOpsDeploymentManager extends React.Component {
       return;
     }
 
-    const getGitOpsRepo = this.props.getGitOpsRepoQuery?.getGitOpsRepo;
-    if (!getGitOpsRepo) {
+    if (!this.state.gitops) {
```
salahalsaleh, line 293: since gitops initial value is `{}`, this will never be true, you can use lodash `isEmpty` function

salahalsaleh: example: https://github.com/replicatedhq/kots/blob/bb2be4d67d912a0bdb074c9f2cafbff2dd3d6d2c/kotsadm/web/src/components/apps/AppSnapshotDetail.jsx#L7

kotsadm/web/src/components/gitops/GitOpsDeploymentManager.jsx
laverya: copied both of these from getAppsList above

kotsadm/web/src/components/gitops/GitOpsDeploymentManager.jsx
laverya: copied both of these from getAppsList above

salahalsaleh: hmmm, might not as big of a problem as i think. can keep it 👍

kotsadm/web/src/components/gitops/GitOpsDeploymentManager.jsx
salahalsaleh: @laverya getAppsList uses the returned value, the caller if this function does not

PR #989, opened 2020-08-22
Move saveSnapshotConfig to Go

areed:


kotsadm/pkg/handlers/snapshots.go
```diff
@@ -542,3 +496,114 @@ func GetVeleroStatus(w http.ResponseWriter, r *http.Request) {
...
+		logger.Error(err)
+		responseBody.Error = "failed to decode request body"
+		JSON(w, 400, responseBody)
+		return
+	}
+
+	app, err := store.GetStore().GetApp(requestBody.AppID)
+	if err != nil {
+		logger.Error(err)
+		responseBody.Error = "Internal Server Error"
```
salahalsaleh, line 535: how about a more descriptive error message, like `failed to get app`? (same comment for other occurrences in this function)

kotsadm/pkg/handlers/snapshots.go
```diff
@@ -542,3 +496,114 @@ func GetVeleroStatus(w http.ResponseWriter, r *http.Request) {
...
+		JSON(w, 500, responseBody)
```
salahalsaleh, line 536: we've been trying to use the status codes from the `http` package
```suggestion
		JSON(w, http.StatusInternalServerError, responseBody)
```

kotsadm/web/src/components/apps/AppSnapshotSchedule.jsx
```diff
@@ -164,28 +164,49 @@ class AppSnapshotSchedule extends Component {
...
+        this.setState({ updatingSchedule: false, updateConfirm: true });
+        setTimeout(() => {
+          this.setState({ updateConfirm: false })
+        }, 3000);
+      })
+      .catch((err) => {
+        console.log(err);
         this.setState({
-          message: msg,
+          message: "Failed to connect to API",
```
salahalsaleh, line 205: you can check for `err` and use `err.message` and if it does not exists, use `"Failed to connect to API"`. something like
```suggestion
          message: err ? err.message : "Failed to connect to API",
```

PR #998, opened 2020-08-25
Migrate cancelRestore to go api

emosbaugh:


kotsadm/pkg/handlers/restore.go
```diff
@@ -189,6 +189,36 @@ func GetRestoreStatus(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	appSlug := mux.Vars(r)["appSlug"]
+
+	foundApp, err := store.GetStore().GetAppFromSlug(appSlug)
+	if err != nil {
+		err = errors.Wrap(err, "failed to get app from app slug")
+		logger.Error(err)
+		w.WriteHeader(500)
```
salahalsaleh, line 208: let's use the status codes from the http package
```suggestion
		w.WriteHeader(http.StatusInternalServerError)
```

emosbaugh: love it! will change

PR #1004, opened 2020-08-25
move createGitOpsRepo mutation to go

salahalsaleh:


kotsadm/pkg/gitops/gitops.go
```diff
@@ -134,10 +143,8 @@ func GetDownstreamGitOps(appID string, clusterID string) (*GitOpsConfig, error)
 		if splitKey[2] == "repoUri" {
 			if string(val) == repoURI {
 				// this is the provider we want
-				idx, err := strconv.ParseInt(splitKey[1], 10, 64)
-				if err != nil {
-				}
-				provider, publicKey, privateKey, repoURI, err := gitOpsConfigFromSecretData(idx, secret.Data)
+				idx, _ := strconv.ParseInt(splitKey[1], 10, 64)
```
divolgin, line 146: @marccampbell this here looks like it should fail on error and not ignore it.

kotsadm/pkg/handlers/gitops.go
```diff
@@ -316,3 +325,34 @@ func GetGitOpsRepo(w http.ResponseWriter, r *http.Request) {
 
 	JSON(w, http.StatusOK, gitOpsConfig)
 }
+
+func CreateGitOps(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Access-Control-Allow-Origin", "*")
```
divolgin, line 330: checkout the `handleOptionsRequest` function for this

kotsadm/pkg/gitops/gitops.go
marccampbell: I don't disagree with your comment, but why me?  :)

salahalsaleh: @marccampbell probably because you wrote this block of code 🙂 . i'll update.

marccampbell: thanks @salahalsaleh.

divolgin: @marccampbell the red block looks like you wrote it.  but I only looked one commit back.

PR #1005, opened 2020-08-25
create s3 bucket in Go

divolgin:


kotsadm/pkg/apiserver/bootstrap.go
```diff
@@ -0,0 +1,57 @@
...
+
+func bootstrap() error {
+	if err := ensureS3Bucket(); err != nil {
+		return errors.Wrap(err, "failed to ensure s3 bucket")
+	}
+
+	return nil
+}
+
+func ensureS3Bucket() error {
```
marccampbell, line 23: Can this be in the S3PGStore, in an init function?  Not all stored need this.

divolgin: yeah...

PR #1007, opened 2020-08-26
Fixing logout loop

marccampbell:


kotsadm/pkg/handlers/logout.go
```diff
@@ -19,14 +19,20 @@ func Logout(w http.ResponseWriter, r *http.Request) {
...
+
 		logger.Error(err)
 		w.WriteHeader(401)
 		return
 	}
 
 	// we don't currently have roles, all valid tokens are valid sessions
 	if sess == nil || sess.ID == "" {
-		w.WriteHeader(401)
+		JSON(w, 204, "")
```
emosbaugh, line 35: I would think this would return double quotes as the response body rather than no content

PR #1010, opened 2020-08-26
Migrate getKotsDownstreamHistory to go api

emosbaugh:


kotsadm/pkg/apiserver/server.go
```diff
@@ -102,6 +102,7 @@ func Start() {
 	// Apps
 	r.Path("/api/v1/apps").Methods("OPTIONS", "GET").HandlerFunc(handlers.ListApps)
 	r.Path("/api/v1/apps/app/{appSlug}").Methods("OPTIONS", "GET").HandlerFunc(handlers.GetApp)
+	r.Path("/api/v1/app/{appSlug}/cluster/{clusterSlug}").Methods("OPTIONS", "GET").HandlerFunc(handlers.GetAppVersionHistory) // NOTE: why do we need cluster slug here?
```
emosbaugh, line 105: how can i avoid this? clusterSlug

kotsadm/web/src/components/apps/AppVersionHistory.jsx
```diff
@@ -80,10 +88,75 @@ class AppVersionHistory extends Component {
...
+          errorMsg: `Unexpected status code: ${res.status}`,
+          displayErrorModal: true,
+        });
+        return;
+      }
+      const response = await res.json();
+      const versionHistory = response.versionHistory;
+
+      if (isAwaitingResults(versionHistory)) {
+        setTimeout(() => this.fetchKotsDownstreamHistory(), 2000);
```
emosbaugh, line 139: any suggestions to avoid this?

trying to replace this code https://github.com/replicatedhq/kots/pull/1010/files#diff-0104f4f9b1ac0ae7899cbddf9481296bL920-L923

kotsadm/pkg/apiserver/server.go
salahalsaleh: 1- in the api: list the downstreams for the app and get the first one. remove clusterSlug from the path
2- in the UI: don't send it?

kotsadm/web/src/components/apps/AppVersionHistory.jsx
salahalsaleh: yes, use the `Repeater` class, and stop it when `isAwaitingResults` becomes false

salahalsaleh: although what you have also works

kotsadm/web/src/components/UploadAirgapBundle.jsx
```diff
@@ -361,7 +361,7 @@ class UploadAirgapBundle extends React.Component {
 
   getApp = async (slug) => {
     try {
-      const res = await fetch(`${window.env.API_ENDPOINT}/apps/app/${slug}`, {
+      const res = await fetch(`${window.env.API_ENDPOINT}/app/${slug}`, {
```
salahalsaleh, line 364: route is not updated in the api

kotsadm/web/src/components/apps/AppConfig.jsx
```diff
@@ -63,7 +63,7 @@ class AppConfig extends Component {
 
     try {
       const { slug } = this.props.match.params;
-      const res = await fetch(`${window.env.API_ENDPOINT}/apps/app/${slug}`, {
+      const res = await fetch(`${window.env.API_ENDPOINT}/app/${slug}`, {
```
salahalsaleh, line 66: route is not updated in the api

kotsadm/web/src/components/apps/AppSnapshotRestore.jsx
```diff
@@ -28,7 +28,6 @@ class AppSnapshotRestore extends Component {
   componentDidUpdate(lastProps) {
     const { match } = this.props;
     if (match.params.id !== lastProps.match.params.id) {
-      this.state.fetchRestoreDetailJob.stop();
```
salahalsaleh, line 31: why not? if it's not stopped, `.start` won't do anything and it will keep running the old instance with the old values.

salahalsaleh: hmmm, might not be a problem since everything is read from `props`. but stopping it won't hurt.

kotsadm/web/src/components/apps/AppVersionHistory.jsx
```diff
@@ -63,10 +62,19 @@ class AppVersionHistory extends Component {
...
+    versionHistoryJob: new Repeater(),
+    loadingVersionHistory: true,
+    versionHistory: [],
+    errorTitle: "",
+    errorMsg: "",
+    displayErrorModal: false,
   }
 
   componentDidMount() {
+    this.startVersionHistoryJob();
```
salahalsaleh, line 76: just curious, why not just call this immediately?
```
this.state.versionHistoryJob.start(this.fetchKotsDownstreamHistory, 2000);
```

salahalsaleh: also, this should not run when the component mounts, only when `isAwaitingResults` is `true`. and should be stopped when `isAwaitingResults` becomes `false`

kotsadm/web/src/components/apps/AppVersionHistory.jsx
```diff
@@ -80,8 +88,76 @@ class AppVersionHistory extends Component {
...
+          errorTitle: "Failed to get version history",
+          errorMsg: `Unexpected status code: ${res.status}`,
+          displayErrorModal: true,
+        });
+        return;
+      }
+      const response = await res.json();
+      const versionHistory = response.versionHistory;
+
+      if (!isAwaitingResults(versionHistory)) {
```
salahalsaleh, line 141: this should be something like:
```
if (isAwaitingResults) {
  this.startVersionHistoryJob();
} else {
  this.startVersionHistoryJob();
}
```

kotsadm/web/src/components/apps/AppVersionHistory.jsx
```diff
@@ -434,12 +510,14 @@ class AppVersionHistory extends Component {
...
     await this.props.makeCurrentVersion(match.params.slug, version);
-    await this.props.data.refetch();
+    await this.fetchKotsDownstreamHistory();
     this.setState({ versionToDeploy: null });
 
     if (this.props.updateCallback) {
       this.props.updateCallback();
     }
+
+    this.startVersionHistoryJob();
```
salahalsaleh, line 520: why?

salahalsaleh: if it's needed, it will be called from `fetchKotsDownstreamHistory`

kotsadm/web/src/components/apps/AppVersionHistory.jsx
```diff
@@ -542,7 +620,7 @@ class AppVersionHistory extends Component {
           if (this.props.updateCallback) {
             this.props.updateCallback();
           }
-          this.props.data.refetch();
+          this.startVersionHistoryJob();
```
salahalsaleh, line 623: this is not equivalent to `this.props.data.refetch();`. this should be:
```
this.fetchKotsDownstreamHistory();
```

kotsadm/web/src/components/apps/AppVersionHistory.jsx
```diff
@@ -587,9 +665,9 @@ class AppVersionHistory extends Component {
         this.props.refreshAppData();
         const response = await res.json();
         if (response.availableUpdates === 0) {
-          if (!find(this.props.data?.getKotsDownstreamHistory, { parentSequence: response.currentAppSequence })) {
+          if (!find(this.state.versionHistory, { parentSequence: response.currentAppSequence })) {
             // version history list is out of sync - most probably because of automatic updates happening in the background - refetch list
-            this.props.data?.refetch();
+            this.startVersionHistoryJob();
```
salahalsaleh, line 670: same here

kotsadm/web/src/components/UploadAirgapBundle.jsx
emosbaugh: good catch. i tried to update this but it didnt work and then i tried to roll it back but apparently i didnt

PR #1014, opened 2020-08-26
move snapshot scheduler to go api

salahalsaleh:


kotsadm/pkg/snapshot/backup.go
```diff
@@ -26,32 +26,36 @@ import (
...
+	backup, err := createApplicationBackup(context.TODO(), a, isScheduled)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create application backup")
 	}
 
 	// uncomment to create disaster recovery snapshots
-	// if err := createAdminConsoleBackup(context.TODO()); err != nil {
-	// 	return errors.Wrap(err, "failed to create admin console backup")
+	// if err := createAdminConsoleBackup(context.TODO(), isScheduled); err != nil {
+	// 	return nil, errors.Wrap(err, "failed to create admin console backup")
```
divolgin, line 37: i like how you fixed commented out code here :)

salahalsaleh: 🙂

kotsadm/pkg/snapshotscheduler/snapshotscheduler.go
```diff
@@ -0,0 +1,153 @@
...
+		logger.Infof("Not yet time to snapshot app %s", a.ID)
+		return nil
+	}
+
+	hasUnfinished, err := snapshot.HasUnfinishedBackup(a.ID)
+	if err != nil {
+		return errors.Wrap(err, "failed to to check if app has unfinished backups")
+	}
+	if hasUnfinished {
+		logger.Infof("Postponing scheduled snapshot for %s because one is in progress", a.ID)
```
divolgin, line 106: I can't think of any reason why we'd want to do this....  Potentially this will block all backups.

salahalsaleh: I agree. It will block all _scheduled_ backups, yes. Just trying to keep the logic from TS.

PR #1016, opened 2020-08-26
use golang api for `kurl` graphql endpoint (lists nodes)

laverya:


kotsadm/pkg/apiserver/server.go
```diff
@@ -174,6 +174,7 @@ func Start() {
 	r.Path("/api/v1/kurl/generate-node-join-command-master").Methods("OPTIONS", "POST").HandlerFunc(handlers.GenerateNodeJoinCommandMaster)
 	r.Path("/api/v1/kurl/nodes/{nodeName}/drain").Methods("OPTIONS", "POST").HandlerFunc(handlers.DrainNode)
 	r.Path("/api/v1/kurl/nodes/{nodeName}").Methods("OPTIONS", "DELETE").HandlerFunc(handlers.DeleteNode)
+	r.Path("/api/v1/kurl/nodes/").Methods("OPTIONS", "GET").HandlerFunc(handlers.GetKurlNodes)
```
emosbaugh, line 177: why the trailing slash?

kotsadm/pkg/kurl/kurl_nodes.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
+	kuberneteserrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	metrics "k8s.io/metrics/pkg/client/clientset/versioned"
+)
+
+type KurlNodes struct {
```
emosbaugh, line 16: wanna move all this to `pkg/kurl/types`?

kotsadm/pkg/handlers/kurl_get.go
```diff
@@ -1 +1,42 @@
...
+	}
+
+	client, err := k8s.Clientset()
+	if err != nil {
+		logger.Error(err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	mc, err := k8s.Metricsset()
```
marccampbell, line 28: I think this should be inline, none of the other specialized clientsets are wrapped in that k8s helper, and this is the only use of that function

kotsadm/pkg/kurl/kurl_nodes.go
```diff
@@ -0,0 +1,116 @@
...
+func GetNodes(client kubernetes.Interface, metrics *metrics.Clientset) (KurlNodes, error) {
+	nodes, err := client.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return KurlNodes{}, errors.Wrap(err, "list nodes")
+	}
+
+	toReturn := KurlNodes{}
+
+	for _, node := range nodes.Items {
+		cpuCapacity, _ := strconv.ParseFloat(node.Status.Capacity.Cpu().String(), 64)
```
marccampbell, line 55: why ignore the errors?

kotsadm/pkg/kurl/kurl_nodes.go
```diff
@@ -0,0 +1,116 @@
...
+	DiskPressure   bool `json:"diskPressure"`
+	PidPressure    bool `json:"pidPressure"`
+	Ready          bool `json:"ready"`
+}
+
+// GetNodes will get a list of nodes with stats
+func GetNodes(client kubernetes.Interface, metrics *metrics.Clientset) (KurlNodes, error) {
+	nodes, err := client.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return KurlNodes{}, errors.Wrap(err, "list nodes")
```
marccampbell, line 49: Why not return a pointer so that this can be nil?

kotsadm/pkg/kurl/kurl_nodes.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	kurlConf, err := ReadConfigMap(client)
+	if err != nil {
+		if kuberneteserrors.IsNotFound(err) {
+			toReturn.IsKurlEnabled = false
+			return toReturn, nil
+		}
+		return KurlNodes{}, errors.Wrap(err, "get kurl config")
+	} else {
```
emosbaugh, line 84: no need for else here

kotsadm/pkg/kurl/kurl_nodes.go
```diff
@@ -0,0 +1,116 @@
...
+		toReturn.HA = parsedBool
+	}
+
+	return toReturn, nil
+}
+
+func findNodeConditions(conditions []v1.NodeCondition) NodeConditions {
+	discoveredConditions := NodeConditions{}
+	for _, condition := range conditions {
+		if condition.Type == "MemoryPressure" {
```
emosbaugh, line 102: switch?

kotsadm/web/src/components/apps/ClusterNodes.jsx
```diff
@@ -22,6 +21,38 @@ export class ClusterNodes extends Component {
...
+  componentDidMount() {
+    this.getNodeStatus();
+  }
+
+  getNodeStatus = async () => {
+    try {
+      const res = await fetch(`${window.env.API_ENDPOINT}/kurl/nodes`, {
+        headers: {
+          "Authorization": Utilities.getToken(),
+          "Content-Type": "application/json",
```
emosbaugh, line 33: a get request does not have a content type

kotsadm/pkg/kurl/kurl_nodes.go
laverya: because I forgot I hadn't handled them 🤦

kotsadm/pkg/apiserver/server.go
laverya: thought I removed that, whoops

kotsadm/pkg/kurl/kurl_nodes.go
laverya: would that be a new package?

emosbaugh: Ya. I like the pattern and we seem to be using it

kotsadm/pkg/kurl/kurl_nodes.go
```diff
@@ -0,0 +1,148 @@
...
+	toReturn := KurlNodes{}
+
+	for _, node := range nodes.Items {
+		cpuCapacity := CapacityAvailable{}
+		memoryCapacity := CapacityAvailable{}
+		podCapacity := CapacityAvailable{}
+
+		memoryCapacity.Capacity = float64(node.Status.Capacity.Memory().Value()) / 1000000000 // capacity in GB
+
+		cpuCapacity.Capacity, err = strconv.ParseFloat(node.Status.Capacity.Cpu().String(), 64)
```
emosbaugh, line 66: I assume these are all type quantity. Why convert to string?

laverya: If there's a better way to convert to float, I'd love to hear it
Or I'll just try returning the decimal type

emosbaugh: I thought maybe you could get it from AsInt64() but idk

kotsadm/web/src/components/apps/ClusterNodes.jsx
```diff
@@ -22,6 +21,38 @@ export class ClusterNodes extends Component {
...
+        if (res.status === 401) {
+          Utilities.logoutUser();
+          return;
+        }
+        console.log("failed to get node status list, unexpected status code", res.status);
+        return;
+      }
+      const response = await res.json();
+      this.setState({
+        kurl: response,
```
salahalsaleh, line 47: should add `kurl` to `state` as well

kotsadm/pkg/kurl/kurl_nodes.go
```diff
@@ -0,0 +1,148 @@
...
+	kurlConf, err := ReadConfigMap(client)
+	if err != nil {
+		if kuberneteserrors.IsNotFound(err) {
+			toReturn.IsKurlEnabled = false
+			return &toReturn, nil
+		}
+		return nil, errors.Wrap(err, "get kurl config")
+	}
+
+	toReturn.IsKurlEnabled = true
```
salahalsaleh, line 118: I know this was ported from TS, but having `IsKurlEnabled` feels weird. Isn't this query only called if kurl is enabled? and the UI only shows the `Cluster Management` tab if kurl is enabled. https://github.com/replicatedhq/kots/blob/0117fba630a0c0b48959ea407564f549833f86f8/kotsadm/web/src/components/shared/NavBar.jsx#L168

salahalsaleh: Unless we want to show some basic info for non-kurl clusters as well in the future.

kotsadm/pkg/kurl/kurl_nodes.go
laverya: Hmmm
.String() gives a value of `8` in my testing thus far
Maybe AsInt64() would give 8000 instead?

kotsadm/pkg/kurl/kurl_nodes.go
laverya: That was pretty much my thought
A lot of the data is available for kurl or non-kurl clusters

kotsadm/web/src/components/apps/ClusterNodes.jsx
laverya: what do you mean?

salahalsaleh: add something like `kurl: null` here https://github.com/replicatedhq/kots/blob/master/kotsadm/web/src/components/apps/ClusterNodes.jsx#L16

kotsadm/web/src/components/apps/ClusterNodes.jsx
```diff
@@ -22,6 +21,38 @@ export class ClusterNodes extends Component {
...
+      });
+      return response;
```
salahalsaleh, line 49: this return is not needed (but won't hurt), just pointing it out.

PR #1031, opened 2020-08-27
cleanup node API objects

divolgin:


pkg/kotsadm/main.go
```diff
@@ -371,14 +362,18 @@ func ensureKotsadm(deployOptions types.DeployOptions, clientset *kubernetes.Clie
...
+	if err := ensureApplicationMetadata(deployOptions, clientset); err != nil {
+		return errors.Wrap(err, "failed to ensure custom branding")
 	}
 
 	if err := ensureOperator(deployOptions, clientset); err != nil {
 		return errors.Wrap(err, "failed to ensure operator")
 	}
 
+	if err := removeNodeAPI(&deployOptions, clientset); err != nil {
+		log.Error(errors.Errorf("Failed to remove deprecated API: %v", err))
```
marccampbell, line 374: technically it's not deprecated, it's just removed.

divolgin: ha!  well, if on error we fail to remove it... :)

PR #1041, opened 2020-08-28
Redirect to login page on 401

divolgin:


kotsadm/pkg/handlers/logout.go
```diff
@@ -11,22 +11,30 @@ import (
 func Logout(w http.ResponseWriter, r *http.Request) {
 	sess, err := session.Parse(r.Header.Get("Authorization"))
 	if err != nil {
+		if store.GetStore().IsNotFound(err) {
+			JSON(w, http.StatusNoContent, "")
```
emosbaugh, line 15: doesnt this return content? why not `w.WriteHeader(http.StatusNoContent)`?

kotsadm/pkg/handlers/logout.go
```diff
@@ -11,22 +11,30 @@ import (
...
+			return
+		}
+		if r.Header.Get("Authorization") == "" {
+			JSON(w, http.StatusNoContent, "")
+			return
+		}
 		logger.Error(err)
```
marccampbell, line 22: @divolgin how would you feel about not logging this. It's common when switching clusters, so it's confusing to see this error in logs.

kotsadm/pkg/apiserver/server.go
```diff
@@ -102,6 +101,11 @@ func Start() {
 	* Session auth routes
 	**********************************************************************/
 
+	sessionAuthQuietRouter := r.PathPrefix("").Subrouter()
+	sessionAuthQuietRouter.Use(handlers.RequireValidSessionQuietMiddleware)
+
+	sessionAuthQuietRouter.Path("/api/v1/ping").Methods("GET").HandlerFunc(handlers.Ping)
```
emosbaugh, line 107: can you remove this comment? https://github.com/replicatedhq/kots/blob/master/kotsadm/pkg/handlers/ping.go#L18

kotsadm/pkg/handlers/logout.go
```diff
@@ -11,22 +11,30 @@ import (
 func Logout(w http.ResponseWriter, r *http.Request) {
 	sess, err := session.Parse(r.Header.Get("Authorization"))
 	if err != nil {
+		if store.GetStore().IsNotFound(err) {
+			JSON(w, http.StatusNoContent, "")
+			return
+		}
+		if r.Header.Get("Authorization") == "" {
```
emosbaugh, line 18: why not move this to the top before session.Parse?

kotsadm/pkg/handlers/logout.go
divolgin: if you switch clusters, it should hit the store's IsNotFound

PR #1047, opened 2020-09-01
Implementing list versions after

marccampbell:


kotsadm/pkg/store/s3pg/version_store.go
```diff
@@ -442,5 +442,45 @@ func (s S3PGStore) GetAppVersion(appID string, sequence int64) (*versiontypes.Ap
...
 
 func (s S3PGStore) GetAppVersionsAfter(appID string, sequence int64) ([]*versiontypes.AppVersion, error) {
-	return nil, errors.New("not implemented")
+	db := persistence.MustGetPGSession()
+	query := `select sequence, created_at, status, applied_at, kots_installation_spec from app_version where app_id = $1 and sequence > $2`
+	rows, err := db.Query(query, appID, sequence)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to query")
+	}
+
```
divolgin, line 451: `defer rows.Close()`

kotsadm/pkg/store/s3pg/version_store.go
```diff
@@ -442,5 +442,45 @@ func (s S3PGStore) GetAppVersion(appID string, sequence int64) (*versiontypes.Ap
...
+	var deployedAt sql.NullTime
+
+	var installationSpec string
+
+	versions := []*versiontypes.AppVersion{}
+
+	for rows.Next() {
+		v := versiontypes.AppVersion{}
+		if err := rows.Scan(&v.Sequence, &v.CreatedOn, &status, &deployedAt, &installationSpec); err != nil {
+			if err == sql.ErrNoRows {
```
divolgin, line 462: I don't think `ErrNoRows` happens in the loop, but if it does, how about just `return versions, nil`?  Looks like caller should not fail if no versions found.

PR #1052, opened 2020-09-02
Airgap bundle chunk upload

salahalsaleh:


kotsadm/pkg/handlers/airgap.go
```diff
@@ -89,49 +262,116 @@ func updateAppFromAirgap(w http.ResponseWriter, r *http.Request) {
...
+			isUploadComplete = false
+		} else if err != nil {
+			return false, errors.Wrap(err, "failed to os state file")
+		}
+	}
+
+	return isUploadComplete, nil
+}
+
+func createAirgapBundleFromChunks(airgapBundlePath string, uploadedFileName string, chunksDir string, totalChunks int64) error {
```
salahalsaleh, line 340: i'm not sure how long this will take for really big files, might be a better idea to write the chunks directly to the file and calculate offsets to determine if the file upload was complete.

kotsadm/pkg/handlers/airgap.go
```diff
@@ -53,7 +71,172 @@ func ResetAirgapInstallStatus(w http.ResponseWriter, r *http.Request) {
...
+	airgapBundleChunk, _, err := r.FormFile("file")
+	if err != nil {
+		logger.Error(err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	// create airgap bundle file if not exists
+	airgapBundlePath := getAirgapBundlePath(resumableIdentifier)
+	_, err = os.Stat(airgapBundlePath)
```
divolgin, line 157: This should be in a mutex

kotsadm/web/src/components/AirgapUploadProgress.jsx
```diff
@@ -214,7 +214,7 @@ class AirgapUploadProgress extends React.Component {
               <div className="flex alignItems--center u-marginTop--20">
                 <span className="u-fontWeight--bold u-fontSize--normal u-color--tundora u-marginRight--10">{percentage}</span>
                 {progressBar}
-                {uploadComplete ? <span className="u-marginLeft--10 icon checkmark-icon" /> : <span className="u-fontWeight--medium u-fontSize--normal u-color--tundora u-marginLeft--10">{formatByteSize(total)}</span>}
+                {uploadComplete ? <span className="u-marginLeft--10 icon checkmark-icon" /> : <span className="u-fontWeight--medium u-fontSize--normal u-color--tundora u-marginLeft--10">{formatByteSize(progress)}</span>}{/* TODO: fix this */}
```
GraysonNull, line 217: What does `/*TODO: fix this */` mean?

GraysonNull (approved):

kotsadm/web/src/components/AirgapUploadProgress.jsx
salahalsaleh: ah nice, good thing i added this, totally forgot about it. this used to show the total size of the bundle i think

kotsadm/pkg/handlers/airgap.go
```diff
@@ -90,48 +288,100 @@ func updateAppFromAirgap(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	return isUploadComplete
+}
+
+func cleanUp(uploadedFileIdentifier string, totalChunks int64) error {
+	var i int64
+	for i = 1; i <= totalChunks; i++ {
+		chunkKey := getChunkKey(uploadedFileIdentifier, i)
+		delete(uploadedAirgapBundleChunks, chunkKey)
```
divolgin, line 376: Is `uploadedFileIdentifier` unique across different uploads?  It might be better to clean up this map when upload starts, at the same time file is created.

salahalsaleh: it's generated based on the file name and size. if it's clean on start, then there won't be resumable uploads, no? (in case the upload was interrupted)

salahalsaleh: it's always the same for the same file name and size

salahalsaleh: it's always the same for the same file name and size

divolgin (approved):

GraysonNull (approved):

GraysonNull (approved):

GraysonNull (approved):

kotsadm/pkg/apiserver/server.go
```diff
@@ -137,6 +137,7 @@ func Start() {
 	sessionAuthRouter.Path("/api/v1/app/{appSlug}/versions").Methods("GET").HandlerFunc(handlers.GetAppVersionHistory)
 
 	// Airgap
+	sessionAuthRouter.Path("/api/v1/app/airgap").Methods("POST", "PUT").HandlerFunc(handlers.LegacyUploadAirgapBundle) // Legacy route only used by scripts
```
divolgin: let's call this "backwards compatibility" instead of legacy.

kotsadm/pkg/handlers/airgap.go
```diff
@@ -443,3 +444,115 @@ func getKurlRegistryCreds() (hostname string, username string, password string,
 
 	return
 }
+
+// Legacy airgap upload handler (only used by scripts)
+func LegacyUploadAirgapBundle(w http.ResponseWriter, r *http.Request) {
```
divolgin, line 449: Let's call this `UploadAirgapBundle` and drop legacy from everything else

divolgin (approved):

divolgin (approved):

PR #1061, opened 2020-09-03
Speed up gitops key generation by pre-seeding

emosbaugh:


kotsadm/cmd/kotsadm/main.go
```diff
@@ -1,7 +1,14 @@
...
-import "github.com/replicatedhq/kots/kotsadm/cmd/kotsadm/cli"
+import (
+	"math/rand"
+	"time"
+
+	"github.com/replicatedhq/kots/kotsadm/cmd/kotsadm/cli"
+)
 
 func main() {
+	rand.Seed(time.Now().UnixNano())
```
laverya, line 11: why not put the rand.Seed in the gitops init function? https://github.com/replicatedhq/kots/pull/1061/files#diff-861ad608adfe5caaf02438963a6b1c02R16

laverya: instead of spread through the code

emosbaugh: idk. typically i feel like it should be left up to the consumer of the package

laverya (approved):

kotsadm/pkg/gitops/key.go
```diff
@@ -0,0 +1,43 @@
...
+	if key != nil {
+		return key, nil
+	}
+	return rsa.GenerateKey(r, 4096)
+}
+
+func generatePreseedPrivateKey() {
+	preseedPrivateKeyMu.Lock()
+	defer preseedPrivateKeyMu.Unlock()
+	preseedPrivateKey, _ = rsa.GenerateKey(r, 4096)
```
salahalsaleh, line 42: why not handle/log the error here?

emosbaugh: i could panic? this is in a goroutine in init() and it would be hard to surface this to the user

laverya: also, the failure mode here is 'generate another cert later, and return the error there if present'

PR #1062, opened 2020-09-03
Create collector and analyzer from supportbundle

areed:


pkg/kotsutil/kots.go
```diff
@@ -231,6 +243,15 @@ func (o KotsKinds) Marshal(g string, v string, k string) (string, error) {
 					return "", errors.Wrap(err, "failed to encode preflight")
 				}
 				return string(b.Bytes()), nil
+			case "SupportBundle":
```
areed, line 246: This isn't used anywhere but added for consistency

divolgin (approved):

PR #1070, opened 2020-09-04
Gitops remove modal error

emosbaugh:


kotsadm/web/src/components/apps/AppGitops.jsx
```diff
@@ -105,33 +102,22 @@ class AppGitops extends Component {
...
         if (res.status === 400) {
-          msg = `Unable to authenticate. Please make sure you have added your deployment key to the git repo.`;
-        }
-        try {
-          const data = await res.json();
-          if (data?.error) {
-            console.log(`Failed to test gitops connection: ${data.error}`);
+          const response = await res.json();
+          if (response?.error) {
+            console.log(response?.error);
```
salahalsaleh, line 110: this error is not surfaced in the UI

kotsadm/web/src/components/apps/AppGitops.jsx
```diff
@@ -105,33 +102,22 @@ class AppGitops extends Component {
           Utilities.logoutUser();
           return;
         }
-        let msg = `Unexpected status code: ${res.status}`;
+        this.props.refetch();
```
salahalsaleh, line 105: moving this here would only refetch if `!res.ok`

kotsadm/web/src/components/apps/AppGitops.jsx
```diff
@@ -105,33 +102,22 @@ class AppGitops extends Component {
...
+          throw new Error(`authentication failed`);
         }
-        this.setState({
-          errorTitle: "Failed to test connection",
-          errorMsg: msg,
-          displayErrorModal: true,
-        });
-        this.props.refetch();
-        return;
+        throw new Error(`unexpected status code: ${res.status}`);
```
salahalsaleh, line 114: i think we should keep handling the error here instead of throwing it because this function is also called when clicking `Try again`

kotsadm/web/src/components/apps/AppGitops.jsx
emosbaugh: its pretty technical

kotsadm/web/src/components/apps/AppGitops.jsx
emosbaugh: I do handle it in the outer try catch

kotsadm/web/src/components/apps/AppGitops.jsx
emosbaugh: that is how it was before https://github.com/replicatedhq/kots/blame/5ccc4e16229c9066ed33fb3224697406c14e7398/kotsadm/web/src/components/apps/AppGitops.jsx#L109

salahalsaleh (approved):

PR #1112, opened 2020-09-18
Headless embedded cluster installs

divolgin:


cmd/kots/cli/install.go
```diff
@@ -173,6 +173,15 @@ func InstallCmd() *cobra.Command {
 				}
 			}
 
+			isKurl, err := kotsadm.IsKurl(kubernetesConfigFlags)
+			if err != nil {
+				return errors.Wrap(err, "failed to check kURL")
+			}
+
+			if isKurl && deployOptions.Namespace == "default" {
```
marccampbell, line 181: ~"default"~ `metav1.NamespaceDefault`

cmd/kots/cli/install.go
```diff
@@ -173,6 +173,15 @@ func InstallCmd() *cobra.Command {
 				}
 			}
 
+			isKurl, err := kotsadm.IsKurl(kubernetesConfigFlags)
+			if err != nil {
+				return errors.Wrap(err, "failed to check kURL")
```
marccampbell, line 178: This will fail if the user's local kubeconfig doesn't have permission to list nodes. Does this break the minimalRBAC flag use?

kotsadm/pkg/airgap/airgap.go
```diff
@@ -74,12 +78,17 @@ func CreateAppFromAirgap(pendingApp *types.PendingApp, airgapBundlePath string,
 		return errors.Wrap(err, "failed to set task status")
 	}
 
-	// we seem to need a lot of temp dirs here... maybe too many?
-	archiveDir, err := version.ExtractArchiveToTempDirectory(airgapBundlePath)
-	if err != nil {
-		return errors.Wrap(err, "failed to extract archive")
+	archiveDir := airgapPath
+	if strings.HasSuffix(strings.ToLower(airgapPath), ".airgap") {
```
marccampbell, line 82: Can we use `filepath.Ext` here instead?

kotsadm/pkg/automation/automation.go
```diff
@@ -84,48 +92,150 @@ func AutomateInstall() error {
...
-		}
-		_, err = online.CreateAppFromOnline(&pendingApp, upstreamURI, true)
+		airgapData, err := getAirgapData(clientset, verifiedLicense)
 		if err != nil {
-			logger.Error(err)
+			logger.Error(errors.Wrapf(err, "failed to load airgap data for %s", verifiedLicense.Spec.AppSlug))
 			continue
 		}
 
+		if skipImagePush && airgapData != nil {
```
marccampbell, line 124: The logic here .. i think it's right, but i think a comment above explaining these `if` blocks will be really useful for future-us.

kotsadm/pkg/automation/automation.go
```diff
@@ -84,48 +92,150 @@ func AutomateInstall() error {
...
+				LicenseData: string(license),
+			}
+
+			err = airgap.CreateAppFromAirgap(&pendingApp, airgapFilesDir, registryHost, namespace, username, password, true)
+			if err != nil {
+				logger.Error(errors.Wrap(err, "failed to create airgap app"))
+				continue
+			}
+		} else {
+			pendingApp := onlinetypes.PendingApp{
```
marccampbell, line 164: same here, i've lost context of the else.  can you explain more about the use case this is handling (in a comment?)

kotsadm/pkg/automation/automation.go
```diff
@@ -84,48 +92,150 @@ func AutomateInstall() error {
...
+	return nil
+}
+
+func getAirgapData(clientset kubernetes.Interface, license *kotsv1beta1.License) (map[string][]byte, error) {
+	labels := []string{
+		"kots.io/automation=airgap",
+		fmt.Sprintf("kots.io/app=%s", license.Spec.AppSlug),
+	}
+	configMaps, err := clientset.CoreV1().ConfigMaps(os.Getenv("POD_NAMESPACE")).List(context.TODO(), metav1.ListOptions{
+		LabelSelector: strings.Join(labels, ","),
```
marccampbell, line 202: Instead of `strings.Join`, i think we should use `labels.SelectorFromSet`

kotsadm/pkg/automation/automation.go
```diff
@@ -84,48 +92,150 @@ func AutomateInstall() error {
...
+	return result, nil
+}
+
+func deleteAirgapData(clientset kubernetes.Interface, license *kotsv1beta1.License) error {
+	labels := []string{
+		"kots.io/automation=airgap",
+		fmt.Sprintf("kots.io/app=%s", license.Spec.AppSlug),
+	}
+	configMaps, err := clientset.CoreV1().ConfigMaps(os.Getenv("POD_NAMESPACE")).List(context.TODO(), metav1.ListOptions{
+		LabelSelector: strings.Join(labels, ","),
```
marccampbell, line 228: `labels.SelectorFromSet`

pkg/kotsadm/main.go
```diff
@@ -363,57 +423,85 @@ func ensureKotsadm(deployOptions types.DeployOptions, clientset *kubernetes.Clie
...
 		// if there's a license, we write it as a secret and kotsadm will
 		// find it on startup and handle installation
-		if err := ensureLicenseSecret(&deployOptions, clientset); err != nil {
+		updated, err := ensureLicenseSecret(&deployOptions, clientset)
+		if err != nil {
 			return errors.Wrap(err, "failed to ensure license secret")
 		}
+
+		if updated {
+			restartKotsadmAPI = true
```
marccampbell, line 432: I think secrets roll the pod when updated, but configmaps don't. We should rely on built-in k8s restarting and not restart on our own if all of the changes applied will result in k8s restarting the pod, or else we will likely end up with a weird scenario where it's restarted twice and fail

marccampbell (commented):
Mostly lgtm, a few nit-picky comments about library usage vs strings.

Generally, this feels like a ton of complexity in added if statements, and somewhat complex if statements. Bigger question: what can we do to start reducing this complexity and branching logic in the installer to make it more maintainable?

pkg/kotsadm/main.go
divolgin: both of these are optional and neither is mounted in the pod.  we try to read these in the automation package on startup.

cmd/kots/cli/install.go
divolgin: I updated this function to not error if we can't list nodes

divolgin:
@marccampbell I agree about complexity.  I think duplicating code is a good alternative here.  So we would have dedicated functions for all of the permutations for online/airgap+existing/embedded+headless/interactive.   The downside is that we have to maintain multiple copies of similar code.  Upside is reduced complexity and reduced risk of breaking all permutations while working on one of them.

marccampbell (approved):

PR #1117, opened 2020-09-21
Making createVersion atomic

divolgin:


kotsadm/pkg/gitops/types/gitops_inetrface.go
```diff
@@ -0,0 +1,5 @@
+package types
```
marccampbell, line 1: Spelling error in filename

kotsadm/pkg/store/s3pg/version_store.go
```diff
@@ -267,7 +272,114 @@ func (s S3PGStore) GetAppVersionArchive(appID string, sequence int64) (string, e
...
+
+		previousArchiveDir = previousDir
+	}
+
+	downstreams, err := s.ListDownstreamsForApp(appID)
+	if err != nil {
+		return int64(0), errors.Wrap(err, "failed to list downstreams")
+	}
+
+	if err := secrets.ReplaceSecretsInPath(filesInDir); err != nil {
```
marccampbell, line 309: can this _somehow_ get out of the store?  it feels better in the version package to me

kotsadm/pkg/store/s3pg/version_store.go
```diff
@@ -267,7 +272,114 @@ func (s S3PGStore) GetAppVersionArchive(appID string, sequence int64) (string, e
...
+			}
+			if t {
+				downstreamStatus = "pending_config"
+			}
+		}
+
+		diffSummary, diffSummaryError := "", ""
+		if currentSequence != nil {
+			// diff this release from the last release
+			diff, err := downstream.DiffAppVersionsForDownstream(d.Name, filesInDir, previousArchiveDir, kotsKinds.KustomizeVersion())
```
marccampbell, line 350: this is weird. i would expect the downstream to import stores, not the other way.

kotsadm/pkg/store/s3pg/version_store.go
```diff
@@ -267,7 +272,114 @@ func (s S3PGStore) GetAppVersionArchive(appID string, sequence int64) (string, e
...
+			} else {
+				b, err := json.Marshal(diff)
+				if err != nil {
+					diffSummaryError = errors.Wrap(err, "failed to marshal diff").Error()
+				}
+				diffSummary = string(b)
+			}
+		}
+
+		commitURL, err := gitops.CreateGitOpsDownstreamCommit(appID, d.ClusterID, int(newSequence), filesInDir, d.Name)
```
marccampbell, line 362: this is wrong, the store shouldn't call gitops.

marccampbell (changes requested):
It's better.  But there are still a couple of architecture issues, namely the store calling downstream and gitops.  

Could the store shouldn't be calling these, because the same logic (make a gitops PR) is going to have to be implemented in each store over time.

How can we continue what you've started and make it so that the non-store work happens before/after the store completes its responsibility of saving the data?

kotsadm/pkg/store/s3pg/version_store.go
divolgin: Looking at the implementation of `DiffAppVersionsForDownstream` I'm going to suggest that we move it out of the downstream package.  It calls `kustomize` and then diff.  technically it does nothing with downstreams.

kotsadm/pkg/store/s3pg/version_store.go
divolgin: this can move out to the caller.  `filesInDir` is available before this function is called.

kotsadm/pkg/store/s3pg/version_store.go
divolgin: I think the correct way to solve this is to create gitops commit asynchronously.  Similar to how operator applies downstreams.

marccampbell (approved):

PR #1118, opened 2020-09-22
Handle intermittent internet failure during airgap bundle upload

salahalsaleh:
The behavior will be as follows:

## In case of internet failure or unable to reach the api for some reason
1- The Admin Console will try reconnecting to the api and will timeout after 10 seconds if it couldn't (intervals of 1 second). During the "reconnecting" phase, the progress bar will be left as is and a message indicating connectivity issues will be shown to the user.
2- If reconnected successfully and there was no data loss in the api, uploading will resume normally and the progress bar will continue from where it left off.
3- If reconnected successfully but there was some data loss in the api, uploading will be automatically restarted and the progress bar will be reset to 0.

## Another improvement included in this PR
When resuming uploads, the progress bar will instantly be set to where it left off instead of showing small incremental changes.

### DIFF
--- a/kotsadm/pkg/apiserver/server.go
+++ b/kotsadm/pkg/apiserver/server.go
@@ -138,6 +138,7 @@ func Start() {
 
 	// Airgap
 	sessionAuthRouter.Path("/api/v1/app/airgap").Methods("POST", "PUT").HandlerFunc(handlers.UploadAirgapBundle) // Backwards compatibility route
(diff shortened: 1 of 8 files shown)


GraysonNull (approved):

divolgin (approved):

markpundsack:
@salahalsaleh What's the user experience here?

salahalsaleh:
@markpundsack I've updated the PR description to explain the behavior.

PR #1120, opened 2020-09-22
return GIT_SHA as kots version if empty

salahalsaleh:


divolgin (approved):

kotsadm/Makefile
```diff
@@ -24,6 +24,10 @@ define GIT_SHA
 endef
 endif
 
+ifeq ($(KOTS_VERSION),)
+KOTS_VERSION := $(GIT_SHA)
```
MikaelSmith, line 28: Did this fix your problem? I suspect your GitHub Actions are actually failing to fetch tags when this runs. https://github.com/replicatedhq/kots/blob/master/.github/workflows/deploy.yaml#L14-L15 is one option for fixing that, or you can set `fetch-depth: 0` on the checkout action as noted in https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches.

PR #1121, opened 2020-09-22
Exec into ekco to purge node

areed:
Ekco can handle cleanup of both Rook 1.0 and 1.4 and we don't want to duplicate that code in kots.

Also implement isConnected for nodes and fix a panic.


kotsadm/pkg/kurl/kurl_nodes.go
```diff
@@ -138,20 +143,16 @@ func getNodeMetrics(nodeIP string) (*statsv1alpha1.Summary, error) {
...
-					Certificates:       []tls.Certificate{cert},
-					InsecureSkipVerify: true,
-				},
+		client.Transport = &http.Transport{
+			TLSClientConfig: &tls.Config{
+				Certificates:       []tls.Certificate{cert},
+				InsecureSkipVerify: true,
 			},
 		}
-		r, err = client.Get(fmt.Sprintf("https://%s:10250/stats/summary", nodeIP))
```
areed, line 149: When this returns an error `r` is nil and only the err in this script gets overwritten. This was causing the `defer r.Body.Close()` call on line 158 below to panic.

salahalsaleh (approved):

salahalsaleh (approved):

PR #1110, opened 2020-09-17
Slim kurl-proxy and use NODE_PORT

MikaelSmith:
The environment variable `NODE_PORT` is already defined and configured, but it was never used. Instead the port `8800` was hard-coded. Uses `NODE_PORT` to set the listening port, defaulting to `8800`.

Uses multi-stage build to only ship the `kurl_proxy` binary. Omits the symbol table, debug information, and DWARF table with linker flags.

PR #1122, opened 2020-09-23
Update 'go build' link flags

MikaelSmith:
Reverts Dockerfile changes from https://github.com/replicatedhq/kots/pull/1110 because they didn't apply to the production Dockerfile.

Adds Makefile.build to set common build variables such as SHELL, GOPROXY, and LDFLAGS. Updates Makefiles to include it. Adds `-s -w` to all `go build` invocations for smaller binaries by omitting the symbol table, debug information, and DWARF table.

### DIFF
--- a/Makefile.build
+++ b/Makefile.build
@@ -1,0 +1,34 @@
+export GO111MODULE=on
+export GOPROXY=https://proxy.golang.org
+export CGO_ENABLED=0
+
+SHELL := /bin/bash -o pipefail
+VERSION_PACKAGE = github.com/replicatedhq/kots/pkg/version
+VERSION ?=`git describe --tags --dirty`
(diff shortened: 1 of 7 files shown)


MikaelSmith:
I didn't try to make any new Dockerfile changes because there seem to be consistent choices about what's included (debian slim, curl, ca-certificates, git) that I don't understand.

Makefile.build
```diff
@@ -0,0 +1,34 @@
+export GO111MODULE=on
+export GOPROXY=https://proxy.golang.org
+export CGO_ENABLED=0
```
MikaelSmith, line 3: This was potentially new rather than a refactor, in case anything starts breaking.

PR #1131, opened 2020-09-24
report deployed/downstream app version as well

salahalsaleh:


pkg/upstream/replicated.go
```diff
@@ -428,13 +428,16 @@ func listPendingChannelReleases(replicatedUpstream *ReplicatedUpstream, license
...
-	sequence := cursor.Cursor
-	if license.Spec.ChannelName != cursor.ChannelName {
+	sequence := currentCursor.Cursor
+	if license.Spec.ChannelName != currentCursor.ChannelName {
 		sequence = ""
 	}
 
 	urlValues := url.Values{}
 	urlValues.Set("channelSequence", sequence)
+	urlValues.Add("channelName", currentCursor.ChannelName)
```
salahalsaleh, line 438: should we get this from the license instead? in case the customer synced the license before checking for updates, this could be different.

divolgin: i think this should come from license.  this would make `channelName ` consistent with `channelSequence` I think

divolgin (approved):

PR #1132, opened 2020-09-24
docker build can't access files outside of context

divolgin:


salahalsaleh (approved):

salahalsaleh (changes requested):
handle importing Makefile.build in kotsadm and kots root dir

salahalsaleh (approved):

PR #1135, opened 2020-09-24
Make sure common annotations are there on initial app install

divolgin:


pkg/midstream/write.go
```diff
@@ -58,6 +58,11 @@ func (m *Midstream) WriteMidstream(options WriteOptions) error {
 		return errors.Wrap(err, "failed to write patches")
 	}
 
+	if m.Kustomization.CommonAnnotations == nil {
+		m.Kustomization.CommonAnnotations = make(map[string]string)
+	}
+	m.Kustomization.CommonAnnotations["kots.io/app-slug"] = options.AppSlug
+
```
emosbaugh, line 65: maybe add a comment that merge kustomization is ignored on headless install?

emosbaugh (approved):

pkg/midstream/write.go
divolgin: this is actually happens if initial config screen is bypassed.  so an app with no required configuration will also have this problem.

PR #1136, opened 2020-09-25
use channel id instead of name

salahalsaleh:


pkg/upstream/replicated.go
```diff
@@ -71,7 +72,7 @@ type ChannelRelease struct {
 }
 
 func (this ReplicatedCursor) Equal(other ReplicatedCursor) bool {
-	return this.ChannelName == other.ChannelName && this.Cursor == other.Cursor
+	return this.ChannelID == other.ChannelID && this.ChannelName == other.ChannelName && this.Cursor == other.Cursor
```
divolgin, line 75: I think this is a candidate for breaking backwards compatibility.  Maybe `ChannelName` comparison should be used as a fallback if one of channel IDs is not available.

salahalsaleh: i agree, i don't see this function used anywhere in the code, only in a test file, but i'll update

divolgin (approved):

PR #1139, opened 2020-09-28
fix push images screen

salahalsaleh:


divolgin (approved):

divolgin (approved):

pkg/kotsadm/push_images.go
```diff
@@ -272,9 +272,10 @@ func TagAndPushAppImages(imagesDir string, options types.PushImagesOptions) ([]k
 
 		reportWriter := options.ProgressWriter
 		if options.LogForUI {
-			reportWriter := reportWriterWithProgress(imageFiles, options.ProgressWriter)
-			defer reportWriter.Write([]byte(fmt.Sprintf("+status.flush:\n")))
-			defer reportWriter.Close()
+			wc := reportWriterWithProgress(imageFiles, options.ProgressWriter)
+			reportWriter = wc.(io.Writer)
```
MikaelSmith, line 276: It's not clear that anything at all happened here. Would you mind explaining what this changed?

salahalsaleh: @MikaelSmith A new `reportWriter` variable was created inside this if statement block instead of reassigning the existing `reportWriter` which is used to write the "pushing images" progress messages.

MikaelSmith: Ah I see it now. Thanks.

PR #1147, opened 2020-10-01
Version card/version history redesign

jgruica:
Screen Shot 2020-10-01 at 3 50 21 PM
Screen Shot 2020-10-01 at 3 48 58 PM
Screen Shot 2020-10-01 at 3 49 45 PM
Screen Shot 2020-10-02 at 8 55 44 AM
Screen Shot 2020-10-02 at 8 35 15 AM


kotsadm/web/src/components/modals/UpdateCheckerModal.jsx
```diff
@@ -128,7 +128,7 @@ export default class UpdateCheckerModal extends React.Component {
         className="Modal SmallSize"
       >
         <div className="u-position--relative flex-column u-padding--20">
-          <span className="u-fontSize--largest u-fontWeight--bold u-color--tuna u-marginBottom--15">Configure automatic updates</span>
+          <span className="u-fontSize--largest u-fontWeight--bold u-color--tuna u-marginBottom--15">Automatic updates checks</span>
```
GraysonNull, line 131: I think this should say "Configure automatic update checks"

kotsadm/web/src/components/apps/DashboardCard.jsx
```diff
@@ -168,36 +262,39 @@ export default class DashboardCard extends React.Component {
...
             : showOnlineUI ?
-              <div className="flex alignItems--center">
-                <button className="btn primary blue u-marginTop--10" onClick={isUpdateAvailable ? redirectToDiff : onCheckForUpdates}>{isUpdateAvailable ? "Show Update" : "Check for update"}</button>
-                <span className="icon settings-small-icon u-marginLeft--5 u-cursor--pointer u-marginTop--10" onClick={this.props.showUpdateCheckerModal} data-tip="Configure automatic updates"></span>
-                <ReactTooltip effect="solid" className="replicated-tooltip" />
+              <div className="flex1 flex-column" style={{ flexGrow: 1 }}>
+                {this.renderVersionAvailable(downstream)}
+                <div className="flex alignItems--center">
+                  <button className="btn primary blue u-marginTop--10" onClick={isUpdateAvailable ? redirectToDiff : onCheckForUpdates}>Check for update</button>
+                  <span className="icon settings-small-icon u-marginLeft--5 u-cursor--pointer u-marginTop--10" onClick={this.props.showUpdateCheckerModal} data-tip="Configure automatic updates"></span>
```
GraysonNull, line 290: Let's change this tooltip to be "Configure automatic update checks"

kotsadm/web/src/components/modals/UpdateCheckerModal.jsx
markpundsack: What about something smaller, like "Configure checks"?

markpundsack: If it's only on hover, then we can get away with more verbose description, if that's clearer.

markpundsack: Oh wait, this is in the modal title itself. Yeah, let's go with verbose.

markpundsack:
In the last image, how come the all versions row for the currently deploying version doesn't have a status icon? The app banner does. 

Likewise, how come the older version doesn't show any status at all?

@GraysonNull Thinking about the deploying icon, it looks like a grayed out do-not-enter sign. Can you come up with a better icon indicating ongoing deployment?

GraysonNull:
Yeah agreed, it's not a great icon for that. How about just showing the loading spinner in place of an icon? I think that would make sense to show that there plus would give some good feedback that something is actually happening.

markpundsack:
Oh interesting, you mean an animated spinner. That's kinda brilliant.

GraysonNull:
Yeah exactly, seems kinda like a perfect use for it.

jgruica:
@GraysonNull updated code and screenshots with deploying spinner

GraysonNull (approved):

markpundsack (approved):

PR #1156, opened 2020-10-03
send reporting info on install as well

salahalsaleh:


kotsadm/pkg/reporting/reporting.go
```diff
@@ -0,0 +1,121 @@
...
+	if di != nil {
+		r.DownstreamCursor = di.DownstreamCursor
+		r.DownstreamChannelID = di.DownstreamChannelID
+		r.DownstreamChannelName = di.DownstreamChannelName
+	}
+
+	// get kubernetes cluster version
+	k8sVersion, err := getK8sVersion()
+	if err != nil {
+		logger.Error(errors.Wrap(err, "failed to get k8s version"))
```
divolgin, line 45: these should be info level

divolgin (approved):

divolgin (approved):

PR #1159, opened 2020-10-05
Fix missing collectors and image rewriting

divolgin:


kotsadm/pkg/supportbundle/supportbundle.go
```diff
@@ -151,6 +151,13 @@ func CreateRenderedSpec(appID string, sequence int64, origin string, inCluster b
 				Name: "default-supportbundle",
 			},
 		}
+
+		if kotsKinds.Collector != nil {
```
divolgin, line 155: this is needed for support bundles defined like this:
```
apiVersion: troubleshoot.replicated.com/v1beta1
kind: Collector
```

marccampbell (approved):

PR #1168, opened 2020-10-09
Wait for the correct kotsadm pod to start running

divolgin:


pkg/kotsadm/kotsadm.go
```diff
@@ -58,13 +59,28 @@ func getKotsadmYAML(deployOptions types.DeployOptions) (map[string][]byte, error
 func waitForKotsadm(deployOptions *types.DeployOptions, clientset *kubernetes.Clientset) error {
 	start := time.Now()
 
+	// When upgrading, the old pod is still running, so this function will terminate instantly if we don't check version.
+	expectedImage := fmt.Sprintf("%s/kotsadm:%s", kotsadmRegistry(deployOptions.KotsadmOptions), kotsadmTag(deployOptions.KotsadmOptions))
```
marccampbell, line 63: This feels like the wrong approach.  Once we've deployed the new deployment, can't we use a higher order Kubernetes API to monitor the rollout of the update?

marccampbell: @divolgin does `--wait` for condition = ready handle this? I'm guessing there's a LOT more handled in this rolling upgrade, and a lot of edge cases we are missing with this current implementation...

marccampbell (approved):

PR #1171, opened 2020-10-12
Don't leak tmp files

divolgin:


marccampbell:
@divolgin can we add a CodeLingoBot rule to detect not cleaning this up?

marccampbell (approved):

pkg/base/replicated.go
```diff
@@ -166,6 +166,8 @@ func renderReplicated(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (
 		if err != nil {
 			return nil, errors.Wrap(err, "failed to create temp file")
 		}
+		defer os.RemoveAll(tmpFile.Name())
```
divolgin, line 169: @CodeLingoBot capture defer remove after ioutil.TempFile call

kotsadm/pkg/handlers/preflight.go
```diff
@@ -176,14 +206,21 @@ func GetPreflightCommand(w http.ResponseWriter, r *http.Request) {
...
 
-	archivePath, err := store.GetStore().GetAppVersionArchive(foundApp.ID, sequence)
+	archivePath, err := ioutil.TempDir("", "kotsadm")
 	if err != nil {
-		logger.Error(errors.Wrap(err, "failed to get app archive"))
+		logger.Error(errors.Wrap(err, "failed to create temp dir"))
 		w.WriteHeader(http.StatusInternalServerError)
 		return
 	}
 	defer os.RemoveAll(archivePath)
```
divolgin, line 215: @CodeLingoBot capture defer remove after ioutil.TempDir call

PR #1173, opened 2020-10-13
Disaster recovery snapshots proposal

divolgin:


design/disaster-recovery-snapshots.md
```diff
@@ -0,0 +1,117 @@
...
+Similarly, restore spec supports the same `labelSelector` field to select specific resources to restore.
+Using different `labelSelector` values allows to restore application and Admin Console separately.
+
+## Detailed Design
+
+### Backup
+
+- All resources (Admin Console and application) will have an additional label: `kots.io/backup: velero`
+- All application resources will have an additional label: `kots.io/app-slug: <slug>`
+
```
divolgin, line 42: forgot to updating this section... there's one more label that's missing.

design/disaster-recovery-snapshots.md
```diff
@@ -0,0 +1,128 @@
...
+## Goals
+
+1. Use Velero to create Admin Console snapshot.
+1. Make CLI based disaster recovery possible.
+1. Propose a way to enable UI based recovery process.
+
+## Non Goals
+
+1. Improving application deployment process to avoid the `field is immutable` errors.
```
markpundsack, line 15: I don't know what that means.

design/disaster-recovery-snapshots.md
```diff
@@ -0,0 +1,128 @@
...
+1. Performing incremental backups of docker images.
```
markpundsack, line 16: Maybe include a non-goal, for now, of backing up and restoring things outside of the admin console. e.g. for  kURL installation, there's a registry which Kots needs, but isn't backed up. At some point we're going to need to solve that, of course.

design/disaster-recovery-snapshots.md
```diff
@@ -0,0 +1,128 @@
...
+Disaster recovery snapshots must include Admin Console data, and potentially all resources.
+Due to this requirement, `velero.io/exclude-from-backup` label can no longer be used.
+The `includedNamespaces` filter also cannot be used because Admin Console and application can be deployed into the same namespace, which will cause problems during the restore process.
+
+## High-Level Design
+
+Velero backup spec supports the `labelSelector` field.
+All resources, application and Admin Console, can be labeled for inclusion in snapshots.
+Similarly, restore spec supports the same `labelSelector` field to select specific resources to restore.
+Using different `labelSelector` values allows to restore application and Admin Console separately.
```
markpundsack, line 34: This is great!

Is it a non-goal to be able to restore individual apps in a multi-app scenario? The labelSelector might help do that if it were to become a goal.

design/disaster-recovery-snapshots.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Detailed Design
+
+### Backup
+
+- All resources (Admin Console and application) will have an additional label:
+    - `kots.io/backup: velero`
```
markpundsack, line 41: If you're using `kots.io/backup`, do you really need `velero`? Velero is an implementation detail.

e.g. maybe use `kots.io/backup: kotsadm` instead, and then applications can use `kots.io/backup: application` or even just `kots.io/backup: <app-slug>`. 

Assuming the goal is to be able to differentiate admin console content from application content, that should be sufficient.

Does `matchLabels` only work with a single label? Is that why you want to structure the labels this way?

markpundsack: Do applications ever create more objects that need to be backed up, that Kots wouldn't be tracking?

design/disaster-recovery-snapshots.md
```diff
@@ -0,0 +1,128 @@
...
+### Restore
+
+- Admin Console restore will be created with `labelSelector` set as in the example below:
+
+        apiVersion: velero.io/v1
+        kind: Restore
+        spec:
+          labelSelector:
+            matchLabels:
+              kots.io/kotsadm: "true"
```
markpundsack, line 63: Does that label already exist?

design/disaster-recovery-snapshots.md
```diff
@@ -0,0 +1,128 @@
...
+
+- Application restore will be created with `labelSelector` set as in the example below:
+
+        apiVersion: velero.io/v1
+        kind: Restore
+        spec:
+          labelSelector:
+            matchLabels:
+              kots.io/app-slug: <slug>
```
markpundsack, line 72: I like the ability to select all objects related to an app in other ways, not just for backup/restore. When we add the ability to delete an app, for example, we should be able to make use of the same label (assuming apps don't create new objects that aren't tracked).

markpundsack: It could possibly be used to check if we're about to overwrite an existing object with an application-specific version. Say, if two apps both try to install nginx in the same namespace. Not sure what we'd expect the user to do with that information, but it could at least be a warning.

design/disaster-recovery-snapshots.md
```diff
@@ -0,0 +1,128 @@
...
+
+- Restore can be performed using Velero CLI:
+
+        velero restore create --from-backup <backup name> -l kots.io/backup=velero
```
markpundsack, line 76: Clarify that this would be a full restore. It's different than the Admin Console restore mentioned above. AIUI, the prior example would restore *just* the admin console, but this would restore the admin console and all apps.

design/disaster-recovery-snapshots.md
```diff
@@ -0,0 +1,128 @@
...
+
+        velero restore create --from-backup <backup name> -l kots.io/app-slug=<slug>
```
markpundsack, line 78: If someone would do this, would we lose the version history counter adjustment?

design/disaster-recovery-snapshots.md
markpundsack: Actually, since all snapshotted objects have that `kots.io/backup` label, do you need the specify `-l` at all? A full restore would just be to restore `--from-backup`.

design/disaster-recovery-snapshots.md
```diff
@@ -0,0 +1,128 @@
...
+
+- `kots install --from-backup` command will be implemented as a wrapper for the above commands
+    - `kots restore` is a potential alternative, but the UI-based approached discussed next will also need a `kots install` command.
```
markpundsack, line 81: Semantically, I do think of restoring as different than installing.

design/disaster-recovery-snapshots.md
```diff
@@ -0,0 +1,128 @@
...
+
+
+### Potential blockers
+
+When adding new labels, the next deployment can potentially fail with the `field is immutable` errors.
+A workaround is to delete the asset that causes the error prior to deploying the new version.
```
markpundsack, line 87: Do you mean that when we add the `kots.io/backup` labels, we'll have to delete objects and recreate them? Is that risky? It must cause downtime.

design/disaster-recovery-snapshots.md
```diff
@@ -0,0 +1,128 @@
...
+
+### UI-based restores
+
+In order to allow a UI-based disaster recovery flow, Admin Console needs to be installed first.
```
markpundsack, line 91: That assumes, of course, that the only UI is the Admin Console. We could always make a different UI if that was important.

design/disaster-recovery-snapshots.md
```diff
@@ -0,0 +1,128 @@
...
+When restoring the Admin Console resources, `kotsadm` pod will be restarted leaving UI in a potentially bad state.
+For example, once data is restored, the logged in user's session will be invalidated.
+Also receiving restore status will be impossible if the API is not available or session is invalid.
```
markpundsack, line 94: I feel like there's got to be some prior art here. Like some other process, outside of kotsadm, that manages the restore. Possibly with the hypothetical restore UI.

design/disaster-recovery-snapshots.md
```diff
@@ -0,0 +1,128 @@
...
+
+In order to avoid the above problems, the following can be implemented:
+
+
+1. Only Postgres data, configmaps, and secrets will be backed up (i.e. pods will not have the `kots.io/backup=velero` label).
+1. There will be a sidecar pod created (`kotsadm-backup`) that will have the `kots.io/backup=velero` label
+    - `kotsadm-backup` will run pre-snapshot hooks to backup Postgres data.
+    - `kotsadm-backup` will run post-restore hook to restore Postgres data
+    - `kotsadm-backup` will be deleted once backup or restore has completed.
+1. `kots install` command will allow installing Admin Console only.
```
markpundsack, line 105: We could augment `kots` with commands to list available snapshots, both for admin console and apps. Make a good enough CLI interface, and the need for a web UI is reduced.

design/disaster-recovery-snapshots.md
```diff
@@ -0,0 +1,128 @@
...
+
+
+Because Admin Console pods will not be restored, we need to ensure that the version installed is compatible with the the data stored in the snapshot.
+To accomplish this, the following will be implemented:
+
+
+1. All Velero backup objects will have the `kots.io/version` label.
+1. During the restore, this label will be used to filter out backups that have non-matching versions.
+1. During the restore, users will be informed that a different version of Admin Console needed if they want to restore a different, incompatible backup.
```
markpundsack, line 114: This feels like it would be cumbersome. We don't make it easy to install older versions of Kots Admin Console today. Maybe we're expecting that someone will have the CLI at the version they need, but a user could have multiple applications, each at different versions, and it's painful to switch between versions. It would be far preferable to have the CLI restore process be able to handle this. At worst, by being able to install older versions, at best by making it not matter.

design/disaster-recovery-snapshots.md
```diff
@@ -0,0 +1,128 @@
...
+
+
+### kURL registry
+
+Airgapped embedded instances will also have images stored in local kURL registry.
+These images can be backed up and restored using the `kotsadm-backup` sidecar pod.
```
markpundsack, line 120: That's great! Is that the only thing kURL needs to backup/restore? What about data stored in PVCs?

design/disaster-recovery-snapshots.md
markpundsack: Should we make it explicit that restoring individual apps is a non-goal?

If it *is* a goal, then the big challenge I have is that you restore admin console and all apps will show up interface, even if they're non-existent. 

Related, should we make it a requirement that when restoring the admin console, you get all apps? At most with a user guide for people that really know what they're doing if somehow admin console gets corrupt, but their app isn't and they don't want to restore the app. But certainly by default, restoring the admin console should do a full restore.

design/disaster-recovery-snapshots.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Alternatives Considered
+
+Storing Admin Console in its own separate snapshot has been considered.
+
+## Security Considerations
+
+None
```
markpundsack, line 128: Some questions:

I'd like some detail on how we expect snapshots to be configured. Should they be configured at the instance level or at the app level? I believe today it's a big half-and-half. If someone triggers a backup of one app, should all apps be backed up? If so, treating this as an instance-level snapshot may make more sense. Just like most backup systems, you back up the whole thing, even if you only want to restore one thing later. 

In the future, for a really large enterprise user with 10,000 applications, that may be a big cumbersome. But for all practical purposes today, it would be fine.

But being able to trigger an app-level snapshot is nice as it's common to do right before making a major change.

So, if we perpetuate a prior design where triggering an application-specific snapshot also snapshots the admin console, but none of the other apps, what would we do if someone restores that admin console snapshot? Would we only restore the one app saved with it? Or would we dig for a prior snapshot that restores the rest of the apps? Even if that all works, it would be very hard for end-users to understand. It may also be dangerous in that Admin Console will show a record of a certain version running which may be different than was restored. Even if we do the magic adjustment to point to the right place, it might be confusing to the user why their app lost a version. For disaster recovery, that may be better than nothing. And it depends how old the backup us.

Maybe we shouldn't let people disaster recovery restore from a partial snapshot. Treat that like the intent it had, which was an app-level rollback mechanism. Maybe even call them different things. Two different features, app snapshots and cluster backups (although that name is misleading...).

Or, just have every snapshot trigger cluster-wide snapshot.

For registry backup, should we support restoring to a different place? One of the use cases people want is to use backup/restore not just to recover to a new cluster, but to recover to a different *type* of cluster. Maybe start with a single-node kURL cluster, then restore to a *real* cluster. That could be a multi-node HA cluster, but it may be EKS with the registry as ECR.

Does this help us at all towards being able to move apps from one multi-app cluster to another? e.g. After restore, trigger an internal delete of the records for the apps not copied over?

markpundsack:
This is a great start. I'd just like to see details about what we trigger for snapshots, when, and where that lives (app-level or instance-level) and the interaction with app-level snapshots and restores which still has value.

design/disaster-recovery-snapshots.md
salahalsaleh: yes

design/disaster-recovery-snapshots.md
markpundsack: Minio
Prometheus
Logs?

PR #1184, opened 2020-10-21
distaster recovery

salahalsaleh:
### To create an instance backup:
```
kubectl kots backup create --namespace <namespace>
```

### To restore an instance backup:
Note: backup command will print out the backup name.
```
kubectl kots restore create --from-backup <backup-name>
```

### To list backups
```
kubectl kots backup ls
```

### To list restores
```
kubectl kots restore ls
```


cmd/kots/cli/backup-create.go
```diff
@@ -0,0 +1,38 @@
...
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			v := viper.GetViper()
+
+			instanceBackupOptions := snapshot.InstanceBackupOptions{
+				Namespace:             v.GetString("namespace"),
+				KubernetesConfigFlags: kubernetesConfigFlags,
+			}
+			if err := snapshot.InstanceBackup(instanceBackupOptions); err != nil {
+				return errors.Cause(err)
```
divolgin, line 28: The problem with using `Cause` here is that it could return something like "EOF" without any other context.  If there are some human readable errors, we could create a type of those and strip the error stack for all other types.

cmd/kots/cli/backup-list.go
```diff
@@ -0,0 +1,34 @@
...
+		Long:          ``,
+		SilenceUsage:  true,
+		SilenceErrors: false,
+		PreRun: func(cmd *cobra.Command, args []string) {
+			viper.BindPFlags(cmd.Flags())
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			backups, err := snapshot.ListBackups()
+			if err != nil {
+				return errors.Cause(err)
```
divolgin, line 24: Same here.

kotsadm/pkg/handlers/backup.go
```diff
@@ -148,3 +148,31 @@ func DeleteKotsadmBackup(w http.ResponseWriter, r *http.Request) {
...
+}
+
+type CreateInstanceBackupResponse struct {
+	Success    bool   `json:"success"`
+	BackupName string `json:"backupName,omitempty"`
+	Error      string `json:"error,omitempty"`
+}
+
+func CreateInstanceBackup(w http.ResponseWriter, r *http.Request) {
+	createInstanceBackupResponse := CreateInstanceBackupResponse{
```
divolgin, line 162: Needs token auth check, see `requireValidKOTSToken`

kotsadm/pkg/k8s/kotsadm.go
```diff
@@ -0,0 +1,36 @@
...
+
+	kotsadmDeployment, err := client.AppsV1().Deployments(namespace).Get(context.TODO(), "kotsadm", metav1.GetOptions{})
+	if err != nil {
+		return "", errors.Wrap(err, "failed to get kotsadm deployment")
+	}
+
+	apiContainerIndex := -1
+	for i, container := range kotsadmDeployment.Spec.Template.Spec.Containers {
+		if container.Name == "kotsadm" {
+			apiContainerIndex = i
```
divolgin, line 24: or `return container.Image`. i'm ok with this :)

kotsadm/pkg/snapshot/backup.go
```diff
@@ -168,43 +207,43 @@ func createAdminConsoleBackup(ctx context.Context, isScheduled bool) error {
 	veleroBackup := &velerov1.Backup{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:         "",
-			GenerateName: "kotsadm-",
+			GenerateName: "instance-",
```
divolgin, line 210: I'm assuming this naming was discussed with @markpundsack

pkg/disasterrecovery/disasterrecovery.go
```diff
@@ -0,0 +1,163 @@
...
+
+		absFilename := filepath.Join(archiveDir, "overlays", "midstream", LabelTransformerFileName)
+
+		if err := ioutil.WriteFile(absFilename, drLabelTransformerYAML, 0644); err != nil {
+			return errors.Wrap(err, "failed to write disaster recovery label transformer yaml file")
+		}
+
+		k.Transformers = append(k.Transformers, LabelTransformerFileName)
+
+		if err := k8sutil.WriteKustomizationToFile(*k, filepath.Join(archiveDir, "overlays", "midstream", "kustomization.yaml")); err != nil {
```
divolgin, line 157: This belongs in the midstream package.  Is it possible to move it there?

kotsadm/pkg/snapshot/backup.go
salahalsaleh: Hmmm, It was not actually. @markpundsack Any objections/suggestions?

markpundsack: This is for the prefix of the backup name? I guess that works. We've been talking about instance vs app snapshots. We've also talked about full vs partial backups. It's no longer only Kots Admin Console, so it makes sense not to call it that. But another option could be simply `kots-`. Instance is such an ambiguous term. But, I don't have strong objections.

markpundsack: BTW, do we expect that if someone has backups of different Kots instances into the same S3 bucket, they use different paths? When showing a list of backups, do we show all backups in that bucket/path? Is the name meaningful at all for identifying Kots backups?

markpundsack: Or (and I hope this isn't the case) does there need to be a record of each backup in the cluster?

salahalsaleh: Currently, the name is not meaningful at all to identify different instances of kots backups. However, we do have annotations added to the Backup object to help identify that.

salahalsaleh: Might be a good idea to have those as labels instead of annotations

pkg/disasterrecovery/disasterrecovery.go
salahalsaleh: @divolgin I assume you're referring to the whole function, not just this step, right?

divolgin (approved):

divolgin (approved):

PR #1193, opened 2020-10-29
Don't use ConfigMap for airgap app spec

divolgin:


kotsadm/pkg/automation/automation.go
```diff
@@ -213,6 +393,9 @@ func getAirgapData(clientset kubernetes.Interface, license *kotsv1beta1.License)
 	result := map[string][]byte{}
 	for _, configMap := range configMaps.Items {
 		for key, value := range configMap.Data {
+			if key == "wait-for-airgap-app" { // do this better
```
marccampbell, line 396: Should we now? Would it be less brittle if we just assumed that the absence of the key/configmap means "wait"?

divolgin: this is just skipping a bool value because bse64 decode works on those too....  it's file, file, file, bool!

marccampbell (approved):

PR #1196, opened 2020-10-29
Set DISABLE_OUTBOUND_CONNECTIONS on --airgap flag

zee-hussain:
Sets `DISABLE_OUTBOUND_CONNECTIONS` on `--airgap` flag and `--kotsadm-registry` instead of just `--kotsadm-registry`

### DIFF
--- a/cmd/kots/cli/install.go
+++ b/cmd/kots/cli/install.go
@@ -166,7 +166,7 @@ func InstallCmd() *cobra.Command {
 				ApplicationMetadata:       applicationMetadata,
 				License:                   license,
 				ConfigValues:              configValues,
-				Airgap:                    v.GetBool("airgap"),
+				Airgap:                    isAirgap,
 				ProgressWriter:            os.Stdout,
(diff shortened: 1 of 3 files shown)


cmd/kots/cli/install.go
```diff
@@ -166,6 +166,7 @@ func InstallCmd() *cobra.Command {
 				ApplicationMetadata:       applicationMetadata,
 				License:                   license,
 				ConfigValues:              configValues,
+				Airgap:                    v.GetBool("airgap"),
```
salahalsaleh, line 169: There is already a variable `IsAirgap` defined above that also takes under consideration the `airgap-bundle` flag, and it would be a good idea to rename `Airgap` field in the DeployOptions to `IsAirgap`
```suggestion
				Airgap:                    isAirgap,
```

salahalsaleh (changes requested):
Thanks for the PR! I just a have a small suggestion.

laverya (approved):

salahalsaleh (approved):

PR #1202, opened 2020-10-30
disaster recovery scheduled snapshots

salahalsaleh:


kotsadm/pkg/downstream/types/types.go
```diff
@@ -11,6 +11,8 @@ type Downstream struct {
 	ClusterSlug     string `json:"slug"`
 	Name            string `json:"name"`
 	CurrentSequence int64  `json:"currentSequence"`
+	SnapshotSchedule      string         `json:"snapshotSchedule,omitempty"`
```
divolgin, line 14: looks like this needs a `go fmt`

kotsadm/pkg/k8s/kotsadm.go
```diff
@@ -13,6 +13,8 @@ func FindKotsadmImage(namespace string) (string, error) {
 		return "", errors.Wrap(err, "failed to get k8s client set")
 	}
 
+	namespace = "default" // TODOOOOO: REMOVE THIS
```
divolgin, line 16: do the to do!

divolgin (approved):

PR #1215, opened 2020-11-04
disaster recovery kotsadm applications restore

salahalsaleh:


kotsadm/pkg/snapshot/backup.go
```diff
@@ -180,6 +182,7 @@ func CreateInstanceBackup(ctx context.Context, cluster *downstreamtypes.Downstre
 			// no version is deployed for this app yet
 			continue
 		}
+		appsSequences[a.Slug] = strconv.FormatInt(parentSequence, 10)
```
divolgin, line 185: It would be better not to use `FormatInt` here, and just make it a `map[string]int`.

kotsadm/pkg/socketservice/socketservice.go
```diff
@@ -515,9 +531,29 @@ func checkRestoreComplete(a *apptypes.App, restore *velerov1.Restore) error {
...
+			if !ok || s == "" {
+				return errors.New("instance backup is missing sequence annotation")
+			}
+			sequenceStr = s
+		} else {
+			s, ok := backupAnnotations["kots.io/app-sequence"]
+			if !ok || s == "" {
+				return errors.New("backup is missing sequence annotation")
+			}
+			sequenceStr = s
```
divolgin, line 556: `ParseInt` can go inside this `else` block if new annotation is `map[string]int`

pkg/snapshot/backup.go
```diff
@@ -111,8 +115,21 @@ func CreateInstanceBackup(options CreateInstanceBackupOptions) error {
 		return errors.New(backupResponse.Error)
 	}
 
+	// wait for backup to complete
```
divolgin, line 118: Can we make this optional with a `wait` flag defaulted to `false`.

pkg/snapshot/restore.go
```diff
@@ -129,8 +135,54 @@ func CreateInstanceRestore(options CreateInstanceRestoreOptions) (*velerov1.Rest
...
+
+	// initiate kotsadm applications restore
+	err = initiateKotsadmApplicationsRestore(options.BackupName, kotsadmNamespace, kotsadmPodName, options.KubernetesConfigFlags, log)
+	if err != nil {
+		log.FinishSpinnerWithError()
+		return nil, errors.Wrap(err, "failed to restore kotsadm applications")
+	}
+
+	// wait for applications restore to finish
+	err = waitForKotsadmApplicationsRestore(options.BackupName, kotsadmNamespace, kotsadmPodName, options.KubernetesConfigFlags, log)
```
divolgin, line 174: can this be optional as well?

kotsadm/pkg/socketservice/socketservice.go
salahalsaleh: i had it that way, then changed it, dunno why :). will update

divolgin (approved):

divolgin (approved):

PR #1220, opened 2020-11-05
sync kotsadm-s3 secret on kurl restores

salahalsaleh:


pkg/kotsutil/kurl.go
```diff
@@ -59,3 +62,63 @@ func GetKurlRegistryCreds() (hostname string, username string, password string,
...
+
+		_, err = clientset.CoreV1().Secrets("default").Update(context.TODO(), s3Secret, metav1.UpdateOptions{})
+		if err != nil {
+			return errors.Wrap(err, "failed to update s3 secret from minio secret")
+		}
+
+		return nil
+	}
+
+	return errors.New("unsupported object store")
```
divolgin, line 123: does endpoint have any secrets in it?  could we add it to the error?

divolgin (approved):

pkg/kotsutil/kurl.go
salahalsaleh: it does not, good idea

divolgin (approved):

PR #1225, opened 2020-11-06
add 'kubectl kots gitops-info -n ns' to get gitops providers and keypairs

laverya:
this is a hidden command

This also moves the 'gitops' package into kots from kotsadm in order to be accessible to the CLI


pkg/gitops/gitops.go
```diff
@@ -677,6 +683,75 @@ func CreateGitOpsCommit(gitOpsConfig *GitOpsConfig, appSlug string, appName stri
...
+
+		newIndex, err := strconv.ParseInt(splitKey[1], 10, 64)
+		if err != nil {
+			continue
+		}
+		indexes[newIndex] = struct{}{}
+	}
+
+	encryptionKeySecret, err := clientset.CoreV1().Secrets(namespace).Get(context.TODO(), "kotsadm-encryption", metav1.GetOptions{})
+	if kuberneteserrors.IsNotFound(err) {
```
salahalsaleh, line 721: i think this should fail if the encryption key is not found

salahalsaleh:
why move `pkg/gitops/clone.go`, `pkg/gitops/key.go` and `pkg/gitops/types/gitops_interface.go`? i think they should stay in kotsadm as the functions/types defined there are all related to kotsadm functionality, not kots.

cmd/kots/cli/gitops-info.go
```diff
@@ -0,0 +1,50 @@
...
+			v := viper.GetViper()
+			log := logger.NewLogger()
+
+			namespace := v.GetString("namespace")
+			if namespace == "" {
+				log.Info("A namespace is required")
+				os.Exit(1)
+			}
+
+			keypairs, err := gitops.GetGitopsKeypairs(namespace)
```
salahalsaleh, line 33: i think this should be an API call instead of moving the kotsadm packages to kots

salahalsaleh, line 33: i think this should be an API call instead of moving the kotsadm packages to kots

salahalsaleh (changes requested):
i think this should be an API call instead of moving the kotsadm packages to kots

AmberAlston:
@laverya should this PR still be open?

PR #1226, opened 2020-11-06
Redeploy on restore

salahalsaleh:
Restoring application backups in the old format will cause the disaster recovery labels to be removed. This will redeploy the restored version upon restore which will make sure the labels are present.


kotsadm/pkg/socketservice/socketservice.go
```diff
@@ -680,3 +678,24 @@ func (s *SocketService) undeployApp(a *apptypes.App, d *downstreamtypes.Downstre
...
 	return nil
 }
+
+// RedeployAppVersion will force trigger a redeploy of the app version, even if it's currently deployed
+// if clusterSocket is nil, a redeploy to all the cluster sockets (downstreams - which theoratically should always be 1) will be triggered
+func RedeployAppVersion(appID string, sequence int64, clusterSocket *ClusterSocket) error {
+	socketMtx.Lock()
+	defer socketMtx.Unlock()
+
+	if err := version.DeployVersion(appID, sequence); err != nil {
```
divolgin, line 688: Does this call need to be inside the locked section?

salahalsaleh: hmmm, no it doesn't _need_ to. i'll update

divolgin (approved):

PR #1233, opened 2020-11-09
ability to redeploy failed deployment

jgruica:
Screen Shot 2020-11-09 at 1 42 20 PM

### DIFF
--- a/kotsadm/pkg/handlers/redeploy.go
+++ b/kotsadm/pkg/handlers/redeploy.go
@@ -1,0 +1,58 @@
+package handlers
+
+import (
+	"net/http"
+	"strconv"
+
+	"github.com/gorilla/mux"
+	"github.com/pkg/errors"
+	"github.com/replicatedhq/kots/kotsadm/pkg/downstream"
+	"github.com/replicatedhq/kots/kotsadm/pkg/logger"
+	"github.com/replicatedhq/kots/kotsadm/pkg/store"
+	"github.com/replicatedhq/kots/kotsadm/pkg/socketservice"
+)
+
+func RedeployAppVersion(w http.ResponseWriter, r *http.Request) {
+	appSlug := mux.Vars(r)["appSlug"]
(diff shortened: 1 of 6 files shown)


kotsadm/pkg/handlers/redeploy.go
```diff
@@ -0,0 +1,58 @@
...
+		return
+	}
+
+	if err := socketservice.RedeployAppVersion(a.ID, int64(sequence), nil); err != nil {
+		logger.Error(err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	JSON(w, 204, "")
```
salahalsaleh, line 56:
```suggestion
	JSON(w, http.StatusNoContent, "")
```

salahalsaleh (approved):

kotsadm/web/src/components/apps/AppVersionHistoryRow.jsx
```diff
@@ -42,29 +42,35 @@ function renderVersionAction(version, nothingToCommitDiff, app, history, deployV
...
           {needsConfiguration ?
             "Configure" :
             downstream.currentVersion?.sequence == undefined ?
               "Deploy" :
-              isRollback ?
-                "Rollback" :
-                isCurrentVersion ?
-                  "Deployed" :
-                  "Deploy"
+              isRedeploy ?
```
GraysonNull, line 65: Can we set this label a different way? This is really hard to grok when reading.

GraysonNull (changes requested):

kotsadm/web/src/components/apps/AppVersionHistoryRow.jsx
jgruica: Open for ideas :) 
I thought moving this logic in separate function and calling it here but that's not changing the label. It can just clean up these ternary operators and make code cleaner.

GraysonNull (approved):

PR #1266, opened 2020-11-19
Preserve app snapshot label selectors

divolgin:


kotsadm/pkg/snapshot/backup.go
```diff
@@ -736,3 +767,15 @@ func downloadBackupLogs(veleroNamespace, backupName string) ([]types.SnapshotErr
 	errs, warnings, execs, err := parseLogs(gzipReader)
 	return errs, warnings, execs, err
 }
+
+func mergeLabelSelector(kots metav1.LabelSelector, app metav1.LabelSelector) (metav1.LabelSelector, error) {
+	for k, v := range app.MatchLabels {
+		if _, ok := kots.MatchLabels[k]; ok {
+			return metav1.LabelSelector{}, errors.Errorf("application label %s is already defined by kotsadm", k)
```
salahalsaleh, line 774: shouldn't this just `continue`?

divolgin: but how would we resolve this?  I guess we can check if values are the same, but what if they are not?

salahalsaleh (approved):

PR #1273, opened 2020-11-20
Update messaging when app cannot be removed

divolgin:
```
$ kubectl kots remove myapp-1 -n default
  • Removing application myapp-1
Error: failed to remove app, command may be retried with --force flag: application myapp-1 is deployed and cannot be removed
```

### DIFF
--- a/cmd/kots/cli/remove.go
+++ b/cmd/kots/cli/remove.go
@@ -23,7 +23,7 @@ func RemoveCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:           "remove [slug]",
 		Short:         "Remove an application from console",
-		Long:          `Remove application identified by slug from Admin Console.`,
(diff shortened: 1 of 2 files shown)


markpundsack:
Hmm... not quite. 

As a user, here's what I need to know. Why did the command fail, what can I do about it, and why would that be different?

e.g. Remove only removes references to an application and this application has already been deployed. Use `--force` to remove references anyway, leaving the app deployed.

Maybe "references" is the wrong word there. Could say "configuration". Or "removes Admin Console configuration for an application".

divolgin:
```
$ kubectl kots remove myapp-1 -n default
  • Removing application myapp-1 reference from Admin Console
Error: Application is already deployed. Re-run the command with --force flag to remove application reference anyway.
```

markpundsack (approved):

PR #1289, opened 2020-11-30
treat image pull 429s the same as unauthorized

laverya:


laverya:
Downside - I believe this requires the vendor to have configured a docker hub image pull account

laverya:
We may actually want to do the opposite and treat 429s the same as 200s

AmberAlston:
@divolgin bump; old PR still awaiting review

laverya:
Coming back to this - treating 429s (ratelimited) the same as unauthorized would cause kotsadm to configure those images to use the replicated image pull proxy, instead of failing the install as happens currently. This would work _if_ the vendor had configured a docker hub account for that kots app, but not otherwise.

The alternative, treating 429s the same as a 200, would allow installs but then (presumably) result in those pods ending up in imagePullBackoff until someone fixed it, but that is something that the end user can (theoretically) do themselves.

emosbaugh:
Closing this in favor of designing a solution to deal with docker hub rate limiting.

PR #1303, opened 2020-12-03
IDP config redact

emosbaugh:


kotsadm/pkg/handlers/identity.go
```diff
@@ -174,6 +188,15 @@ func getDexConnectorInfo(request ConfigureIdentityServiceRequest) (*DexConnector
 			},
 		}
 
+		// un-redact
+		if c.ClientSecret == "" {
```
salahalsaleh, line 197: this should use a mask i think instead of an empty string

salahalsaleh (approved):

PR #1306, opened 2020-12-04
Session issued at and various fixes

emosbaugh:


kotsadm/pkg/store/ocistore/session_store.go
```diff
@@ -88,6 +83,11 @@ func (s OCIStore) GetSession(id string) (*sessiontypes.Session, error) {
 		return nil, errors.Wrap(err, "failed to unmarshal session")
 	}
 
+	// sessions created before this change will not have IssuedAt
+	if session.IssuedAt.IsZero() {
+		session.IssuedAt = session.ExpiresAt.AddDate(0, 0, -14)
```
salahalsaleh, line 88: this won't have any effect though

kotsadm/pkg/store/s3pg/session_store.go
```diff
@@ -71,6 +68,13 @@ func (s S3PGStore) GetSession(id string) (*sessiontypes.Session, error) {
 		session.Roles = metadata.Roles
 	}
 
+	// sessions created before this change will not have IssuedAt
+	if issuedAt.Valid {
+		session.IssuedAt = issuedAt.Time
+	} else {
+		session.IssuedAt = session.ExpiresAt.AddDate(0, 0, -14)
```
salahalsaleh, line 75: ditto

salahalsaleh (approved):

kotsadm/pkg/store/ocistore/session_store.go
emosbaugh: sure but id like to use this in the future to expire sessions. say an admin reconfigures the identity provider. you can invalidate all sessions with issued at prior to reconfiguration.

PR #1314, opened 2020-12-04
validate identity config connection

salahalsaleh:
Image 2020-12-04 at 3 51 25 PM


pkg/identity/config.go
```diff
@@ -4,9 +4,12 @@ import (
 	"context"
 	"net/http"
```
emosbaugh, line 6: no newline here

emosbaugh (approved):

PR #1333, opened 2020-12-10
confirm modals for deploy, redeploy, and rollback

GraysonNull:


kotsadm/web/src/components/apps/AppVersionHistory.jsx
```diff
@@ -314,28 +317,66 @@ class AppVersionHistory extends Component {
...
+      this.setState({
+        displayConfirmDeploymentModal: true,
+        confirmIsRedeploy: false,
+        confirmIsRollback: true,
+        versionToDeploy: version,
+      });
+    } else {
+      this.setState({
+        displayConfirmDeploymentModal: true,
+        confirmIsRedeploy: true,
```
salahalsaleh, line 364: how about also setting `confirmIsRollback` to `false` here? you could also use string type and merge those into 1 variable

kotsadm/web/src/components/apps/AppVersionHistoryRow.jsx
```diff
@@ -82,7 +82,7 @@ function renderVersionAction(version, nothingToCommitDiff, app, history, deployV
         <button
           className={classNames("btn", { "secondary dark": isRollback, "secondary blue": isSecondaryBtn, "primary blue": isPrimaryButton })}
           disabled={version.status === "deploying"}
-          onClick={() => needsConfiguration ? history.push(`/app/${app.slug}/config/${version.sequence}`) : deployVersion(version)}
+          onClick={() => needsConfiguration ? history.push(`/app/${app.slug}/config/${version.sequence}`) : isRollback ? deployVersion(version, true) : deployVersion(version)}
```
salahalsaleh, line 85:
```suggestion
          onClick={() => needsConfiguration ? history.push(`/app/${app.slug}/config/${version.sequence}`) : deployVersion(version, isRollback)}
```

GraysonNull: I'm not sure it's this simple. the `deployVersion` function is passed to it and can be different. I only want to pass that var if it is rollback

salahalsaleh: ah i see.

salahalsaleh (approved):

PR #1334, opened 2020-12-10
Generic identity deploy for vendor app

emosbaugh:


salahalsaleh (approved):
LGTM. We're gonna have to make configuring Postgres generic as well If we decide to use it for the applications as well.

PR #1341, opened 2020-12-10
Deploy identity service alongside application

emosbaugh:


pkg/midstream/identity.go
```diff
@@ -0,0 +1,175 @@
...
+	if err != nil {
+		return "", errors.Wrap(err, "failed to render identity service")
+	}
+
+	kustomization := kustomizetypes.Kustomization{
+		TypeMeta: kustomizetypes.TypeMeta{
+			APIVersion: "kustomize.config.k8s.io/v1beta1",
+			Kind:       "Kustomization",
+		},
+		CommonLabels: map[string]string{
```
salahalsaleh, line 45: Using common labels will break subsequent deployments when adding additional labels later, because kustomize also tries to inject those labels into `matchLabels` which is immutable. We used label transformers to resolve this for disaster recovery. https://github.com/replicatedhq/kots/blob/359ea2ab3007b19dc69ac02accda824e6d79145f/pkg/disasterrecovery/disasterrecovery.go#L23

emosbaugh: thanks

PR #1344, opened 2020-12-11
add settings tab to configure app level identity provider details

GraysonNull:
Screen Shot 2020-12-11 at 4 39 23 PM

 * my Application's name here is "App Name", it's not a hard coded string or bug, it will show the actual application name.

### DIFF
--- a/kotsadm/web/src/components/apps/AppDetailPage.jsx
+++ b/kotsadm/web/src/components/apps/AppDetailPage.jsx
@@ -21,6 +21,7 @@ import SubNavBar from "@src/components/shared/SubNavBar";
 import SidebarLayout from "../layout/SidebarLayout/SidebarLayout";
 import SideBar from "../shared/SideBar";
 import Loader from "../shared/Loader";
+import AppRegistrySettings from "./AppRegistrySettings";
 import AppSettings from "./AppSettings";
 import AppGitops from "./AppGitops";
 import AppSnapshots from "./AppSnapshots";
@@ -263,7 +264,8 @@ class AppDetailPage extends Component {
       match,
       appsList,
       rootDidInitialAppFetch,
-      appName
+      appName,
+      isIdentityServiceSupported
     } = this.props;
 
     const {
@@ -290,7 +292,6 @@ class AppDetailPage extends Component {
     } else {
       this.state.getAppJob.stop();
     }
-
 
     return (
       <div className="WatchDetailPage--wrapper flex-column flex1 u-overflow--auto">
@@ -338,6 +339,7 @@ class AppDetailPage extends Component {
                     className="flex"
                     activeTab={match.params.tab || "app"}
                     watch={app}
+                    isIdentityServiceSupported={isIdentityServiceSupported}
                     isVeleroInstalled={isVeleroInstalled}
                   />
                   <Switch>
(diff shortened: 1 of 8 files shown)


markpundsack:
Since it's the only setting, maybe we should call it Access; matching the global one? Or would that just add confusion?

"Keep these settings in sync" is ambiguous, implying if I make a change here it will sync bi-directionally. I'd think it would make more sense to simply "use" those settings and hide the fields (or show them but disabled, but really, they shouldn't need to see them.)

You seem to have dropped the connector selection, to pick between openid and other options.

You have "Admin console URL", but that should be the app URL, right? Ideally we'd fetch that from the application.yaml. Likewise, identity address would most likely be derived from that, and hidden. Most customers would append `/dex`. I suppose we need to figure out if we need to allow vendors to specify a different URL.

GraysonNull:
> You seem to have dropped the connector selection, to pick between openid and other options.

It's still there, this is just the top part of the screenshot because that's the only part that's different

GraysonNull:
I'll update the copy on the checkbox to not refer to it as sync and the admin console URL to be "App URL".

There's like a good deal of logic that will need to be put into place here regarding what fields to show/hide or show disabled but my thinking was to just have everything there initially for @emosbaugh to dev with and then myself of Jelena can go in and remove/disable fields as it becomes clear which ones we do and do not need to show here .

kotsadm/web/src/components/apps/AppDetailPage.jsx
```diff
@@ -422,6 +424,14 @@ class AppDetailPage extends Component {
                     <Route exact path="/app/:slug/snapshots/:id/restore" render={() =>
                       <AppSnapshotRestore app={app} />
                     } />
+                    {isIdentityServiceSupported &&
+                      <Route exact path="/app/:slug/settings" render={() =>
+                        <AppSettings
```
salahalsaleh, line 429: don't wanna be picky, but why not call it `AppIdentityServiceSettings` for example?

kotsadm/web/src/config-ui/subNavConfig.js
```diff
@@ -90,5 +90,13 @@ export default [
     displayRule: (watch, isVeleroInstalled) => {
       return watch.allowSnapshots && isVeleroInstalled;
     }
+  },
+  {
+    tabName: "settings",
```
salahalsaleh, line 95: ditto

sgalsaleh (dismissed):

markpundsack:
For the checkbox, we'll also have to think through scenarios where the admin console is set up to use a provider that the app doesn't have support for. One suggestion is to gray out the checkbox with a note (possibly on hover).

kotsadm/web/src/config-ui/subNavConfig.js
GraysonNull: This one is "Settings" because ultimately other items will migrate over to this tab and it won't be just app identity settings.

kotsadm/web/src/components/apps/AppDetailPage.jsx
GraysonNull: Yeah, will update the component name

AmberAlston:
@GraysonNull should this PR still be open?

emosbaugh:
@GraysonNull ping on this

PR #1346, opened 2020-12-12
License in the app table is the current license, not the one in version bundle

divolgin:


pkg/upstream/replicated.go
```diff
@@ -315,30 +313,30 @@ func parseReplicatedURL(u *url.URL) (*ReplicatedUpstream, error) {
 	return &replicatedUpstream, nil
 }
 
-func getSuccessfulHeadResponse(replicatedUpstream *ReplicatedUpstream, license *kotsv1beta1.License) (*kotsv1beta1.License, error) {
```
divolgin, line 318: This change doesn't do anything, but I kept it because this way it's not misleading.

salahalsaleh (approved):

PR #1352, opened 2020-12-15
Add proxy options for dex deployment

divolgin:


kotsadm/pkg/handlers/identity.go
```diff
@@ -203,7 +203,12 @@ func ConfigureIdentityService(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if err := identity.Deploy(r.Context(), clientset, namespace, identityConfig, *ingressConfig, &registryOptions); err != nil {
+	proxyEnv := map[string]string{
+		"HTTP_PROXY":  os.Getenv("HTTP_PROXY"),
+		"HTTPS_PROXY": os.Getenv("HTTPS_PROXY"),
+		"NO_PROXY":    os.Getenv("NO_PROXY"),
```
emosbaugh, line 209: just wondering what is the difference here and why you dont include lowercase?

emosbaugh (approved):

kotsadm/pkg/handlers/identity.go
divolgin: we only set upper case ones in the kotsadm pod

PR #1369, opened 2020-12-16
Identity openshift support

emosbaugh:


cmd/kots/cli/pull.go
```diff
@@ -53,6 +53,7 @@ func PullCmd() *cobra.Command {
 					Username:  v.GetString("registry-username"),
 					Password:  v.GetString("registry-password"),
 				},
+				IsOpenShift:        v.GetBool("is-openshift"),
```
emosbaugh, line 56: is this the best way when its from the cli?

divolgin: is this in case the `IsOpenShift` function doesn't work?

divolgin: i would think you can get this info automatically

divolgin (approved):

PR #1370, opened 2020-12-17
RBAC groups UI

jgruica:
Screen Shot 2020-12-18 at 1 05 37 PM

Screen Shot 2020-12-16 at 4 26 25 PM


kotsadm/web/src/scss/components/forms/input.scss
```diff
@@ -17,6 +17,7 @@
     border: 1px solid #C4C7CA;
     padding: 6px 10px 7px;
     height: 30px;
+    width: 300px;
```
GraysonNull, line 20: This feels scary. Is this going to affect other inputs across the admin console?

salahalsaleh (approved):

markpundsack:
I think we should make these radio selections rather than checkboxes. Or possibly add some vendor config to let them choose.

GraysonNull:
They were radios originally but after a conversation with @salahalsaleh it was made known to me that you can choose multiple roles for app level access. The global access is still radios since you can have only one role. If you can in fact select more than one role for a group at the app level I don't think radios would be the right solution.

markpundsack:
The question is, what should we offer? Just because the data structure technically allows something doesn't make it reasonable. 

Having said that, I suspect anything we come up with will be "wrong" for some vendors, so what we'll probably end up with is letting the vendors specify what they need. But I'd like to start with something reasonable (and easy and intuitive).

Some of the common patterns are:
1. Users can be part of only one role at a time, and roles have a sort order, such that higher roles take priority. e.g. if you match a group for more than one role, you get the highest role.
2. Users can be part of more than one role, but only by virtue of belonging to more than one group. Each group belongs to only one role. You get the union of permissions that each role provides.
3. Each group can have multiple roles, and each user can have multiple groups. Users get the union of permissions.
4. Like the above, but with complicated math to allow for restrictive roles, not just additive roles.
5. Like above, but using fine grained roles and calling them "permissions". So you pick and choose exactly which permissions you want.

One of the reasons I'm avoiding the checkbox concept is that it starts to look a lot like a permissions system rather than a role system. They're both valuable,  but we shouldn't deliver one when we're thinking we're delivering another. The reason companies choose roles is because it's simpler to reason about. Permissions, like AWA IAM users or Salesforce, are notoriously painful to use.

AmberAlston:
Recapping convo from this morning's stand-up - it likely makes the most sense to start with radio buttons and see if multi-select via checkboxes is necessary as we get feedback on the BETA functionality

jgruica:
Closing this pr, made cleaner one: https://github.com/replicatedhq/kots/pull/1396

PR #1391, opened 2020-12-21
don't validate identity service connection in airgapped mode

salahalsaleh:


emosbaugh:
There may have already been discussion around this but I feel like even though it's airgap this still needs to be validated.

PR #1392, opened 2020-12-21
Use correct init containers during upgrades

divolgin:


marccampbell (approved):

kotsadm/pkg/apiserver/bootstrap.go
```diff
@@ -47,11 +47,16 @@ func bootstrapClusterToken() error {
...
 }
 
-// After snapshot restore, we need to create dex db for each app.
-// But the password has to match the one in the app's secret.
-// The secret is restored after kotsadm comes up, but we can get it from
-// the app's archive files.
 func bootstrapIdentity() error {
+	err := identity.CreateDexPostgresDatabase("dex", "dex", os.Getenv("DEX_PGPASSWORD"))
+	if err != nil {
+		return errors.Wrap(err, "failed to create identity db")
```
marccampbell, line 53: so this will fail, even if identity is not enabled.  would it be possible to know if identity is enabled, and then the failure could be ignored?  or does that create too many weird & hard to support scenarios?

salahalsaleh (approved):

kotsadm/pkg/apiserver/bootstrap.go
divolgin: no way to know at this point in the execution... but my thinking is that that's just some DB commands.  if they fail, everything _should_ fail.

PR #1393, opened 2020-12-21
log when not pushing images in airgap cli install

laverya:
if someone runs `kubectl kots install --airgap-bundle ./bundle.airgap --license-file ./mylicense.yaml --namespace mynamespace` or similar _without_ `--kotsadm-registry`, `--registry-username` etc, we should make it clear that the app images will not be pushed.

(I'm not sure if there's any valid scenarios in which you would want to do this, and I figure the "I'm missing args" log line is better than making it an error until we figure that out)

### DIFF
--- a/pkg/kotsadm/main.go
+++ b/pkg/kotsadm/main.go
@@ -164,6 +164,10 @@ func Deploy(deployOptions types.DeployOptions) error {
 	}
 
 	log := logger.NewLogger()
+	if deployOptions.AirgapRootDir != "" && deployOptions.KotsadmOptions.OverrideRegistry == "" {
+		log.Info("not pushing airgapped app images as no registry was provided")
+	}
+
 	if !deployOptions.ExcludeAdminConsole {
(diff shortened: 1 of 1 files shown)


divolgin (approved):

AmberAlston (approved):
Andrew and I chatted about this as part of my approval

laverya:
An example set of logs with this:

```
  • Extracting airgap bundle
Extracting /var/folders/bb/_kf6ckjd2rzd_z0bwytz3rr00000gn/T/kotsadm-airgap004544079/airgap.yaml
...removed for brevity...
Extracting /var/folders/bb/_kf6ckjd2rzd_z0bwytz3rr00000gn/T/kotsadm-airgap004544079/images/docker-archive/quay.io/prometheus/node-exporter/v0.18.1
  • Deploying Admin Console
    not pushing airgapped app images as no registry was provided

    • Creating namespace ✓
    • Waiting for datastore to be ready
...
```

PR #1395, opened 2020-12-22
Config screen messages

jgruica:
Screen Shot 2021-01-04 at 4 10 00 PM
Screen Shot 2021-01-04 at 4 10 07 PM

Screen Shot 2021-01-04 at 4 10 17 PM
Screen Shot 2021-01-04 at 4 10 31 PM
Screen Shot 2021-01-04 at 4 11 50 PM

### DIFF
--- a/kotsadm/web/src/components/apps/AppConfig.jsx
+++ b/kotsadm/web/src/components/apps/AppConfig.jsx
@@ -5,6 +5,7 @@ import PropTypes from "prop-types";
 import classNames from "classnames";
 import Helmet from "react-helmet";
 import debounce from "lodash/debounce";
+import size from "lodash/size";
 import map from "lodash/map";
 import Modal from "react-modal";
 import Loader from "../shared/Loader";
@@ -73,7 +74,7 @@ class AppConfig extends Component {
         const app = await res.json();
         this.setState({ app });
       }
-    } catch(err) {
+    } catch (err) {
       console.log(err);
     }
   }
@@ -131,10 +132,10 @@ class AppConfig extends Component {
   handleSave = async () => {
     this.setState({ savingConfig: true, configError: "" });
 
-    const { fromLicenseFlow, history, match } = this.props;
+    const { fromLicenseFlow, history } = this.props;
     const sequence = this.getSequence();
     const slug = this.getSlug();
-    const createNewVersion = !fromLicenseFlow && match.params.sequence == undefined;
+    const createNewVersion = !fromLicenseFlow;
 
     fetch(`${window.env.API_ENDPOINT}/app/${slug}/config`, {
       method: "PUT",
@@ -232,7 +233,7 @@ class AppConfig extends Component {
         "Accept": "application/json",
       },
       method: "POST",
-      body: JSON.stringify({"configGroups":groups, "sequence": sequence}),
+      body: JSON.stringify({ "configGroups": groups, "sequence": sequence }),
     }).then(async (response) => {
       if (!response.ok) {
         if (response.status == 401) {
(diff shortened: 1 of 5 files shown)


salahalsaleh (approved):

kotsadm/web/src/components/apps/AppConfig.jsx
```diff
@@ -308,7 +355,7 @@ class AppConfig extends Component {
           :
           <div className="ConfigError--wrapper flex-column u-marginTop--20 u-marginBottom--auto alignItems--center">
             {configError && <span className="u-color--chestnut u-marginBottom--20 u-fontWeight--bold">{configError}</span>}
-            <button className="btn secondary blue" disabled={!changed && !fromLicenseFlow} onClick={this.handleSave}>{fromLicenseFlow ? "Continue" : "Save config"}</button>
+            <button className="btn secondary blue" disabled={!changed && !fromLicenseFlow || this.checkIsCurrentOrPastVersion(app)} onClick={this.handleSave}>{fromLicenseFlow ? "Continue" : "Save config"}</button>
```
salahalsaleh, line 358: how does this behave when installing an app?

kotsadm/web/src/components/apps/AppConfig.jsx
```diff
@@ -271,6 +272,50 @@ class AppConfig extends Component {
...
+      return null;
+    }
+  }
+
+  checkIsCurrentOrPastVersion = (app) => {
+    const { match } = this.props;
+    const sequence = parseInt(match.params.sequence);
+    const currentSequence = app?.downstreams[0]?.currentVersion?.parentSequence;
+
+    if (sequence <= currentSequence) {
```
salahalsaleh, line 311: will this work for the initial config screen when installing an app?

jgruica: yes. because at that point we don't have anything in params. url looks like: `/app/:appSlug/config`.
All these changes are visible when viewing config for specific sequence. `/app/:appSlug/config/:sequence"

kotsadm/web/src/components/apps/AppConfig.jsx
jgruica: this will be false. because there's no sequence to compare

kotsadm/web/src/components/apps/AppConfig.jsx
```diff
@@ -271,6 +272,50 @@ class AppConfig extends Component {
...
+        </div>
+      )
+    } else {
+      return null;
+    }
+  }
+
+  checkIsCurrentOrPastVersion = (app) => {
+    const { match } = this.props;
+    const sequence = parseInt(match.params.sequence);
```
salahalsaleh, line 308: when sequence is undefined during installation, won't `parseInt` fail here?

salahalsaleh: how about using
```
const sequence = this.getSequence();
```

jgruica: it will return NaN. and comparison will return false. 
but i can add check just in case to parse it only if params sequence exists

salahalsaleh (approved):

kotsadm/web/src/components/apps/AppVersionHistoryRow.jsx
```diff
@@ -75,12 +84,22 @@ function renderVersionAction(version, nothingToCommitDiff, app, history, deployV
 
   const isSecondaryBtn = isPastVersion || needsConfiguration || isRedeploy && !isRollback;
   const isPrimaryButton = !isSecondaryBtn && !isRedeploy && !isRollback;
+  let tooltipTip;
+  if (checkIsCurrentOrPastVersion(currentDownstreamVersion, version)) {
```
divolgin, line 88: This should allow editing the latest version, not the one currently deployed.  This is current behavior.

kotsadm/web/src/components/apps/AppConfig.jsx
```diff
@@ -131,10 +132,10 @@ class AppConfig extends Component {
   handleSave = async () => {
     this.setState({ savingConfig: true, configError: "" });
 
-    const { fromLicenseFlow, history, match } = this.props;
+    const { fromLicenseFlow, history } = this.props;
     const sequence = this.getSequence();
     const slug = this.getSlug();
-    const createNewVersion = !fromLicenseFlow && match.params.sequence == undefined;
```
divolgin, line 137: not string on JS here, but should it be `===` here and also is checking for `undefined` future-proof?

kotsadm/web/src/components/apps/AppVersionHistoryRow.jsx
```diff
@@ -46,7 +47,15 @@ function deployButtonStatus(downstream, version, app) {
   }
 }
 
-function renderVersionAction(version, nothingToCommitDiff, app, history, deployVersion) {
+function checkIsCurrentOrPastVersion(latestVersion, version) {
```
divolgin, line 50: maybe `isLatestVersion` is a better name and flip the condition to match name, `true => false` and `false => true`

divolgin: or change it to `isEditable`?

divolgin (approved):

PR #1420, opened 2020-12-31
kots get apps command

divolgin:
```
$ kubectl kots get app -n default
SLUG            STATUS
app1            ready
app2            missing
```

```
$ kubectl kots get app -n default -o json
[
    {
        "slug": "app1",
        "state": "missing"
    },
    {
        "slug": "app2",
        "state": "ready"
    }
]
```


marccampbell:
I don't really like the `kubectl kots app ls` syntax. This feels like a Docker-style CLI, but it's in Kubernetes. The Kubernetes convention for CLIs is verb-noun, not noun-verb.

How about making this `kubectl kots get apps` to list the apps, then we could have a `describe app` and other verbs that are commonly used in built-in kubectl commands, lowering the learning curve?

marccampbell (approved):

salahalsaleh (approved):

PR #1428, opened 2021-01-05
Use app encryption key to bootstrap dex DB

divolgin:


kotsadm/pkg/apiserver/bootstrap.go
```diff
@@ -96,7 +93,6 @@ func bootstrapIdentity() error {
 			return errors.Wrapf(err, "failed to get stat identity config file for app %s", app.Slug)
 		}
 
-		decode := scheme.Codecs.UniversalDeserializer().Decode
```
salahalsaleh, line 99: any reason this was moved 10 lines above? :)

salahalsaleh (approved):

kotsadm/pkg/apiserver/bootstrap.go
```diff
@@ -106,6 +102,25 @@ func bootstrapIdentity() error {
 		}
 		identityConfig := obj.(*kotsv1beta1.IdentityConfig)
 
+		installationFile := filepath.Join(currentArchivePath, "upstream", "userdata", "installation.yaml")
+		installationData, err := ioutil.ReadFile(installationFile)
```
salahalsaleh, line 106: we should check if this does not exist

divolgin: this should always exist.  if it doesn't we can't proceed.

kotsadm/pkg/apiserver/bootstrap.go
divolgin: grouping... :srug:

kotsadm/pkg/apiserver/bootstrap.go
salahalsaleh, line 106: this can be used instead
```
installation, err := kotsutil.LoadInstallationFromPath(filepath.Join(appDir, "upstream", "userdata", "installation.yaml"))
if err != nil {
	return "", errors.Wrap(err, "failed to load installation from path")
}
```

salahalsaleh (approved):

PR #1445, opened 2021-01-14
Removing Resource limits

yuvraj9:
We have been using kots in multiple deployments. We found out that CPU is throttling a lot in some of our deployments. Due to this new releases are failing to deploy with this error:-
```failed to run kustomize: kustomize stderr: ""```

Attaching few screenshots of CPU throttling and memory going up.
**Pod names are in screenshots**

Screenshot 2021-01-14 at 10 34 18 PM
Screenshot 2021-01-14 at 10 34 38 PM
Screenshot 2021-01-14 at 10 35 11 PM

I read few articles that CPU throtlling is majorly due to limits applied. So I am removing those limits.

### DIFF
--- a/pkg/kotsadm/kotsadm_objects.go
+++ b/pkg/kotsadm/kotsadm_objects.go
@@ -529,10 +529,6 @@ func kotsadmDeployment(deployOptions types.DeployOptions) *appsv1.Deployment {
 							},
 							Env: env,
 							Resources: corev1.ResourceRequirements{
-								Limits: corev1.ResourceList{
-									"cpu":    resource.MustParse("500m"),
-									"memory": resource.MustParse("500Mi"),
-								},
 								Requests: corev1.ResourceList{
 									"cpu":    resource.MustParse("100m"),
 									"memory": resource.MustParse("100Mi"),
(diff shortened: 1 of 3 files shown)


divolgin:
@yuvraj9 thanks for the PR.  I think completely removing the limits isn't the best solution here.  Either the limits need to be raised or maybe there is a performance issue that we need to look into.  Do you mind providing server specs where you saw this problem and some steps to reproduce?

yuvraj9:
@divolgin I am completely removing the limits because if you see 1st screenshot cpu utilisation is under requests but still CPU is throttling.

Regarding server specs we run different type of nodes ranging from - 4cpu-16gb to 8cpu-64gb. We have seen this issue 10+ times in different instances.

yuvraj9:
@divolgin We use all helm charts. To reproduct you can deploy helm charts in your cluster. Number of helm charts we have is - 36 helm charts. We also have some kubernetes manifests which will be around - 10 to 15.

yuvraj9:
@divolgin any update?

jala-dx:
@yuvraj9 Seems like you are running into this issue ? https://github.com/kubernetes/kops/issues/8954 . Does your VM has the kernel patch mentioned in the PR?

AmberAlston:
@yuvraj9 Are you still experiencing this issue, or did the kernel patch mentioned above help resolve?

yuvraj9:
I would have to check again. To fix this we wrote some custom scripts - to remove load from kots by deleting previous releases.

PR #1464, opened 2021-02-09
Making version app creation more atomic

divolgin:


marccampbell (approved):
Interface changes look 👍

PR #1465, opened 2021-02-10
Snapshots - handle minimal RBAC

salahalsaleh:
Image 2021-02-10 at 2 54 04 PM


cmd/kots/cli/backup-create-rbac-permissions.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"github.com/replicatedhq/kots/pkg/snapshot"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+func BackupCreateRBACPermissionsCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:           "create-rbac-permissions",
+		Short:         "Creates the necessary minimal RBAC permissions that enables the Admin Console to access Velero.",
```
divolgin, line 12: This doesn't look velero specific.  How about
```
Creates the necessary minimal RBAC permissions for the Admin Console.
```

pkg/snapshot/velero.go
```diff
@@ -0,0 +1,74 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+func CreateRBACPermissions(kotsadmNamespace string) error {
+	veleroNamespace, err := DetectVeleroNamespace()
+	if err != nil {
+		return errors.Wrap(err, "failed to detect velero namespace")
+	}
+	if veleroNamespace == "" {
+		return errors.New("velero not found")
```
divolgin, line 22: How about logging this on INFO level instead of aborting.  Again, to not make this function depend on Velero's existence.

cmd/kots/cli/backup-create-rbac-permissions.go
salahalsaleh: It is velero specific though, this cli command will find the velero namespace and create the role and rolebinding there

pkg/snapshot/velero.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	if err := kotsadm.EnsureKotsadmRoleBinding(veleroNamespace, kotsadmNamespace, clientset); err != nil {
+		return errors.Wrap(err, "failed to ensure kotsadm rolebinding")
+	}
+
+	return nil
+}
+
+func DetectVeleroNamespace() (string, error) {
```
divolgin, line 46: This looks like a function that's already implemented elsewhere.

pkg/snapshot/velero.go
salahalsaleh: i intentionally made this a sub cli command in the `backup` command and made it velero specific so that users won't abuse/misuse it.

pkg/snapshot/velero.go
salahalsaleh: yes, it is already defined in `kotsadm`, but it's already part of the refactor in my other NFS pr, and didn't want to refactor kotsadm again to reduce conflicts.

cmd/kots/cli/backup-create-rbac-permissions.go
divolgin: then `create-rbac-permissions` is not the right name for the command.  Also it applies all roles and binding, not just velero ones

AmberAlston:
@salahalsaleh Suggested copy updates below. Minimal RBAC is a term that isn't really meaningfully to the enterprise customer, so I've updated the copy to be more clear for the end-user

Velero Namespace Access Required 

We’ve detected that the Admin Console is running with minimal role-based-access-control (RBAC) privileges, meaning that the Admin Console is limited to a single namespace. To use the snapshots functionality, the Admin Console requires access to the velero namespace. Please run the following command to provide the Admin Console with the necessary permissions to access velero.

divolgin (approved):

AmberAlston (approved):
Approving now that content changes were updated

divolgin (approved):

PR #1473, opened 2021-02-10
Update go mods and add operator in check

marccampbell:


go.mod
```diff
@@ -53,6 +53,7 @@ require (
 	github.com/russellhaering/goxmldsig v1.1.0 // indirect
 	github.com/segmentio/ksuid v1.0.3
 	github.com/sergi/go-diff v1.1.0
+	github.com/shirou/gopsutil v0.0.0-20190901111213-e4ec7b275ada
```
laverya, line 56: was this addition intentional?

laverya (approved):

go.mod
marccampbell: I ran `go mod tidy` and it was automatically added. Should we be concerned?

divolgin (approved):

PR #1494, opened 2021-02-12
Allow skipping RBAC spec when instlling/upgrading Admin Console

divolgin:


cmd/kots/cli/admin-console-upgrade.go
```diff
@@ -96,9 +97,11 @@ func AdminConsoleUpgradeCmd() *cobra.Command {
 	cmd.Flags().String("registry-password", "", "password to use to authenticate with the registry")
 	cmd.Flags().String("kotsadm-namespace", "", "set to override the namespace of kotsadm image. this may create an incompatible deployment because the version of kots and kotsadm are designed to work together")
 	cmd.Flags().String("wait-duration", "2m", "timeout out to be used while waiting for individual components to be ready.  must be in Go duration format (eg: 10s, 2m)")
+	cmd.Flags().Bool("ensure-rbac", true, "when set, kots will create the roles and rolebindings necessary to manage applications")
 	cmd.Flags().MarkHidden("force-upgrade-kurl")
 	cmd.Flags().MarkHidden("kotsadm-tag")
 	cmd.Flags().MarkHidden("kotsadm-namespace")
+	cmd.Flags().MarkHidden("ensure-rbac")
```
emosbaugh, line 104: does this need to be hidden?

cmd/kots/cli/install.go
```diff
@@ -399,6 +400,9 @@ func InstallCmd() *cobra.Command {
 	cmd.Flags().MarkHidden("with-dockerdistribution")
 	cmd.Flags().MarkHidden("storage-base-uri-plainhttp")
 
+	cmd.Flags().Bool("ensure-rbac", true, "when set, kots will create the roles and rolebindings necessary to manage applications")
+	cmd.Flags().MarkHidden("ensure-rbac")
```
emosbaugh, line 404: does this need to be hidden?

emosbaugh (approved):

cmd/kots/cli/install.go
divolgin: I think we should keep this hidden and possibly remove it once we have a proper fix

emosbaugh: tbh i think this is a pretty good solution. we can do a better job with minimal rbac support but this seems different.

PR #1313, opened 2020-12-04
Fix kotsadm dev workflow

MikaelSmith:

PR #1500, opened 2021-02-12
add 'kots upload --deploy --skip-preflights'

### LINKED ISSUE #1201: A `--deploy` option for `kots upload`
When uploading changes, I sometimes want to immediately deploy them. It'd be helpful to have a `kots upload --deploy` option to trigger that on the uploaded code, rather than the roundabout step of running `kots upstream upgrade --deploy` after and hoping it doesn't pull changes from upstream.

MikaelSmith:
Adds `--deploy` and `--skip-preflights` flags to `kots upload`. This enables uploading local changes and immediately deploying them, which is useful for a CLI-only workflow.

Resolves #1201.

### DIFF
--- a/pkg/handlers/upload.go
+++ b/pkg/handlers/upload.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/pkg/errors"
 	"github.com/replicatedhq/kots/pkg/kotsutil"
 	"github.com/replicatedhq/kots/pkg/logger"
 	"github.com/replicatedhq/kots/pkg/preflight"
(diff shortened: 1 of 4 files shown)


MikaelSmith:
I'll remove commits from https://github.com/replicatedhq/kots/pull/1313 once I've tested it.

MikaelSmith:
This seems to work as I'd expect.

MikaelSmith:
An alternative would be a separate command to deploy the current latest sequence, with the option to ignore preflight failures.

dexhorthy:
🤩

pkg/handlers/upload.go
```diff
@@ -141,10 +143,20 @@ func (h *Handler) UploadExistingApp(w http.ResponseWriter, r *http.Request) {
...
+		if err := preflight.Run(a.ID, a.Slug, newSequence, a.IsAirgap, archiveDir); err != nil {
+			logger.Error(err)
+			w.WriteHeader(500)
+			return
+		}
+	}
+
+	if uploadExistingAppRequest.Deploy {
+		if err := version.DeployVersion(a.ID, newSequence); err != nil {
+			logger.Error(errors.Wrap(err, "failed to deploy latest version"))
```
MikaelSmith, line 156: This was based on https://github.com/replicatedhq/kots/blob/v1.33.2/pkg/updatechecker/updatechecker.go#L259, but it's the only place in this file that uses `errors.Wrap`. So not sure if it makes sense to include.

PR #1507, opened 2021-02-16
Proposal for CLI to manage app config

divolgin:


design/apply-config-from-cli.md
```diff
@@ -0,0 +1,68 @@
...
+ --config-value
+ --config-value-from-file
+ --deploy
+ ```
+
+ | Flag | Description |
+| :---- | ----------- |
+| `--config-values` | This is a path to a config file compiant with KOT's `kind: ConfigValues` format.  By default, the contents of this file will be used to replace existing app config in its entirety.  This property can be used to delete config items that are no longer used. |
+| `--merge` | When this parameter is specified, only the keys included in the `--config-values` file will be replaced and all other values will be preserved.  This can be used to avoid creating multiple app versions when setting one value at a time when more than one config value needs to be changed. |
+| `--config-key` | This is the name of the config key whose value will be replaced.  Either `--config-value` or `--config-value-from-file` is required with this parameter.  `--config-values` cannot be used in conjunction with this parameter. |
```
markpundsack, line 56: Why not just something like `--set KEY=VALUE`? That seems more intuitive than separate args for key and value. I get that having them separate does allow you to do `--config-value-from-file`, which is kind of nice. But using `--set` also allows you to use it multiple times, like Helm does. How important is `--config-value-from-file` (as opposed to using `--config-values` with `--merge`?

design/apply-config-from-cli.md
```diff
@@ -0,0 +1,68 @@
...
+ --merge
+ --config-key
+ --config-value
+ --config-value-from-file
+ --deploy
+ ```
+
+ | Flag | Description |
+| :---- | ----------- |
+| `--config-values` | This is a path to a config file compiant with KOT's `kind: ConfigValues` format.  By default, the contents of this file will be used to replace existing app config in its entirety.  This property can be used to delete config items that are no longer used. |
```
markpundsack, line 54: `--config-values` is confusable with `--config-value`. How about `--config-file` instead? Or even just `--config`, especially if we use `--set` for individual config?

markpundsack: Feature creep, but it would be nice to support a URL for values as well, especially with `--merge` as a vendor could provide some sample configuration, while allowing people to leave other config alone.

design/apply-config-from-cli.md
```diff
@@ -0,0 +1,68 @@
...
+There will be two modes of operation:
+1. Apply an entire configuration to the app.
+1. Apply a config change to a subset of values.  This is mostly a convenience method since this can be achieved with the first mode.
+
+## Detailed Design
+
+The following command will be added to CLI:
+
+```
+kubectl kots config <app-slug> set [parameters]
```
markpundsack, line 38: Why is `set` needed here? Is that because we're leaving room for a `get` and `export`?

It conflicts with my use of `--set` below, but if we do keep set, then I'd suggest we should simply support key=value pairs after the `set` command without having to use `--` args at all.

Lastly, should we follow k8s verb-noun and use `kubectl kots set config`?

divolgin: > Why is set needed here? Is that because we're leaving room for a get and export?
yes
> Lastly, should we follow k8s verb-noun and use kubectl kots set config
yes, we can do this

design/apply-config-from-cli.md
divolgin: how about `--from-file`?

design/apply-config-from-cli.md
divolgin: We currently don't enforce key format, so I didn't want to get into parsing these.   "From file" is nice for multi-line config items like PEM certs as people usually struggle with these (it's a platform thing).  We can do `KEY=VALUE`  and have `--key` + `--from-file` as well.

markpundsack: Gotcha, that makes sense. Multi-line values are always a pain, so if that's common, it makes to support it from the beginning.

design/apply-config-from-cli.md
markpundsack: Hmm... interesting. Since the command is `set config`, I guess the `config` part of the arguments is redundant. Are you suggesting `kots set config --from-file config.yaml` to load the whole thing, or is `--from-file` for setting an individual value? 

Maybe put it all together and see how it works. What's it look like if we support:
1. key=value
2. individual values from file
3. whole config from file
4. k8s-style verb-then-noun

design/apply-config-from-cli.md
```diff
@@ -0,0 +1,85 @@
...
+| `--skip-preflights` | By default preflight checks will run on config update. This parameter has the same function as the one in `kots upstream upgrade` command. |
+
+### Examples
+
+```
+kubectl kots set config myapp --from-file /path/to/local/config.yaml
+```
+
+```
+kubectl kots set config myapp --key config-item-name --from-file /path/to/config/file/value.txt
```
markpundsack, line 68: Ooh, this is ambiguous. `--from-file` is only a single value if combined with `--key`, but otherwise is gvk-formatted?

design/apply-config-from-cli.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+```
+kubectl kots set config myapp config-item-name="config item value"
+```
+
+```
+kubectl kots set config myapp --key config-item-name --value "config item value"
```
markpundsack, line 76: It's a little weird to have two ways of doing the exact same thing.

design/apply-config-from-cli.md
```diff
@@ -0,0 +1,85 @@
...
+There will be two modes of operation:
+1. Apply an entire configuration to the app.
+1. Apply a config change to a subset of values.  This is mostly a convenience method since this can be achieved with the first mode.
+
+## Detailed Design
+
+The following command will be added to CLI:
+
+```
+kubectl kots set config <app-slug> [parameters]
```
markpundsack, line 38:
```suggestion
kubectl kots set config [options] <app-slug> [KEY_1=VAL_1 ... KEY_2=VAL_2]
```

design/apply-config-from-cli.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+The following flags will be supported, in addition to all higher level flags:
+
+```
+ --key
+ --value
+ --from-file <path to config file>
```
markpundsack, line 46:
```suggestion
 --value-from-file <path to value file>
 --config-file <path to config file>
```

PR #1515, opened 2021-02-17
Don't look like app install failed when not using port forward

divolgin:


cmd/kots/cli/install.go
```diff
@@ -328,6 +328,10 @@ func InstallCmd() *cobra.Command {
 				log.ActionWithoutSpinner("")
 				log.ActionWithoutSpinner("To access the Admin Console, run kubectl kots admin-console --namespace %s", namespace)
 				log.ActionWithoutSpinner("")
+			} else {
+				log.ActionWithoutSpinner("")
+				log.ActionWithoutSpinner("Done")
```
divolgin, line 333: Without this, the last thing printed is an ambiguous error, making it look like install failed.

salahalsaleh (approved):

cmd/kots/cli/install.go
emosbaugh: Why is that EOF logged?

PR #1520, opened 2021-02-17
Move more from kotsadm/pkg into kots/pkg

marccampbell:
This has no effect on the product, but is a refactor moving almost everything from kotsadm/pkg into kots/pkg in preparation for a single project workspace


marccampbell:
@divolgin @salahalsaleh moved the web and skaffold to the root here too

divolgin (approved):

PR #1532, opened 2021-02-18
Update metadata configmap on app spec change

jgruica:


divolgin (approved):
lgtm, though it would be good to have all objects files in the new objects package.

divolgin (approved):

PR #1537, opened 2021-02-19
handle unexpected config map deletions/creations

jala-dx:


pkg/apiserver/server.go
```diff
@@ -183,42 +182,26 @@ func generateKotsadmID() error {
 		return nil
 	}
 	clusterID := clusters[0].ClusterID
-
-	// Write a Query to set/get an Event from the Store
-	isKotsadmIDGenerated, err := store.GetStore().IsKotsadmIDGenerated()
+	isKotsadmIDInDB, err := store.GetStore().IsKotsadmIDGenerated()
```
salahalsaleh, line 185: why rename it? `isKotsadmIDGenerated` sounds better

pkg/apiserver/server.go
```diff
@@ -183,42 +182,26 @@ func generateKotsadmID() error {
...
 	if err != nil {
-		return errors.Wrap(err, "Failed to check kotsadm id generation")
```
salahalsaleh, line 190: we should wrap the error

pkg/apiserver/server.go
```diff
@@ -49,7 +49,7 @@ func Start() {
 	}
 
 	if err := generateKotsadmID(); err != nil {
-		logger.Infof("failed to generate kots admin id", err)
+		logger.Infof("kots adminid:", err)
```
salahalsaleh, line 52:
```suggestion
		logger.Infof("Failed to generate kotsadm id:", err)
```

pkg/apiserver/server.go
```diff
@@ -183,42 +182,26 @@ func generateKotsadmID() error {
...
-		}
-		//generate guid and use that as clusterId to identify that as a different install
-		clusterID = ksuid.New().String()
-		_, err = s3pg.CreateAdminIDConfigMap(clusterID)
-		if err != nil {
-			return errors.Wrap(err, "failed to to create config map")
-		}
+	cmpExists, err := s3pg.IsKotsadmIDConfigMapPresent()
+	if err != nil {
+		return err
```
salahalsaleh, line 191: should use errors.wrap

pkg/apiserver/server.go
```diff
@@ -183,42 +182,26 @@ func generateKotsadmID() error {
...
+		// id exists and so as configmap, noop
+	}
+	if err == nil {
 		err = store.GetStore().SetIsKotsadmIDGenerated()
-		if err != nil {
-			return errors.Wrap(err, "failed to set admin event status")
-		}
 	}
-	return nil
+	return err
```
salahalsaleh, line 206: it's gonna be hard to know where the error came from, it would be better to wrap the error and return it after each function instead

pkg/store/s3pg/reporting_store.go
```diff
@@ -33,25 +33,15 @@ func (s S3PGStore) GetReportingInfo(appID string) *upstreamtypes.ReportingInfo {
...
-		clusterID = ksuid.New().String()
-		r.ClusterID = clusterID
-		_, err = CreateAdminIDConfigMap(clusterID)
-		if err != nil {
-			logger.Errorf("Failed to to create config map %v", err)
-		}
+		r.ClusterID = ksuid.New().String()
+		CreateAdminIDConfigMap(r.ClusterID)
 	} else {
-		logger.Errorf("Config map check error %v", err)
```
salahalsaleh, line 54: we should still log this so that we know something went wrong

pkg/store/s3pg/reporting_store.go
```diff
@@ -163,17 +153,15 @@ func getKotsadmIDConfigMap() (*corev1.ConfigMap, error) {
...
-	if existingConfigmap != nil {
-		return existingConfigmap, nil
-	}
-	return nil, nil
+	return existingConfigmap, nil
 }
 
 // CreateAdminIDConfigMap creates an id for an kotsadm instance and stores in configmap
-func CreateAdminIDConfigMap(clusterID string) (*corev1.ConfigMap, error) {
+func CreateAdminIDConfigMap(kotsadmID string) error {
```
salahalsaleh, line 160:
```suggestion
func CreateKotsadmIDConfigMap(kotsadmID string) error {
```

pkg/store/s3pg/reporting_store.go
```diff
@@ -205,15 +188,30 @@ func IsKotsadmIDConfigMapPresent() (bool, error) {
...
 	if err != nil && !kuberneteserrors.IsNotFound(err) {
 		return false, errors.Wrap(err, "failed to get configmap")
 	} else if kuberneteserrors.IsNotFound(err) {
 		return false, nil
 	}
+	return true, nil
+}
+
+// UpdateAdminIDConfigMap creates an id for an kotsadm instance and stores in configmap
+func UpdateAdminIDConfigMap(kotsadmID string) error {
```
salahalsaleh, line 201:
```suggestion
func UpdateKotsadmIDConfigMap(kotsadmID string) error {
```

pkg/apiserver/server.go
markpundsack: FWIW, is ID generated is kind of a weird name and I'm not sure it accurately reflects what this means. But renaming between the function and the variable seems wrong no matter what.

salahalsaleh: @markpundsack how about `IsKotsadmIDMigrated`?

pkg/store/s3pg/reporting_store.go
jala-dx: Here we are replacing it with "generation of guid"  if incase k8 fails create configmap (as per our previous discussion).

pkg/apiserver/server.go
jala-dx: let us go with IsKotsadmIDMigrated

salahalsaleh (approved):

PR #1539, opened 2021-02-19
Move tasks to a configmap, out of postgres

marccampbell:
I am migrating any rows from Postgres into the ConfigMap. It's unlikely that this will be needed, but there is possibly an edge case where an application upgrade is in progress and the Admin Console is upgrades before it completes. In this case, the migration isn't going to add a new resumable functionality, but it will keep the user experience consistent.

### DIFF
--- a/pkg/store/kotsstore/kots_store.go
+++ b/pkg/store/kotsstore/kots_store.go
@@ -19,6 +19,9 @@ import (
 	kotss3 "github.com/replicatedhq/kots/pkg/s3"
 	troubleshootscheme "github.com/replicatedhq/troubleshoot/pkg/client/troubleshootclientset/scheme"
 	veleroscheme "github.com/vmware-tanzu/velero/pkg/generated/clientset/versioned/scheme"
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


pkg/store/kotsstore/kots_store.go
```diff
@@ -194,3 +197,53 @@ func (c KOTSStore) GetClientset() (*kubernetes.Clientset, error) {
...
+		configmap := corev1.ConfigMap{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: "v1",
+				Kind:       "ConfigMap",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      name,
+				Namespace: os.Getenv("POD_NAMESPACE"),
+				Labels: map[string]string{
+					"owner": "kotsadm",
```
divolgin, line 220: This doesn't look like a real label.  Everything else uses [these](https://github.com/replicatedhq/kots/blob/3f5e3166c8bb14666ba696eaac6ebbefc4310f6f/pkg/kotsadm/types/constants.go#L16)

divolgin (approved):

divolgin (approved):

divolgin (approved):

PR #1556, opened 2021-02-22
Migrate support bundles to native k8s objects

marccampbell:
In the pg version of KOTS, there are three tables, three types:
- Pending Support BUndle
- Support Bundle 
- Support Bundle Analysis

This PR keeps the types the same, but moves the storage. Now there will be 1 secret per bundle/analysis. If theres no analysis, the spec will have an empty analysis.

The secrets for bundles and analysis are named `supportbundle-<id>`
In addition to the common KOTS labels, these objects have some additional labels:

```
kots.io/kind: supportbundle
kots.io/appId: <value>
```

Pending support bundles are secrets, with common labels, the labels above, but `kots.io/kind` is set to `pendingsupportbundle`.

Additionally, there's some cleanup. We had GetSupportBundle and GetSupportBundleForSlug. But Slugs and IDs were the same, so this was two code paths that was brittle and hopeful to not diverge. Cleaned this up.

Finally, the support bundle analysis was stored as a string in PG and then painfully unmarshaled in the Get function of the store. This is handled one final time in the migration, but the insights are stored in native format now.

### DIFF
--- a/pkg/store/kotsstore/supportbundle_store.go
+++ b/pkg/store/kotsstore/supportbundle_store.go
@@ -1,6 +1,7 @@
 package kotsstore
 
 import (
+	"context"
 	"database/sql"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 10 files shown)


salahalsaleh (approved):

pkg/store/kotsstore/supportbundle_store.go
```diff
@@ -60,90 +69,238 @@ func (s KOTSStore) ListSupportBundles(appID string) ([]*supportbundletypes.Suppo
...
+	}
 
-	id := ""
-	if err := row.Scan(&id); err != nil {
-		if err == sql.ErrNoRows {
-			return nil, nil
-		}
-		return nil, errors.Wrap(err, "failed to scan id")
+	if err := clientset.CoreV1().Secrets(os.Getenv("POD_NAMESPACE")).Delete(context.TODO(), fmt.Sprintf("pendingsupportbundle-%s", id), metav1.DeleteOptions{}); err != nil {
+		return errors.Wrap(err, "failed to delete")
```
divolgin, line 205: This should check IsNotFound

pkg/store/kotsstore/supportbundle_store.go
```diff
@@ -60,90 +69,238 @@ func (s KOTSStore) ListSupportBundles(appID string) ([]*supportbundletypes.Suppo
...
+			Data: map[string][]byte{
+				"bundle":     bundleMarshaled,
+				"analysis":   analysisMarshaled,
+				"redactions": redactionsMarshaled,
+			},
+		}
+
+		_, err = clientset.CoreV1().Secrets(os.Getenv("POD_NAMESPACE")).Create(context.TODO(), &secret, metav1.CreateOptions{})
+		if err != nil {
+			return errors.Wrap(err, "failed to create support bundle secret")
```
divolgin, line 148: This should just continue if secret already exists

divolgin (approved):
A couple of comments on migrations.  But also looks like this breaks the sort order on the UI.

divolgin (approved):

PR #1561, opened 2021-02-23
PROPOSAL: editing existing config 

divolgin:


design/edit-existing-release-config.md
```diff
@@ -0,0 +1,98 @@
...
+The contents of the `base` spec are then generated from `upstream` and `config.yaml`.
+The proposal is to allow modifying application version archives and versioning `config.yaml` within the archive.
+
+## Detailed Design
+
+1. `ConfigValues` will be versioned within the `upstream/userdata/` directory.
+   - Config version sequence will be zero-based.
+   - Annotation `kots.io/config-sequence` will be added to ConfigValues spec.
+   - Current configuration will be stored in `upstream/userdata/config.yaml` (same file as now)
+   - **All** configurations will be stored in files with version numbers included in the files: `config-N.yaml`
```
marccampbell, line 43: This feels _slightly_ brittle. Sequencing things is hard and if there's a gap, things will break? An alternative might be to make config.yaml a multi-doc yaml, with each version ORDERED with the newest at the bottom?

marccampbell:
@divolgin I follow and it seems like a good approach. I had a comment about the sequence in filenames.

What happens when:

1. Install application
2. Check for updates and a new update is available (not installed)
3. Re-configure the 1st (active) version.

Does this change propagate to a new config version in all future (pending) versions?


---

Also, when changing the config of an active version, does it instantly get deployed? If the config change results in new images (optional database, etc), that could take a bit to rewrite and push images. Is there a multi-stage process or any plans to make the user aware if the "Save" button at the bottom is really save, or save and deploy? My understanding of the proposal is that it will be Save if not editing the "active" version, but "Save and Deploy" if editing the active version.

design/edit-existing-release-config.md
divolgin: I'd like to keep keep config.yaml as is because it's an easy place to copy config from to use with kots CLI.
How about adding config-history.yaml, which will have a new kind: an array of configs.

divolgin:
> Does this change propagate to a new config version in all future (pending) versions?
Good question.  It would make sense to add it as a config change to all pending versions.  Alternatively we can do when happens now: create new version based on latest release with new config.

> Also, when changing the config of an active version, does it instantly get deployed?
I think it should just save and then the Redeploy button needs to be clicked afterwards.

> Is there a multi-stage process?
There isn't.  We only push images on initial install and when registry info is updated.  But this proposal needs to be updated to handle registry updates

PR #1575, opened 2021-02-26
better handling for non-existant app slugs in backups

salahalsaleh:


divolgin (changes requested):
Use `store.GetStore().IsNotFound(err)`

divolgin (approved):

PR #1581, opened 2021-02-28
Multi-doc helm files only include first document

emosbaugh:


emosbaugh:
@jala-dx I fixed a bunch of issues here. This code probably needs to get copied over to the airgap builder.

divolgin (approved):

PR #1616, opened 2021-03-10
Back to the dashboard button on error modal

jgruica:
Screen Shot 2021-03-10 at 10 06 35 AM

### DIFF
--- a/web/src/components/PreflightResultErrors.jsx
+++ b/web/src/components/PreflightResultErrors.jsx
@@ -151,6 +151,7 @@ class PreflightResultErrors extends Component {
             toggleErrorModal={this.toggleErrorModal}
             err={errorTitle}
             errMsg={errorMsg}
+            appSlug={this.props.appSlug}
           />}
       </div>
     );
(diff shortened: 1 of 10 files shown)


web/src/components/PreflightResultErrors.jsx
```diff
@@ -151,6 +151,7 @@ class PreflightResultErrors extends Component {
             toggleErrorModal={this.toggleErrorModal}
             err={errorTitle}
             errMsg={errorMsg}
+            appSlug={this.props.appSlug}
```
GraysonNull, line 154: Everywhere else this is using `this.props.match.params.slug`. Is this component being passed that value or something or should this also use params?

jgruica: this component is subcomponent of PreflightResultPage and there I'm passing this.props.match.params.slug
https://github.com/replicatedhq/kots/pull/1616/files#diff-55df8e3fa9c9b98697e9c8f63f1e24f2320ac043ad728aa267e2d6a2bb742842R186

GraysonNull: great, thanks for clarifying!

PR #1629, opened 2021-03-14
support backup ordered resources in full snapshots

salahalsaleh:


pkg/kotsadmsnapshot/backup.go
```diff
@@ -177,7 +177,7 @@ func CreateInstanceBackup(ctx context.Context, cluster *downstreamtypes.Downstre
 	includedNamespaces := []string{kotsadmNamespace}
 	excludedNamespaces := []string{}
 	backupAnnotations := map[string]string{}
-	backupTTL := metav1.Duration{}
```
salahalsaleh, line 180: this will always be overridden by the default `720h` value and can only be configured by the customer not the vendor. docs updated here: https://github.com/replicatedhq/kots.io/pull/436/commits/b0cfca9ce5344a1e9127733cd03e5f85ed64eba5#diff-27cf6040c53d708d1cd8133792a964b18cc21ffb4065881af76b55254a992660R36

divolgin (approved):

PR #1638, opened 2021-03-19
Kots and app install metrics

salahalsaleh:


kustomize/overlays/dev/deployment.yaml
```diff
@@ -69,6 +69,10 @@ spec:
             #   value: "true"
             - name: KURL_PROXY_TLS_CERT_PATH
               value: /etc/kurl-proxy/ca/tls.crt
+            # - name: KURL_INSTALL_ID
```
divolgin, line 72: dead yaml

pkg/metrics/install.go
```diff
@@ -0,0 +1,134 @@
...
+func getEndpoint(license *kotsv1beta1.License) string {
+	if license != nil {
+		if isDevEndpoint(license.Spec.Endpoint) {
+			// cluster ip services are not resolvable from the cli ...
+			return DevEndpoint
+		}
+		return license.Spec.Endpoint
+	}
+
+	devCfgFile := fmt.Sprintf("%s/src/github.com/replicatedhq/kots/pkg/metrics/dev-cfg.json", os.Getenv("GOPATH"))
```
divolgin, line 117: This is weird... How about setting endpoint at build time using the -X option.  With goreleaser it will be empty string, which would default to prod.

pkg/metrics/install.go
```diff
@@ -0,0 +1,134 @@
...
+	StartedAt   time.Time `json:"started"`
+	FinishedAt  time.Time `json:"finished"`
+	FailedAt    time.Time `json:"failed"`
+	KotsVersion string    `json:"kots_version"`
+	Cause       string    `json:"cause"`
+}
+
+func InitInstallMetrics(license *kotsv1beta1.License, disableOutboundConnections bool) *InstallMetrics {
+	if disableOutboundConnections {
+		return nil
```
divolgin, line 36: `InstallMetrics` has exported fields, which can't be private because of json tags... So I feel it would be better to keep `disableOutboundConnections` as an internal flag instead of returning nil here.

pkg/reporting/app.go
```diff
@@ -0,0 +1,151 @@
...
+		r.ClusterID = configMap.Data["id"]
+	} else {
+		// configmap is missing for some reason, recreate with new guid, this will appear as a new instance in the report
+		r.ClusterID = ksuid.New().String()
+		k8s.CreateKotsadmIDConfigMap(r.ClusterID)
+	}
+
+	di, err := getDownstreamInfo(appID)
+	if err != nil {
+		logger.Infof("failed to get downstream info: %v", err.Error())
```
divolgin, line 83: reporting functions shouldn't log above debug level.  i'd prefer if they didn't log anything.

pkg/socketservice/socketservice.go
```diff
@@ -377,10 +377,15 @@ func processDeploySocketForApp(clusterSocket *ClusterSocket, a *apptypes.App) er
...
 			},
 		}
-		err := appstatus.Set(a.ID, defaultReadyState, time.Now())
+		err := store.GetStore().SetAppStatus(a.ID, defaultReadyState, time.Now())
 		if err != nil {
 			return errors.Wrap(err, "failed to set app status")
 		}
+		go func() {
+			if err := reporting.SendAppInfo(a.ID); err != nil {
+				logger.Debugf("failed to send app info: %#v", err)
```
divolgin, line 386: how about no logging for this?

divolgin (approved):

PR #1645, opened 2021-03-22
use primary and secondary in language

GraysonNull:


marccampbell:
@GraysonNull I don't know if we should drop the old API endpoints without a deprecation. It's not a documented API, but we do point people at it sometimes. How about leaving both the old and new endpoints in the code for a few versions before removing the old endpoints?

GraysonNull:
Ah i didn't realize those kurl endpoints were public facing. Yeah agreed we shouldn't change them then. I'll duplicate them with updated language for now and we can phase them out after a few versions like you suggested. Thanks

jgruica (approved):

salahalsaleh (approved):

PR #1650, opened 2021-03-24
Cache session and task status data

divolgin:


pkg/store/kotsstore/kots_store.go
```diff
@@ -194,10 +201,36 @@ func (s KOTSStore) IsNotFound(err error) bool {
 		return true
 	}
 
+	if kuberneteserrors.IsNotFound(err) {
```
salahalsaleh, line 204: whooh, i thought we had this already, hope this didn't cause subtle bugs

pkg/store/kotsstore/kots_store.go
```diff
@@ -194,10 +201,36 @@ func (s KOTSStore) IsNotFound(err error) bool {
...
+
+	if strings.Contains(err.Error(), "connection refused") {
+		return true
+	}
+
+	if strings.Contains(err.Error(), "request timed out") {
+		return true
+	}
+
+	return true
```
salahalsaleh, line 224:
```suggestion
	return false
```

pkg/store/kotsstore/kots_store.go
divolgin: indeed.  but this feels like good time to add this.

salahalsaleh (approved):

PR #1652, opened 2021-03-24
use s3url endpoint and region to detect the location of s3 bucket

jala-dx:


pkg/snapshot/store.go
```diff
@@ -369,8 +370,16 @@ func updateGlobalStore(ctx context.Context, store *types.Store, kotsadmNamepsace
 	}
 
 	if store.AWS != nil {
+		var endpointURL string
```
salahalsaleh, line 373: why is this variable needed? could probably just do
```
kotsadmVeleroBackendStorageLocation.Spec.Config = map[string]string{
	"region": store.AWS.Region,
	"s3Url":  resolvedEndpoint.URL,
}
```

pkg/snapshot/store.go
```diff
@@ -788,8 +797,16 @@ func GetGlobalStore(ctx context.Context, kotsadmNamepsace string, kotsadmVeleroB
...
+	var domainName string
+	u, err := url.Parse(endpoint)
+	if err == nil {
+		splitHost := strings.Split(u.Hostname(), ".")
+		if len(splitHost) > 2 {
+			domainName = splitHost[len(splitHost)-2]
+		}
+	}
+	// if it is not s3 compatible, check if the endpoint domain matches amazonaws
+	if !isS3Compatible || domainName == "amazonaws" {
```
divolgin, line 809: Instead of splitting host name, do this `strings.HasSuffix(host, ".amazonaws.com")`

pkg/snapshot/store.go
```diff
@@ -788,8 +797,16 @@ func GetGlobalStore(ctx context.Context, kotsadmNamepsace string, kotsadmVeleroB
 
 func mapAWSBackupStorageLocationToStore(kotsadmVeleroBackendStorageLocation *velerov1.BackupStorageLocation, store *types.Store) error {
 	endpoint, isS3Compatible := kotsadmVeleroBackendStorageLocation.Spec.Config["s3Url"]
-
-	if !isS3Compatible {
+	var domainName string
+	u, err := url.Parse(endpoint)
+	if err == nil {
```
salahalsaleh, line 802: the endpoint must be a valid url, so i suggest just returning an error here if `err` is not nil

pkg/snapshot/store.go
```diff
@@ -788,8 +797,16 @@ func GetGlobalStore(ctx context.Context, kotsadmNamepsace string, kotsadmVeleroB
...
+		splitHost := strings.Split(u.Hostname(), ".")
+		if len(splitHost) > 2 {
+			domainName = splitHost[len(splitHost)-2]
+		}
+	}
+	// if it is not s3 compatible, check if the endpoint domain matches amazonaws
```
divolgin, line 808: the comment should explain why we are doing this

pkg/snapshot/store.go
```diff
@@ -808,7 +825,7 @@ func mapAWSBackupStorageLocationToStore(kotsadmVeleroBackendStorageLocation *vel
 	}
 
 	// check if using file system store
-	u, err := url.Parse(endpoint)
+	u, err = url.Parse(endpoint)
```
salahalsaleh, line 828: this is not needed anymore, could just use the parsed url above

pkg/snapshot/store.go
divolgin, line 802: `err != nil` needs to be handled

pkg/snapshot/store.go
divolgin, line 828: isn't this already done in the same scope?

pkg/snapshot/store.go
```diff
@@ -788,8 +795,13 @@ func GetGlobalStore(ctx context.Context, kotsadmNamepsace string, kotsadmVeleroB
 
 func mapAWSBackupStorageLocationToStore(kotsadmVeleroBackendStorageLocation *velerov1.BackupStorageLocation, store *types.Store) error {
 	endpoint, isS3Compatible := kotsadmVeleroBackendStorageLocation.Spec.Config["s3Url"]
-
-	if !isS3Compatible {
+	// check if using file system store
```
salahalsaleh, line 798: this comment should stay where it was :)

pkg/snapshot/store.go
```diff
@@ -788,8 +795,13 @@ func GetGlobalStore(ctx context.Context, kotsadmNamepsace string, kotsadmVeleroB
...
+	u, err := url.Parse(endpoint)
+	if err != nil {
+		return errors.Wrap(err, "failed to parse s3 url")
+	}
+	// without endpoint, the ui has no logic to figure if it is amazon-s3 or other-s3 compatible storages
+	if !isS3Compatible || strings.HasSuffix(endpoint, ".amazonaws.com") {
```
salahalsaleh, line 804: this should use `u.Hostname` instead of endpoint

salahalsaleh (approved):

divolgin (approved):

PR #1679, opened 2021-03-27
kotsadm and app redact spec lazy migration

salahalsaleh:


pkg/api/handlers/types/types.go
```diff
@@ -27,7 +27,6 @@ type ResponseApp struct {
 	CreatedAt         time.Time  `json:"createdAt"`
 	UpdatedAt         *time.Time `json:"updatedAt"`
 	LastUpdateCheckAt string     `json:"lastUpdateCheckAt"`
-	BundleCommand     []string   `json:"bundleCommand"`
```
salahalsaleh, line 30: this is not used

divolgin (approved):

PR #1682, opened 2021-03-29
also ignore EOF network errors in non-critical paths

divolgin:


salahalsaleh (approved):

pkg/store/kotsstore/kots_store.go
```diff
@@ -221,6 +221,10 @@ func canIgnoreEtcdError(err error) bool {
 		return true
 	}
 
+	if strings.Contains(err.Error(), "EOF") {
```
marccampbell, line 224: @divolgin can we use `errors.Is` instead?

https://golang.org/pkg/errors/#example_Is

divolgin: Would love to, but I don't know which error it is without a way to reproduce it.  Also, this covers any EOF error.

PR #1701, opened 2021-03-31
respect kubeconfig flag when getting k8s or velero clientset

salahalsaleh:


cmd/kots/cli/root.go
```diff
@@ -21,8 +22,8 @@ func RootCmd() *cobra.Command {
 
 	cobra.OnInitialize(initConfig)
 
-	kubernetesConfigFlags = genericclioptions.NewConfigFlags(false)
-	kubernetesConfigFlags.AddFlags(cmd.PersistentFlags())
+	k8sutil.KubernetesConfigFlags = genericclioptions.NewConfigFlags(false)
```
divolgin, line 25: i think this would be better off in the `init` function in `k8sutil` package. and also since this is the only place `KubernetesConfigFlags` is used directly, it could be un-exported.

divolgin (approved):

divolgin (approved):

PR #1702, opened 2021-03-31
feat(kotsadm): use default branch for gitops when none provided

DanStough:
Modifying the initialization step in Gitops for the kotsadm backend to interpret an empty branch name as using the default (frontend work in separate PR). 

Init is always called on update or create from the frontend so it is relatively safe operation. Also from a performance standpoint, we can use the existing close with talking to the remote twice. The only downside is that changing the default branch will require a manual update, but imagining this to be rare for customers.

Tested on Github and Gitlab. Since it's vanilla git, not too worried about other compatibility.

### DIFF
--- a/pkg/handlers/gitops.go
+++ b/pkg/handlers/gitops.go
@@ -143,7 +143,8 @@ func (h *Handler) InitGitOpsConnection(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if err := gitops.TestGitOpsConnection(downstreamGitOps); err != nil {
+	defaultBranchName, err := gitops.TestGitOpsConnection(downstreamGitOps)
+	if err != nil {
 		logger.Infof("Failed to test gitops connection: %v", err)
 
 		if err := gitops.SetGitOpsError(a.ID, d.ClusterID, err.Error()); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/handlers/gitops.go
```diff
@@ -154,6 +155,18 @@ func (h *Handler) InitGitOpsConnection(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// If a branch is not provided, use the default branch
+	if downstreamGitOps.Branch == "" {
+		err := gitops.UpdateDownstreamGitOps(a.ID, d.ClusterID, downstreamGitOps.RepoURI, defaultBranchName,
```
salahalsaleh, line 160: i feel like this should be part of the "update app gitops settings" workflow instead of "test app gitops settings" workflow.

pkg/handlers/gitops.go
```diff
@@ -154,6 +155,18 @@ func (h *Handler) InitGitOpsConnection(w http.ResponseWriter, r *http.Request) {
...
+			downstreamGitOps.Path, downstreamGitOps.Format, downstreamGitOps.Action)
+		if err != nil {
+			logger.Error(err)
```
salahalsaleh, line 163: gitops error needs to also be set, like in here https://github.com/replicatedhq/kots/pull/1702/files#diff-43f6a5e0d4ac97148500c550a148da148479ad4ee1a4e9c5809ea84174e4aa2bR150

pkg/handlers/gitops.go
DanStough: Yes, I agree, but it would require a bigger refactor to test the repo while we perform a create or update.  Normally setting up a Gitops repo executes the following flow /Create->/Update ->/Init. We could combine them into two independent /create and /update APIs. Would like to know how @GraysonNull feels about doing this refactor on the frontend.

salahalsaleh (approved):

pkg/handlers/gitops.go
DanStough: Per our conversations we'll end up keeping this as a relatively low-risk change and plan to update the flow/UI later.

PR #1703, opened 2021-04-01
do not re-deploy app on snapshot restore

salahalsaleh:


pkg/socketservice/socketservice.go
```diff
@@ -610,10 +610,16 @@ func checkRestoreComplete(clusterSocket *ClusterSocket, a *apptypes.App, restore
...
 
-		logger.Info(fmt.Sprintf("restore complete, re-deploying version %d", sequence))
+		logger.Info(fmt.Sprintf("restore complete, marking version %d as deployed", sequence))
 
-		if err := RedeployAppVersion(a.ID, sequence, clusterSocket); err != nil {
-			return errors.Wrap(err, "failed to redeploy app version")
+		// mark the sequence as deployed both in the socket history and db
+		// so that the admin console does not try to re-deploy it
+		socketMtx.Lock()
+		clusterSocket.LastDeployedSequences[a.ID] = sequence
```
divolgin, line 618: Shouldn't this be done after DeployVersion is called?

salahalsaleh: It won't matter because "restoreInProgressName" will block the deploy loop and it's reset at the end of this function. And there's no right order here because the deploy loop checks for both, how would switching the order make a difference?

divolgin (approved):

PR #1705, opened 2021-04-01
Decouple license from airgap bundle

jala-dx:


pkg/automation/automation.go
```diff
@@ -124,6 +125,12 @@ LICENSE_LOOP:
 			continue
 		}
 
+		// check for the airgap flag in the annotations
+		objMeta := licenseSecret.GetObjectMeta()
+		annotations := objMeta.GetAnnotations()
+		if annotations["kots.io/airgap"] == "true" {
+			continue
```
salahalsaleh, line 132: won't this break the workflow of actually providing an airgap bundle?

pkg/automation/automation.go
```diff
@@ -379,6 +386,7 @@ func NeedToWaitForAirgapApp() (bool, error) {
 }
 
 func getAirgapData(clientset kubernetes.Interface, license *kotsv1beta1.License) (map[string][]byte, error) {
+
```
salahalsaleh, line 389: can we keep the current code format of not having an empty line at the beginning of the function?

pkg/handlers/app.go
```diff
@@ -20,6 +20,51 @@ import (
 	"github.com/replicatedhq/kots/pkg/version"
 )
 
+func (h *Handler) GetPendingApps(w http.ResponseWriter, r *http.Request) {
+	sess := session.ContextGetSession(r)
```
salahalsaleh, line 24: this is not needed, it's already handled in the session middleware (should be removed from `ListApps` handler as well)

pkg/handlers/app.go
```diff
@@ -20,6 +20,51 @@ import (
 	"github.com/replicatedhq/kots/pkg/version"
 )
 
+func (h *Handler) GetPendingApps(w http.ResponseWriter, r *http.Request) {
```
salahalsaleh, line 23: this function seems to be returning 1 pending app, maybe the name should be `GetPendingApp` instead of `GetPendingApps`?

pkg/handlers/app.go
```diff
@@ -20,6 +20,51 @@ import (
...
+		if store.GetStore().IsNotFound(err) {
+			w.WriteHeader(http.StatusNotFound)
+		} else {
+			logger.Error(errors.Wrap(err, "failed to get pending app"))
+			w.WriteHeader(http.StatusInternalServerError)
+		}
+		return
+	}
+
+	defaultRoles := rbac.DefaultRoles() // TODO (ethan): this should be set in the handler
```
salahalsaleh, line 42: this is not needed here, it's used in `ListApps` to check access for each app separately, but here we're only returning 1 pending app and access control should already be handled in the session middleware

pkg/kotsadm/objects/license_objects.go
```diff
@@ -6,7 +6,13 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func LicenseSecret(namespace string, appSlug string, license string) *corev1.Secret {
+func LicenseSecret(namespace string, appSlug string, isAirgap bool, license string) *corev1.Secret {
+	annotations := map[string]string{
+		"kots.io/airgap": "false",
```
salahalsaleh, line 11: this logic can be replaced with the follow (and then be part of the secret definition instead)
```
annotations := map[string]string{
  "kots.io/airgap": fmt.Sprintf("%t", isAirgap),
}
```

web/src/Root.jsx
```diff
@@ -128,6 +128,38 @@ class Root extends Component {
...
+          "Content-Type": "application/json",
+        },
+        method: "GET",
+      });
+      if (!res.ok) {
+        if (res.status === 401) {
+          Utilities.logoutUser();
+          return;
+        }
+        if (res.status === 404) {
```
salahalsaleh, line 145: can we have this logged as an "unexpected status" as well? no need for this check

web/src/Root.jsx
```diff
@@ -328,7 +360,7 @@ class Root extends Component {
                   }} />
                   <ProtectedRoute path="/:slug/preflight" render={props => <PreflightResultPage {...props} logo={this.state.appLogo} appName={this.state.selectedAppName} appsList={appsList} fromLicenseFlow={true} refetchAppsList={this.getAppsList} />} />
                   <ProtectedRoute exact path="/:slug/config" render={props => <AppConfig {...props} fromLicenseFlow={true} refetchAppsList={this.getAppsList} />} />
-                  <Route exact path="/secure-console" render={props => <SecureAdminConsole {...props} logo={this.state.appLogo} appName={this.state.selectedAppName} onLoginSuccess={this.getAppsList} fetchingMetadata={this.state.fetchingMetadata} />} />
+                  <Route exact path="/secure-console" render={props => <SecureAdminConsole {...props} logo={this.state.appLogo} appName={this.state.selectedAppName} pendingApp={this.getPendingApps} onLoginSuccess={this.getAppsList} fetchingMetadata={this.state.fetchingMetadata} />} />
```
salahalsaleh, line 363: can we name this as if it's a function? something like `getPendingApp`?

web/src/components/SecureAdminConsole.jsx
```diff
@@ -34,10 +34,13 @@ class SecureAdminConsole extends React.Component {
...
 
         const apps = await this.props.onLoginSuccess();
+        const pendingApps = await this.props.pendingApp();
         this.setState({ authLoading: false });
         if (apps.length > 0) {
           this.props.history.replace(`/app/${apps[0].slug}`);
-        } else {
+        }else if (pendingApps?.slug) {
+          this.props.history.replace(`/${pendingApps.slug}/airgap`);
+        }else {
```
salahalsaleh, line 43:
```suggestion
        } else {
```

web/src/components/SecureAdminConsole.jsx
```diff
@@ -34,10 +34,13 @@ class SecureAdminConsole extends React.Component {
         }
 
         const apps = await this.props.onLoginSuccess();
+        const pendingApps = await this.props.pendingApp();
         this.setState({ authLoading: false });
         if (apps.length > 0) {
           this.props.history.replace(`/app/${apps[0].slug}`);
-        } else {
+        }else if (pendingApps?.slug) {
```
salahalsaleh, line 41:
```suggestion
        } else if (pendingApps?.slug) {
```

salahalsaleh, line 41: again, it's a single app, not apps, so the name `pendingApps` is misleading here

pkg/kotsadm/objects/license_objects.go
jala-dx: looks neat, but the annotations seems to be map[string]string- https://pkg.go.dev/k8s.io/apimachinery@v0.20.4/pkg/apis/meta/v1#ObjectMeta.Annotations

salahalsaleh: the sprintf will convert it to a string

web/src/Root.jsx
jala-dx: this is to address a corner case where the license got deleted in the DB.

pkg/kotsadm/objects/license_objects.go
jala-dx: yes indeed:) https://play.golang.org/p/d0sV-jkzPXn fixed now

divolgin (approved):

divolgin (approved):

web/src/components/SecureAdminConsole.jsx
```diff
@@ -34,9 +34,12 @@ class SecureAdminConsole extends React.Component {
         }
 
         const apps = await this.props.onLoginSuccess();
+        const pendingApp = await this.props.pendingApp();
```
salahalsaleh, line 37: this function name is confusing, does indicate if it's a function or a variable
```suggestion
        const pendingApp = await this.props.getPendingApp();
```

web/src/Root.jsx
```diff
@@ -128,6 +128,38 @@ class Root extends Component {
...
+          return;
+        }
+ 
+        console.log("failed to get pending apps, unexpected status code", res.status);
+        return;
+      }
+      const response = await res.json();
+      const app = response.app;
+      this.setState({
+        pendingapp: app,
```
salahalsaleh, line 155: this is not used

web/src/Root.jsx
```diff
@@ -128,6 +128,38 @@ class Root extends Component {
...
+      if (!res.ok) {
+        if (res.status === 401) {
+          Utilities.logoutUser();
+          return;
+        }
+        if (res.status === 404) {
+          return;
+        }
+ 
+        console.log("failed to get pending apps, unexpected status code", res.status);
```
salahalsaleh, line 149:
```suggestion
        console.log("failed to get pending app, unexpected status code", res.status);
```

salahalsaleh (approved):

PR #1714, opened 2021-04-06
read-only registry / disable image push

divolgin:


cmd/kots/cli/admin-console-push-images.go
```diff
@@ -56,6 +56,7 @@ func AdminPushImagesCmd() *cobra.Command {
 			}
 
 			options := kotsadmtypes.PushImagesOptions{
+				KotsadmTag: v.GetString("kotsadm-tag"),
```
divolgin, line 59: add for local testing

pkg/store/kotsstore/version_store.go
```diff
@@ -353,11 +352,6 @@ func (s *KOTSStore) createAppVersion(tx *sql.Tx, appID string, currentSequence *
 		previousArchiveDir = previousDir
 	}
 
-	registryInfo, err := s.GetRegistryDetailsForApp(appID)
```
divolgin, line 356: none of this was used for anything

pkg/store/ocistore/version_store.go
```diff
@@ -341,11 +340,6 @@ func (s *OCIStore) CreateAppVersion(appID string, currentSequence *int64, filesI
 		previousArchiveDir = previousDir
 	}
 
-	registryInfo, err := s.GetRegistryDetailsForApp(appID)
```
divolgin, line 344: none of this was used for anything

salahalsaleh:
Overview understanding of the PR: This PR will add the ability to disable pushing application images to the provided registry when installing an application using either the CLI or the UI, and it assumes that the customer will take care of pushing the images manually before the installation process. This PR will also add the ability to disable pushing the images when changing the registry using the UI, and it also assumes that the customer will handle that manually before doing so, or before deploying the new version.

cmd/kots/cli/upstream-upgrade.go
```diff
@@ -99,9 +99,16 @@ func UpstreamUpgradeCmd() *cobra.Command {
 				}
 
 				imagesRootDir := filepath.Join(airgapRootDir, "images")
```
salahalsaleh, line 101: this can be moved to be inside this if statement here https://github.com/replicatedhq/kots/pull/1714/files#diff-450f22c924117dc9b5e725632d49bd946038c4d9505484ee4c72a72f01fd168cR102

pkg/airgap/airgap.go
```diff
@@ -33,12 +33,24 @@ import (
 	"k8s.io/client-go/kubernetes/scheme"
 )
 
+type CreateAirgapAppOpts struct {
```
salahalsaleh, line 36: <3

salahalsaleh, line 36: ❤️

pkg/handlers/license.go
```diff
@@ -342,6 +343,7 @@ func (h *Handler) UploadNewLicense(w http.ResponseWriter, r *http.Request) {
 		uploadLicenseResponse.Success = true
 		uploadLicenseResponse.Slug = a.Slug
 		uploadLicenseResponse.NeedsRegistry = false
+		uploadLicenseResponse.RegistryIsReadOnly = false
```
salahalsaleh, line 346: why always `false`? isn't this still supported in online installations?

pkg/image/builder.go
```diff
@@ -346,7 +351,7 @@ func copyOneImage(srcRegistry, destRegistry registry.RegistryOptions, image stri
 		}
 	}
 
-	if dryRun {
+	if !copyImages {
```
salahalsaleh, line 354: would've been nicer to not have to pass this flag to this function since the function name is already `copyOneImage`, and it's a bit confusing that in some cases it doesn't _actually_ copy the image but only rewrite it. i understand that the previous logic in this function is needed here, but that could be moved to another common function i guess.

pkg/kotsadm/push_images.go
```diff
@@ -61,20 +69,35 @@ func ExtractAirgapImages(archive string, destDir string, progressWriter io.Write
...
 
+		if header.Name == "." {
+			continue
+		}
+
 		if err != nil {
 			return errors.Wrap(err, "failed to read tar header")
 		}
 
+		if excludeImages && header.Typeflag == tar.TypeDir {
```
salahalsaleh, line 87: IIRC, `tar.TypeDir` will only indicate if it's a directory but not if the file is "in a directory". e.g.:

/path/to -> header.Typeflag will be tar.TypeDir
/path/to/file -> header.Typeflag might/will _not_ be tar.TypeDir

pkg/store/kotsstore/version_store.go
salahalsaleh: oof

pkg/store/kotsstore/version_store.go
```diff
@@ -366,35 +360,6 @@ func (s *KOTSStore) createAppVersion(tx *sql.Tx, appID string, currentSequence *
...
-		configValuesSpec, err := kotsKinds.Marshal("kots.io", "v1beta1", "ConfigValues")
-		if err != nil {
-			return int64(0), errors.Wrap(err, "failed to marshal configvalues spec")
-		}
-		identityConfigSpec, err := kotsKinds.Marshal("kots.io", "v1beta1", "IdentityConfig")
-		if err != nil {
-			return int64(0), errors.Wrap(err, "failed to marshal identityconfig spec")
-		}
-
-		configOpts := kotsconfig.ConfigOptions{
```
salahalsaleh, line 386: i remember this being used for checking if the config contains any required items that are not set. where's that logic now?

pkg/store/ocistore/version_store.go
```diff
@@ -354,35 +348,6 @@ func (s *OCIStore) CreateAppVersion(appID string, currentSequence *int64, filesI
 	for _, d := range downstreams {
 		// there's a small chance this is not optimal, but no current code path
 		// will support multiple downstreams, so this is cleaner here for now
-		licenseSpec, err := kotsKinds.Marshal("kots.io", "v1beta1", "License")
```
salahalsaleh, line 357: same here

web/src/components/shared/AirgapRegistrySettings.jsx
```diff
@@ -145,10 +148,17 @@ class AirgapRegistrySettings extends Component {
   handleFormChange = (field, val) => {
     let nextState = {};
     nextState[field] = val;
+
+    if (this.props.app?.isAirgap && field === "isReadOnly" && val === false) {
```
salahalsaleh, line 152: js could be tricky that way, how about this?
```suggestion
    if (this.props.app?.isAirgap && field === "isReadOnly" && !val) {
```

pkg/handlers/license.go
divolgin: good catch. it's not used. we get registry info on the next page separately.

pkg/image/builder.go
divolgin: renamed it....

pkg/kotsadm/push_images.go
divolgin: right, that's why `isFileInDir` variable is used. technically we don't need to continue once we see a directory, but just in case that changes.

pkg/store/kotsstore/version_store.go
divolgin: i didn't remove any code here.  this is all unused variables.

pkg/store/ocistore/version_store.go
divolgin: same here :)

pkg/store/kotsstore/version_store.go
salahalsaleh: after discussion: this needs to be put back and modified to fix required config items in a separate pr

pkg/image/builder.go
salahalsaleh: better :)

salahalsaleh (approved):

salahalsaleh (approved):

PR #1715, opened 2021-04-06
fix(kotsadm): missing azure snapshot option in settings UI

DanStough:
Fix for the Azure Velero plugin name. I opted to modify the existing entry instead of adding a new one since I think we are recommending >1.5.

Also changed the ordering to alphabetical per @AmberAlston (also works because cloud providers end up first).


DanStough:
Screenshot
image

jgruica (approved):

PR #1717, opened 2021-04-06
Add host flags to join scripts

areed:
This allows all host preflights to run.


salahalsaleh (approved):

salahalsaleh:
Overview understanding of the PR: this PR will add primary/secondary host flags to join scripts displayed in the UI because they enable additional preflight checks in kURL, and they are optional.

PR #1721, opened 2021-04-07
require configuration before deployment when there are non-set required config items

salahalsaleh:
this pr will add (put back) the ability to block the deployment of new versions if they have non-set required config items until these items are configured. this pr will preserve the functionality of showing the config screen by default during the installation process, even if all required config items are set and have values (except for automated installs, which might still override/bypass this behavior later if no required items are unset).

### DIFF
--- a/pkg/online/online.go
+++ b/pkg/online/online.go
@@ -19,7 +19,6 @@ import (
 	"github.com/replicatedhq/kots/pkg/preflight"
 	"github.com/replicatedhq/kots/pkg/pull"
 	"github.com/replicatedhq/kots/pkg/redact"
-	registrytypes "github.com/replicatedhq/kots/pkg/registry/types"
 	"github.com/replicatedhq/kots/pkg/reporting"
 	"github.com/replicatedhq/kots/pkg/store"
 	"github.com/replicatedhq/kots/pkg/supportbundle"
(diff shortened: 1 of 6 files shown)


divolgin:
This change will block app version from being created if config has any unset required items.  Version creation will be terminated in a transactional manner as supported at this time.

pkg/store/kotsstore/version_store.go
```diff
@@ -362,11 +368,21 @@ func (s *KOTSStore) createAppVersion(tx *sql.Tx, appID string, currentSequence *
 		// will support multiple downstreams, so this is cleaner here for now
 
 		downstreamStatus := "pending"
-		if currentSequence == nil && kotsKinds.Config != nil {
+		if currentSequence == nil && kotsKinds.Config != nil { // always block on config (if exists) for the initial version even if all required items are already set and have values
 			downstreamStatus = "pending_config"
 		} else if kotsKinds.Preflight != nil && !skipPreflights {
 			downstreamStatus = "pending_preflight"
 		}
+		if currentSequence != nil { // only block and check if the version needs configuration for later versions (not the initial one) since the config is always required for the initial version
```
divolgin, line 376: OCI stores need the same

pkg/online/online.go
```diff
@@ -185,38 +186,11 @@ func CreateAppFromOnline(pendingApp *types.PendingApp, upstreamURI string, isAut
...
-
-		configOptions := kotsadmconfig.ConfigOptions{
-			ConfigSpec:         configSpec,
-			ConfigValuesSpec:   configValuesSpec,
-			LicenseSpec:        licenseSpec,
-			IdentityConfigSpec: identityConfigSpec,
-			// TODO: are there ever registry settings here?
-		}
-		needsConfig, err := kotsadmconfig.NeedsConfiguration(configOptions)
+		registrySettings := registrytypes.RegistrySettings{} // TODO: are there ever registry settings here?
```
divolgin, line 189: `store.GetStore().GetRegistryDetailsForApp(appID)` will return this.  If registry info is used in a template, in a required config item, it will be blank without registry info.

divolgin (approved):

PR #1722, opened 2021-04-07
feat(cli): commands for configuring snapshots storage destination

DanStough:
Mostly @salahalsaleh's work, but had some git, err, "issues" with my rebase and it was easier for me to cherrypick on a new branch.

Added CLI support for configuring backup before app is installed for AWS, GCP and Azure providers. I broke Azure out into subcommands because I anticipate having to add other auth methods from the docs that require different fields. I'd like to do this for AWS and GCP too depending on input.

### DIFF
--- a/cmd/kots/cli/velero.go
+++ b/cmd/kots/cli/velero.go
@@ -3,6 +3,7 @@ package cli
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"strings"
 	"time"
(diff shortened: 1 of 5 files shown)


cmd/kots/cli/velero.go
```diff
@@ -139,7 +621,6 @@ func VeleroConfigureNFSCmd() *cobra.Command {
 
 	cmd.Flags().String("nfs-path", "", "the path that is exported by the NFS server")
 	cmd.Flags().String("nfs-server", "", "the hostname or IP address of the NFS server")
-	cmd.Flags().StringP("namespace", "n", "", "the namespace in which kots/kotsadm is installed")
```
DanStough, line 142: These are masked by the global flag.

cmd/kots/cli/velero.go
```diff
@@ -77,12 +87,484 @@ func VeleroEnsurePermissionsCmd() *cobra.Command {
...
+			if veleroStatus == nil {
+				return errors.New("velero namespace not found")
+			}
+
+			if !veleroStatus.ContainsPlugin("velero-plugin-for-aws") {
+				return errors.New("velero does not have the 'velero-plugin-for-aws' installed; " +
+					"consult https://kots.io/kotsadm/snapshots/overview/ for install instructions`)")
+			}
+
+			if v.GetString("bucket") == "" {
```
divolgin, line 197: how about `MarkFlagRequired`?

DanStough: Yeah, I thought about that but turns out it's just for shell completion. I decided not to add this since I anticipate the shell completion won't play well with a krew plugin. Let me know if I'm mistaken though on either count.

DanStough: Updated per our slack convo!

divolgin (approved):

PR #1723, opened 2021-04-07
Back off deploy loop error

emosbaugh:


divolgin:
This PR creates a global map of loggers for each app that deploy loop uses to keep track of last time an error was logged.  Errors are considered the same if the error text matches exactly.  And only last error is remembered, so if error changes every time, every error will be logged.  The period of silence is doubled every time the error is logged until the max of 30 minutes is reached.

divolgin (approved):

pkg/socketservice/socketservice.go
```diff
@@ -144,33 +147,53 @@ func deployLoop() {
...
+				}
+				appDeployLoopErrorBackoff[a.ID].OnError(err, func() {
+					logger.Error(errors.Wrapf(err, "failed to run deploy loop for app %s in cluster %s", a.ID, clusterSocket.ClusterID))
+				})
+			} else if deployed {
+				logger.Infof("Deploy success for app %s in cluster %s", a.ID, clusterSocket.ClusterID)
 			}
 		}
+
+		time.Sleep(5 * time.Second)
```
salahalsaleh, line 164: why override the seconds here? and why 5? it's already defined in this step https://github.com/replicatedhq/kots/pull/1723/files#diff-0a0a367f8346de283aa12dd1378bea8f6e6aa93d025aec20683bbf757b912fe7R123

pkg/socketservice/socketservice.go
```diff
@@ -144,33 +147,53 @@ func deployLoop() {
...
+	if err := deployVersionForApp(clusterSocket, a, deployedVersion); err != nil {
+		return false, errors.Wrap(err, "failed to deploy version")
 	}
+	return true, nil
+}
 
-	if value, ok := clusterSocket.LastDeployedSequences[a.ID]; ok && value == deployedVersion.ParentSequence {
-		// this version is already the currently deployed version
-		return nil
+func getDeployedVersionForApp(clusterSocket *ClusterSocket, a *apptypes.App) (*downstreamtypes.DownstreamVersion, error) {
```
salahalsaleh, line 189: i don't understand why this function is needed. also, returning `nil` if there's a restore in progress seems unintuitive given the function name.

pkg/socketservice/socketservice.go
emosbaugh: i meant to take this out. thanks

pkg/socketservice/socketservice.go
emosbaugh: i removed the extra function

salahalsaleh (approved):

PR #1754, opened 2021-04-08
Kubernetes dependency v0.20.5

emosbaugh:


emosbaugh:
Resolves #1615 #1639 #1641 #1644

emosbaugh:
Backoff didnt reset the time when outside of the backoff period.

emosbaugh:
Holding off on this because it will introduce conflicts

PR #1761, opened 2021-04-08
Remove dockerfile web

emosbaugh:
We no longer deploy the web as a separate component. It's built and then copied in to the api container and served there.


emosbaugh:
Resolves #1698

divolgin (approved):

PR #1763, opened 2021-04-09
registry image preflights

divolgin:
Adding Private Image check that will show up when read-only private registry is used.

This uses similar pattern to injected support bundle collectors.  When there is a private registry read-only configured, we add a new collector (`Registry`) and analyzer (`RegistryAnalyze`).   The images are populated from the `knownImages` field in the Installation spec, which includes all images found in release spec and was created for a similar purpose in support bundle.  Image pull secret is also generated from registry settings. 

Screen Shot 2021-04-08 at 5 28 31 PM

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -37,9 +37,9 @@ require (
 	github.com/go-logfmt/logfmt v0.5.0
 	github.com/go-logr/logr v0.3.0 // indirect
 	github.com/go-logr/zapr v0.2.0 // indirect
+	github.com/go-test/deep v1.0.7
 	github.com/gofrs/uuid v3.3.0+incompatible // indirect
 	github.com/golang/mock v1.4.4
-	github.com/google/go-containerregistry v0.4.1 // indirect
 	github.com/google/gofuzz v1.1.0
 	github.com/google/uuid v1.1.2
 	github.com/googleapis/gnostic v0.5.1 // indirect
@@ -64,7 +64,7 @@ require (
 	github.com/pierrec/lz4 v2.2.6+incompatible // indirect
 	github.com/pkg/errors v0.9.1
 	github.com/replicatedhq/kurl/kurlkinds v0.0.0-20210227025942-3b373149acb7
-	github.com/replicatedhq/troubleshoot v0.10.12
+	github.com/replicatedhq/troubleshoot v0.10.16
 	github.com/replicatedhq/yaml/v3 v3.0.0-beta5-replicatedhq
 	github.com/robfig/cron v1.1.0
 	github.com/robfig/cron/v3 v3.0.1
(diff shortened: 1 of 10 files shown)


go.mod
```diff
@@ -140,4 +140,7 @@ replace (
 	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.20.2
 	k8s.io/sample-cli-plugin => k8s.io/sample-cli-plugin v0.20.2
 	k8s.io/sample-controller => k8s.io/sample-controller v0.20.2
+	sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.6.5
 )
+
+replace github.com/replicatedhq/troubleshoot => ../troubleshoot
```
divolgin, line 146: to be removed when new troubleshoot release is out

pkg/preflight/preflight.go
```diff
@@ -303,3 +304,47 @@ func CreateRenderedSpec(appID string, sequence int64, origin string, inCluster b
...
+				{
+					Fail: &troubleshootv1beta2.SingleOutcome{
+						When:    "missing > 0",
+						Message: "Application uses images that cannot be found in the private registry",
+					},
+				},
+				{
+					Warn: &troubleshootv1beta2.SingleOutcome{
+						When:    "errors > 0",
+						Message: "Availability of application images in the private registru could not be verified.",
```
salahalsaleh, line 339:
```suggestion
						Message: "Availability of application images in the private registry could not be verified.",
```

pkg/store/ocistore/app_store.go
```diff
@@ -211,15 +211,14 @@ func (s *OCIStore) CreateApp(name string, upstreamURI string, licenseData string
...
 	app := apptypes.App{
-		ID:                 id,
-		Name:               name,
-		IconURI:            "",
-		CreatedAt:          time.Now(),
-		Slug:               slugProposal,
-		UpstreamURI:        upstreamURI,
-		License:            licenseData,
-		InstallState:       installState,
-		RegistryIsReadOnly: registryIsReadOnly,
```
salahalsaleh, line 222: should be removed from the other store as well

pkg/preflight/preflight.go
```diff
@@ -303,3 +304,47 @@ func CreateRenderedSpec(appID string, sequence int64, origin string, inCluster b
...
+	}
+
+	preflight.Spec.Collectors = append(preflight.Spec.Collectors, &troubleshootv1beta2.Collect{
+		Registry: &troubleshootv1beta2.Registry{
+			Images: images,
+		},
+	})
+	preflight.Spec.Analyzers = append(preflight.Spec.Analyzers, &troubleshootv1beta2.Analyze{
+		Registy: &troubleshootv1beta2.RegistryAnalyze{
+			AnalyzeMeta: troubleshootv1beta2.AnalyzeMeta{
```
divolgin, line 326: @AmberAlston need a review for this text in this analyzer.  This is the last row in the screenshot attached in description.

AmberAlston: I'm good with the current description text. I'd like to see "Private Images" updated to "Private Registry Images Available"

salahalsaleh (approved):

salahalsaleh (approved):

PR #1766, opened 2021-04-10
remove namespace from registry in secrets

divolgin:


salahalsaleh (approved):

DanStough:
FWIW I also ended up testing this and the secrets look good now both was you can install (w and w/o `kotsadm-namespace`)

DanStough (approved):

PR #1773, opened 2021-04-14
make cache instead of make kotsadm

divolgin:


salahalsaleh (approved):

salahalsaleh:
This PR will make it so that developers won't have to run `make kotsadm` after every change in the api (a change that was introduced before to reduce the amount of time it takes skaffold to rebuild images even when there were no changes to the kotsadm image). Instead, this moves all the time consuming (not as frequently changed) logic which includes downloading binaries and dependencies from the kotsadm dev Dockerfile to a separate "cache" Dockerfile whose image will only need to be updated with `make cache` when the relevant binaries or dependencies change (which happens much less frequently and has less side effects/consequences when not updated).

PR #1785, opened 2021-04-19
Required registry namespace

jgruica:
Screen Shot 2021-04-19 at 9 58 29 AM

### DIFF
--- a/pkg/handlers/registry.go
+++ b/pkg/handlers/registry.go
@@ -81,6 +81,14 @@ func (h *Handler) UpdateAppRegistry(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/handlers/registry.go
```diff
@@ -81,6 +81,14 @@ func (h *Handler) UpdateAppRegistry(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if updateAppRegistryRequest.Namespace == "" {
+		err := errors.New("registry namespace is required")
+		logger.Error(err)
+		updateAppRegistryResponse.Error = err.Error()
+		JSON(w, http.StatusInternalServerError, updateAppRegistryResponse)
```
divolgin, line 88: this should be http.StatusBadRequest

PR #1794, opened 2021-04-21
Fix Marshaling EntitlementValue

MikaelSmith:
Fixes marshaling EntitlementValue so that values are correctly marshaled. Without this change values were always an empty object (`{}`). This change fixed it, and seems to be consistent with all examples in https://golang.org/pkg/encoding/json/.

### DIFF
--- a/kotskinds/apis/kots/v1beta1/license_types.go
+++ b/kotskinds/apis/kots/v1beta1/license_types.go
@@ -51,7 +51,7 @@ func (entitlementValue *EntitlementValue) Value() interface{} {
 	return entitlementValue.StrVal
 }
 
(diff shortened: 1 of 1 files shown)


laverya (approved):

MikaelSmith:
https://github.com/golang/go/issues/22967 for context.

laverya:
EntitlementValue is used here: https://github.com/replicatedhq/kots/pull/1794/files#diff-45d1a621cdb46de1abb5419783da11678f70f5eb6cf24c6f75bf4fe130596ca1L96

Which is not a reference, and thus the marshaller interface shouldn't use a reference either

PR #1802, opened 2021-04-26
Use official minio image

divolgin:


cmd/kots/cli/admin-console-upgrade.go
```diff
@@ -108,5 +112,15 @@ func AdminConsoleUpgradeCmd() *cobra.Command {
 	cmd.Flags().MarkHidden("ensure-rbac")
 	cmd.Flags().MarkHidden("airgap-upload-parallelism")
 
+	// options for the alpha feature of using a reg instead of s3 for storage
+	cmd.Flags().String("storage-base-uri", "", "an s3 or oci-registry uri to use for kots persistent storage in the cluster")
```
divolgin, line 116: Parity with install CLI options

pkg/kotsadm/objects/postgres_objects.go
```diff
@@ -31,7 +31,7 @@ func PostgresStatefulset(deployOptions types.DeployOptions, size resource.Quanti
 		}
 	} else if deployOptions.KotsadmOptions.OverrideRegistry != "" {
 		// if there is a registry specified, use the postgres image there and not the one from docker hub - even though there's not a username/password specified
-		image = fmt.Sprintf("%s/postgres:%s", kotsadmversion.KotsadmRegistry(deployOptions.KotsadmOptions), kotsadmversion.KotsadmTag(deployOptions.KotsadmOptions))
+		image = fmt.Sprintf("%s/postgres:%s", kotsadmversion.KotsadmRegistry(deployOptions.KotsadmOptions), strings.Split(image, ":")[1])
```
divolgin, line 34: @laverya tag here should always be the original postgres tag, right?

pkg/snapshot/filesystem.go
```diff
@@ -241,8 +241,7 @@ func ensureFileSystemMinioDeployment(ctx context.Context, clientset kubernetes.I
 }
 
 func fileSystemMinioDeploymentResource(clientset kubernetes.Interface, secretChecksum string, deployOptions FileSystemDeployOptions, registryOptions kotsadmtypes.KotsadmOptions) (*appsv1.Deployment, error) {
-	kotsadmTag := kotsadmversion.KotsadmTag(kotsadmtypes.KotsadmOptions{}) // default tag
-	image := fmt.Sprintf("kotsadm/minio:%s", kotsadmTag)
+	image := "minio/minio:RELEASE.2021-04-06T23-11-00Z"
```
divolgin, line 244: @salahalsaleh I think this is the equivalent of the original code, but how does this work with private registries and airgap?

pkg/kotsadm/objects/postgres_objects.go
laverya: I believe so, but I know it's been the kotsadm tag at certain times in the past.  It needs to match the code in [push-images](https://kots.io/kots-cli/admin-console/push-images/), for sure. (and if you make it use the actual upstream tag, you'll need to ensure that CI pushes that tag to ttl.sh and not `2h`)

pkg/snapshot/filesystem.go
salahalsaleh: @divolgin https://github.com/replicatedhq/kots/pull/1802/files?w=1#diff-3ee38938c503a8ac3c5e47bf4596b2936ef10e22475a40a457123f20844985dcR249

pkg/kotsadm/main.go
```diff
@@ -225,10 +229,6 @@ func Deploy(deployOptions types.DeployOptions) error {
 		}
 	}
 
-	if err := ensureStorage(deployOptions, clientset, log); err != nil {
```
divolgin, line 228: ensureStorage now calls this

salahalsaleh (approved):

PR #1806, opened 2021-04-27
Move code with SQL to KOTSStore

divolgin:


pkg/store/store.go
```diff
@@ -10,6 +11,7 @@ var (
 )
 
 var _ Store = (*kotsstore.KOTSStore)(nil)
+var _ Store = (*ocistore.OCIStore)(nil)
```
divolgin, line 14: enforce interface implementation

salahalsaleh (approved):

salahalsaleh (approved):

PR #1807, opened 2021-04-29
feat(cli): add support for self-signed s3 providers for snapshots

DanStough:


pkg/s3/s3.go
```diff
@@ -90,18 +91,19 @@ func CreateS3BucketUsingAPod(ctx context.Context, clientset kubernetes.Interface
...
-		if err := json.Unmarshal([]byte(line), &createBucketPodOutput); err != nil {
+		if logError = json.Unmarshal([]byte(line), &createBucketPodOutput); logError != nil {
 			continue
 		}
 
 		break
 	}
 
-	if !createBucketPodOutput.Success {
-		return errors.Wrapf(err, "failed to create bucket, please check %s pod logs for more details", createBucketPod.Name)
```
DanStough, line 104: This error was out of scope, so there wouldn't be any failure reported and the pod would stick around

deploy/s3-bucket-head.sh
```diff
@@ -1,14 +1,25 @@
...
+fi
+
+TMP_S3_HOST=$(echo "$TMP_S3_ENDPOINT" | awk -F/ '{print $3}')
+export TMP_S3_HOST
+export TMP_S3_S3CMD_FLAGS="--access_key=$TMP_S3_ACCESS_KEY_ID --secret_key=$TMP_S3_SECRET_ACCESS_KEY --host=$TMP_S3_HOST --host-bucket=$TMP_S3_BUCKET_NAME.$TMP_S3_HOST $TMP_S3_S3CMD_SSL_FLAGS"
 
-if s3cmd $TMP_S3_S3CMD_FLAGS ls s3://$TMP_S3_BUCKET_NAME 2>&1 | grep -q 'NoSuchBucket'
+echo "Running with flags: --access_key=***** --secret_key=***** --host=$TMP_S3_HOST --host-bucket=$TMP_S3_BUCKET_NAME.$TMP_S3_HOST $TMP_S3_S3CMD_SSL_FLAGS"
+
+if timeout 5 s3cmd $TMP_S3_S3CMD_FLAGS ls s3://$TMP_S3_BUCKET_NAME
```
DanStough, line 19: Added the timeout here. If you type in the wrong IP address s3cmd will block and keep retrying.

cmd/kots/cli/velero.go
```diff
@@ -352,6 +353,19 @@ func VeleroConfigureOtherS3Cmd() *cobra.Command {
 				return errors.New("velero namespace not found")
 			}
 
+			var caCertData []byte
+			caCertFile := v.GetString("cacert")
+			if caCertFile != "" {
+				realPath, err := filepath.Abs(caCertFile)
```
salahalsaleh, line 359: why is this needed?

deploy/s3-bucket-head.sh
salahalsaleh: can we increase the timeout a bit? 5 seconds seem very low

pkg/s3/s3.go
```diff
@@ -90,18 +91,19 @@ func CreateS3BucketUsingAPod(ctx context.Context, clientset kubernetes.Interface
...
+	if !createBucketPodOutput.Success || logError != nil {
```
salahalsaleh, line 105: won't `logError` always be NOT nil? since we're echoing invalid json in the script and unmarshalling these messages will fail?

pkg/s3/s3.go
```diff
@@ -142,18 +144,19 @@ func HeadS3BucketUsingAPod(ctx context.Context, clientset kubernetes.Interface,
...
+		if logError = json.Unmarshal([]byte(line), &headBucketPodOutput); logError != nil {
 			continue
 		}
 
 		break
 	}
 
-	if !headBucketPodOutput.Success {
-		return errors.Wrapf(err, "failed to head bucket, please check %s pod logs for more details", headBucketPod.Name)
+	if !headBucketPodOutput.Success || logError != nil {
```
salahalsaleh, line 158: ditto

cmd/kots/cli/velero.go
DanStough: TBH I got this directly from the Velero CLI. My assumption is that it's because `Abs` also runs [Clean](https://golang.org/pkg/path/filepath/#Clean) which can handle some corner case inputs. I tested and it works without, but I'm inclined to keep unless there are strong objections.

deploy/s3-bucket-head.sh
DanStough: 15 per our slack thread 👍

cmd/kots/cli/velero.go
salahalsaleh: sounds good

pkg/s3/s3.go
```diff
@@ -142,18 +144,19 @@ func HeadS3BucketUsingAPod(ctx context.Context, clientset kubernetes.Interface,
 	headBucketPodOutput := HeadBucketPodOutput{}
 
 	scanner := bufio.NewScanner(bytes.NewReader(logs))
+	var logError error
```
salahalsaleh, line 147: i believe this var isn't needed anymore, right?

pkg/s3/s3.go
```diff
@@ -90,18 +91,19 @@ func CreateS3BucketUsingAPod(ctx context.Context, clientset kubernetes.Interface
 	createBucketPodOutput := CreateBucketPodOutput{}
 
 	scanner := bufio.NewScanner(bytes.NewReader(logs))
+	var logError error
```
salahalsaleh, line 94: ditto

salahalsaleh (approved):

salahalsaleh (approved):

PR #1810, opened 2021-04-29
Allow editing deployed version config

divolgin:
Allow editing currently deployed and latest version.  Editing in this context means config can be edited and when saved, it will result in a new version being created.  Only version 0 can be modified directly during the initial install.

### DIFF
--- a/web/src/components/apps/AppConfig.jsx
+++ b/web/src/components/apps/AppConfig.jsx
@@ -400,22 +400,13 @@ class AppConfig extends Component {
     }
   }
 
-  checkIsCurrentOrPastVersion = (app) => {
+  isConfigReadOnly = (app) => {
     const { match } = this.props;
     if (!match.params.sequence) return false;
     const sequence = parseInt(match.params.sequence);
-    let latestSequence;
-    if (app?.downstreams[0]?.pendingVersions.length > 0) {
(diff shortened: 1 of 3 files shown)


web/src/components/apps/AppConfig.jsx
```diff
@@ -400,22 +400,13 @@ class AppConfig extends Component {
...
-      latestSequence = app?.downstreams[0]?.currentVersion?.parentSequence;
-    }
-
-    if (sequence < latestSequence) {
-      return true;
-    } else {
-      return false;
-    }
+    const isCurrentVersion = app.downstreams[0]?.currentVersion?.sequence === sequence;
+    const isLatestVersion = app.currentVersion.sequence == sequence;
```
GraysonNull, line 408: loose equal here vs strict?

web/src/components/apps/AppVersionHistoryRow.jsx
```diff
@@ -73,6 +65,7 @@ function renderVersionAction(version, latestVersion, nothingToCommitDiff, app, h
   }
 
   const isCurrentVersion = version.sequence === downstream.currentVersion?.sequence;
+  const isLatestVersion = version.sequence == latestVersion.sequence;
```
GraysonNull, line 68: same here.

salahalsaleh (approved):

GraysonNull (approved):

PR #1813, opened 2021-04-29
Ability to update registry and license after a deployed version is edited

divolgin:
After an old deployed version is edited, this enables license sync to get the license up to date and saving registry to point old version at the right registry.

### DIFF
--- a/pkg/handlers/config.go
+++ b/pkg/handlers/config.go
@@ -21,7 +21,9 @@ import (
 	kotsadmconfig "github.com/replicatedhq/kots/pkg/kotsadmconfig"
 	"github.com/replicatedhq/kots/pkg/kotsutil"
 	"github.com/replicatedhq/kots/pkg/logger"
+	"github.com/replicatedhq/kots/pkg/midstream"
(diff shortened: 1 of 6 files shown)


pkg/handlers/config.go
```diff
@@ -301,6 +318,26 @@ func (h *Handler) CurrentAppConfig(w http.ResponseWriter, r *http.Request) {
...
+		return true, nil
+	}
+
+	downstreams, err := store.GetStore().ListDownstreamsForApp(app.ID)
+	if err != nil {
+		return false, errors.Wrap(err, "failed to get downstreams")
+	}
+
+	for _, d := range downstreams {
+		if d.CurrentSequence == sequence {
```
salahalsaleh, line 333:
```suggestion
		if d.ParentSequence == sequence {
```

divolgin: There's no ParentSequence in this one.  I think that is the right field.  It's just sequence: https://github.com/replicatedhq/kots/blob/051058090f62305445d9761930eb9b0775839774/pkg/version/version.go#L104

salahalsaleh (approved):

PR #1819, opened 2021-05-05
Check for deployed sequence before reporting preflights

jgruica:
**Problems:** 
- Go routine was running after new version was deployed, and it was reporting “missing” status for older versions. 
- Sending preflight reporting data was called before version is deployed.

**Fixes:**
 - Added check to see is deployed sequence changed.
 - Send preflight reporting data after version is deployed.

### DIFF
--- a/pkg/handlers/deploy.go
+++ b/pkg/handlers/deploy.go
@@ -96,7 +96,7 @@ func (h *Handler) DeployAppVersion(w http.ResponseWriter, r *http.Request) {
 	// preflights reports
 	go func() {
 		if request.IsSkipPreflights || request.ContinueWithFailedPreflights {
-			if err := reporting.SendPreflightInfo(a.ID, int64(sequence), request.IsSkipPreflights, request.IsCLI); err != nil {
+			if err := reporting.ReportAppInfo(a.ID, int64(sequence), request.IsSkipPreflights, request.IsCLI); err != nil {
 				logger.Debugf("failed to send preflights data to replicated app: %v", err)
 				return
 			}
--- a/pkg/handlers/deploy.go
+++ b/pkg/handlers/deploy.go
@@ -41,7 +41,6 @@ type UpdateUndeployResultRequest struct {
 type DeployAppVersionRequest struct {
 	IsSkipPreflights             bool `json:"isSkipPreflights"`
 	ContinueWithFailedPreflights bool `json:"continueWithFailedPreflights"`
-	IsAirgap                     bool `json:"isAirgap"`
 	IsCLI                        bool `json:"isCli"`
 }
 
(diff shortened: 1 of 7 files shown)


pkg/reporting/preflight.go
```diff
@@ -75,6 +75,18 @@ func SendPreflightInfo(appID string, sequence int, isSkipPreflights bool, isCLI
...
 				return
 			}
+
+			parentSequence, err := store.GetStore().GetCurrentParentSequence(appID, clusterID)
+			if err != nil {
+				logger.Debugf("failed to get downstream parent sequence: %v", err.Error())
+				return
+			}
+
+			if (parentSequence != int64(sequence)) {
```
salahalsaleh, line 85:
```suggestion
			if parentSequence != int64(sequence) {
```

pkg/reporting/preflight.go
```diff
@@ -75,6 +75,18 @@ func SendPreflightInfo(appID string, sequence int, isSkipPreflights bool, isCLI
...
+				logger.Debugf("deployed sequence has changed")
```
salahalsaleh, line 86:
```suggestion
				logger.Debug("deployed sequence has changed")
```

pkg/preflight/preflight.go
```diff
@@ -143,32 +145,37 @@ func Run(appID string, appSlug string, sequence int64, isAirgap bool, archiveDir
...
 
 	logger.Debug("automatically deploying first app version")
 
 	// note: this may attempt to re-deploy the first version but the operator will take care of
 	// comparing the version to current
 
-	return version.DeployVersion(appID, sequence)
+	err = version.DeployVersion(appID, sequence)
+	if err !=  nil {
+		return false, errors.Wrap(err, "no deploted version")
```
salahalsaleh, line 175:
```suggestion
		return false, errors.Wrap(err, "failed to deploy version")
```

salahalsaleh (approved):

pkg/handlers/deploy.go
```diff
@@ -41,7 +41,6 @@ type UpdateUndeployResultRequest struct {
 type DeployAppVersionRequest struct {
 	IsSkipPreflights             bool `json:"isSkipPreflights"`
 	ContinueWithFailedPreflights bool `json:"continueWithFailedPreflights"`
-	IsAirgap                     bool `json:"isAirgap"`
```
divolgin, line 44: Will this now try to report in airgapped installs?

pkg/preflight/preflight.go
```diff
@@ -108,20 +108,22 @@ func Run(appID string, appSlug string, sequence int64, isAirgap bool, archiveDir
...
 			if err != nil {
 				err = errors.Wrap(err, "failed to deploy first version")
 				logger.Error(err)
 				return
 			}
 
-			if err := reporting.SendPreflightInfo(appID, int(sequence), false, false); err != nil {
-				logger.Debugf("failed to send preflights data to replicated app: %v", err)
-				return
+			if isDeployed {
```
divolgin, line 118: Why is preflight info not reported if version is not deployed?  Deploying is not required for preflight checks to have a status.

pkg/handlers/deploy.go
divolgin: Never mind, I see where this check moved

salahalsaleh: it won't https://github.com/replicatedhq/kots/pull/1819/files#diff-706f7e57e6f7361d82c9a8f1b0c3ddb08c18e88d3ee384cf20964b0ca86254c2R60

pkg/reporting/preflight.go
```diff
@@ -75,7 +85,19 @@ func SendPreflightInfo(appID string, sequence int, isSkipPreflights bool, isCLI
...
 			}
-			if s.Sequence == int64(sequence) && s.State == appstatustypes.StateReady {
+
+			currentDeployedSequence, err := store.GetStore().GetCurrentParentSequence(appID, clusterID)
+			if err != nil {
+				logger.Debugf("failed to get downstream parent sequence: %v", err.Error())
+				return
+			}
+
+			if currentDeployedSequence != sequence {
```
divolgin, line 95: If I'm calling `SendPreflightInfo` for a sequence, I expect it to report.  Why is this check needed?

pkg/preflight/preflight.go
salahalsaleh: the main goal is to report the status of the application when the user skips/waits for preflights when deploying

pkg/reporting/preflight.go
salahalsaleh: since the app can take some time to become ready, during that time the user can deploy another version which will make this whole goroutine irrelevant because it's reading a status for a different version. this check will simply not report that case.

pkg/preflight/preflight.go
salahalsaleh: the naming could be better i agree

divolgin (approved):

PR #1833, opened 2021-05-12
license upload supports installation from multi-part yaml

GraysonNull:


cmd/kots/cli/install.go
```diff
@@ -167,6 +167,8 @@ func InstallCmd() *cobra.Command {
 				Context:                   v.GetString("context"),
 				SharedPassword:            sharedPassword,
 				ApplicationMetadata:       applicationMetadata,
+				AppSlug:                   appSlugFromArgs(args[0]),
```
divolgin, line 170: Seems like `UpstreamURI` is all we need since `args[0]` is the same thing

cmd/kots/cli/install.go
```diff
@@ -651,3 +652,7 @@ func getHttpProxyEnv(v *viper.Viper) map[string]string {
 	return env
 
 }
+
+func appSlugFromArgs(arg string) string {
```
divolgin, line 656: this is not used

PR #1834, opened 2021-05-13
chore: add app slug to application metadata

DanStough:
Adding the app slug to metadata so that for multi-app licenses will know which app to install.

Need to remove the ttl.sh image testing but grayson wanted to test with this branch. Making a draft until it is removed.

### DIFF
--- a/cmd/kots/cli/install.go
+++ b/cmd/kots/cli/install.go
@@ -167,6 +167,7 @@ func InstallCmd() *cobra.Command {
 				Context:                   v.GetString("context"),
 				SharedPassword:            sharedPassword,
 				ApplicationMetadata:       applicationMetadata,
+				UpstreamURI:               upstream,
 				License:                   license,
 				ConfigValues:              configValues,
 				Airgap:                    isAirgap,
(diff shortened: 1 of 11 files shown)


pkg/kotsadm/objects/kotsadm_objects.go
```diff
@@ -618,8 +618,9 @@ func KotsadmDeployment(deployOptions types.DeployOptions) *appsv1.Deployment {
 					},
 					Containers: []corev1.Container{
 						{
-							Image:           fmt.Sprintf("%s/kotsadm:%s", kotsadmversion.KotsadmRegistry(deployOptions.KotsadmOptions), kotsadmversion.KotsadmTag(deployOptions.KotsadmOptions)),
-							ImagePullPolicy: corev1.PullIfNotPresent,
+							// TODO (dans): MUST UNDO THE FOLLOWING TWO LINES!!!
```
divolgin, line 621: yes, indeed

cmd/kots/cli/install.go
```diff
@@ -167,6 +167,7 @@ func InstallCmd() *cobra.Command {
 				Context:                   v.GetString("context"),
 				SharedPassword:            sharedPassword,
 				ApplicationMetadata:       applicationMetadata,
+				AppSlug:                   appSlugFromArgs(args[0]),
```
divolgin, line 170: How about just saving it at as `UpstreamURI` since this could also be a helm chart instead of kots app

divolgin (approved):

PR #1836, opened 2021-05-14
Deploy alpine based postrges on OpenShift 4.2 and later

divolgin:
OpenShift now automatically adds the random user to the passwd file, which is required for intidb to work in alpine image: https://access.redhat.com/articles/4859371

```
The OpenShift run-time CRI-O (starting from OpenShift 4.2 onward) now inserts the random user for the container into /etc/passwd.
Removing the requirement to insert the random user manually into /etc/passwd completely.
```


go.mod
```diff
@@ -85,11 +88,11 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 	gotest.tools v2.2.0+incompatible
 	helm.sh/helm/v3 v3.5.3
-	k8s.io/api v0.20.5
+	k8s.io/api v0.21.0-rc.0
```
emosbaugh, line 91: why rc?
https://github.com/kubernetes/api/tree/v0.21.0

emosbaugh (approved):

PR #1838, opened 2021-05-14
generating default cert if none found

jeffreygolden:
If a bad TLS certificate / key are uploaded, kurl proxy may become unavailable.  Currently there is no recovery from this other than manually creating a new kotsadm-tls secret and restarting the kurl_proxy container.

With this PR, kurl_proxy will check for the kotsadm-tls secret up front.  If it is not found, kurl_proxy will recreate a default self-signed certificate and save it as the kotsadm-tls secret.

To recover from a bad TLS certificate, admins can delete the kotsadm-tls secret and restart the kurl_proxy container.  This will reset TLS back to default settings.

This PR includes a kots command, "kots reset-tls" that will automatically delete the kotsadm-tls secret, restart the kurl_proxy container, and watch for the kotsadm-tls secret to be recreated.


cmd/kots/cli/reset-tls.go
```diff
@@ -0,0 +1,148 @@
...
+			err = checkTLSSecret(namespace, time.Second*10)
+			if err != nil {
+				return err
+			}
+
+			return nil
+		},
+	}
+
+	cmd.Flags().StringP("namespace", "n", "default", "namespace in which kots/kotsadm is installed")
```
divolgin, line 62: This shouldn't be needed.  "namespace" is included with kubectl flags.

cmd/kots/cli/reset-tls.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/replicatedhq/kots/pkg/k8sutil"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	kuberneteserrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func ResetTLSCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:           "reset-tls",
```
divolgin, line 19: "reset-tls [namespace]"

cmd/kots/cli/reset-tls.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+	}
+
+	if existingSecret.Name != "" {
+		err = clientset.CoreV1().Secrets(namespace).Delete(context.TODO(), existingSecret.Name, metav1.DeleteOptions{})
+		if err != nil {
+			return errors.Wrap(err, "failed to delete secret")
+		}
+	} else {
+		fmt.Print("kotsadm-tls secret not found - restarting kurl_proxy pod")
```
divolgin, line 86: I'm not sure this message is needed.  The pod will always be restarted as long as this function returns nil.

kurl_proxy/cmd/main.go
```diff
@@ -78,6 +78,17 @@ func main() {
 	}
 	secrets := clientset.CoreV1().Secrets(namespace)
 
+	_, err = secrets.Get(context.Background(), tlsSecretName, metav1.GetOptions{})
+	if err != nil {
+		log.Print("tls secret not found")
```
divolgin, line 83: if everything works, this will look like an error even if it isn't.  how about "creating tls secret"

cmd/kots/cli/reset-tls.go
```diff
@@ -0,0 +1,148 @@
...
+
+	existingSecret, err := clientset.CoreV1().Secrets(namespace).Get(context.TODO(), "kotsadm-tls", metav1.GetOptions{})
+	if err != nil {
+		if !kuberneteserrors.IsNotFound(err) {
+			return errors.Wrap(err, "failed to lookup secret")
+		}
+	}
+
+	if existingSecret.Name != "" {
+		err = clientset.CoreV1().Secrets(namespace).Delete(context.TODO(), existingSecret.Name, metav1.DeleteOptions{})
```
divolgin, line 81: This secret is mounted in kotsadm.  Does this need additional handling?

https://github.com/replicatedhq/kURL/blob/master/addons/kotsadm/1.41.1/kotsadm.yaml#L71-L73

jeffreygolden: Shouldn't need anything, no.  Deleting the secret doesn't impact the pod in any way, the pod still has the secret mounted and nothing changes until the pod is redeployed.

divolgin: Shouldn't the pod have the new secret?

jeffreygolden: If kurl_proxy boots up and the secret doesn't exist, it creates the self-signed cert and saves it as the kotsadm-tls secret.  All we have to do here is delete the previous secret to remove any problems and redeploy kurl_proxy.

jeffreygolden: References to this in the kots deployment appear to be old artifacts.  They have been removed in this PR.

divolgin (approved):

PR #1852, opened 2021-05-20
Scan postgres images on all branches

divolgin:
All other images are scanned on all branches


DanStough (approved):
Do we need need to worry about forks at all by disabling this? e.g. https://github.com/replicatedhq/kURL/blob/master/.github/workflows/test-addon-pr.yaml#L12

divolgin:
yeah, probably... but only because this flow pushes to dockerhub and forks won't have credentials.   and that's unrelated to this change.

PR #1857, opened 2021-05-26
Jeffgolden/ch33919/kots should detect helm version

jeffreygolden:
Adding a detection method for Helm v3.

All Helm V3 charts should include the key:value "apiVersion: v2" in the Chart.yaml file.  This key is used by Helm itself to validate whether a chart requires Helm v2 or Helm v3, and not providing the key is unsupported by Helm.

If the Helm version is not explicitly provided in HelmChart kind -> helmVersion, Kots will now inspect Chart.yaml when rendering the helm chart.  Kots will use Helm v3 if Chart.yaml -> apiVersion is set to "v2".  If an explicit Helm version is provided this check is skipped and the provided version is always used.

As a note, Helm API version 2 is the same as "Helm v3".

### DIFF
--- a/pkg/base/helm.go
+++ b/pkg/base/helm.go
@@ -151,14 +151,11 @@ func checkChartForVersion(file *upstreamtypes.UpstreamFile) (string, error) {
 	if err != nil {
 		return "", errors.Wrap(err, "failed to unmarshal chart.yaml")
 	}
-	// note: helm API version 2 is equivilent to Helm V3
+	// note: helm API v2 is equivilent to Helm V3
 	if version, ok := chartValues["apiVersion"]; ok && strings.EqualFold(version.(string), "v2") {
 		return "v3", nil
 	}
-	if _, ok := chartValues["dependencies"]; ok {
-		return "v3", nil
-	}
 
-	// if no determination is made, assume v2 until default changes
+	// if no determination is made, assume v2
 	return "v2", nil
 }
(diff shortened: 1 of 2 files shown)


pkg/base/helm.go
```diff
@@ -34,6 +35,11 @@ func RenderHelm(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (*Base,
 			}
 		}
 
+		// check chart.yaml for Helm version if a helm version has not been explicitly provided
+		if strings.EqualFold(fileName, "Chart.yaml") && renderOptions.HelmVersion == "" {
+			renderOptions.HelmVersion, err = checkChartForVersion(&file)
```
divolgin, line 40: this `err` is never checked.  but also, how about making `checkChartForVersion` return "v2" instead of error?  also I think we need to log info level message with helm version used for rendering.

jeffreygolden: I don't suspect we'll hit chart.yaml files that can't be unmarshaled, but it will be easier to troubleshoot if we ever hit that case and we can find an error explaining why it defaulted to v2.  How about the following?
``` Go
if strings.EqualFold(fileName, "Chart.yaml") && renderOptions.HelmVersion == "" {
	renderOptions.HelmVersion, err = checkChartForVersion(&file)
	if err != nil {
		renderOptions.Log.Errorf("could not determine helm version (will use helm v2 by default): %v", err)
	} else {
		renderOptions.Log.Info("rendering with Helm %v", renderOptions.HelmVersion)
	}
}
```

divolgin: I think it should be info in both cases.  If it's an error, it should fail.

jeffreygolden: switched to log.Info for both cases

divolgin (approved):

PR #1859, opened 2021-05-27
UI stuck in loading state when updating settings - fix

jgruica:


pkg/snapshot/velero.go
```diff
@@ -379,6 +393,51 @@ func getVersion(nameSpace string) (string, error) {
...
+		return "", errors.Wrap(err, "failed to list velero pods before restarting")
+	}
+
+	for _, pod := range veleroPods.Items {
+		if pod.Status.Phase == corev1.PodRunning {
+			return pod.Name, nil
+		}
+	}
+
+	return "", errors.New("unable to find velero pod")
```
divolgin, line 415: I think this needs some consistency.  If `getVeleroPod` finds 0 pods, it will fail.  But if `getResticPods` finds 0 pods, it will succeed.

pkg/snapshot/velero.go
```diff
@@ -379,6 +393,51 @@ func getVersion(nameSpace string) (string, error) {
...
+
+	veleroPods, err := clientset.CoreV1().Pods(namespace).List(context.TODO(), metav1.ListOptions{
+		LabelSelector: labels.SelectorFromSet(veleroLabels).String(),
+	})
+	if err != nil {
+		return "", errors.Wrap(err, "failed to list velero pods before restarting")
+	}
+
+	for _, pod := range veleroPods.Items {
+		if pod.Status.Phase == corev1.PodRunning {
```
divolgin, line 410: When pod is running, it's not quite ready until containers are also running. `pod.Status.ContainerStatuses` will have that info.

web/src/components/snapshots/SnapshotSettings.jsx
```diff
@@ -116,15 +117,15 @@ class SnapshotSettings extends Component {
 
 
       if (this.state.checkForVeleroAndRestic) {
-        if (this.state.snapshotSettings?.isVeleroRunning !== lastState.snapshotSettings?.isVeleroRunning) {
-          this.setState({ veleroVersionUpdated: this.state.snapshotSettings?.isVeleroRunning });
+        if (this.state.snapshotSettings?.veleroPod !== lastState.snapshotSettings?.veleroPod && !isEmpty(this.state.snapshotSettings?.veleroPod)) {
```
divolgin, line 120: I don't think this is right.  The value should be compare to the state before the Save button was clicked, not to the state since the last check.

jgruica: It is like that...flag on line 119 becomes true when save button is clicked.

divolgin: cool

web/src/components/snapshots/SnapshotSettings.jsx
```diff
@@ -116,19 +117,21 @@ class SnapshotSettings extends Component {
...
+          this.setState({ veleroUpdated: true });
         }
 
-        if (this.state.snapshotSettings?.isResticRunning !== lastState.snapshotSettings?.isResticRunning) {
-          this.setState({ resticVersionUpdated: this.state.snapshotSettings?.isResticRunning });
+        if (!isEqual(this.state.snapshotSettings?.resticPods, lastState.snapshotSettings?.resticPods) && !isEmpty(this.state.snapshotSettings?.resticPods)) {
```
divolgin, line 124: We should probably sort arrays before calling `isEqual` on them

divolgin (approved):

PR #1870, opened 2021-06-03
UI handles preflights skipped with unattended installs

divolgin:
Screen Shot 2021-06-02 at 8 54 00 PM

### DIFF
--- a/pkg/store/kotsstore/migrations.go
+++ b/pkg/store/kotsstore/migrations.go
@@ -36,6 +36,10 @@ func (s *KOTSStore) RunMigrations() {
 		logger.Error(errors.Wrap(err, "failed to migrate app_spec"))
 	}
 
+	if err := s.migrateSkippedPreflights(); err != nil {
+		logger.Error(errors.Wrap(err, "failed to migrate skipped preflights"))
+	}
+
 	// migrate postgrtes data from postgres
 	if err := s.migrateSessionsFromPostgres(); err != nil {
 		logger.Error(errors.Wrap(err, "failed to migrate sessions"))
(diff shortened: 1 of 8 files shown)


web/src/components/snapshots/SnapshotStorageDestination.jsx
```diff
@@ -354,7 +355,7 @@ class SnapshotStorageDestination extends Component {
         await this.snapshotProviderGoogle();
         break;
       case "other":
-        const s3CompatibleFieldErrors = this.validateSnapshotProviderS3Compatible();
```
divolgin, line 357: This change is unrelated.  JSX compiler was complaining about it.

GraysonNull (approved):

pkg/store/kotsstore/migrations.go
```diff
@@ -36,6 +36,10 @@ func (s *KOTSStore) RunMigrations() {
 		logger.Error(errors.Wrap(err, "failed to migrate app_spec"))
 	}
 
+	if err := s.migrationSkippedPreflights(); err != nil {
```
salahalsaleh, line 39:
```suggestion
	if err := s.migrateSkippedPreflights(); err != nil {
```

web/src/components/PreflightResultPage.jsx
```diff
@@ -312,9 +312,10 @@ class PreflightResultPage extends Component {
     const { slug } = this.props.match.params;
     const { showSkipModal, showWarningModal, preflightResultData, errorMessage } = this.state;
 
-    const stopPolling = !!preflightResultData?.result;
+    const preflightSkipped = preflightResultData?.skipped;
+    const stopPolling = (preflightResultData?.result || preflightSkipped);
     let preflightJSON = {};
-    if (stopPolling) {
+    if (stopPolling && !preflightSkipped) {
```
salahalsaleh, line 318: how about doing this instead?
```suggestion
    if (preflightResultData?.result) {
```

salahalsaleh (approved):

web/src/components/PreflightResultPage.jsx
divolgin: hmm, you don't think this is easier to read?

PR #1877, opened 2021-06-03
Don't deploy versions that are missing required configurations

divolgin:
When new updates are available, we download them asynchronously and they show up as "Pending config" on dashboard
Screen Shot 2021-06-03 at 2 31 12 PM

When there are no updates, we attempt to deploy the latest known version, which now produces this message

Screen Shot 2021-06-03 at 2 30 57 PM

### DIFF
--- a/pkg/updatechecker/updatechecker.go
+++ b/pkg/updatechecker/updatechecker.go
@@ -16,6 +16,8 @@ import (
 	kotspull "github.com/replicatedhq/kots/pkg/pull"
 	"github.com/replicatedhq/kots/pkg/reporting"
 	"github.com/replicatedhq/kots/pkg/store"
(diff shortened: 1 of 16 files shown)


pkg/updatechecker/updatechecker.go
```diff
@@ -243,19 +251,29 @@ func CheckForUpdates(appID string, deploy bool, skipPreflights bool, isCLI bool)
 		}
 
 		latestVersion := allVersions[len(allVersions)-1]
-		downstreams, err := store.GetStore().ListDownstreamsForApp(a.ID)
 		if err != nil {
```
salahalsaleh, line 254: any reason why this wasn't moved as well?

divolgin: woops!

salahalsaleh (approved):

PR #1878, opened 2021-06-07
Danstough/ch28072/config option for inputting an array of fields

DanStough:


design/variadic_config_alt.md
```diff
@@ -0,0 +1,637 @@
...
+
+    # NEW!
+    # This is a repeatable Config Item
+    - name: "static_files"
+      type: file
+      title: "Static Assets"
+      repeatable: true      # NEW! Tells the UI/Kots to expect an array
+      minimumCount: 3       # NEW! Not sure if this is needed here but including for discussion
+      repeatValues:         # NEW! Returned to the API filled in from the CLI/console
+      - "encoded file value one"
```
sennerholm, line 145: I think that we will need to have some kind of "ID" as the other repeatValues in the example to have the possibility to add the filename from an end user perspective.

design/variadic_config_alt.md
```diff
@@ -0,0 +1,637 @@
...
+          image: httpd
+          volumeMounts:
+            - mountPath: /var/www/html
+              name: secret-assets
+              readOnly: true
+      volumes:
+      - name: secret-assets
+        projected:
+          sources:
+          # repl{{ range $index, $value := ConfigOptionList "static_files" }}
```
sennerholm, line 311: From an end user perspective we need to have a way to "select" what filename the static_file should be exposed as, for example file1: index.html, file2: favicon.ico ...

design/variadic_config.md
```diff
@@ -0,0 +1,513 @@
...
+    # This is a repeatable Config Item
+    - name: "static_files"
+      type: file
+      title: "Static Assets"
+      repeatable: true      # NEW! Tells the UI/Kots to expect an array
+      minimumCount: 3       # NEW! Validation 
+      templates:            # NEW! This desclares how the array of values will be used
+        - apiVersion: v1    # By targeting a YAML path, we can clone this node for each element of the item.
+          kind: Deployment
+          name: my-deploy
```
sgalsaleh, line 148: suggestion: add the ability to specify a namespace, since we might need to distinguish resources with the same name but in different namespaces.

design/variadic_config.md
```diff
@@ -0,0 +1,513 @@
...
+          yamlPath: spec.template.spec.volumes[0].projected.sources
+        - apiVersion: v1    # By targeting a resource file, we can clone the whole file for each element.
+          kind: Secret
+          prefix: secret-
```
sgalsaleh, line 152: ditto

design/variadic_config.md
```diff
@@ -0,0 +1,513 @@
...
+      title: "Kafka Broker IDs"
+      repeatName: Broker
+      repeatable: true
+      default: ""
+      templates:
+      - apiVersion: v1
+        kind: kafka.banzaicloud.io/v1alpha1
+        name: KafkaTopic
+        yamlPath: spec.brokers[0]
+      valuesByGroup:
```
sgalsaleh, line 227: 👍

sgalsaleh (approved):

design/variadic_config_alt.md
```diff
@@ -0,0 +1,637 @@
+# Variadic Config Proposal
+
+Vendors require the ability to dynamically create resources as part of install configuration.
+One common use case is installing operators, where the customers need to create dynamic resources, like instances of an application or service, that are unknown until install time.
```
dexhorthy, line 4:
```suggestion
One common use case is installing operators, where the customers need to create dynamic resources, like instances of an application or service, the number of which is unknown until install time.
```

design/variadic_config.md
```diff
@@ -0,0 +1,513 @@
...
+This proposal outlines a plan to support dynamic/variadic application configuration to facilitate dynamic resource creation. 
+
+## Goals
+
+Two Main Business-Driver Goals
+1. Vendors can create "template" resources in the broadest sense; define resources once and they can be used _N_ times as needed by their application.
+1. Vendors can extend resources with _N_ additional configuration properties, like environment variables or volume mounts.
+
+Additional Technical Goals
+1. Maintain last mile kustomization of of all resources.
```
dexhorthy, line 15:
```suggestion
1. Maintain last mile kustomization of all resources.
```

pkg/base/replicated.go
```diff
@@ -37,11 +38,18 @@ func renderReplicated(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (
...
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to create new config context template builder")
 	}
 
+	config, _, _, _, err := findConfigAndLicense(u, renderOptions.Log)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to find config file")
+	}
+
+	actualizedConfig, err := kotsconfig.TemplateConfigObjects(config, itemValues, nil, template.LocalRegistry{}, nil, nil)
```
divolgin, line 51: missing error check

pkg/base/variadic.go
```diff
@@ -0,0 +1,339 @@
...
+	"k8s.io/client-go/kubernetes/scheme"
+)
+
+// Known issues and TODOs:
+// This currently only addresses variadic items.  Variadic groups are not included yet and may require changes to these functions.
+// Test_renderReplicated may sometimes fail because the resulting array can be out of order.  A more specific results check could be executed instead.
+// getVariadicGroupsForTemplate should be split into subfunctions to make it easier to read
+// The last element in the YamlPath must be an array
+
+func processVariadicConfig(u *upstreamtypes.UpstreamFile, config *kotsv1beta1.Config) error {
```
divolgin, line 22: Looks like this function will panic if upstream file is not a yaml file, like a helm chart.

pkg/base/variadic.go
```diff
@@ -0,0 +1,339 @@
...
+	// get upstreamFile gvk
+	decode := scheme.Codecs.UniversalDeserializer().Decode
+	_, gvk, err := decode(u.Content, nil, nil)
```
divolgin, line 25: `err` is not used

pkg/base/variadic.go
jeffreygolden: added some safety handling here.  Files that cannot be unmarshaled as yaml, do not have valid `apiVersion` or `kind` fields, or have a name or namespace that is not a string will be skipped for variadic processing.  These files would not match any variadic template targets if we continued processing them anyway.

design/variadic_config_alt.md
jeffreygolden: This has been added in with the updated design

pkg/base/variadic_test.go
```diff
@@ -0,0 +1,330 @@
...
+		{
+			name: "metadata in the wrong spot",
+			content: []byte(`
+apiVersion: kots.io/v1beta1 
+kind: Config 
+data: 
+  creationTimestamp: null
+  metadata:
+    name: config-sample
+    namespace: test`),
```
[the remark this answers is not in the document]
jeffreygolden: We don't, that test expects a graceful error.  The test ensures this code doesn't panic when given malformed yaml.

pkg/handlers/config_test.go
```diff
@@ -0,0 +1,383 @@
...
+			req := require.New(t)
+			updatedValues, err := updateAppConfigValues(test.values, test.configGroups, "")
+			req.NoError(err)
+
+			req.Equal(test.want, updatedValues)
+		})
+	}
+}
+
+func Test_mergeConfigValues(t *testing.T) {
```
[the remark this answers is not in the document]
jeffreygolden: "mergeConfigValues" is used optionally when values should not be deleted.  The function can't remove any existing values.

design/variadic_config.md
```diff
@@ -0,0 +1,520 @@
...
+      type: file
+      title: "Static Assets"
+      repeatable: true      # NEW! Tells the UI/Kots to expect an array
+      minimumCount: 3       # NEW! Validation 
+      templates:            # NEW! This desclares how the array of values will be used
+        - apiVersion: v1    # By targeting a YAML path, we can clone this node for each element of the item.
+          kind: Deployment
+          name: my-deploy
+          namespace: myApp
+          yamlPath: spec.template.spec.volumes[0].projected.sources
```
jeffreygolden, line 150: Generated files will not have a static name to target here.  Additional options could be provided as a way to target a file.
`namePrefix: ` could be used to target generated files starting with the prefix string.  This would be used instead of name, and validation should ensure only one field is provided.

Additionally, for manifests that do not have a name (IE jobs), this template could use Selectors to find the config.  This would be a larger lift and should be in a separate feature PR.

pkg/base/variadic_test.go
```diff
@@ -0,0 +1,330 @@
...
+			configOptionName: "secret",
+			valueName:        "secret-1",
+			target:           "repl{{ ConfigOptionFilename \"secret\" }}",
+			want:             "repl{{ ConfigOptionFilename \"secret-1\" }}",
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			req := require.New(t)
+			result := generateTargetValue(test.configOptionName, test.valueName, test.target)
```
divolgin, line 269: I think this function is very unstable.  Also templates can be quoted in a variety of ways.  Some examples:
```
value: repl{{ ConfigOptionFilename "secret" }}
value: "repl{{ ConfigOptionFilename "secret" }}"
value: "repl{{ ConfigOptionFilename 'secret' }}"
value: "repl{{ ConfigOptionFilename `secret` }}
```

I think we should explore other options, for example, another template can be used

```
value: "repl{{ ConfigOptionFilename 'secret-[[repl i]]' }}"
```

kotskinds/apis/kots/v1beta1/config_types.go
```diff
@@ -31,32 +31,49 @@ type ConfigChildItem struct {
...
 	// ValueCmd    *ConfigItemCmd         `json:"value_cmd,omitempty"`
 	// DataCmd     *ConfigItemCmd         `json:"data_cmd,omitempty"`
 }
 
+type RepeatTemplate struct {
+	APIVersion string `json:"apiVersion"`
+	Kind       string `json:"kind"`
+	Name       string `json:"name"`
+	Namespace  string `json:"namespace"`
+	YamlPath   string `json:"yamlPath"`
```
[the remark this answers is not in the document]
jeffreygolden: Namespace and YamlPath are optional, omitempty is now set

design/variadic_config.md
```diff
@@ -0,0 +1,520 @@
...
+    # NEW!
+    # Values for a template
+    <repeat-config-item-unique-name>:
+      repeatableItem: <Config item Name>
+      default: <value>
+      value: <value> 
+    # Example
+    port-<short guid>:
+      repeatableItem: port
+      value: 80
```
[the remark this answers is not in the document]
jeffreygolden: It should?  it might change later when config groups are built out

pkg/base/variadic_test.go
jeffreygolden: Fixed with subtemplating, as discussed.

pkg/base/variadic_test.go
```diff
@@ -0,0 +1,339 @@
...
+		configOptionName string
+		valueName        string
+		target           string
+		want             interface{}
+	}{
+		{
+			configOptionName: "secret",
+			valueName:        "secret-123",
+			target:           "repl{{ ConfigOption repl[[ .secret ]] }}",
+			want:             "repl{{ ConfigOption \"secret-123\" }}",
```
divolgin, line 251: template shouldn't return quoted values.  if quotes are wanted, they should be added outside of the template by person writing yaml

divolgin (dismissed):

sgalsaleh (dismissed):

PR #1883, opened 2021-06-09
feat: ability to provide dockerhub username and password to increase rate limit

sgalsaleh:


pkg/image/builder.go
```diff
@@ -527,6 +527,13 @@ func IsPrivateImage(image string) (bool, error) {
 
 		sysCtx := types.SystemContext{DockerDisableV1Ping: true}
 
+		if dockerHubRegistry.Username != "" && dockerHubRegistry.Password != "" {
```
divolgin, line 530: I think you need to also check domain is docker.io.   This will get host name from parsed reference
```
		registryHost := reference.Domain(ref.DockerReference())
```

pkg/image/builder.go
```diff
@@ -527,7 +528,8 @@ func IsPrivateImage(image string, dockerHubRegistry registry.RegistryOptions) (b
 
 		sysCtx := types.SystemContext{DockerDisableV1Ping: true}
 
-		if dockerHubRegistry.Username != "" && dockerHubRegistry.Password != "" {
+		registryHost := reference.Domain(ref.DockerReference())
+		if isRateLimited && strings.HasSuffix(registryHost, "docker.io") && dockerHubRegistry.Username != "" && dockerHubRegistry.Password != "" {
```
divolgin, line 532: should probably be ".docker.io" to ensure only subdomains of docker.io match.

pkg/template/config_context.go
```diff
@@ -268,7 +268,7 @@ func (ctx ConfigCtx) localImageName(imageRef string) string {
 	// Not airgap and no local registry.  Rewrite images that are private only.
 
 	if ctx.app == nil || !ctx.app.Spec.ProxyPublicImages {
-		isPrivate, err := image.IsPrivateImage(imageRef)
+		isPrivate, err := image.IsPrivateImage(imageRef, registry.RegistryOptions{}) // TODO: pass docker registry
```
divolgin, line 271: do the todo?

divolgin (approved):

divolgin (approved):

PR #1888, opened 2021-06-15
fix: preserve defaults for editable config items during live rendering

sgalsaleh:
We already render the default values when creating a new config context (new builder), we should apply those rendered values to the config instead of/before having them be rendered later when all of the config is rendered in one shot.


divolgin (approved):

laverya:
This makes unit tests break,  and they're tests I'm pretty confident in

sgalsaleh:
@laverya defaults were not respected/preserved if the item is editable, only if the item is read-only

laverya (approved):

PR #1890, opened 2021-06-15
Write helm charts as sub-bases in kustomization

emosbaugh:
This will write charts as bases in kustomization. The is pre-work for kustomizing charts individually.

```
apiVersion: kustomize.config.k8s.io/v1beta1
bases:
- charts/grafana
- charts/kube-state-metrics
- charts/prometheus-node-exporter
- crds
kind: Kustomization
resources:
- templates/alertmanager/alertmanager.yaml
- templates/alertmanager/psp-role.yaml
- templates/alertmanager/psp-rolebinding.yaml
- templates/alertmanager/psp.yaml
- ```

### DIFF
--- a/cmd/kots/cli/pull.go
+++ b/cmd/kots/cli/pull.go
@@ -57,6 +57,7 @@ func PullCmd() *cobra.Command {
 				HTTPSProxyEnvValue: v.GetString("https-proxy"),
 				NoProxyEnvValue:    v.GetString("no-proxy"),
 				IncludeMinio:       v.GetBool("with-minio"),
+				NativeHelmInstall:  v.GetBool("native-helm-install"),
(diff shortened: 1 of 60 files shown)


jala-dx (commented):
@emosbaugh - Can I  request for these specific unit test cases:  

1. That takes input a "sample chart that has a **dependent** chart" and spits out the **_kustomization.yaml_** for each of those nested charts
2. The _**final directory structure**_ that will be sent to the operator. (with or without Chart.yaml to begin with)..(Does not have to be accurate to begin with)..

emosbaugh:
Added Chart.yaml and Chart.lock

Screen Shot 2021-06-15 at 5 09 30 PM

PR #1892, opened 2021-06-16
Remove minio and migrate data

sgalsaleh:


DanStough (approved):
No notes!

deploy/backup.sh
```diff
@@ -2,12 +2,12 @@
...
-export S3_DIR=/backup/s3/
-export S3_HOST=`echo $S3_ENDPOINT | awk -F/ '{print $3}'`
-rm -rf $S3_DIR
-mkdir -p $S3_DIR
-s3cmd --access_key=$S3_ACCESS_KEY_ID --secret_key=$S3_SECRET_ACCESS_KEY --host=$S3_HOST --no-ssl --host-bucket=$S3_BUCKET_NAME.$S3_HOST sync s3://$S3_BUCKET_NAME $S3_DIR
+# this is to work around the fact that Velero does not support backing up volumes of type hostpath (which could be the case in k3s clusters for example)
+# we copy the contents to the /backup directory where it gets restored later during the restore process to /kotsadmdata
+# ref: https://github.com/vmware-tanzu/velero/discussions/3378
+rm -rf /backup/kotsadmdata
+cp -R /kotsadmdata /backup
```
divolgin, line 13: How about adding
`-a` for archive mode to preserve all file attributes
`-f` for force overwrite
`-v` to output files being copied

Same for restore script

sgalsaleh: the `kotsadmdata` volume should always be empty on restore, but the other flags are good suggestions. i'll add `-f` just in case as well.

divolgin (approved):

divolgin (approved):

divolgin (approved):

pkg/filestore/store.go
```diff
@@ -0,0 +1,84 @@
...
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+const (
+	ArchivesDir = "/kotsadmdata/archives"
```
marccampbell, line 14: nit:  `kotsadm` isn't a term in the product, can this just be `kotsdata` ?

sgalsaleh: hmm, maybe. the pod name is `kotsadm-0`, the bucket name that we were using in minio was also called `kotsadm`. we use the term `kotsadm` extensively. i can change it if you have a strong opinion about it.

divolgin (approved):

PR #1912, opened 2021-06-23
Continue editing config button fix

jgruica:


web/src/components/apps/AppConfig.jsx
```diff
@@ -540,23 +547,23 @@ class AppConfig extends Component {
...
-      </div>
+  { gettingConfigErrMsg &&
+  <ErrorModal
+    errorModal={displayErrorModal}
+    toggleErrorModal={this.toggleErrorModal}
+    err={errorTitle}
+    errMsg={gettingConfigErrMsg}
+    tryAgain={this.getConfig}
+  />}
+      </div >
```
sgalsaleh, line 566:
```suggestion
      </div>
```

web/src/components/apps/AppConfig.jsx
```diff
@@ -540,23 +547,23 @@ class AppConfig extends Component {
...
-          <ErrorModal
-            errorModal={displayErrorModal}
-            toggleErrorModal={this.toggleErrorModal}
-            err={errorTitle}
-            errMsg={gettingConfigErrMsg}
-            tryAgain={this.getConfig}
-          />}
-      </div>
+  { gettingConfigErrMsg &&
+  <ErrorModal
```
sgalsaleh, line 559: indentation feels weird here

web/src/components/apps/AppConfig.jsx
```diff
@@ -431,6 +431,13 @@ class AppConfig extends Component {
     this.setState({ displayErrorModal: !this.state.displayErrorModal });
   }
 
+  navigateToUpdatedConfig = (app) => {
+    this.setState({ showNextStepModal: false });
+
+    const pendingVersions = app?.downstreams[0]?.pendingVersions;
```
sgalsaleh, line 437: how could you guarantee that pendingVersions include the latest version?

jgruica: because after you save the changes the new version is created and that's added to the top of the downstreams pending versions.

sgalsaleh: yeah, but that's coming from props, and it's a separate api request that updates the app, the downstream, and the pending versions, so there could be a delay between when the modal shows up, and the pending versions being updated, right?

jgruica: handler to save config is updating the app downstreams. modals shows after changes are saved, not while changes are saving.

sgalsaleh (approved):

PR #1919, opened 2021-06-29
Allow using multiple path components in registry namespace

divolgin:


cmd/kots/cli/admin-console-push-images.go
```diff
@@ -40,11 +40,6 @@ func AdminPushImagesCmd() *cobra.Command {
 			airgapArchive := args[0]
 			endpoint := args[1]
 
-			endpointParts := strings.Split(endpoint, "/")
-			if len(endpointParts) != 2 {
```
sgalsaleh, line 44: i think we should still check that `endpointParts` is greater than 1, otherwise fail?

sgalsaleh (approved):

sgalsaleh (approved):

cmd/kots/cli/admin-console-push-images.go
divolgin: allowing 0 or more parts now.

PR #1922, opened 2021-06-30
Don't allow installing airgap bundles from wring channel

divolgin:


pkg/pull/pull.go
```diff
@@ -221,6 +221,13 @@ func Pull(upstreamURI string, pullOptions PullOptions) (string, error) {
 			return "", errors.Wrap(err, "failed to parse license from file")
 		}
 
+		if fetchOptions.License.Spec.ChannelID != airgap.Spec.ChannelID {
```
divolgin, line 224: Intentionally not checking for old style bundles with no channel ID because why?

sgalsaleh (approved):

PR #1932, opened 2021-07-02
feat: ability to opt-in to remove minio

sgalsaleh:


pkg/filestore/s3_store.go
```diff
@@ -0,0 +1,147 @@
...
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "NotFound" {
+			logger.Debug("object store is ready")
+			return nil
+		}
+
+		select {
+		case <-time.After(period):
+			continue
+		case <-ctx.Done():
+			return errors.Wrap(err, "failed to find valid object store")
```
divolgin, line 100: `err` is nil here.  `ctx.Err()` will return context error, but could also be nil.   so how about:
```
errors.Errorf("failed to find valid object store: %s", ctx.Err())
```

pkg/kotsadm/objects/kotsadm_objects.go
```diff
@@ -663,7 +664,550 @@ func KotsadmDeployment(deployOptions types.DeployOptions) *appsv1.Deployment {
...
+}
+
+func KotsadmStatefulSet(deployOptions types.DeployOptions, size resource.Quantity) (*appsv1.StatefulSet, error) {
+	securityContext := &corev1.PodSecurityContext{
+		RunAsUser: util.IntPointer(1001),
+		FSGroup:   util.IntPointer(1001),
+	}
+	if deployOptions.IsOpenShift {
+		// we have to specify a pod security context here because if we don't, here's what will happen:
+		// the kotsadm service account is associated with a role/clusterrole that has wildcard privilages,
```
divolgin, line 741: privileges

divolgin (approved):

divolgin (approved):

PR #1948, opened 2021-07-09
Allow removing charts from app

divolgin:
Remove files from "charts" sub-directories in overlays if corresponding files are missing from base.  Based on file system layout, which is less than ideal, but it works well.

### DIFF
--- a/kotsadm/operator/pkg/applier/process.go
+++ b/kotsadm/operator/pkg/applier/process.go
@@ -3,6 +3,7 @@ package applier
 import (
 	"io/ioutil"
 	"os/exec"
+	"sync"
 
 	"github.com/pkg/errors"
 )
@@ -25,8 +26,23 @@ func Run(cmd *exec.Cmd) ([]byte, []byte, error) {
 		return nil, nil, errors.Wrap(err, "failed to start commmand")
 	}
 
-	stdout, _ := ioutil.ReadAll(stdoutReader)
-	stderr, _ := ioutil.ReadAll(stderrReader)
+	var stdout, stderr []byte
+	var wg sync.WaitGroup
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		stdout, _ = ioutil.ReadAll(stdoutReader)
+	}()
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		stderr, _ = ioutil.ReadAll(stderrReader)
+	}()
+
+	// cmd.Wait() must be called after all readers have completed
+	wg.Wait()
 
 	err = cmd.Wait()
 	return stdout, stderr, err
--- a/kotsadm/operator/pkg/client/deploy.go
+++ b/kotsadm/operator/pkg/client/deploy.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
-	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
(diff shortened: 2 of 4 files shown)


kotsadm/operator/pkg/applier/process.go
```diff
@@ -25,8 +26,23 @@ func Run(cmd *exec.Cmd) ([]byte, []byte, error) {
 		return nil, nil, errors.Wrap(err, "failed to start commmand")
 	}
 
-	stdout, _ := ioutil.ReadAll(stdoutReader)
-	stderr, _ := ioutil.ReadAll(stderrReader)
+	var stdout, stderr []byte
+	var wg sync.WaitGroup
+
+	wg.Add(1)
```
divolgin, line 32: Unrelated.  But I think `stdoutReader` got blocked at some point because `stderrReader` buffer got filled up and the command never terminated.

kotsadm/operator/pkg/client/deploy.go
```diff
@@ -514,6 +514,30 @@ func (c *Client) installHelm(helmDir string, namespace string) (*commandResult,
 	return result, nil
 }
 
+func (c *Client) uninstallWithHelm(helmDir string, namespace string, charts []string) error {
+	version := "3.4.2"
+	if os.Getenv("KOTSADM_TARGET_NAMESPACE") != "" {
+		namespace = os.Getenv("KOTSADM_TARGET_NAMESPACE")
```
emosbaugh, line 520: Don't you have to use the namespace in the kots.io hemlchart if set?

pkg/pull/pull.go
```diff
@@ -696,7 +747,7 @@ func writeDownstreams(options PullOptions, overlaysDir string, m *midstream.Mids
 		}
 
 		for _, mid := range helmMidstreams {
-			helmMidstream := mid
+			helmMidstream := mid // bug? this object contains pointers which are not deep-copied here
```
emosbaugh, line 750: does it mutate that variable?

pkg/pull/pull.go
```diff
@@ -391,21 +391,24 @@ func Pull(upstreamURI string, pullOptions PullOptions) (string, error) {
 	}
 
 	helmMidstreams := []midstream.Midstream{}
-	for _, base := range helmBases {
+	for _, helmBase := range helmBases {
 		writeMidstreamOptions := commonWriteMidstreamOptions
-		writeMidstreamOptions.MidstreamDir = filepath.Join(base.GetOverlaysDir(writeBaseOptions), "midstream", base.Path)
-		writeMidstreamOptions.BaseDir = filepath.Join(u.GetBaseDir(writeUpstreamOptions), base.Path)
-
-		helmBase := base
```
emosbaugh, line 399: are you sure you can remove this?

pkg/pull/pull.go
```diff
@@ -391,21 +391,24 @@ func Pull(upstreamURI string, pullOptions PullOptions) (string, error) {
...
 		helmMidstream, err := writeMidstream(writeMidstreamOptions, pullOptions, u, &helmBase, fetchOptions.License, identityConfig, u.GetUpstreamDir(writeUpstreamOptions), log)
 		if err != nil {
-			return "", errors.Wrapf(err, "failed to write helm midstream %s", base.Path)
+			return "", errors.Wrapf(err, "failed to write helm midstream %s", helmBase.Path)
 		}
 
 		helmMidstreams = append(helmMidstreams, *helmMidstream)
 	}
 
+	err = removeUnusedHelmOverlays(writeMidstreamOptions.MidstreamDir, writeMidstreamOptions.BaseDir)
```
emosbaugh, line 407: wont this just be set to the last in the loop above?

pkg/pull/pull.go
divolgin: I renamed `base` variable to not be the same as `base` package.  It's now created as `helmBase`.

pkg/pull/pull.go
divolgin: nah, that variable is in the loop scope.

pkg/pull/pull.go
divolgin: `mid.Base.Path` is the same value on each iteration.  There's something going on here that's not quite right, but it produces the right result.  I think we are rendering the same files multiple times.  I didn't really get to the bottom of it, but that line looked suspicious to me.

kotsadm/operator/pkg/client/deploy.go
divolgin: Good catch.  Updated to use the same logic as `kubectl apply` commands.

sgalsaleh (approved):

PR #1949, opened 2021-07-09
Don't allow toggling helm install flag

divolgin:


pkg/kotsutil/kots.go
```diff
@@ -416,6 +416,42 @@ func LoadKotsKindsFromPath(fromDir string) (*KotsKinds, error) {
...
+			}
+
+			contents, err := ioutil.ReadFile(path)
+			if err != nil {
+				return err
+			}
+
+			decoded, gvk, err := decode(contents, nil, nil)
+			if err != nil {
+				return nil
```
emosbaugh, line 439: wrap this with the path since its in a loop

pkg/kotsutil/kots.go
```diff
@@ -416,6 +416,42 @@ func LoadKotsKindsFromPath(fromDir string) (*KotsKinds, error) {
...
+				return err
+			}
+
+			if info.IsDir() {
+				return nil
+			}
+
+			contents, err := ioutil.ReadFile(path)
+			if err != nil {
+				return err
```
emosbaugh, line 434: wrap this with the path since its in a loop

pkg/pull/pull.go
```diff
@@ -269,6 +275,24 @@ func Pull(upstreamURI string, pullOptions PullOptions) (string, error) {
...
+	}
+
+	for _, prevChart := range prevHelmCharts {
+		for _, newChart := range newHelmCharts {
+			if prevChart.Spec.Chart.Name != newChart.Spec.Chart.Name {
+				continue
+			}
+			if prevChart.Spec.UseCLIInstall != newChart.Spec.UseCLIInstall {
+				log.FinishSpinnerWithError()
+				return "", errors.Errorf("deployment method for chart %s has changed", newChart.Spec.Chart.Name)
```
emosbaugh, line 291: i thought you just cant change it from UseCLIInstall to ! UseCLIInstall

divolgin: Toggling is untested.  Future feature?

pkg/kotsutil/kots.go
divolgin: not an error

sgalsaleh (approved):

pkg/pull/pull.go
emosbaugh: So this works on new installs only?

divolgin: Yes, that was the intention.

PR #1955, opened 2021-07-09
fix: allow separate namespaces for helm chart resources

morningvera:
Fixes this [issue](https://app.clubhouse.io/replicated/story/33490/helmchart-namespace-field-overrides-custom-values-specified-in-the-chart). Kustomize was previously overwriting all namespaces. This PR adds the spec namespace (if it's defined) to all the helm resources. If a helm resource already has a defined namespace (e.g. from the values), the namespace is left as is. If no custom namespaces are defined, the resources are installed in the same namespace as the app.

### DIFF
--- a/pkg/base/helm.go
+++ b/pkg/base/helm.go
@@ -9,10 +9,10 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/ghodss/yaml"
 	"github.com/pkg/errors"
 	upstreamtypes "github.com/replicatedhq/kots/pkg/upstream/types"
 	"github.com/replicatedhq/kots/pkg/util"
+	"gopkg.in/yaml.v2"
 	"helm.sh/helm/v3/pkg/strvals"
 )
 
@@ -270,4 +270,32 @@ func checkChartForVersion(file *upstreamtypes.UpstreamFile) (string, error) {
 
 	// if no determination is made, assume v2
 	return "v2", nil
+}
+
+// insert namespace if it's defined in the spec and not already present in the manifests
+func insertHelmNamespace(baseFiles []BaseFile, renderOptions *RenderOptions) ([]BaseFile, error) {
+	if renderOptions.Namespace != "" {
+		for i, v := range baseFiles {
+			var manifest map[string]interface{}
+			err := yaml.Unmarshal(v.Content, &manifest)
+			if err != nil {
+				return nil, errors.Wrap(err, "failed to unmarshal helm manifest")
+			}
+
+			if metadata, ok := manifest["metadata"].(map[interface{}]interface{}); ok {
+				if _, ok := metadata["namespace"]; !ok {
+					metadata["namespace"] = renderOptions.Namespace
+				} else {
+					continue // don't bother overwriting the yaml if a namespace already exists
+				}
+			}
+
+			baseFiles[i].Content, err = yaml.Marshal(manifest)
+			if err != nil {
+				return nil, errors.Wrap(err, "failed to encode helm manifest")
+			}
+		}
+	}
+
+	return baseFiles, nil
 }
--- a/pkg/base/helm_test.go
+++ b/pkg/base/helm_test.go
@@ -67,6 +67,195 @@ func Test_checkChartForVersion(t *testing.T) {
 	}
 }
 
+func Test_RenderHelm(t *testing.T) {
+	type args struct {
+		upstream      *upstreamtypes.Upstream
+		renderOptions *RenderOptions
+	}
+	tests := []struct {
+		name    string
+		args    args
+		want    *Base
+		wantErr bool
+	}{
+		{
+			name: "helm v3 namespace insertion",
+			args: args{
+				upstream: &upstreamtypes.Upstream{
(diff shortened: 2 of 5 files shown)


pkg/base/helm.go
```diff
@@ -74,6 +74,32 @@ func RenderHelm(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (*Base,
 		return nil, errors.Errorf("unknown helmVersion %s", renderOptions.HelmVersion)
 	}
 
+	// add namespace if it's defined in the spec and not already present in the manifests
+	if renderOptions.Namespace != "" {
+		for i, v := range rendered {
+			var manifest map[string]interface{}
+			err := yaml.Unmarshal(v.Content, &manifest)
+			if err != nil {
+				return nil, errors.Wrap(err, "failed to unmarshal helm manifest")
```
morningvera, line 83: if marshalling or unmarshalling fails, is it better to return an error or just skip adding the namespace?

pkg/base/write.go
```diff
@@ -132,7 +132,6 @@ func (b *Base) writeBase(options WriteOptions) error {
 			APIVersion: "kustomize.config.k8s.io/v1beta1",
 			Kind:       "Kustomization",
 		},
-		Namespace:             b.Namespace,
```
sgalsaleh, line 135: we should keep this since it can affect non-helm resources. you probably want to get rid of this instead: https://github.com/replicatedhq/kots/blob/b1dd65d608efab4c02d51c97ca26e407e8e5aad4/pkg/base/replicated.go#L124

pkg/base/helm.go
```diff
@@ -74,6 +74,32 @@ func RenderHelm(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (*Base,
...
+			if metadata, ok := manifest["metadata"].(map[interface{}]interface{}); ok {
+				if _, ok := metadata["namespace"]; !ok {
+					metadata["namespace"] = renderOptions.Namespace
+				} else {
+					continue // don't bother overwriting the yaml if a namespace already exists
+				}
+			}
+
+			b := new(bytes.Buffer)
+			err = yaml.NewEncoder(b).Encode(manifest)
```
sgalsaleh, line 95: could use yaml.Marshal instead for consistency sake

pkg/base/helm.go
```diff
@@ -74,6 +74,32 @@ func RenderHelm(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (*Base,
 		return nil, errors.Errorf("unknown helmVersion %s", renderOptions.HelmVersion)
 	}
 
+	// add namespace if it's defined in the spec and not already present in the manifests
+	if renderOptions.Namespace != "" {
+		for i, v := range rendered {
+			var manifest map[string]interface{}
```
sgalsaleh, line 80: you could use this struct instead: https://github.com/replicatedhq/kots/blob/a7fb3d422442db7094f28bd1788e375adadaa526/pkg/base/base.go#L34

morningvera:
Jeff mentioned that `yaml.Unmarshal` doesn't work with multidocs. Need to test that some more to make sure this PR doesn't break anything.

pkg/base/helm.go
morningvera: I considered that but it omits the spec which means we'd lose it when marshalling. Should I add the spec to that type instead of unmarshalling to `map[string]interface{}`? I'd also worry that it'll omit other things that I haven't thought of yet so `map[string]interface{}` seemed a bit safer.

sgalsaleh: nope, you have a good point ... you can keep it as is

pkg/base/helm.go
morningvera: done

pkg/base/helm.go
morningvera: leaving the errors as is but let me know if something else is preferred

morningvera:
I ended up moving the namespace insert to a func and calling it in helm_v3 and helm_v2 to avoid the multidoc issue. Also had to add manifest splitting and merging to helm_v2. I've tested both with a few multidocs with and without namespaces passed from the values. Everything is working as expected now.

sgalsaleh (approved):

divolgin (approved):

divolgin (approved):

pkg/base/helm_test.go
```diff
@@ -67,6 +67,205 @@ func Test_checkChartForVersion(t *testing.T) {
...
+				AdditionalFiles: []BaseFile{
+					{
+						Path:    "Chart.yaml",
+						Content: []byte("name: test-chart\nversion: 0.1.0"),
+					},
+				},
+			},
+		},
+		{
+			name: "helm v3 namespace insertion with multidoc",
```
sgalsaleh, line 215: nit: this has the same name as the previous test

sgalsaleh (approved):

pkg/base/helm_test.go
morningvera: copy/paste, good catch!

sgalsaleh (approved):

PR #1962, opened 2021-07-14
Deduplicated app resources

divolgin:
Chart directories still have the original layout, with all files intact.  Deduplication happens in the top level kustomization file, so its layout reverted to the original one.

Screen Shot 2021-07-14 at 1 44 04 PM

### DIFF
--- a/pkg/base/write.go
+++ b/pkg/base/write.go
@@ -36,7 +36,7 @@ func (b *Base) WriteBase(options WriteOptions) error {
 		}
 	}
 
-	if err := b.writeBase(options); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/base/write.go
```diff
@@ -36,7 +36,7 @@ func (b *Base) WriteBase(options WriteOptions) error {
 		}
 	}
 
-	if err := b.writeBase(options); err != nil {
+	if _, _, err := b.writeBase(options, true); err != nil {
```
jala-dx, line 39: is the flag passed true by default? will it ever be false?

jala-dx (approved):

PR #1964, opened 2021-07-15
Implement cli method to get deployed version

jala-dx:
Screen Shot 2021-07-14 at 10 06 57 PM

### DIFF
--- a/cmd/kots/cli/get.go
+++ b/cmd/kots/cli/get.go
@@ -151,14 +151,22 @@ func getAppsCmd(cmd *cobra.Command, args []string) error {
 
 	printableApps := make([]print.App, 0)
 	for _, app := range apps.Apps {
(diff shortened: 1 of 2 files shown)


divolgin (changes requested):
This will always print the same version, not the deployed version.  Instead of adding a new API call, this should use the existing call.  This information is already available [here](https://github.com/replicatedhq/kots/blob/96a75bcfc8b1167202cddb0871f4d18e72f12742/pkg/handlers/app.go#L171).

divolgin:
Version label should be available in [downstream version](https://github.com/replicatedhq/kots/blob/96a75bcfc8b1167202cddb0871f4d18e72f12742/pkg/handlers/app.go#L238)

divolgin (approved):

PR #1969, opened 2021-07-16
fix: allow separate namespaces for helm chart resources

morningvera:
Another fix for this [https://app.clubhouse.io/replicated/story/33490/helmchart-namespace-field-overrides-custom-values-specified-in-the-chart](url). Previous method was failing for cluster scoped resources like CRDs. Refactored to use kustomize instead of manually adding the namespace.


pkg/base/helm.go
```diff
@@ -271,3 +274,105 @@ func checkChartForVersion(file *upstreamtypes.UpstreamFile) (string, error) {
 	// if no determination is made, assume v2
 	return "v2", nil
 }
+
+// insert namespace if it's defined in the spec and not already present in the manifests
+func kustomizeHelmNamespace(baseFiles []BaseFile, renderOptions *RenderOptions) ([]BaseFile, error) {
```
morningvera, line 279: this got very complicated. since there are potentially duplicates (i.e. patches), i still had to filter those and add them to `kustomization.yaml` correctly. that means those patches are already applied so i chose to not pass them on to the next step. for resources that have a hardcoded namespace, duplicates are maintained and added to base like before.

pkg/base/helm.go
```diff
@@ -271,3 +274,105 @@ func checkChartForVersion(file *upstreamtypes.UpstreamFile) (string, error) {
...
+		defer os.RemoveAll(chartsPath)
+
+		resources := map[string]BaseFile{}
+		var kustomizeResources []string
+		var kustomizePatches []kustomizetypes.PatchStrategicMerge
+		foundGVKNamesMap := map[string]bool{}
+		for _, baseFile := range baseFiles {
+			// write temp files for manifests that need a namespace
+			gvk, manifest := GetGVKWithNameAndNs(baseFile.Content, renderOptions.Namespace)
+			if manifest.APIVersion == "" && manifest.Kind == "" && manifest.Metadata.Name == "" {
```
morningvera, line 295: since this is unmarshalling into a struct now, it has to check for the common k8s resource fields to identify invalid manifests. this also means that "valid" yaml will be filtered out but that's probably okay considering they'll fail in deployment anyway. let me know if this is a problem though.

pkg/base/helm.go
```diff
@@ -271,3 +274,105 @@ func checkChartForVersion(file *upstreamtypes.UpstreamFile) (string, error) {
...
+				}
+
+				if _, err := tmpFile.Write(baseFile.Content); err != nil {
+					return nil, errors.Wrapf(err, "failed to write temp file %v content", tmpFile.Name())
+				}
+				if err := tmpFile.Close(); err != nil {
+					return nil, errors.Wrapf(err, "failed to close temp file %v", tmpFile.Name())
+				}
+
+				if !foundGVKNamesMap[gvk] || gvk == "" {
```
morningvera, line 315: since there are potentially duplicates (i.e. patches), i still had to filter those and add them to `kustomization.yaml` correctly. that means those patches are already applied so i chose to not pass them on to the next step. let me know if we prefer passing them on anyway. for resources that have a hardcoded namespace, duplicates are maintained and added to base like before.

pkg/base/helm.go
```diff
@@ -271,3 +274,105 @@ func checkChartForVersion(file *upstreamtypes.UpstreamFile) (string, error) {
...
+			// write temp files for manifests that need a namespace
+			gvk, manifest := GetGVKWithNameAndNs(baseFile.Content, renderOptions.Namespace)
+			if manifest.APIVersion == "" && manifest.Kind == "" && manifest.Metadata.Name == "" {
+				updatedBaseFiles = append(updatedBaseFiles, baseFile)
+				continue // ignore invalid resources
+			}
+
+			if manifest.Metadata.Namespace == "" {
+				splitPath := strings.Split(baseFile.Path, "/")
+				name := splitPath[len(splitPath)-1]
```
sgalsaleh, line 302: you can use `filepath.Base` instead to get the file name from the full path.

example: https://play.golang.org/p/DzlCV-HC-r

sgalsaleh, line 302: what if two files have the same name but the exist in different subdirectories/levels?

pkg/base/helm.go
```diff
@@ -271,3 +274,105 @@ func checkChartForVersion(file *upstreamtypes.UpstreamFile) (string, error) {
...
+		}
+		defer os.RemoveAll(chartsPath)
+
+		resources := map[string]BaseFile{}
+		var kustomizeResources []string
+		var kustomizePatches []kustomizetypes.PatchStrategicMerge
+		foundGVKNamesMap := map[string]bool{}
+		for _, baseFile := range baseFiles {
+			// write temp files for manifests that need a namespace
+			gvk, manifest := GetGVKWithNameAndNs(baseFile.Content, renderOptions.Namespace)
```
sgalsaleh, line 294: looking at this function's definition, it seems like it would override the namespace of manifests with no namespace with the provided namespace here.

pkg/base/helm.go
morningvera: normally there could be name collisions but `ioutil.TempFile` handles that by appending a random number to the end of each file name. something similar happens with multidoc manifests and i made sure that works.

pkg/base/helm.go
morningvera: i think that's intended behavior though. since we're adding that namespace to manifests with no namespace, it makes sense to me to add it to the gvk. if we didn't do that, splitting the manifests and replacing the `baseFile.Content` would fail after the kustomize step (since that second gvk would include the new namespace).

.github/workflows/build-test.yaml
```diff
@@ -12,11 +12,16 @@ jobs:
...
+      
+      - uses: imranismail/setup-kustomize@v1
+        with:
+          kustomize-version: "3.5.4"
 
       - name: setup env
         run: |
           echo "GOPATH=$(go env GOPATH)" >> $GITHUB_ENV
           echo "$(go env GOPATH)/bin" >> $GITHUB_PATH
+          cp /usr/local/bin/kustomize $(go env GOPATH)/bin/kustomize3.5.4
```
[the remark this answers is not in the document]
morningvera: ended up using the [kustomize install script](https://github.com/kubernetes-sigs/kustomize/blob/master/hack/install_kustomize.sh)

go.mod
```diff
@@ -97,14 +97,13 @@ require (
...
 	k8s.io/helm v2.14.3+incompatible
-	k8s.io/kube-openapi v0.0.0-20210323165736-1a6458611d18 // indirect
 	k8s.io/kubelet v0.0.0
 	k8s.io/kubernetes v1.20.5
 	k8s.io/utils v0.0.0-20210305010621-2afb4311ab10
 	rsc.io/letsencrypt v0.0.3 // indirect
 	sigs.k8s.io/application v0.8.3
 	sigs.k8s.io/controller-runtime v0.8.3
-	sigs.k8s.io/kustomize/api v0.8.5
+	sigs.k8s.io/kustomize/api v0.3.3
```
morningvera, line 106: Had to downgrade this because kustomize v3.5.4 uses this version of the api

sgalsaleh: dependabot is not gonna be happy about this :)

pkg/base/helm.go
```diff
@@ -271,3 +274,104 @@ func checkChartForVersion(file *upstreamtypes.UpstreamFile) (string, error) {
...
+
+	var updatedBaseFiles []BaseFile
+	var kustomizeResources []string
+	var kustomizePatches []kustomizetypes.PatchStrategicMerge
+	resources := map[string]BaseFile{}
+	foundGVKNamesMap := map[string]bool{}
+	for _, baseFile := range baseFiles {
+		// write temp files for manifests that need a namespace
+		gvk, manifest := GetGVKWithNameAndNs(baseFile.Content, renderOptions.Namespace)
+		if manifest.APIVersion == "" && manifest.Kind == "" && manifest.Metadata.Name == "" {
```
sgalsaleh, line 298: i feel like this should be or'ed instead of and'ed

morningvera: done

go.mod
morningvera: reverted this to v0.8.5 since the replace is sufficient

pkg/base/helm_test.go
```diff
@@ -90,11 +90,11 @@ func Test_RenderHelm(t *testing.T) {
 						},
 						{
 							Path:    "templates/deploy-1.yaml",
-							Content: []byte("file: 1\nkind: Deployment\nmetadata:\n  name: deploy-1\n  namespace: test-one"),
+							Content: []byte("apiVersion: v1\nkind: Deployment\nmetadata:\n  name: deploy-1\n  namespace: test-one"),
```
sgalsaleh, line 93: deployments are `apps/v1` not `v1` i think :)

sgalsaleh (approved):

pkg/base/helm_test.go
morningvera: hah fair enough 😄

pkg/base/helm.go
```diff
@@ -271,3 +274,104 @@ func checkChartForVersion(file *upstreamtypes.UpstreamFile) (string, error) {
...
+			updatedBaseFiles = append(updatedBaseFiles, baseFile)
+			continue // ignore invalid resources
+		}
+
+		if manifest.Metadata.Namespace == "" {
+			name := filepath.Base(baseFile.Path)
+			tmpFile, err := ioutil.TempFile(chartsPath, name)
+			if err != nil {
+				return nil, errors.Wrapf(err, "failed to write temp file %v", tmpFile.Name())
+			}
```
divolgin, line 308: Since this is going to be a long running service, this should also defer close to prevent file handle leaks.  It's ok to call `Close` twice on a file.

divolgin (approved):

pkg/base/helm.go
morningvera: fixed

divolgin (approved):

PR #1983, opened 2021-07-19
Grayson/variadic config files

GraysonNull:


web/src/components/config_render/ConfigFileInput.jsx
```diff
@@ -19,9 +22,15 @@ export default class ConfigFileInput extends React.Component {
...
-    if (this.props.multiple) {
-      if (this.props.multi_data && this.props.multi_data.length) {
-        return this.props.multi_data.join(", ");
+    if (this.props.repeatable) {
+      if (this.props.valuesByGroup && Object.keys(this.props.valuesByGroup[this.props.groupName]).length) {
+        const filenames= [];
+        Object.keys(this.props.valuesByGroup[this.props.groupName]).map((key) => {
+          filenames.push(key);
+        });
+        if (filenames.length > 0) {
```
sgalsaleh, line 31: if this is false, the function doesn't return anything. is that expected?

sgalsaleh (approved):

web/src/components/config_render/ConfigFileInput.jsx
GraysonNull: Yeah I think it should be fine. There are null checks and an empty array check where this value is being read.

PR #1990, opened 2021-07-22
Fix kots pull command

divolgin:
- Fixes pull command to not look for Helm charts
- Makes test build components in parallel
- Adds a test for kots pull command

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -20,51 +20,49 @@ jobs:
         shell: bash
 
       - uses: actions/checkout@v2
-
       - run: make ci-test kots
       - uses: actions/upload-artifact@v2.2.4
         with:
           name: kots
           path: bin/kots
 
+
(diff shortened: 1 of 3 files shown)


.github/workflows/build-test.yaml
```diff
@@ -474,3 +474,26 @@ jobs:
           kubectl -n $APP_NAME logs -p deployment/kotsadm-operator
           kill $ADMIN_CONSOLE_PID
           exit $EXIT_CODE
+
+
+  validate-kots-pull:
```
sgalsaleh, line 479: this doesn't validate the result, but it's better than nothing

sgalsaleh (approved):

.github/workflows/build-test.yaml
divolgin: i'm gonna add kubectl apply at least

sgalsaleh (approved):

PR #1994, opened 2021-07-23
fix(operator): surface deployment errors

DanStough:
Errors weren't being returned from `ensureResourcesPresent`. Tried to simplify this to remove the send in multiple places. Could be simplified further if we always send back the result w/ stdout,stderr, but I kept the existing functionality where those are only returned on error.

### DIFF
--- a/kotsadm/operator/pkg/client/client.go
+++ b/kotsadm/operator/pkg/client/client.go
@@ -231,38 +231,43 @@ func (c *Client) registerHandlers(socketClient *socket.Client) error {
 
 		log.Println("received a deploy request for", args.AppSlug)
 
-		var applyResult, helmResult *commandResult
-		var applyError, helmError error
+		var deployRes *deployResult
+		var helmResult *commandResult
+		var deployError, helmError error
 		defer func() {
-			if applyError != nil {
-				applyResult = &commandResult{
-					hasErr:      true,
-					multiStdout: [][]byte{},
-					multiStderr: [][]byte{[]byte(applyError.Error())},
-				}
+			var dryRunResult, applyResult *commandResult
+
+			// only send results if we have an error
+			if deployRes.dryRunResult.hasErr {
+				dryRunResult = &deployRes.dryRunResult
 			}
-			if helmError != nil {
-				helmResult = &commandResult{
-					hasErr:      true,
-					multiStdout: [][]byte{},
-					multiStderr: [][]byte{[]byte(helmError.Error())},
-				}
+			if deployRes.applyResult.hasErr {
+				applyResult = &deployRes.applyResult
+			}
+			if helmResult != nil && !helmResult.hasErr {
+				helmResult = &commandResult{}
 			}
 
-			err := c.sendResult(args, nil, applyResult, helmResult)
+			err := c.sendResult(args, dryRunResult, applyResult, helmResult)
 			if err != nil {
 				log.Printf("failed to report result: %v", err)
 			}
 		}()
 
-		applyResult, applyError = c.deployManifests(args)
-		if applyError != nil {
-			log.Printf("falied to deploy manifests: %v", applyError)
+		deployRes, deployError = c.deployManifests(args)
+		if deployError != nil {
+			deployRes = &deployResult{}
+			deployRes.applyResult.hasErr = true
+			deployRes.applyResult.multiStderr = [][]byte{[]byte(deployError.Error())}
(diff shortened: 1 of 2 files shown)


emosbaugh (approved):
i approve but you should get someone else's eyes on this one as im not familiar with the code

kotsadm/operator/pkg/client/client.go
```diff
@@ -231,38 +231,41 @@ func (c *Client) registerHandlers(socketClient *socket.Client) error {
 
 		log.Println("received a deploy request for", args.AppSlug)
 
-		var applyResult, helmResult *commandResult
-		var applyError, helmError error
+		var deployRes *deployResult
+		var dryRunResult, applyResult, helmResult *commandResult
```
sgalsaleh, line 235: can we move `dryRunResult` and `applyResult` to be inside the deferred function so that it's a bit clearer?

kotsadm/operator/pkg/client/client.go
```diff
@@ -316,12 +319,7 @@ func (c *Client) deployManifests(args ApplicationManifests) (*commandResult, err
 	c.imagePullSecret = args.ImagePullSecret
 	c.watchedNamespaces = args.AdditionalNamespaces
 
-	result, err := c.ensureResourcesPresent(args)
```
sgalsaleh, line 319: i like the old format since it has an additional error wrap and could be helpful when debugging

kotsadm/operator/pkg/client/client.go
```diff
@@ -231,38 +231,41 @@ func (c *Client) registerHandlers(socketClient *socket.Client) error {
...
-		applyResult, applyError = c.deployManifests(args)
-		if applyError != nil {
-			log.Printf("falied to deploy manifests: %v", applyError)
+		deployRes, deployError = c.deployManifests(args)
+		if deployError != nil {
+			deployRes = &deployResult{}
+			deployRes.applyResult.hasErr = true
+			deployRes.applyResult.multiStderr = [][]byte{[]byte(deployError.Error())}
+			log.Printf("falied to deploy manifests: %v", deployError)
 			return
```
sgalsaleh, line 261:
```suggestion
			return nil
```

kotsadm/operator/pkg/client/client.go
```diff
@@ -231,38 +231,41 @@ func (c *Client) registerHandlers(socketClient *socket.Client) error {
...
 		}
 
 		helmResult, helmError = c.deployHelmCharts(args)
 		if helmError != nil {
+			helmResult = &commandResult{}
+			helmResult.hasErr = true
+			helmResult.multiStderr = [][]byte{[]byte(helmError.Error())}
 			log.Printf("falied to deploy helm charts: %v", helmError)
 			return
```
sgalsaleh, line 270:
```suggestion
			return nil
```

sgalsaleh (approved):

kotsadm/operator/pkg/client/client.go
DanStough: Does this makes sense if we don't have return argument for this function?

sgalsaleh: oh, no it doesn't 🤦 . github diff did not help there for sure.

kotsadm/operator/pkg/client/client.go
```diff
@@ -231,38 +231,43 @@ func (c *Client) registerHandlers(socketClient *socket.Client) error {
...
 			}
-			if helmError != nil {
-				helmResult = &commandResult{
-					hasErr:      true,
-					multiStdout: [][]byte{},
-					multiStderr: [][]byte{[]byte(helmError.Error())},
-				}
+			if helmResult != nil && !helmResult.hasErr {
+				helmResult = &commandResult{}
 			}
```
DanStough, line 249: Changed the logic here.

kotsadm/operator/pkg/client/client.go
```diff
@@ -318,7 +323,7 @@ func (c *Client) deployManifests(args ApplicationManifests) (*commandResult, err
 
 	result, err := c.ensureResourcesPresent(args)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to deploy")
+		return nil, errors.Wrap(err, "failed to execute manifest deployment")
```
jala-dx, line 326: this log, I would leave it as before.

jala-dx (approved):

PR #1996, opened 2021-07-23
chore: add kotsadm-operator skaffold cache

DanStough:
Signed-off-by: Dan Stough <dans@replicated.com>


kotsadm/operator/Dockerfile.skaffold
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.16 as deps
+FROM kotsadm-operator:cache AS builder
```
divolgin, line 1: This should use multistage to reduce image size.  We don't need all dependencies in the running image; just the operator binary.  Similar to this https://github.com/replicatedhq/kots/blob/master/hack/dev/Dockerfile.skaffold#L86

divolgin (approved):

PR #2009, opened 2021-07-28
Adding testing framework for rendering upstream to base

emosbaugh:
This pull request attempts to fill a gap in KOTS test coverage. We have very little coverage of combinations of application yaml and upstream configuration.

Currently kots has unit tests and e2e tests. Unit tests for renderReplicated can cover the same but are cumbersome to write because they are in code. E2e tests also can cover this but are slow.

This adds a testing framework for testing kots rendering of upstream yaml to base yaml. To add tests, the developer must include yaml files in the directories upstream and base, which include input yaml and expected output yaml respectively.


pkg/base/render/render_test.go
```diff
@@ -0,0 +1,243 @@
...
+
+		// TODO: helm bases
+
+		tests = append(tests, test)
+	}
+	require.NoError(t, err)
+
+	for _, tt := range tests {
+		t.Run(tt.Name, func(t *testing.T) {
+			template.TestingDisableKurlValues = true
```
jeffreygolden, line 85: quick comment here to describe what effect this setting has

pkg/template/kurl_context.go
```diff
@@ -15,6 +15,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+var (
+	TestingDisableKurlValues = false
```
jeffreygolden, line 19: Comment describing use case for this global var

emosbaugh:
[ch36201]

sgalsaleh (approved):

divolgin (approved):

PR #2014, opened 2021-07-28
feat: add cli command for generating admin console manifests

morningvera:
Adds a new CLI command for generating admin console manifests ([CH36021](https://app.clubhouse.io/replicated/story/36021/a-cli-command-or-a-flag-for-the-kots-pull-command-to-generate-only-the-admin-console-components-as-kustomizable-yaml-manifests)).

### DIFF
--- a/pkg/upstream/admin-console.go
+++ b/pkg/upstream/admin-console.go
@@ -32,7 +32,7 @@ type UpstreamSettings struct {
 	IncludeMinio           bool
 }
 
-func generateAdminConsoleFiles(renderDir string, options types.WriteOptions) ([]types.UpstreamFile, error) {
+func GenerateAdminConsoleFiles(renderDir string, options types.WriteOptions) ([]types.UpstreamFile, error) {
 	if _, err := os.Stat(path.Join(renderDir, "admin-console")); os.IsNotExist(err) {
 		settings := &UpstreamSettings{
 			SharedPassword:         options.SharedPassword,
--- a/cmd/kots/cli/admin-console.go
+++ b/cmd/kots/cli/admin-console.go
@@ -83,6 +83,7 @@ func AdminConsoleCmd() *cobra.Command {
 	cmd.AddCommand(AdminConsoleUpgradeCmd())
 	cmd.AddCommand(AdminPushImagesCmd())
 	cmd.AddCommand(GarbageCollectImagesCmd())
+	cmd.AddCommand(AdminGenerateManifestsCmd())
 
 	return cmd
 }
--- a/cmd/kots/cli/admin-generate-manifests.go
+++ b/cmd/kots/cli/admin-generate-manifests.go
@@ -1,0 +1,71 @@
+package cli
+
+import (
+	"io/ioutil"
+	"os"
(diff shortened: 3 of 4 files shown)


pkg/upstream/admin-console.go
```diff
@@ -32,6 +32,29 @@ type UpstreamSettings struct {
...
+		fileRenderPath := path.Join(renderDir, file.Path)
+		d, _ := path.Split(fileRenderPath)
+		if _, err := os.Stat(d); os.IsNotExist(err) {
+			if err := os.MkdirAll(d, 0744); err != nil {
+				return errors.Wrap(err, "failed to mkdir")
+			}
+		}
+
+		if err := ioutil.WriteFile(fileRenderPath, file.Content, 0644); err != nil {
+			return errors.Wrap(err, "failed to write upstream file")
```
sgalsaleh, line 51:
```suggestion
			return errors.Wrapf(err, "failed to write file %s", fileRenderPath)
```

pkg/upstream/admin-console.go
```diff
@@ -32,6 +32,29 @@ type UpstreamSettings struct {
 	IncludeMinio           bool
 }
 
+func WriteAdminConsoleFiles(renderDir string, options types.WriteOptions) error {
```
sgalsaleh, line 35: renderDir won't make sense since we're not really rendering anything, maybe:
```suggestion
func WriteAdminConsoleFiles(outputDir string, options types.WriteOptions) error {
```

sgalsaleh, line 35: i don't think this function belongs in the upstream package IMO, i would just dump the functions logic in the cli command itself

sgalsaleh (approved):

pkg/upstream/admin-console.go
```diff
@@ -32,6 +32,29 @@ type UpstreamSettings struct {
...
+	adminConsoleFiles, err := generateAdminConsoleFiles(renderDir, options)
+	if err != nil {
+		return errors.Wrap(err, "failed to generate admin console files")
+	}
+
+	for _, file := range adminConsoleFiles {
+		fileRenderPath := path.Join(renderDir, file.Path)
```
sgalsaleh, line 42: same for `fileRenderPath`, we're not rendering anything, so maybe just `filePath`?

pkg/upstream/admin-console.go
sgalsaleh, line 35: i feel like this function doesn't belong in the upstream package IMO, i would just dump this logic in the cli function. but i'll leave the decision up to you.

morningvera: yeah i thought about that but the method that it calls was private to the package. i'm fine with moving it and making that method public though.

morningvera: to be fair, it kinda is rendering YAML and `renderDir` is used throughout the admin console generation code. so i'm probably keeping it to be consistent with that.

pkg/upstream/admin-console.go
morningvera: see response above

PR #2017, opened 2021-07-29
feature: migration from minio snapshots to local-volume-provider plugin

DanStough:


pkg/kotsadm/main.go
```diff
@@ -1024,3 +1059,63 @@ func GetKotsadmOptionsFromCluster(namespace string, clientset kubernetes.Interfa
...
+	}
+
+	if !veleroStatus.ContainsPlugin("local-volume-provider") {
+		log.Info("velero local-volume-provider plugin is not installed - migration cannot be performed")
+		log.Info("")
+		log.Info("---TO INSTALL the local-volume-provider plugin---")
+		log.Info("For existing cluster installations, complete the installation instructions here:")
+		log.Info("https://github.com/replicatedhq/local-volume-provider")
+		log.Info("")
+		log.Info("If you're seeing this message on an embedded cluster installation, please contact your vendor for support.")
```
DanStough, line 1091: WIP

DanStough:
CLI Output (minus the "Checking Time")
image

cmd/kots/cli/velero.go
```diff
@@ -611,8 +607,7 @@ func VeleroConfigureAzureCmd() *cobra.Command {
 				return errors.New("velero namespace not found")
 			}
 
-			// init containers are names differently starting in velero 1.6
-			if !veleroStatus.ContainsPlugin("velero-plugin-for-microsoft-azure") || !veleroStatus.ContainsPlugin("velero-velero-plugin-for-microsoft-azure") {
```
sgalsaleh, line 615: is this no longer the case?

DanStough: I modified `ContainsPlugin` to search for the substring. Not sure how this was working in airgap with a custom registry namespace.

pkg/handlers/snapshots.go
```diff
@@ -187,16 +208,17 @@ func (h *Handler) UpdateGlobalSnapshotSettings(w http.ResponseWriter, r *http.Re
...
 		KotsadmNamespace: kotsadmNamespace,
 		RegistryOptions:  &registryOptions,
+		IsMinioDisabled:  globalSnapshotSettingsResponse.IsMinioDisabled,
 	}
 	updatedStore, err := kotssnapshot.ConfigureStore(r.Context(), options)
 	if err != nil {
 		if _, ok := errors.Cause(err).(*kotssnapshot.InvalidStoreDataError); ok {
 			logger.Error(err)
-			globalSnapshotSettingsResponse.Error = "invalid store data"
+			globalSnapshotSettingsResponse.Error = "invalid store data: " + err.Error()
```
sgalsaleh, line 221: would suggest using `fmt.Sprintf` here instead of `+`

pkg/kotsadm/kotsadm.go
```diff
@@ -252,72 +253,6 @@ func ensureKotsadmClusterRoleBinding(serviceAccountNamespace string, clientset *
...
-
-	return nil
-}
-
-func updateKotsadmRoleBinding(existing, desiredRoleBinding *rbacv1.RoleBinding) *rbacv1.RoleBinding {
-	existing.Subjects = desiredRoleBinding.Subjects
-
-	return existing
-}
-
```
DanStough, line 320: Moved to a sub-package to prevent circular dependencies.

pkg/kotsadm/secrets.go
```diff
@@ -383,24 +383,3 @@ func getAPIClusterToken(namespace string, clientset *kubernetes.Clientset) (stri
...
-		}
-
-		_, err := clientset.CoreV1().Secrets(namespace).Create(context.TODO(), secret, metav1.CreateOptions{})
-		if err != nil {
-			return errors.Wrap(err, "failed to create private kotsadm registry secret")
-		}
-	}
-
-	return nil
-}
```
DanStough, line 406: moved

pkg/snapshot/filesystem.go
```diff
@@ -139,6 +165,8 @@ func ensureFileSystemConfigMap(ctx context.Context, clientset kubernetes.Interfa
 func fileSystemConfigMapResource(fileSystemConfig types.FileSystemConfig) *corev1.ConfigMap {
 	data := map[string]string{}
 
+	// TODO: put in the migration status here
+
```
DanStough, line 169: I'll delete this

pkg/snapshot/filesystem.go
```diff
@@ -719,8 +750,10 @@ func fileSystemMinioConfigPod(clientset kubernetes.Interface, deployOptions File
 		}
 	}
 
-	kotsadmTag := kotsadmversion.KotsadmTag(kotsadmtypes.KotsadmOptions{}) // default tag
-	image := fmt.Sprintf("kotsadm/kotsadm:%s", kotsadmTag)
+	// kotsadmTag := kotsadmversion.KotsadmTag(kotsadmtypes.KotsadmOptions{}) // default tag
+	// image := fmt.Sprintf("kotsadm/kotsadm:%s", kotsadmTag)
+	// TODO: REMOVE ME
+	image := "localhost:32000/kotsadm:v1.49.0-10-gaff83f7e8-dirty@sha256:ad62a01b7b076883c538f1480242bdf57fc79100cda306a241e12171c453ae07"
```
DanStough, line 756: I'll change this when I'm done testing - this is how I am using the new check script that looks for writable permissions.

pkg/kotsadm/minio.go
```diff
@@ -115,3 +119,162 @@ func ensureMinioService(namespace string, clientset *kubernetes.Clientset) error
...
+
+	start := metav1.Now()
+	success := false
+	defer func() {
+		if !success {
+			err := snapshot.RevertToMinioFS(context.TODO(), deployOptions.Namespace, veleroNamespace, previousBsl)
+			if err != nil {
+				log.Error(errors.Wrap(err, "Could not restore minio backup storage location"))
+			}
+			log.Info("Minio backup storage location restored")
```
sgalsaleh, line 224: feels like this needs to be in an `else` statement

pkg/kotsadm/minio.go
```diff
@@ -115,3 +119,162 @@ func ensureMinioService(namespace string, clientset *kubernetes.Clientset) error
...
+		}
+	}()
+
+	// New bucket name will be assigned during configuration
```
sgalsaleh, line 228: why not just use the same bucket?

pkg/snapshot/filesystem.go
```diff
@@ -139,6 +165,8 @@ func ensureFileSystemConfigMap(ctx context.Context, clientset kubernetes.Interfa
 func fileSystemConfigMapResource(fileSystemConfig types.FileSystemConfig) *corev1.ConfigMap {
 	data := map[string]string{}
 
+	// TODO: put in the migration status here
```
sgalsaleh, line 168: should the be handled now?

pkg/snapshot/filesystem.go
```diff
@@ -817,7 +849,59 @@ func getFileSystemResetWarningMsg(fileSystemConfig types.FileSystemConfig) strin
 	return fmt.Sprintf("The %s directory was previously configured by a different minio instance.\nProceeding will re-configure it to be used only by the minio instance we deploy to configure the file system, and any other minio instance using this location will no longer have access.\nIf you are attempting to fully restore a prior installation, such as a disaster recovery scenario, this action is expected.", path)
 }
 
-func GetCurrentFileSystemConfig(ctx context.Context, namespace string) (*types.FileSystemConfig, error) {
+func GetCurrentFileSystemConfig(ctx context.Context, namespace string, isMinioDisabled bool) (*types.FileSystemConfig, error) {
+
```
sgalsaleh, line 853: would love to keep the current pattern of not having the function begin with a new line :)

pkg/snapshot/filesystem.go
```diff
@@ -817,7 +849,59 @@ func getFileSystemResetWarningMsg(fileSystemConfig types.FileSystemConfig) strin
...
+	var fileSystemConfig *types.FileSystemConfig
+
+	if !isMinioDisabled {
+		fileSystemConfig, err := GetCurrentMinioFileSystemConfig(ctx, namespace)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to get current minio file system config")
+		}
+		if fileSystemConfig != nil {
+			return fileSystemConfig, nil
+		}
```
sgalsaleh, line 863: probably need a `return nil, nil` after this

pkg/snapshot/filesystem.go
```diff
@@ -848,3 +932,137 @@ func GetCurrentFileSystemConfig(ctx context.Context, namespace string) (*types.F
...
+	if err != nil {
+		return errors.Wrap(err, "failed to revert to minio backup storage location")
+	}
+	return nil
+}
+
+// DeleteFileSystemMinio cleans up the minio resources for hostpath and nfs snapshot deployments.
+// The secret is not deleted, just in case.
+func DeleteFileSystemMinio(ctx context.Context, kotsadmNamespace string) error {
+	clientset, err := k8sutil.GetClientset()
```
sgalsaleh, line 956: it's always a good idea to just pass down the original clientset to avoid creating new ones on the fly.

pkg/snapshot/store.go
```diff
@@ -277,7 +286,35 @@ func ConfigureStore(ctx context.Context, options ConfigureStoreOptions) (*types.
...
+		store.Internal = nil
+
+		b, err := json.Marshal(options.FileSystem)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to marshal filesystem config")
+		}
+
+		hash := md5.Sum(b)
+		hashId := hex.EncodeToString(hash[:6])
+		store.Bucket = "velero-lvp-" + hashId
```
sgalsaleh, line 308: why not use a fixed bucket name? won't this start using a different bucket if we add more options to the file system config later?

pkg/snapshot/store.go
```diff
@@ -929,6 +1037,17 @@ func BuildStoreFileSystem(ctx context.Context, clientset kubernetes.Interface, k
 	return &storeFileSystem, nil
 }
 
+func BuildLvpStoreFileSystem(ctx context.Context, clientset kubernetes.Interface, kotsadmNamespace string, config *types.FileSystemConfig) (*types.StoreFileSystem, error) {
+	storeFileSystem := types.StoreFileSystem{}
+
+	if config.NFS != nil && config.NFS.Path == "" {
```
sgalsaleh, line 1043: what about hostpath?

pkg/snapshot/store.go
```diff
@@ -1194,7 +1334,37 @@ func validateInternal(ctx context.Context, storeInternal *types.StoreInternal, b
 	return nil
 }
 
-func validateFileSystem(ctx context.Context, storeFileSystem *types.StoreFileSystem, bucket string, options ValidateStoreOptions) error {
+// validateLvpFileSystem checks that the fs configuration can be mounted, that the chosen directory is writable and
+// also test for legacy minio files. If minio files are detected, the store is updated with the /velero prefix
+func validateLvpFileSystem(ctx context.Context, store *types.Store, options ValidateStoreOptions) error {
```
sgalsaleh, line 1339: i don't see where this function validates that the bucket exists

web/src/components/snapshots/ConfigureSnapshots.jsx
```diff
@@ -103,10 +103,23 @@ class ConfigureSnapshots extends React.Component {
...
+                            <>
+                              <a href="https://github.com/replicatedhq/local-volume-provider" target="_blank" rel="noopener noreferrer" className="snapshotOptions">
+                                <span style={{ width: "130px" }}>  <span className="icon nfsIcon u-cursor--pointer u-marginRight--5" /> NFS </span>
+                                <span className="icon external-link-icon u-cursor--pointer u-marginLeft--30" /></a>
+                              <a href="https://github.com/replicatedhq/local-volume-provider" target="_blank" rel="noopener noreferrer" className="snapshotOptions">
+                                <span style={{ width: "130px" }}>  <span className="icon hostpathIcon u-cursor--pointer u-marginRight--5" /> Host Path </span>
+                                <span className="icon external-link-icon u-cursor--pointer u-marginLeft--30" /></a>
+                            </>
+                          :
+                            <>
```
sgalsaleh, line 116: whooh, i didn't this was possible

web/src/components/snapshots/SnapshotStorageDestination.jsx
```diff
@@ -904,20 +904,32 @@ class SnapshotStorageDestination extends Component {
...
-              label: "Network File System (NFS)",
-            });
-            availableDestinations.push({
-              value: "hostpath",
-              label: "Host Path",
-            });
+            // Checks for legacy behavior where minio was used for hostpath and nfs
+            if (!snapshotSettings.isMinioDisabled) {
+                availableDestinations.push({
+                value: "nfs",
```
sgalsaleh, line 910: indentation

web/src/components/snapshots/SnapshotStorageDestination.jsx
```diff
@@ -904,20 +904,32 @@ class SnapshotStorageDestination extends Component {
...
 
+        } else if (veleroPlugin.includes("local-volume-provider") && snapshotSettings.isMinioDisabled) {
+            availableDestinations.push({
+                value: "nfs",
+                label: "Network File System (NFS)",
+                });
+                availableDestinations.push({
+                value: "hostpath",
+                label: "Host Path",
+                });
```
sgalsaleh, line 932: indentation

cmd/kots/cli/admin-console-upgrade.go
```diff
@@ -59,6 +59,13 @@ func AdminConsoleUpgradeCmd() *cobra.Command {
 			}
 
 			includeMinio := v.GetBool("with-minio")
+
+			// if the cli flag for kotsadm minio was explicitly passed, disable minio for snapshots
+			includeMinioSnapshots := v.GetBool("with-minio-snapshots")
```
sgalsaleh, line 64: IMO, we should just use the `with-minio` flag for this as well. the customer just wants to NOT use minio, they won't care if it's just the main minio instance vs the one we deploy for snapshots.

web/src/components/snapshots/ConfigureSnapshots.jsx
DanStough: You mean the [fragment](https://reactjs.org/docs/fragments.html)? Picking up the occasional React tip :D

pkg/snapshot/store.go
DanStough: Mostly just forward looking that maybe we will support multiple backup storage locations someday, and this would prevent naming collisions. The bucket name is used as the name of the volume attached to Velero/Restic. LVP plugin should replace the volume mount name if the path changes, so I can stick with a fixed name if you feel strongly about it.

pkg/snapshot/store.go
DanStough: Hmm, good point but I think we would want to require a path for hostpath, where with NFS we can probably assume we can mount to the root of the volume, right? Not sure if we have the FE validation for this though.

cmd/kots/cli/admin-console-upgrade.go
DanStough: I think the previous thought was that it would be confusing to specify the same flag again, but I agree with you the more I think about it. cc @JustinN-Replicated

pkg/kotsadm/minio.go
DanStough: yeah, or a `return`

pkg/kotsadm/minio.go
DanStough: bucket is used like the name/id of a volume, so I wanted to call them something more descriptive than `velero` when we add them to LVP. Maybe there is some future point where there are more than one BSL and I wanted to make them each unique.

DanStough:
Kurl support will be a separate PR. It will be a hidden migration command where noted in the `TODO`.

cmd/kots/cli/velero.go
```diff
@@ -44,6 +48,8 @@ func VeleroCmd() *cobra.Command {
 	cmd.AddCommand(VeleroConfigureNFSCmd())
 	cmd.AddCommand(VeleroConfigureHostPathCmd())
 	cmd.AddCommand(VeleroPrintFileSystemInstructionsCmd())
+	// TODO (dans)
+	// cmd.AddCommand(VeleroMigrationFileSystemCmd())
```
sgalsaleh, line 52: do we still need such a command?

DanStough: Yes, this will be what the addon calls to execute the migration since it doesn't run `upgrade`.

cmd/kots/cli/velero.go
```diff
@@ -830,93 +836,100 @@ func VeleroConfigureHostPathCmd() *cobra.Command {
...
-
-	veleroNamespace, err := snapshot.DetectVeleroNamespace(ctx, clientset, opts.Namespace)
+	veleroStatus, err := snapshot.DetectVelero(ctx, opts.Namespace)
 	if err != nil {
 		return errors.Wrap(err, "failed to detect velero namespace")
 	}
-	if veleroNamespace == "" {
-		c, err := buildPrintableFileSystemVeleroConfig(ctx, clientset, opts.Namespace)
+
+	if (veleroStatus == nil && !opts.IsMinioDisabled) || (veleroStatus != nil && !veleroStatus.ContainsPlugin("plugin-for-aws") && !opts.IsMinioDisabled) {
```
sgalsaleh, line 902: it's a bit hard to figure out what this condition is trying to check for at first. i get it, but it took me a bit

pkg/handlers/snapshots.go
```diff
@@ -176,6 +187,16 @@ func (h *Handler) UpdateGlobalSnapshotSettings(w http.ResponseWriter, r *http.Re
 		}
 	}
 
+	if updateGlobalSnapshotSettingsRequest.FileSystem != nil &&
```
sgalsaleh, line 190: personal nit pick, this can be moved to the if statement above, like so:
```go
if updateGlobalSnapshotSettingsRequest.FileSystem != nil {
    if !globalSnapshotSettingsResponse.IsMinioDisabled {
         ....
    } else {
        ....
    }
}
```

pkg/kotsadm/main.go
```diff
@@ -116,6 +117,12 @@ func Upgrade(clientset *kubernetes.Clientset, upgradeOptions types.UpgradeOption
 		return errors.Wrap(err, "failed to read deploy options")
 	}
 
+	// If user has passed in the flag to migrate minio, save this status as part of the install
+	// Only works if they have not already chose to opt-in
+	if !upgradeOptions.IncludeMinio && deployOptions.IncludeMinioSnapshots {
```
sgalsaleh, line 122: i guess this is not valid now, right? since they're both using the same flag

DanStough: `deployOptions` contains the value from the configmap if they previously opt-ed in. My intent was trying to prevent folks from going backward and specifying `with-minio=true`.

pkg/snapshot/filesystem.go
```diff
@@ -765,7 +794,6 @@ func fileSystemMinioConfigPod(clientset kubernetes.Interface, deployOptions File
 						{
 							Name:      "fs",
 							MountPath: "/fs",
-							ReadOnly:  readOnly,
```
sgalsaleh, line 768: can we keep this if it's still using minio?

cmd/kots/cli/admin-console-upgrade.go
JustinN-Replicated: From standup, I am OK with us consolidating under 1 flag; I will need to follow up with additional details to the email that already went out this week but it should be easy to put together and get sent out.

pkg/snapshot/store.go
sgalsaleh: ah, cool

pkg/snapshot/filesystem.go
DanStough: I think it's a value add to be able to check if the volume is writable even in the Minio case. I've run afoul of this many times testing, and now the user will be notified in the UI. The check will fail with this flag enabled. 

I can revert depending on how strongly you feel.

sgalsaleh (approved):
we'll need to include the plugin images in the airgap bundle, but that can be in a separate PR

cmd/kots/cli/velero.go
```diff
@@ -1200,3 +1199,60 @@ func validateVeleroNamespace(namespace string) error {
...
+			}
+
+			deployOptions, err := kotsadm.ReadDeployOptionsFromCluster(namespace, clientset)
+			if err != nil {
+				return errors.Wrap(err, "failed to read deploy options")
+			}
+
+			if !deployOptions.IncludeMinioSnapshots {
+				// Already migrated, so this is a no-op
+				log.Info("Previous minio preference detected; migration not needed")
```
sgalsaleh, line 1239: is this message right?

sgalsaleh (approved):

cmd/kots/cli/velero.go
DanStough: Changed per our slack convo.

DanStough: test

sgalsaleh: can you hear me now?

PR #2020, opened 2021-07-30
Embedded K8s experiment via kots run

marccampbell:
Caution: This is very experimental!

This PR adds a new command: `kots run <app slug>`

When running this command, the kots binary will start the Admin Console API (and embedded web server) locally on the server (not in a Kubernetes cluster). Additionally, when running `kots run`, kots will start an embedded Kubernetes "cluster" (api-server, controller-manager, scheduler). The intent here is that this is new cluster will be very predictable and not vendor or end-user configurable.

## Why?
Simply put: configuring Kubernetes is complex. It's difficult for someone to manage when they don't have expertise. 
Embedding Kubernetes (technically k3s) provides a very specific, predictble cluster that will run the single application.
This is limited to a single node for now -- it's not a built in limitation, but that's all I'm hoping to support at this time.
Starting KOTS will start the vendored in k8s servers in goroutines. 
There's no separate kubernetes cluster to manage that can fail.
There are no environments to configure as everything is local (no firewalls, it's all localhost. no storage drivers, it's all local).
This is using Kine, so that it's using sqlite instead of etcd for Kubernetes object storage.
Eventually, we could embed the vendor images and manaifests into the binary also, giving a total airgap solution for single node installs (dev, POC, or even small prod) that don't require any extra components.

In addition, this tightly couples the k8s version and distribution to the kots version. 

This does does add ~ 175 MB to the kots binary!

### To accomplish this, a few changes needed to be made in KOTS:
1. postgres is a difficult requirement to expect someone to bring. We aren't using any Postgres specific extensions, so I made this possible to run with a local SQLite database. The standard kots will continue to use Postgres for now.
2. The web ui was just files copied into a container image. I moved this to be embedded in the go binary so that we can have the admin console ui and a single binary distribution.
3. Some changes to stop reading os.Getenv() and pass in as params

## Current status

So, does it work? No, not yet. This PR is very large and I want to give an opportunity for feedback and to merge and continue working in a new PR to keep the changes smaller for review.

What's left to make it completely work?
- [x] Use SQLite instead of Postgres when running from the CLI
- [x] Start kubernetes api server 
- [x] Start controller-manager
- [x] Start scheduler
- [x] Write a kubeconfig file to disk for direct access
- [ ] Write a systemd service to start and manage kots admin console (and cluster) lifecycle
- [ ] Install kubelet
- [ ] Install containerd
- [ ] Install antrea to the cluster
- [ ] Install a host path provisioner for PVCs
- [ ] Configure the admin console to install the app to the cluster
- [ ] Run the current e2e tests in the `kots run` environment to validate that all functionality works


cmd/kots/cli/root.go
```diff
@@ -42,6 +42,7 @@ func RootCmd() *cobra.Command {
 	cmd.AddCommand(AppStatusCmd())
 	cmd.AddCommand(GetCmd())
 	cmd.AddCommand(SetCmd())
+	cmd.AddCommand(RunCmd())
```
emosbaugh, line 45: should we make the command hidden or maybe better something like `kots alpha run`?

pkg/filestore/blob_store.go
```diff
@@ -28,9 +28,7 @@ func (s *BlobStore) Init() error {
 
 func (s *BlobStore) WaitForReady(ctx context.Context) error {
 	// there's no waiting, it's either there or it's not. the pod won't come up if the volume didn't mount.
-	volumeMount := "/kotsadmdata"
-	_, err := os.Stat(volumeMount)
-	return err
+	return nil
```
emosbaugh, line 31: what is the reason for this change?

marccampbell: When running kotsadm on the host, the paths are different, and this won't work.
Also, this isn't really a readiness check. It's not an error condition that's going to resolve over time, this is here just incase we fail to mount this directory. This interface function feels like it shouldn't really exist since all stores don't need a "wait for". Minio needed it, so we copied it over which makes sense. But short of removing that interface and refactoring, this just didn't make sense as a check that we will "wait for" at startup.

cmd/kots/cli/run.go
```diff
@@ -0,0 +1,109 @@
...
+		EnableIdentity:         false,
+	}
+
+	go apiserver.Start(&params)
+
+	return nil
+}
+
+func startK8sAuthnz(ctx context.Context, dataDir string) error {
+	go cluster.StartAuthnzServer()
```
divolgin, line 106: this should probably panic on error or at least log the error

cmd/kots/cli/run.go
```diff
@@ -0,0 +1,111 @@
...
+			if len(args) == 0 {
+				cmd.Help()
+				os.Exit(1)
+			}
+
+			slug := args[0]
+			log := logger.NewCLILogger()
+			log.Info("Running application %s", slug)
+
+			loggerCtx := context.WithValue(context.Background(), "log", log)
```
emosbaugh, line 44: im not sure i agree with this pattern. im not sure context is the best place for DI

cmd/kots/cli/run.go
```diff
@@ -0,0 +1,111 @@
...
+			if err := cluster.Start(ctx, slug, v.GetString("data-dir")); err != nil {
+				return err
+			}
+
+			// wait for interrupt, and stop the cluster when we receive
+			log.Info("The cluster is running. Press ctrl+c to terminate")
+			c := make(chan os.Signal, 1)
+			signal.Notify(c, os.Interrupt, syscall.SIGTERM)
+			<-c
+			cancelFunc()
```
emosbaugh, line 73: i dont think this works how you think it does. the you probably have to catch another signal while the cluster terminates

Makefile
```diff
@@ -29,6 +29,12 @@ fmt:
 vet:
 	go vet $(BUILDFLAGS) ./pkg/... ./cmd/...
 
+.PHONY: run
+run:
+	# make -C web deps build-kotsadm
+	make kots
```
divolgin, line 35: this will need to build a different target

pkg/apiserver/server.go
```diff
@@ -28,27 +28,47 @@ import (
...
+		log.Println("error waiting for ready")
 		panic(err)
 	}
 	cancel()
 
-	if err := bootstrap(); err != nil {
+	if err := bootstrap(BootstrapParams{
+		AutoCreateClusterToken: params.AutocreateClusterToken,
+	}); err != nil {
+		log.Println("error bootstrapping")
```
divolgin, line 56: how about just panic without the extra log line.
```
		panic(errors.Wrap(err, "error bootstrapping"))
```

pkg/base/helm.go
```diff
@@ -300,6 +300,12 @@ func kustomizeHelmNamespace(baseFiles []BaseFile, renderOptions *RenderOptions)
 			continue // ignore invalid resources
 		}
 
+		// ignore crds
```
emosbaugh, line 303: what does this do? why is this in this pr?

pkg/handlers/spa.go
```diff
@@ -30,21 +33,29 @@ func (h SPAHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-		return
+		rr.URL.Path = "/"
 	} else if err != nil {
 		// if we got an error (that wasn't that the file doesn't exist) stating the
 		// file, return a 500 internal server error and stop
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
 
+	fmt.Printf("\n\n%s\n\n", rr.URL.Path)
```
emosbaugh, line 58: debug code?

pkg/persistence/sqlite_schema.go
```diff
@@ -0,0 +1,786 @@
+package persistence
+
+var tables = []string{
+	`apiVersion: schemas.schemahero.io/v1alpha4
```
emosbaugh, line 4: this seems hard to maintain

pkg/apiserver/server.go
```diff
@@ -28,27 +28,47 @@ import (
...
+	if params.EnableIdentity {
+		err := bootstrapIdentity()
+		if err != nil {
+			log.Println("error bootstrapping identity")
+			panic(err)
+		}
 	}
 
 	if err := generateKotsadmID(); err != nil {
+		log.Println("error generating id")
```
emosbaugh, line 71: debug code

emosbaugh (commented):
i dont see anything that obviously affects the current kots product.

pkg/cluster/apiserver.go
```diff
@@ -0,0 +1,90 @@
...
+		fmt.Sprintf("--cert-dir=%s", filepath.Join(dataDir, "kubernetes")),
+		fmt.Sprintf("--tls-cert-file=%s", apiCertFile),
+		fmt.Sprintf("--tls-private-key-file=%s", apiKeyFile),
+	}
+
+	command := app.NewAPIServerCommand(ctx.Done())
+	command.SetArgs(args)
+
+	go func() {
+		logger.Infof("kubernetes api server exited %v", command.Execute())
```
divolgin, line 84: This needs better handling.  Nothing will attempt to restart the server if it exits.

pkg/cluster/scheduler.go
```diff
@@ -0,0 +1,34 @@
...
+	args := []string{
+		fmt.Sprintf("--config=%s", schedulerConfigFile),
+		"--v=2",
+	}
+
+	command := app.NewSchedulerCommand()
+	command.SetArgs(args)
+
+	go func() {
+		logger.Infof("kubernetes scheduler exited %v", command.Execute())
```
divolgin, line 30: this error needs to be nadled.

cmd/kots/cli/run.go
```diff
@@ -0,0 +1,109 @@
...
+
+	cmd.Flags().String("data-dir", cwd, "directory to store admin console, kubernetes, and application data in")
+
+	return cmd
+}
+
+func startKotsadm(ctx context.Context, dataDir string) error {
+	filestore.ArchivesDir = filepath.Join(dataDir, "archives")
+
+	util.PodNamespace = "default"
```
divolgin, line 92: something is odd about this pattern. these variables are set in two places to two different values, yet they are global.

pkg/persistence/sqlite_schema.go
divolgin: can we use the embed trick here?

pkg/base/helm.go
marccampbell: I had to update the kustomize dependency, which changed the functionality. This was needed for tests to pass. Agreed that this is super suspect, but reverting the Krusty update makes it impossible (for me) to mod in k8s

divolgin (approved):

PR #2021, opened 2021-07-31
Adding basic test for rendering upstream helm hooks to base

jala-dx:


pkg/tests/base/render_test.go
```diff
@@ -72,10 +73,11 @@ func TestRenderUpstream(t *testing.T) {
 
 		test.Upstream.Files = upstreamFilesFromDir(t, filepath.Join(path, "upstream"))
 
-		test.WantBase = baseFromDir(t, filepath.Join(path, "base"))
-
-		// TODO: helm bases
-
+		test.WantBase = baseFromDir(t, filepath.Join(path, "base"), false)
+		helmpath := filepath.Join(path, "basehelm")
+		if _, err := os.Stat(helmpath); !os.IsNotExist(err) {
```
sgalsaleh, line 78:
```suggestion
		if _, err := os.Stat(helmpath); err == nil {
```

pkg/tests/base/render_test.go
```diff
@@ -86,24 +88,41 @@ func TestRenderUpstream(t *testing.T) {
...
+						if len(gotBase.Files) > idx && gotBase.Files[idx].Path == tt.WantBase.Files[idx].Path {
+							t.Log("FILE", tt.WantBase.Files[idx].Path)
+							t.Log(diffString(string(gotBase.Files[idx].Content), string(tt.WantBase.Files[idx].Content)))
+						}
+					}
+					t.FailNow()
+				}
+			}
+			if len(tt.WantHelmBase.Files) > 0 {
+				gotHelmBase := gotHelmBases[0]
```
sgalsaleh, line 110: why only validate one base? why not have `tt.WantHelmBase` be an array and validate all bases?

jala-dx: I need to think a way to add a testcase to handle multiple charts. will add a TODO.

sgalsaleh (approved):

pkg/tests/base/render_test.go
```diff
@@ -86,24 +88,42 @@ func TestRenderUpstream(t *testing.T) {
...
+					for idx := range tt.WantHelmBase.Files {
+						if len(gotBase.Files) > idx && gotBase.Files[idx].Path == tt.WantBase.Files[idx].Path {
+							t.Log("FILE", tt.WantHelmBase.Files[idx].Path)
+							t.Log("FILE", gotHelmBases[0].Files[idx].Path)
+							t.Log(diffString(string(gotHelmBases[0].Files[idx].Content), string(tt.WantHelmBase.Files[idx].Content)))
+						}
 					}
+					t.FailNow()
 				}
-				t.FailNow()
```
divolgin, line 103: This test will pass but should fail if first check fails and the second one passes.

pkg/tests/base/render_test.go
```diff
@@ -86,24 +88,42 @@ func TestRenderUpstream(t *testing.T) {
...
+					t.FailNow()
+				}
+			}
+			if len(tt.WantHelmBase.Files) > 0 {
+				// TODO handle multiple charts
+				gotHelmBase := gotHelmBases[0]
+				if !assert.IsEqual(tt.WantHelmBase.Files, gotHelmBase.Files) {
+					t.Log(diffJSON(gotHelmBases, tt.WantHelmBase))
+				}
+				if !assert.IsEqual(tt.WantBase.Files, gotBase.Files) {
```
sgalsaleh, line 115: should this be `WantHelmBase` and `gotHelmBase`?

pkg/tests/base/render_test.go
divolgin: This should assert that `len(gotHelmBases) == 1`

pkg/tests/base/render_test.go
```diff
@@ -86,24 +88,42 @@ func TestRenderUpstream(t *testing.T) {
...
+					for idx := range tt.WantBase.Files {
+						if len(gotBase.Files) > idx && gotBase.Files[idx].Path == tt.WantBase.Files[idx].Path {
+							t.Log("FILE", tt.WantBase.Files[idx].Path)
+							t.Log(diffString(string(gotBase.Files[idx].Content), string(tt.WantBase.Files[idx].Content)))
+						}
+					}
+					t.FailNow()
+				}
+			}
+			if len(tt.WantHelmBase.Files) > 0 {
```
divolgin, line 109: This check shouldn't be needed.  Number of items in `gotHelmBases` should be checked instead

sgalsaleh (approved):

sgalsaleh (approved):

PR #2028, opened 2021-08-04
Refactor login to support additional backends

marccampbell:
This is a refactor to move the login logic out of the user.go and into a store function. We don't touch the database for in-cluster (normal) installations, so this wasn't obvious to be in the store.
Note: there's now a `make kots_experimental` to make this workflow easier to debug.  I've been running:

```
make kots-experimental  && ./bin/kots run test --shared-password=password
```

---

### What this PR does

For kots run (experimental) support, we don't expect to have a k8s api available to write the shared password and invalid counts to. So I moved this to the sqlite database that we are creating on the host filesystem. There's a table called "kotsadm_params" which is just (key/value) and it felt like a good place to stash this data.

There are a few quirks here that were difficult to place this morning. Open to any and all feedback on this implementation. I feel like the isSQLite() functionality to determine if it's db or in-cluster isn't perfect long term, but ok for now (thoughts?). But there's some weird error returning and initial password setting that could use a better look here too.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ MINIO_VERSION := RELEASE.2021-07-27T02-40-15Z
 POSTGRES_VERSION := 10.17-alpine
 
 BUILDFLAGS = -tags='netgo containers_image_ostree_stub exclude_graphdriver_devicemapper exclude_graphdriver_btrfs containers_image_openpgp' -installsuffix netgo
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -46,7 +51,7 @@ release:
 
 .PHONY: mock
 mock:
-	go get github.com/golang/mock/mockgen@v1.4.4
+	go get github.com/golang/mock/mockgen@v1.5.0
```
marccampbell, line 54: this was updated to match what was already in the go.mod file.

divolgin (approved):

PR #2030, opened 2021-08-04
Fix embedded license install

marccampbell:
When using SQLite, the there is not a time type. These are stored as strings. So I've created two new types to handle the conversion between pg's native type and sqlite's non-type. As we continue integrating, these types will have to be use more.

### DIFF
--- a/pkg/persistence/stringtime.go
+++ b/pkg/persistence/stringtime.go
@@ -1,0 +1,62 @@
+package persistence
+
+import "time"
+
+type StringTime struct {
+	Time time.Time
+}
+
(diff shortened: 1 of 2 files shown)


divolgin (approved):
Pretty sure `NullStringTime` can do both jobs :)

marccampbell:
> Pretty sure `NullStringTime` can do both jobs :)

I can change it to just one, but I felt like then it would be slightly sloppier and less known. Today we have two types: NullTime and time.Time, so is a 1:1 mapping better for us?

PR #2032, opened 2021-08-04
Support non-access key based AWS credentials

dylanlingelbach:
Currently kots determines if the Docker repo being used is an ECR repo. If it is, it expects to be provided with an access key and secret to authenticate with the ECR repo. However in some cases, kots will be run with an assumed role or on an EC2 instance with an instance role that doesn't have an access key. The go SDK will use a default credential provider chain if no credentials are specified that will use the instance role or other credentials.

This change allows kots to be installed without an access key when using ECR

I didn't see a contributor guide so happy to take any feedback

Related to feature 31678, although this isn't likely sufficient to truly solve that case

### DIFF
--- a/cmd/kots/cli/admin-console-push-images.go
+++ b/cmd/kots/cli/admin-console-push-images.go
@@ -131,7 +131,9 @@ func getECRLogin(endpoint string, keyID string, accessKey string) (string, strin
 
 func getECRService(accessKeyID, secretAccessKey, zone string) *ecr.ECR {
 	awsConfig := &aws.Config{Region: aws.String(zone)}
-	awsConfig.Credentials = credentials.NewStaticCredentials(accessKeyID, secretAccessKey, "")
+	if (accessKeyID != "" && secretAccessKey != "") {
+		awsConfig.Credentials = credentials.NewStaticCredentials(accessKeyID, secretAccessKey, "")
+	}
 	return ecr.New(session.New(awsConfig))
 }
 
--- a/pkg/docker/registry/ecr.go
+++ b/pkg/docker/registry/ecr.go
@@ -65,7 +65,9 @@ func GetECRBasicAuthToken(ecrEndpoint, username, password string) (string, error
 
 func getECRService(accessKeyID, secretAccessKey, zone string) *ecr.ECR {
 	awsConfig := &aws.Config{Region: aws.String(zone)}
-	awsConfig.Credentials = credentials.NewStaticCredentials(accessKeyID, secretAccessKey, "")
+	if (accessKeyID != "" && secretAccessKey != "") {
+		awsConfig.Credentials = credentials.NewStaticCredentials(accessKeyID, secretAccessKey, "")
+	}
 	return ecr.New(session.New(awsConfig))
 }
 
--- a/pkg/s3/s3.go
+++ b/pkg/s3/s3.go
@@ -52,7 +52,7 @@ func GetConfig() *aws.Config {
 	secretAccessKey := os.Getenv("S3_SECRET_ACCESS_KEY")
 
 	var creds *credentials.Credentials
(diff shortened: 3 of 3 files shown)


divolgin (approved):

pkg/s3/s3.go
```diff
@@ -48,8 +48,16 @@ func GetConfig() *aws.Config {
 		region = "us-east-1"
 	}
 
+	accessKeyID := os.Getenv("S3_ACCESS_KEY_ID")
+	secretAccessKey := os.Getenv("S3_SECRET_ACCESS_KEY")
+
+	creds := credentials.NewStaticCredentials(accessKeyID, secretAccessKey, "")
```
sgalsaleh, line 54: why not check if the accessKey and the secretKey are set first? (like the other two changes)
for example
```go
var creds *credentials.Credentials
if accessKeyID != "" || secretAccessKey != "" {
	creds = credentials.NewStaticCredentials(accessKeyID, secretAccessKey, "")
}
```

pkg/docker/registry/ecr.go
```diff
@@ -65,7 +65,9 @@ func GetECRBasicAuthToken(ecrEndpoint, username, password string) (string, error
 
 func getECRService(accessKeyID, secretAccessKey, zone string) *ecr.ECR {
 	awsConfig := &aws.Config{Region: aws.String(zone)}
-	awsConfig.Credentials = credentials.NewStaticCredentials(accessKeyID, secretAccessKey, "")
+	if (accessKeyID != "" && secretAccessKey != "") {
```
sgalsaleh, line 68: to be consistent with the current pattern:
```suggestion
	if accessKeyID != "" && secretAccessKey != "" {
```

cmd/kots/cli/admin-console-push-images.go
```diff
@@ -131,7 +131,9 @@ func getECRLogin(endpoint string, keyID string, accessKey string) (string, strin
 
 func getECRService(accessKeyID, secretAccessKey, zone string) *ecr.ECR {
 	awsConfig := &aws.Config{Region: aws.String(zone)}
-	awsConfig.Credentials = credentials.NewStaticCredentials(accessKeyID, secretAccessKey, "")
+	if (accessKeyID != "" && secretAccessKey != "") {
```
sgalsaleh, line 134: ditto
```suggestion
	if accessKeyID != "" && secretAccessKey != "" {
```

pkg/s3/s3.go
```diff
@@ -48,8 +48,16 @@ func GetConfig() *aws.Config {
...
+	if (accessKeyID == "" || secretAccessKey == "") {
+		creds = nil
+	}
+
 	s3Config := &aws.Config{
-		Credentials:      credentials.NewStaticCredentials(os.Getenv("S3_ACCESS_KEY_ID"), os.Getenv("S3_SECRET_ACCESS_KEY"), ""),
+		Credentials:      creds,
```
sgalsaleh, line 60: FYI, these credentials are only used to communicate with the s3 object store and are not related to pulling/writing images from/to the registry

cmd/kots/cli/admin-console-push-images.go
dylanlingelbach: Thanks - I've been working in JS too long 😄

pkg/s3/s3.go
dylanlingelbach: Thanks - in what scenario is the s3 store used? It wasn't immediately apparent to me if this was something that would impact our scenario or not

sgalsaleh: the s3 store is currently used for storing the application archives and support bundles. i still think that the changes in the PR are safe and are good to have in general.

sgalsaleh (approved):

sgalsaleh (approved):

PR #2033, opened 2021-08-04
fix: always show embedded workflow when kurl registry is detected

morningvera:
[CH36190](https://app.clubhouse.io/replicated/story/36190/existing-cluster-airgap-bundle-upload-page-sometimes-shows-up-for-embedded-cluster-installs)

### DIFF
--- a/pkg/handlers/app.go
+++ b/pkg/handlers/app.go
@@ -13,6 +13,7 @@ import (
 	"github.com/replicatedhq/kots/pkg/gitops"
 	"github.com/replicatedhq/kots/pkg/logger"
 	"github.com/replicatedhq/kots/pkg/rbac"
+	"github.com/replicatedhq/kots/pkg/registry"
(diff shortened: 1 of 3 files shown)


pkg/handlers/app.go
```diff
@@ -53,12 +54,21 @@ func (h *Handler) GetPendingApp(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	// Carefully now, peek at registry credentials to see if we need to prompt for them
+	hasKurlRegistry, err := registry.HasKurlRegistry()
+	if err != nil {
+		logger.Error(errors.Wrapf(err, "failed to check registry status for pending app %s", papp.Slug))
+		w.WriteHeader(http.StatusInternalServerError)
```
morningvera, line 61: Not 100% sure what the response should be here. A 300 is returned when it fails in UploadNewLicense.

sgalsaleh (approved):

PR #2036, opened 2021-08-04
Wait for the K8s api server to start

marccampbell:
This will wait for the api server to start by polling it's readyz endpoint.
This is necessary to start controlling the startup and verifying that the cluster is ready before the app can be deployed.

Now, when starting the app with `kots run` the output is:

```
  • Starting cluster ✓  
    The cluster is running. Press ctrl+c to terminate
```

### DIFF
--- a/pkg/cluster/scheduler.go
+++ b/pkg/cluster/scheduler.go
@@ -3,13 +3,14 @@ package cluster
 import (
 	"context"
 	"fmt"
+	"sync"
 
 	"github.com/pkg/errors"
 	"github.com/replicatedhq/kots/pkg/logger"
 	"k8s.io/kubernetes/cmd/kube-scheduler/app"
 )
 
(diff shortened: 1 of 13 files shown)


pkg/cluster/scheduler.go
```diff
@@ -31,5 +33,8 @@ func runScheduler(ctx context.Context, dataDir string) error {
 		logger.Infof("kubernetes scheduler exited %v", command.Execute())
 	}()
 
+	time.Sleep(time.Second * 20)
+	wg.Done()
```
divolgin, line 37: I think this is meant to be in the go func above.  not sure using wait groups in synchronous code makes much sense

marccampbell: oops, that sleep was debugging code, removed.
agreed on wait groups. should this be async and wgs, or all sync?  any preference?

divolgin (approved):

PR #2045, opened 2021-08-05
Add CNI (Antrea) to embedded cluster

marccampbell:
This PR embeds a small amount of YAML for Antrea and then applies it after the cluster is running. This uses the dynamic client so that we don't need kubectl installed to make it work.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -26,16 +26,26 @@ func Start(ctx context.Context, slug string, dataDir string) error {
 	}
 
 	// start the scheduler on port 11251 (this is a non-standard port)
-	// wg.Add(1)
-	// if err := runScheduler(ctx, &wg, dataDir); err != nil {
-	// 	return errors.Wrap(err, "start scheduler")
-	// }
+	if err := runScheduler(ctx, dataDir); err != nil {
+		return errors.Wrap(err, "start scheduler")
+	}
 
-	// // start the controller manager on port 11252 (non standard)
-	// wg.Add(1)
-	// if err := runController(ctx, &wg, dataDir); err != nil {
+	// start the controller manager on port 11252 (non standard)
+	// TODO the controller should start
+	// if err := runController(ctx, dataDir); err != nil {
 	// 	return errors.Wrap(err, "start controller")
 	// }
+
+	// because these are all synchoronous, the api is ready and we
+	// can install our addons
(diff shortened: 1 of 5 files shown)


pkg/cluster/controller.go
```diff
@@ -33,7 +36,8 @@ func runController(ctx context.Context, wg *sync.WaitGroup, dataDir string) erro
 	}
 
 	args := []string{
-		"--bind-address=0.0.0.0:11252",
+		"--bind-address=0.0.0.0",
```
marccampbell, line 39: This was failing as the port number is not parseable. The correct flags need to be determined.

pkg/cluster/controller.go
```diff
@@ -33,7 +36,8 @@ func runController(ctx context.Context, wg *sync.WaitGroup, dataDir string) erro
...
+		"--secure-port=11252",
```
marccampbell, line 40: This port is not respected, but the docs show this flag. I'm not sure what the right one is.

pkg/cluster/scheduler.go
```diff
@@ -32,7 +34,31 @@ func runScheduler(ctx context.Context, wg *sync.WaitGroup, dataDir string) error
 		logger.Infof("kubernetes scheduler exited %v", command.Execute())
 	}()
 
-	wg.Done()
+	// watch the readyz endpoint to know when the api server has started
+	stopWaitingAfter := time.Now().Add(time.Minute)
+	for {
+		url := "http://localhost:11251/healthz"
```
marccampbell, line 40: Basic healthz check on scheduler, note that this is not https

pkg/cluster/cluster.go
```diff
@@ -26,16 +26,26 @@ func Start(ctx context.Context, slug string, dataDir string) error {
...
+	// if err := runScheduler(ctx, dataDir); err != nil {
 	// 	return errors.Wrap(err, "start scheduler")
 	// }
 
-	// // start the controller manager on port 11252 (non standard)
-	// wg.Add(1)
-	// if err := runController(ctx, &wg, dataDir); err != nil {
+	// start the controller manager on port 11252 (non standard)
+	// TODO the controller should start
+	// if err := runController(ctx, dataDir); err != nil {
```
marccampbell, line 35: Disabling controllermanager for now since this is not starting. This needs to be figured out.

divolgin (approved):

PR #2050, opened 2021-08-05
update warning message about failing to pull app metadata

sgalsaleh:
update warning message about failing to pull app metadata to include that it might also cause the admin console to not run with minimal rbac privileges if the application requires that.

### DIFF
--- a/cmd/kots/cli/install.go
+++ b/cmd/kots/cli/install.go
@@ -142,7 +142,7 @@ func InstallCmd() *cobra.Command {
 			} else if !v.GetBool("airgap") {
 				applicationMetadata, err = pull.PullApplicationMetadata(upstream)
 				if err != nil {
-					log.Info("Unable to pull application metadata. This can be ignored, but custom branding will not be available in the Admin Console until a license is installed.")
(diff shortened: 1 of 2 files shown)


divolgin (approved):

cmd/kots/cli/install.go
```diff
@@ -142,7 +142,7 @@ func InstallCmd() *cobra.Command {
 			} else if !v.GetBool("airgap") {
 				applicationMetadata, err = pull.PullApplicationMetadata(upstream)
 				if err != nil {
-					log.Info("Unable to pull application metadata. This can be ignored, but custom branding will not be available in the Admin Console until a license is installed.")
+					log.Info("Unable to pull application metadata. This can be ignored, but custom branding will not be available in the Admin Console until a license is installed. This could also cause the Admin Console to not run with minimal role-based-access-control (RBAC) privileges in case the application requires that.")
```
JustinN-Replicated, line 145: Would suggest changing to:
..." This may also cause the Admin Console to run without minimal role-based-access-control (RBAC) privileges in case the application requires that."
When we say 'in case the application requires that.' it sounds like we're implying that their application could require it - is that correct? If so we could change to "which may be required by your application."

JustinN-Replicated (commented):
Added a comment and suggestion(s) for change.

JustinN-Replicated (approved):
Looks good

JustinN-Replicated (approved):

PR #2051, opened 2021-08-05
detect the registry namespace from the registry endpoint flag if provided

sgalsaleh:


cmd/kots/cli/upstream-upgrade.go
```diff
@@ -66,6 +66,15 @@ func UpstreamUpgradeCmd() *cobra.Command {
 				registryUsername := v.GetString("registry-username")
 				registryPassword := v.GetString("registry-password")
 
+				if registryNamespace == "" {
```
sgalsaleh, line 69: inspired from https://github.com/replicatedhq/kots/blob/03faa3388614dfef56ec97f203fd1ed5243e0eb3/cmd/kots/cli/install.go#L605

divolgin (approved):

PR #2052, opened 2021-08-06
Send k8s logs to a file

divolgin:


marccampbell:
I think we should log to the `dataDir/kubernetes/log` or something like that instead of an ephemeral temp. Might be useful for debugging?

marccampbell (approved):

PR #2066, opened 2021-08-09
feat: add json output for version cmd

morningvera:

### DIFF
--- a/cmd/kots/cli/version.go
+++ b/cmd/kots/cli/version.go
@@ -39,7 +39,7 @@ func VersionCmd() *cobra.Command {
 			}
 
 			if output != "json" && output != "" {
-				return errors.Errorf("output format %s not supported", output)
+				return errors.Errorf("output format %s not supported (allowed formats are: json)", output)
 			} else if output == "json" {
 				// marshal JSON
 				outputJSON, err := json.Marshal(versionOutput)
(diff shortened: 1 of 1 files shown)


cmd/kots/cli/version.go
```diff
@@ -1,29 +1,67 @@
...
+				Version: buildversion.Version(),
+			}
 			if err == nil && !isLatest {
-				fmt.Printf("\nVersion %s is available for kots. To install updates, run\n  $ curl https://kots.io/install | bash\n", latestVer)
+				versionOutput.LatestVersion = latestVer
+				versionOutput.InstallLatest = "curl https://kots.io/install | bash"
+			}
+
+			if output != "json" && output != "" {
+				return errors.Errorf("output format %s not supported", output)
```
morningvera, line 42: Should we error here or just default to the text output?

morningvera:
this story also requested a `status` field that indicates success or failure but i'm not sure it makes sense for this one. any thoughts?

cmd/kots/cli/version.go
morningvera: Thinking we should match `kubectl` here so going to leave this as is

sgalsaleh (approved):

divolgin (approved):

PR #2069, opened 2021-08-10
feat: add json output to upload and download cli cmds

morningvera:

### DIFF
--- a/cmd/kots/cli/download.go
+++ b/cmd/kots/cli/download.go
@@ -1,6 +1,8 @@
 package cli
 
 import (
+	"encoding/json"
+	"fmt"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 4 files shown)


cmd/kots/cli/download.go
```diff
@@ -34,21 +43,46 @@ func DownloadCmd() *cobra.Command {
...
+				downloadOutput.UploadCommand = fmt.Sprintf("kubectl kots upload --namespace %s --slug %s %s", v.GetString("namespace"), appSlug, downloadPath)
 			}
 
 			log := logger.NewCLILogger()
+			if output == "json" {
+				outputJSON, err := json.Marshal(downloadOutput)
+				if err != nil {
+					return errors.Wrap(err, "error marshaling JSON")
+				}
+				log.Info(string(outputJSON))
```
sgalsaleh, line 77: i would recommend adding a `return` after this?

morningvera: yeah that would remove the need to silence the logger again so i agree with that, thanks!

PR #2080, opened 2021-08-12
ch:31845: Fix support bundle command for airgap installations

jala-dx:


web/src/components/troubleshoot/GenerateSupportBundle.jsx
```diff
@@ -314,6 +314,37 @@ class GenerateSupportBundle extends React.Component {
                 </div>
               }
               {showRunCommand ?
+                watch.isAirgap ?
```
sgalsaleh, line 317: we still calling this `watch`? :) wow

GraysonNull: haha yeah...still in the code here and there 🤡

GraysonNull:
Screen Shot 2021-08-12 at 5 05 37 PM

divolgin (changes requested):
This should use the spec generated by kotsadm and provide instructions for installing support-bundle binary.

PR #2081, opened 2021-08-12
feat: add json output to backup/restore and upstream cmds

morningvera:

### DIFF
--- a/pkg/handlers/update.go
+++ b/pkg/handlers/update.go
@@ -22,8 +22,9 @@ type AppUpdateCheckRequest struct {
 }
 
 type AppUpdateCheckResponse struct {
-	AvailableUpdates   int64 `json:"availableUpdates"`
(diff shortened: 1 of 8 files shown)


pkg/handlers/update.go
```diff
@@ -22,8 +22,9 @@ type AppUpdateCheckRequest struct {
 }
 
 type AppUpdateCheckResponse struct {
-	AvailableUpdates   int64 `json:"availableUpdates"`
-	CurrentAppSequence int64 `json:"currentAppSequence"`
+	AvailableUpdates   int64  `json:"availableUpdates"`
+	CurrentAppSequence int64  `json:"currentAppSequence"`
+	LatestVersion      string `json:"latestVersion"`
```
morningvera, line 27: Ended up having to add this to the API because getting the latest version is non-trivial if there are updates available. Originally tried to grab the version history after starting the updates but it was a race condition since new releases take a couple seconds to appear in the history.

PR #2083, opened 2021-08-12
feat: add json output for backup and restore cmds

morningvera:

### DIFF
--- a/cmd/kots/cli/backup.go
+++ b/cmd/kots/cli/backup.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"encoding/json"
+	"fmt"
 
 	"github.com/pkg/errors"
 	"github.com/replicatedhq/kots/pkg/logger"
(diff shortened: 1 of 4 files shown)


cmd/kots/cli/backup.go
```diff
@@ -23,20 +26,37 @@ func BackupCmd() *cobra.Command {
...
+			}
+
 			options := snapshot.CreateInstanceBackupOptions{
 				Namespace: namespace,
 				Wait:      v.GetBool("wait"),
+				Silent:    output != "",
 			}
-			if err := snapshot.CreateInstanceBackup(cmd.Context(), options); err != nil {
+			backupRes, err := snapshot.CreateInstanceBackup(cmd.Context(), options)
+			if err != nil && output == "" {
```
sgalsaleh, line 40: what about if there's an error and an output format is specified? i don't see that handled

sgalsaleh: should be like this one :) https://github.com/replicatedhq/kots/pull/2083/files#diff-5e5da03ddf8e728ae099106e99cc9a821f479eeaaedee622042cf98baae91626R56

morningvera: ahhh i had already fixed that but must've lost it while splitting. good catch!

morningvera: fixed

sgalsaleh (approved):

PR #2110, opened 2021-08-24
deploy app using kots run (first pass)

sgalsaleh:
outstanding issues:

1- containerd doesn’t run because it can’t find runc in PATH
2- antrea falls back to using the in-cluster config. we need to point to the right kubeconfig file to use instead.

### DIFF
--- a/pkg/cluster/antrea.go
+++ b/pkg/cluster/antrea.go
@@ -52,23 +52,23 @@ func installCNI(kubeconfigPath string) error {
 		obj := &unstructured.Unstructured{}
 		_, gvk, err := decUnstructured.Decode(buf, nil, obj)
 		if err != nil {
-			return err
+			return errors.Wrap(err, "decode doc")
 		}
 
 		dc, err := discovery.NewDiscoveryClientForConfig(config)
 		if err != nil {
-			return err
+			return errors.Wrap(err, "new discovery client for config")
 		}
 		mapper := restmapper.NewDeferredDiscoveryRESTMapper(memory.NewMemCacheClient(dc))
 
 		dyn, err := dynamic.NewForConfig(config)
 		if err != nil {
-			return err
+			return errors.Wrap(err, "new dynamic config")
 		}
 
 		mapping, err := mapper.RESTMapping(gvk.GroupKind(), gvk.Version)
 		if err != nil {
-			return err
+			return errors.Wrap(err, "rest mapping")
 		}
 
 		var dr dynamic.ResourceInterface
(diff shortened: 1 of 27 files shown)


pkg/cluster/controller.go
```diff
@@ -50,7 +50,7 @@ func runController(ctx context.Context, dataDir string) error {
 		"--use-service-account-credentials=true",
 		"--v=2",
 		fmt.Sprintf("--flex-volume-plugin-dir=%s", dataDir),
-		"--controllers=-service,-route,-cloud-node-lifecycle", // This is equivalent to using ExternalLoops in the ControllerLoopMode type, which excludes cloud specific controlers
+		"--controllers=*,-service,-route,-cloud-node-lifecycle", // This is equivalent to using ExternalLoops in the ControllerLoopMode type, which excludes cloud specific controlers
```
sgalsaleh, line 53: this was disabling all controllers

pkg/persistence/sqlite_schema.go
```diff
@@ -20,7 +20,7 @@ spec:
         constraints:
           notNull: true
       - name: updated_at
-        type: integer
+        type: timestamp
```
sgalsaleh, line 23: these were failing to be converted to time.Time in go. sqlite had a workaround for this by using the `timestamp` column type which will handle it internall.

pkg/cluster/antrea.go
```diff
@@ -82,19 +82,15 @@ func installCNI(kubeconfigPath string) error {
...
 		_, err = dr.Patch(context.Background(), obj.GetName(), types.ApplyPatchType, data, metav1.PatchOptions{
 			FieldManager: "kots",
 		})
-
 		if err != nil {
-			return err
+			return errors.Wrap(err, "patch")
 		}
-
-		return nil
```
sgalsaleh, line 96: this resulted in only having the first yaml document to be applied

pkg/k8sutil/kotsadm.go
```diff
@@ -160,6 +161,11 @@ func UpdateKotsadmIDConfigMap(kotsadmID string) error {
 }
 
 func FindKotsadm(clientset *kubernetes.Clientset, namespace string) (string, error) {
+	if persistence.IsSQlite() {
```
sgalsaleh, line 164: TODO: need to have a separate function for detecting the kots run workflow so that this isn't confusing (applies to other occurrences as well)

pkg/persistence/sqlite_schema.go
marccampbell: @divolgin ...  thoughts on this?

marccampbell (approved):

pkg/persistence/sqlite_schema.go
sgalsaleh: 1st reference: https://github.com/mattn/go-sqlite3/issues/748#issuecomment-538643131

sgalsaleh: 2nd reference: https://github.com/mattn/go-sqlite3/issues/190#issuecomment-186704817

divolgin (approved):

PR #2112, opened 2021-08-24
Move code into pkg dir

marccampbell:


divolgin (approved):

pkg/cluster/init.go
```diff
@@ -10,18 +10,49 @@ import (
...
 	"time"
 
+	"github.com/mholt/archiver"
 	"github.com/pkg/errors"
 	"github.com/replicatedhq/kots/pkg/logger"
 	"go.uber.org/zap"
 )
 
+// Init will download all binaries for the cluster
+func Init(ctx context.Context, dataDir string) error {
```
sgalsaleh, line 29: there's already a `clusterInit` function in the same package, this naming can be confusing, any reason not to have this be part of the other?

marccampbell: maybe I'll rename `init` to `clientInit`?

sgalsaleh (approved):

pkg/store/store.go
```diff
@@ -11,7 +10,6 @@ var (
 )
 
 var _ Store = (*kotsstore.KOTSStore)(nil)
-var _ Store = (*ocistore.OCIStore)(nil)
```
divolgin, line 14: Why not OCIStore?

divolgin (approved):

PR #2128, opened 2021-09-04
Add test for non-helm based render

divolgin:


pkg/tests/base/cases/configcontext/upstream/test-chart-2.yaml
```diff
@@ -0,0 +1,20 @@
...
+kind: HelmChart
+metadata:
+  name: test-chart-2
+spec:
+  # exclude: 'repl{{ ConfigOptionEquals "helm" "no" }}'
+  # chart identifies a matching chart from a .tgz
+  chart:
+    name: test-chart-2
+    chartVersion: 0.1.1
+  useHelmInstall: false
```
jala-dx, line 11: I may be off.. you have a chart with helmhooks but useHelmInstall set to false and added a different namespace ? that is when this regression happens?

jala-dx (approved):

pkg/tests/base/cases/configcontext/upstream/test-chart-2.yaml
divolgin: Removed Helm annotations.

emosbaugh (approved):

PR #2135, opened 2021-09-10
Read analysis from the support bundle

sgalsaleh:


pkg/supportbundle/supportbundle.go
```diff
@@ -237,9 +237,77 @@ func CreateSupportBundleDependencies(appID string, sequence int64, opts types.Tr
...
+
+		relPath, err := filepath.Rel(bundleDir, path)
+		if err != nil {
+			return errors.Wrap(err, "failed to get relative path")
+		}
+
+		// trim the subdirectory from the path too (if any)
+		trimmedRelPath := SupportBundleNameRegex.ReplaceAllString(relPath, "")
+		trimmedRelPath = strings.TrimPrefix(trimmedRelPath, string(os.PathSeparator))
+		if trimmedRelPath == "" {
```
sgalsaleh, line 274: gonna remove this

pkg/supportbundle/supportbundle.go
```diff
@@ -237,9 +237,77 @@ func CreateSupportBundleDependencies(appID string, sequence int64, opts types.Tr
...
+			return nil
+		}
+
+		if trimmedRelPath == "analysis.json" {
```
divolgin, line 278: We should just hardcode absolute path to avoid returning some random file with the same name since there is no schema validation.

divolgin (approved):

divolgin (approved):

divolgin (approved):

PR #2141, opened 2021-09-15
fix: kubectl apply for large files

DanStough:


sgalsaleh:
`kubectl create` is imperative, not declarative. which is not recommended for production environments. so i think we should only resort to it in case `kubectl apply` fails because of a large file.

emosbaugh:
Additionally I think the most minimal change would be best. I'd prefer if the apply failed only then did it fork from the current functionality.

pkg/operator/applier/kubectl.go
```diff
@@ -126,10 +134,79 @@ commonAnnotations:
...
+	if err := ioutil.WriteFile(renderedManifestPath, renderedManifest, 0644); err != nil {
+		return nil, nil, errors.Wrapf(err, "failed to write %s", renderedManifestPath)
+	}
+
+	// Try to create and checkout output to see if it already exists
+	cmd := c.kubectlCreateCommand(renderedManifestPath, targetNamespace, dryRun, wait)
 
 	stdout, stderr, err := Run(cmd)
-	return stdout, stderr, errors.Wrap(err, "failed to run kubectl apply")
+	if !strings.Contains(string(stderr), "already exists") {
```
divolgin, line 151: this falls through to patch if err is nil

pkg/operator/applier/kubectl.go
```diff
@@ -132,6 +136,138 @@ commonAnnotations:
...
+		if err != nil {
+			return combinedStdout, combinedStderr, errors.Wrap(err, "failed to run kubectl create")
+		}
+	}
+
+	return combinedStdout, combinedStderr, nil
+}
+
+func (c *Kubectl) createOrPatchSingleDoc(name string, targetNamespace string, slug string, yamlDoc []byte, dryRun bool, wait bool, annotateSlug bool) ([]byte, []byte, error) {
+	tmp, err := ioutil.TempDir("", "")
```
sgalsaleh, line 168: ca we give this a name?
```suggestion
	tmp, err := ioutil.TempDir("", "kots")
```

pkg/operator/applier/kubectl.go
```diff
@@ -132,6 +136,138 @@ commonAnnotations:
...
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "failed to create temp directory")
+	}
+	defer os.Remove(tmp)
```
sgalsaleh, line 172: RemoveAll is for directories
```suggestion
	defer os.RemoveAll(tmp)
```

pkg/operator/applier/kubectl.go
```diff
@@ -132,6 +136,138 @@ commonAnnotations:
...
+
+	renderedManifestPath := filepath.Join(tmp, "patch.yaml")
+	if err := ioutil.WriteFile(renderedManifestPath, renderedManifest, 0644); err != nil {
+		return nil, nil, errors.Wrapf(err, "failed to write %s", renderedManifestPath)
+	}
+	// Try to create and checkout output to see if it already exists
+	cmd := c.kubectlCreateCommand(renderedManifestPath, targetNamespace, dryRun, wait)
+
+	stdout, stderr, err := Run(cmd)
+	if !strings.Contains(string(stderr), "already exists") || err == nil {
```
sgalsaleh, line 210: i wonder if we could check the returned err for something like
```
kuberneteserrors.IsNotFound(err)
```
using the `k8s.io/apimachinery/pkg/api/errors` package

pkg/operator/applier/kubectl.go
```diff
@@ -132,6 +136,138 @@ commonAnnotations:
...
+// ApplyCreateOrPatch attempts to run a `kubectl apply` on the yaml document. If it fails
+func (c *Kubectl) ApplyCreateOrPatch(targetNamespace string, slug string, yamlDoc []byte, dryRun bool, wait bool, annotateSlug bool) ([]byte, []byte, error) {
+
+	stdout, stderr, err := c.Apply(targetNamespace, slug, yamlDoc, dryRun, wait, annotateSlug)
+	if !strings.Contains(string(stderr), "metadata.annotations: Too long") || err == nil {
+		return stdout, stderr, errors.Wrap(err, "failed attempted kubectl apply")
+	}
+	log.Println("Failed to apply document: metadata too long. Trying to 'create' or 'patch' instead")
+
+	docs, err := splitter.SplitYAML(yamlDoc)
```
sgalsaleh, line 148: kubectl create accepts multi-yaml docs, why do we need to split them here?

sgalsaleh: the current kubectl apply workflow doesn't split them, so i don't think this should

pkg/operator/applier/kubectl.go
sgalsaleh, line 168: can we give this a name?
```suggestion
	tmp, err := ioutil.TempDir("", "kots")
```

pkg/operator/applier/kubectl.go
DanStough: That would be better but since we are exec-ing the kubectl binary, I wasn't sure how to get this error type back. Is there any easy way to parse this back into `kuberneteserrrors`?

pkg/operator/applier/kubectl.go
DanStough: Yes, but patch does not. I didn't want to get into a scenario where create only partially fails (like maybe someone added new manifest to an existing doc), then we have to figure out which documents we need to patch anyway. I also opted to inline them here to minimize the io of writing out this files and the temp dir.

sgalsaleh: ah okay

sgalsaleh (approved):

pkg/operator/applier/kubectl.go
DanStough: hmm, maybe `tmp, err := ioutil.TempDir("", "kots-apply-*")` to be less ambiguous?

pkg/operator/applier/kubectl.go
```diff
@@ -132,6 +136,138 @@ commonAnnotations:
...
+	}
+
+	var combinedStdout, combinedStderr []byte
+	for name, b := range docs {
+		stdout, stderr, err = c.createOrPatchSingleDoc(name, targetNamespace, slug, b, dryRun, wait, annotateSlug)
+		combinedStdout = append(combinedStdout, stdout...)
+		combinedStderr = append(combinedStderr, stderr...)
+
+		if err != nil {
+			return combinedStdout, combinedStderr, errors.Wrap(err, "failed to run kubectl create")
```
emosbaugh, line 160: can you add name to this message

```
errors.Wrapf(err, "failed to run kubectl create for document %s", name)
```

pkg/operator/applier/kubectl.go
```diff
@@ -132,6 +136,138 @@ commonAnnotations:
...
+		return nil, nil, errors.Wrapf(err, "failed to write %s", renderedManifestPath)
+	}
+	// Try to create and checkout output to see if it already exists
+	cmd := c.kubectlCreateCommand(renderedManifestPath, targetNamespace, dryRun, wait)
+
+	stdout, stderr, err := Run(cmd)
+	if !strings.Contains(string(stderr), "already exists") || err == nil {
+		return stdout, stderr, errors.Wrap(err, "failed attempted kubectl create")
+	}
+	log.Printf("Failed to create document %s: already exists. Trying to 'patch' instead", name)
```
emosbaugh, line 213: whats the standard for logging? should we use this?
```
```

pkg/operator/applier/kubectl.go
```diff
@@ -132,6 +136,138 @@ commonAnnotations:
 	return stdout, stderr, errors.Wrap(err, "failed to run kubectl apply")
 }
 
+// ApplyCreateOrPatch attempts to run a `kubectl apply` on the yaml document. If it fails
+func (c *Kubectl) ApplyCreateOrPatch(targetNamespace string, slug string, yamlDoc []byte, dryRun bool, wait bool, annotateSlug bool) ([]byte, []byte, error) {
+
+	stdout, stderr, err := c.Apply(targetNamespace, slug, yamlDoc, dryRun, wait, annotateSlug)
+	if !strings.Contains(string(stderr), "metadata.annotations: Too long") || err == nil {
```
emosbaugh, line 143: this would be more readable if you checked error == nil separate from handling the special error case

pkg/operator/applier/kubectl.go
```diff
@@ -132,6 +136,138 @@ commonAnnotations:
...
+
+	renderedManifestPath := filepath.Join(tmp, "patch.yaml")
+	if err := ioutil.WriteFile(renderedManifestPath, renderedManifest, 0644); err != nil {
+		return nil, nil, errors.Wrapf(err, "failed to write %s", renderedManifestPath)
+	}
+	// Try to create and checkout output to see if it already exists
+	cmd := c.kubectlCreateCommand(renderedManifestPath, targetNamespace, dryRun, wait)
+
+	stdout, stderr, err := Run(cmd)
+	if !strings.Contains(string(stderr), "already exists") || err == nil {
```
emosbaugh, line 210: this would be more readable if you checked error == nil separate from handling the special error case

pkg/operator/applier/kubectl.go
DanStough: I was going off the logging already in `deploy.go`

DanStough: I updated deploy.go and kubectl.go to use the kots package.

emosbaugh (approved):

PR #2146, opened 2021-09-16
Replaced hard coded image tags in kotsadm

murphybytes:
Added go format to generated files, and replaced hard coded tag references in kotsadm


pkg/image/image.go
```diff
@@ -10,6 +10,19 @@ import (
 	kustomizetypes "sigs.k8s.io/kustomize/api/types"
 )
 
+// GetTag extracts the image tag from an image reference
+func GetTag(imageRef string) (string, error) {
+	parts := strings.Split(imageRef, ":")
+	if len(parts) < 2 {
+		return "", fmt.Errorf("malformed image reference %q could not find tag", imageRef)
```
divolgin, line 17: I would argue that errors in this function are reversed.  This line is actually "empty tag", from image like `"nginx"`, which is a valid image name.  And the error below is actually "malformed image name", from image like `"nginx:"`

I get that this function is used just for a small set of well known image names, but since this is an exported function in a generic package, how about using an actual image ref parser, like `alltransports.ParseImageName`?

divolgin (approved):

PR #2148, opened 2021-09-16
chore: enable slack notifications for nightly builds

morningvera:
https://app.shortcut.com/replicated/story/37194/ci-failures-should-notify-the-team-in-slack

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -55,6 +55,15 @@ jobs:
           name: web
           path: ./web/dist
 
+      - name: Notify Slack
+        if: always()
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test.yaml
```diff
@@ -154,6 +154,27 @@ jobs:
           push: true
 
 
+  notify_slack_test:
```
emosbaugh, line 157: i dont think that this is needed in the build-test workflow. only nightly

morningvera: i'm aware, see my PR description above.

emosbaugh (approved):

PR #2176, opened 2021-09-18
get minio, posgres_alpine and dex image tags from env file for airgap build

jala-dx:

Fixes #

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,9 @@
 include Makefile.build
+include .image.env
 CURRENT_USER := $(shell id -u -n)
-MINIO_TAG ?= RELEASE.2021-09-15T04-54-25Z
-POSTGRES_ALPINE_TAG ?= 10.17-alpine
-DEX_TAG ?= v2.28.1
+MINIO_TAG := $(shell echo $(MINIO_TAG))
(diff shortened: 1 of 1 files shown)


emosbaugh:
This doesn't feel right. What's the reason for this?

jala-dx:
> This doesn't feel right. What's the reason for this?

@emosbaugh the image tags have to come from .image.env file instead of hardcoding in the Makefile. Is my assumption incorrect here?

emosbaugh:
This changed a lot since I made my first comment

jala-dx:
> This changed a lot since I made my first comment

Two things changed since you commented on the PR. (Sorry I dint check my email so dint know you commented and force pushed it)

emosbaugh:
Can you fill out the pull request template?

emosbaugh (commented):
I think you should discuss this with @murphybytes . There is a mix of make and bash for setting and retrieving these vars. I think we can clean this up a bit and be consistent

murphybytes:
This won't work.  I've submitted a different that fixes the problem PR https://github.com/replicatedhq/kots/pull/2197

jala-dx:
> This won't work. I've submitted a different that fixes the problem PR #2197

nice . Exporting the .image.env before Makefile gets it is a much better option.

PR #2191, opened 2021-09-21
chore: fix job and took in slack notifications

morningvera:


.github/workflows/release.yaml
```diff
@@ -763,16 +763,18 @@ jobs:
         if: failure()
         uses: 8398a7/action-slack@v3
         with:
-          status: ${{ job.status }}
+          status: custom
```
morningvera, line 766: changing this to custom to condense the build status output. similar to this: https://github.com/8398a7/action-slack#custom-formats-of-your-choice

.github/workflows/release.yaml
```diff
@@ -763,16 +763,18 @@ jobs:
...
           custom_payload: |
             {
               attachments: [{
+                color: '${{ job.status }}' === 'success' ? 'good' : '${{ job.status }}' === 'failure' ? 'danger' : 'warning',
                 text: `${process.env.AS_WORKFLOW}\n${process.env.AS_JOB} of ${process.env.AS_REPO}@${process.env.AS_REF}\n${{ job.status }} in ${process.env.AS_TOOK}`,
               }]
             }
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.KOTS_BUILD_STATUS_SLACK_WEBHOOK_URL }}
+          MATRIX_CONTEXT: ${{ toJson(matrix) }}
```
morningvera, line 777: this is necessary when jobs include the matrix field. however, it's apparently buggy and the docs and GH issues are very vague on when and why it will fail. proposing we try it out and delete if it's still not working.

PR #2192, opened 2021-09-21
fix: check for existing k8s version analyzer

sgalsaleh:
#### What type of PR is this?

kind/bug

fixes an issue where support bundle generation would hang if the analyzer spec includes a k8s version analyzer

no

no

### DIFF
--- a/pkg/supportbundle/analyze.go
+++ b/pkg/supportbundle/analyze.go
@@ -10,7 +10,9 @@ import (
 func InjectDefaultAnalyzers(analyzers []*troubleshootv1beta2.Analyze) []*troubleshootv1beta2.Analyze {
 	analyzers = append(analyzers, getAPIReplicaAnalyzer())
 	analyzers = append(analyzers, getNoGvisorAnalyzer())
-	analyzers = append(analyzers, getIfMissingKubernetesVersionAnalyzer(analyzers))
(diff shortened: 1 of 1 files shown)


pkg/supportbundle/analyze.go
```diff
@@ -10,7 +10,9 @@ import (
 func InjectDefaultAnalyzers(analyzers []*troubleshootv1beta2.Analyze) []*troubleshootv1beta2.Analyze {
 	analyzers = append(analyzers, getAPIReplicaAnalyzer())
 	analyzers = append(analyzers, getNoGvisorAnalyzer())
-	analyzers = append(analyzers, getIfMissingKubernetesVersionAnalyzer(analyzers))
+	if a := getIfMissingKubernetesVersionAnalyzer(analyzers); a != nil {
+		analyzers = append(analyzers, a)
```
emosbaugh, line 14: do you think we should also check for nil here?

https://github.com/replicatedhq/troubleshoot/blob/cb5ddf752f30b68a6855476aa0ec3a889b1552cd/pkg/analyze/analyzer.go#L78

emosbaugh (approved):

pkg/supportbundle/analyze.go
sgalsaleh: i'll pr

emosbaugh:
[sc-37797]

PR #2200, opened 2021-09-22
Jeffgolden/sc 37499/pullsecrets need lowest helmhook weight

jeffreygolden:
#### What type of PR is this?

kind/bug

ImagePullSecrets are generated automatically by default.  If these are created for a helm chart, it's possible the helm resources have Helm Hooks defined that would be processed before the pull secret is generated.  This causes the pull to fail.

This PR attaches a helm hook and weight to image pull secrets.  This ensures the secret is created before any other helm resources with a weight above -99.  A vendor setting a resource weight below -99 will still run into this problem.

```release-note
Fixed a Helm hooks bug where pre-install and pre-upgrade hooks were processed before the image pull secret was deployed.
```

NONE

### DIFF
--- a/pkg/docker/registry/registry.go
+++ b/pkg/docker/registry/registry.go
@@ -108,6 +108,11 @@ func PullSecretForRegistries(registries []string, username, password string, kub
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      secretName,
 			Namespace: kuberneteNamespace,
+			// try to ensure this is created first if using a helm install
+			Annotations: map[string]string{
+				"helm.sh/hook":        "pre-install,pre-upgrade",
+				"helm.sh/hook-weight": "-99",
+			},
 		},
 		Type: corev1.SecretTypeDockerConfigJson,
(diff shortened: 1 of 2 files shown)


morningvera (approved):
lgtm

DanStough (approved):

pkg/docker/registry/registry.go
```diff
@@ -108,6 +108,11 @@ func PullSecretForRegistries(registries []string, username, password string, kub
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      secretName,
 			Namespace: kuberneteNamespace,
+			// try to ensure this is created first if using a helm install
+			Annotations: map[string]string{
+				"helm.sh/hook":        "pre-install,pre-upgrade",
+				"helm.sh/hook-weight": "-99",
```
DanStough, line 114: Why not -99999 to be sure?

jeffreygolden: Compromised for -9999, should take care of just about every use case.

PR #2209, opened 2021-09-23
skip charts directories when finding images

jeffreygolden:
#### What type of PR is this?

kind/bug

KOTS searches for private images in yaml files and writes Kustomization to use the Replicated registry.  When searching for private images, KOTS will search inside of subchart directories and report those images up to the parent chart's Kustomize file.  This resulted in kustomize errors complaining that the image's manifest could not be found, because it belongs to the subchart.

This PR skips searching for images in any directory named "charts", which is used to store subcharts.  The subcharts are still searched independently for private images and each subchart gets it's own Kustomize file to re-write the private image name.

```release-note
Fixed which private images Kustomize targets inside of Helm charts when sub-charts are present.
```

NONE


pkg/image/builder.go
```diff
@@ -101,6 +101,13 @@ func GetPrivateImages(upstreamDir string, checkedImages map[string]ImageInfo, al
 			if err != nil {
 				return err
 			}
+
+			// Skip charts folders containing subcharts.  The kustomization for subcharts will be defined
+			// inside of those charts, and should not be referenced at the parent chart level.
+			if info.IsDir() && info.Name() == "charts" {
+				return filepath.SkipDir
+			}
+
```
jala-dx, line 110: can you please add a testcase for subcharts.

divolgin (approved):

jala-dx (approved):

pkg/image/builder.go
jeffreygolden: I will add one to the next PR, we ran into a separate bug with sub-sub-charts.  If I add it to tests now the tests will fail.

pkg/image/builder.go
```diff
@@ -101,6 +101,13 @@ func GetPrivateImages(upstreamDir string, checkedImages map[string]ImageInfo, al
 			if err != nil {
 				return err
 			}
+
+			// Skip charts folders containing subcharts.  The kustomization for subcharts will be defined
+			// inside of those charts, and should not be referenced at the parent chart level.
+			if info.IsDir() && info.Name() == "charts" {
```
jala-dx, line 107: Can you check if this is needed only if useHelmInstall flag is set?

if info.IsDir() && info.Name() == "charts"  && (useHelmInstall[info.Name()] == true) {

jeffreygolden: The check below this one is `if info.IsDir() && (useHelmInstall[info.Name()] == true)` , but it wasn't excluding subchart folders correctly.  Adding the useHelmInstall check would replicate the existing behavior.

erquhart:
@divolgin will take a look here.

jeffreygolden:
@divolgin  pinging for a re-review since this changed significantly since your first approval

divolgin (approved):

PR #2216, opened 2021-09-25
fix: append additional files correctly for sub charts

morningvera:
#### What type of PR is this?

kind/bug

This fixes an issue that was causing Chart.yaml and Chart.lock to be ignored for Helm sub-charts. With this fix, any level of sub-chart nesting should work now.

It still handles the special case where `base.Path == ""` for the top-level Chart.yaml and Chart.lock. For sub-charts, it does a slightly different check to find the correct additional files. Previously those additional files were being ignored which led to a Helm error when deploying.

Fixes [SH37929](https://app.shortcut.com/replicated/story/37929/helm-sub-sub-charts-are-not-copied-into-base-correctly)

```release-note
Fixed a Helm issue where Chart.yaml and Chart.lock files were being ignored and lost for sub-charts of sub-charts.
```

NONE

### DIFF
--- a/pkg/base/helm.go
+++ b/pkg/base/helm.go
@@ -235,13 +235,26 @@ func removeCommonPrefix(baseFiles []BaseFile) []BaseFile {
 
 func helmChartBaseAppendAdditionalFiles(base Base, u *upstreamtypes.Upstream) Base {
 	for _, upstreamFile := range u.Files {
-		if upstreamFile.Path == path.Join(base.Path, "Chart.yaml") {
+		// need to check if base path is an empty string here to catch just the top level additional files
+		// otherwise, we need to check if the upstream path contains the base path to correctly add files for sub-charts
+		if base.Path == "" && upstreamFile.Path == path.Join(base.Path, "Chart.yaml") {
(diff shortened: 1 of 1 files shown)


morningvera:
sub-chart-fix

pkg/base/helm.go
```diff
@@ -235,13 +235,26 @@ func removeCommonPrefix(baseFiles []BaseFile) []BaseFile {
 
 func helmChartBaseAppendAdditionalFiles(base Base, u *upstreamtypes.Upstream) Base {
 	for _, upstreamFile := range u.Files {
-		if upstreamFile.Path == path.Join(base.Path, "Chart.yaml") {
+		// need to check if base path is an empty string here to catch just the top level additional files
+		// otherwise, we need to check if the upstream path contains the base path to correctly add files for sub-charts
+		if base.Path == "" && upstreamFile.Path == path.Join(base.Path, "Chart.yaml") {
```
sgalsaleh, line 240: nit pick: i would do it the "go" way, something like this at the beginning of the loop
```
if base.Path != "" {
  continue
}
```

sgalsaleh (approved):

divolgin (approved):

PR #2217, opened 2021-09-29
Build Dex image from dockerfile

emosbaugh:
#### What type of PR is this?

kind/chore

Just some cleanup

Fixes #NONE

```release-note
NONE
```

NONE

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ include Makefile.build
 CURRENT_USER := $(shell id -u -n)
 MINIO_TAG ?= RELEASE.2021-09-15T04-54-25Z
 POSTGRES_ALPINE_TAG ?= 10.18-alpine
-DEX_TAG ?= v2.28.1
+DEX_TAG ?= v2.30.0
 LVP_VERSION := v0.1.0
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -2,7 +2,7 @@ include Makefile.build
 CURRENT_USER := $(shell id -u -n)
 MINIO_TAG ?= RELEASE.2021-09-15T04-54-25Z
 POSTGRES_ALPINE_TAG ?= 10.18-alpine
-DEX_TAG ?= v2.28.1
+DEX_TAG ?= v2.30.0
```
DanStough, line 5: Isn't this tag suppose to come from the environment file now? This is just the fallaback, correct?

emosbaugh: yes, fallback

jala-dx: @DanStough the tag is exported at the root, this is a fallback like @emosbaugh says.

emosbaugh: i do think it would be good to change that to `latest` to avoid confusion. not donna do it in this pr

DanStough (approved):

PR #2221, opened 2021-09-30
Move support bundle yaml back to root

jeffreygolden:
support-bundle looks for default spec in the root of the project.  This was recently moved to /pkg/support-bundle, which is causing problems with support bundle generation.

This impacts customer support bundles.

```release-note

```

NONE

### DIFF
--- a/pkg/supportbundle/defaultspec/spec.yaml
+++ b/pkg/supportbundle/defaultspec/spec.yaml
@@ -99,7 +99,6 @@ spec:
         name: kotsadm-replicated-registry # NOTE: this will not live under the kots/ directory like other collectors
         includeValue: false
         key: .dockerconfigjson
-        name: kotsadm-replicated-registry
     - logs:
         collectorName: rook-ceph-agent
         selector:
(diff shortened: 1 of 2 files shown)


marccampbell (approved):

pkg/supportbundle/defaultspec/spec.yaml
```diff
@@ -99,7 +99,6 @@ spec:
         name: kotsadm-replicated-registry # NOTE: this will not live under the kots/ directory like other collectors
         includeValue: false
         key: .dockerconfigjson
-        name: kotsadm-replicated-registry
```
laverya, line 102: why is this being removed?

laverya: oh because it's a duplicate

laverya (approved):

pkg/supportbundle/defaultspec/spec.yaml
jeffreygolden: it's a duplicate key from line 99

PR #2244, opened 2021-10-13
Unblock one chart at a time for midstream

jeffreygolden:
#### What type of PR is this?
kind/bug

When using the Beta Helm functions, charts were not getting a proper Kustomize file or PullSecrets file.  The charts were being skipped by kustomize.  This fixed a previous bug where subcharts were being targeted by parent level Kustomize files, causing a Kustomize error on deploy, but introduced this new regression where the charts weren't being kustomized at all.

This PR splits the difference.  It prevents charts from being kustomized at the Common Midstream level.  It also prevents subcharts from being kustomized at the parent chart level.  Now, all Kustomize files target each chart discretely, at the appropriate directory level.

Added new comments in the code to prevent future confusion around the UseHelmInstall map.  Please let me know if these comments can be clarified further.

```release-note
Helm Charts should get the proper Kustomization for images and pull secrets at each chart level when using the beta Helm Install feature.
```

NONE


jeffreygolden:
I've tested this with three licenses:
1) no Helm charts (received the correct kustomization.yaml file in midstream)
2) single chart, no subcharts (received the correct kustomization.yaml file in midstream/chart/chartname)
3) parent chart with many nested subcharts (received kustomization.yaml files in each chart with only that chart's images listed)

PR #2252, opened 2021-10-15
Update bin kustomize version

emosbaugh:
#### What type of PR is this?

kind/chore

Update kustomize version for latest features and bug fixes.

Additionally this new code will detect kustomize versions in the image from path KOTS_KUSTOMIZE_BIN_DIR rather than hardcoding versions in code.

This also fixes a hypothetical issue if there were multiple kustomize versions where the operator could use an incorrect kustomize version when applying yaml.

Fixes NONE

```release-note
Update Kustomize version from 3.5.4 to 3.10.0
```

https://github.com/replicatedhq/kots.io/pull/590


emosbaugh:
It makes me nervous that we use 3 but the k8s project jumped from 2 to 4

https://github.com/kubernetes/kubernetes/blob/v1.20.11/go.mod#L535

https://github.com/kubernetes/kubernetes/blob/v1.21.0/go.mod#L536

PR #2292, opened 2021-11-03
Adding helm data to downstream report

jeffreygolden:
#### What type of PR is this?
kind/enhancement

Product and Engineering wants to track the number of Native Helm installs as the feature goes to GA.  This will inform priority for future Helm updates.

As KOTS collects downstream data, it will now search for HelmChart KOTS kinds.  If found, the HelmChart's install method (Replicated Helm vs Native Helm) will be recorded.  This data is sent back to Replicated.

```release-note
KOTS telemetry will now include whether installed helm charts leverage our Native Helm feature.
```

NONE

### DIFF
--- a/pkg/kotsutil/kots.go
+++ b/pkg/kotsutil/kots.go
@@ -44,7 +44,7 @@ func init() {
 type KotsKinds struct {
 	KotsApplication kotsv1beta1.Application
 	Application     *applicationv1beta1.Application
-	HelmChart       *kotsv1beta1.HelmChart
+	HelmCharts      []*kotsv1beta1.HelmChart
 
 	Collector     *troubleshootv1beta2.Collector
 	Preflight     *troubleshootv1beta2.Preflight
@@ -419,7 +419,7 @@ func LoadKotsKindsFromPath(fromDir string) (*KotsKinds, error) {
 				case "kots.io/v1beta1, Kind=Installation":
 					kotsKinds.Installation = *decoded.(*kotsv1beta1.Installation)
 				case "kots.io/v1beta1, Kind=HelmChart":
-					kotsKinds.HelmChart = decoded.(*kotsv1beta1.HelmChart)
+					kotsKinds.HelmCharts = append(kotsKinds.HelmCharts, decoded.(*kotsv1beta1.HelmChart))
 				case "troubleshoot.sh/v1beta2, Kind=Collector":
 					kotsKinds.Collector = decoded.(*troubleshootv1beta2.Collector)
 				case "troubleshoot.sh/v1beta2, Kind=Analyzer":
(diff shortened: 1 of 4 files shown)


pkg/kotsutil/kots.go
```diff
@@ -417,6 +418,8 @@ func LoadKotsKindsFromPath(fromDir string) (*KotsKinds, error) {
 					kotsKinds.IdentityConfig = decoded.(*kotsv1beta1.IdentityConfig)
 				case "kots.io/v1beta1, Kind=Installation":
 					kotsKinds.Installation = *decoded.(*kotsv1beta1.Installation)
+				case "kots.io/v1beta1, Kind=HelmChart":
```
morningvera, line 421: i wonder why this wasn't here before now. was it because we don't think of helm charts as kots kinds? i think it's fine to add but might wanna double check with someone else

pkg/reporting/app.go
```diff
@@ -144,6 +144,14 @@ func getDownstreamInfo(appID string) (*types.DownstreamInfo, error) {
 		di.Cursor = deployedKotsKinds.Installation.Spec.UpdateCursor
 		di.ChannelID = deployedKotsKinds.Installation.Spec.ChannelID
 		di.ChannelName = deployedKotsKinds.Installation.Spec.ChannelName
+
+		if len(deployedKotsKinds.HelmCharts) > 0 {
+			for _, chart := range deployedKotsKinds.HelmCharts {
+				if chart.Spec.UseHelmInstall {
+					di.IsNativeHelm = true
```
morningvera, line 151: this just made me realize we allow releases with some charts that use replicated helm and others that use native helm simultaneously 🤯

morningvera (approved):

laverya:
Maybe we should report both the number of native helm charts AND the number of external charts?

laverya (approved):

morningvera (approved):

pkg/reporting/util.go
```diff
@@ -20,6 +20,8 @@ func InjectReportingInfoHeaders(req *http.Request, reportingInfo *types.Reportin
 	req.Header.Set("X-Replicated-AppStatus", reportingInfo.AppStatus)
 	req.Header.Set("X-Replicated-ClusterID", reportingInfo.ClusterID)
 	req.Header.Set("X-Replicated-InstanceID", reportingInfo.InstanceID)
+	req.Header.Set("X-Replicated-ReplHelmInstalls", fmt.Sprint(reportingInfo.Downstream.ReplHelmInstalls))
+	req.Header.Set("X-Replicated-NativeHelmInstalls", fmt.Sprint(reportingInfo.Downstream.NativeHelmInstalls))
```
laverya, line 24: Can we do sprintf with %d here to be explicit?

laverya (approved):

PR #2293, opened 2021-11-03
create helmv3 bookkeeping secrets

laverya:

Fixes #

```release-note

```


pkg/base/helm_v3.go
```diff
@@ -76,6 +82,29 @@ func renderHelmV3(chartName string, chartPath string, vals map[string]interface{
...
+		helmReleaseSecretObj, err := newSecretsObject(rel)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to generate helm secret")
+		}
+
+		renderedSecret, err := k8syaml.Marshal(helmReleaseSecretObj)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to generate helm secret")
+		}
+		baseFiles = append(baseFiles, BaseFile{
```
jeffreygolden, line 102: This is going to show up in the diff for everybody upgrading to the latest version of KOTS, yes?  We should probably include a specific release note for this.

laverya: I should also remove the timestamp (or make it a constant) so that there aren't diffs every update

jeffreygolden (approved):

PR #2299, opened 2021-11-09
Sort releases by semver if available

divolgin:

Fixes #

```release-note

```

### DIFF
--- a/pkg/api/downstream/types/types.go
+++ b/pkg/api/downstream/types/types.go
@@ -3,6 +3,7 @@ package types
 import (
 	"time"
 
+	"github.com/blang/semver"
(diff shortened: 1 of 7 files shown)


pkg/api/downstream/types/types.go
```diff
@@ -18,6 +19,7 @@ type Downstream struct {
 
 type DownstreamVersion struct {
 	VersionLabel             string                             `json:"versionLabel"`
+	Semver                   *semver.Version                    `json:"semver,omitempty"`
```
sgalsaleh, line 22: why not just parse the version label during comparsion? having both Semver and VersionLabel feels off

sgalsaleh (approved):

sgalsaleh (approved):

PR #2311, opened 2021-11-11
Semver auto deploy schedules

sgalsaleh:


pkg/autodeployer/autodeployer.go
```diff
@@ -0,0 +1,262 @@
...
+func Start() error {
+	logger.Debug("starting auto deployer")
+
+	appsList, err := store.GetStore().ListInstalledApps()
+	if err != nil {
+		return errors.Wrap(err, "failed to list installed apps")
+	}
+
+	for _, a := range appsList {
+		if a.IsAirgap {
```
divolgin, line 34: `Configure` also checks this

sgalsaleh: which is fine, can't be too careful :). configure is called by other packages as well

pkg/autodeployer/autodeployer.go
```diff
@@ -0,0 +1,262 @@
...
+	if err != nil {
+		return -1
+	}
+
+	pendingVersions := PendingVersions{}
+
+	for _, v := range versions {
+		switch opts.SemverAutoDeploy {
+		case apptypes.SemverAutoDeployPatch:
+			s, err := semver.ParseTolerant(v.VersionLabel)
```
divolgin, line 219: there should be `v.Semver` already parsed.

pkg/autodeployer/autodeployer.go
sgalsaleh: i can get rid of it

divolgin (approved):

divolgin (approved):

PR #2322, opened 2021-11-12
use the correct cursor and base version when checking for updates

sgalsaleh:


pkg/updatechecker/updatechecker.go
```diff
@@ -207,7 +207,12 @@ func CheckForUpdates(opts CheckForUpdatesOpts) (int64, error) {
 		}
 	}()
 
-	err = store.GetStore().GetAppVersionArchive(a.ID, a.CurrentSequence, archiveDir)
+	baseSequence, err := store.GetStore().GetSequenceForLatestUpdateCursor(a.ID)
```
divolgin, line 210: How about passing `latestLicense.Spec.ChannelID` here for consistency, so that the caller knows that this is tied to a channel.

divolgin (approved):

pkg/updatechecker/updatechecker.go
```diff
@@ -307,51 +283,142 @@ func CheckForUpdates(opts CheckForUpdatesOpts) (int64, error) {
...
+	appVersions, err := store.GetStore().GetAppVersions(appID, clusterID)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to get app versions for app %s", appID)
+	}
+	if len(appVersions.AllVersions) == 0 {
+		return "", errors.Errorf("no app versions found for app %s in downstream %s", appID, clusterID)
+	}
+
+	mockVersion := &downstreamtypes.DownstreamVersion{
+		Sequence: -1, // to id the mocked version and be able to retrieve it later
```
divolgin, line 382: If there is no semver, this will be placed at the bottom of the list.  How about `math.MaxInt64`?

pkg/updatechecker/updatechecker.go
```diff
@@ -307,51 +283,142 @@ func CheckForUpdates(opts CheckForUpdatesOpts) (int64, error) {
...
+	}
+
+	appVersions.AllVersions = append(appVersions.AllVersions, mockVersion)
+	downstreamtypes.SortDownstreamVersions(appVersions)
+
+	var baseVersion *downstreamtypes.DownstreamVersion
+	for i, v := range appVersions.AllVersions {
+		if v.Sequence == -1 {
+			// this is our mocked version, base it off of the previous version in the sorted list (if exists).
+			if i < len(appVersions.AllVersions)-1 {
```
divolgin, line 397: If `i` is the last element in the list, how about using the `i-1` config?  I'm not sure what the best behavior here would be.

divolgin (approved):

divolgin (approved):

PR #2333, opened 2021-11-17
feat: ability to deploy a specific version label during update checks

sgalsaleh:


pkg/updatechecker/updatechecker.go
```diff
@@ -322,189 +271,209 @@ type DownloadUpdateOptions struct {
...
+		if err := deployVersionLabel(opts, clusterID, opts.DeployVersionLabel); err != nil {
+			return errors.Wrapf(err, "failed to deploy version label %s", opts.DeployVersionLabel)
+		}
+		return nil
 	}
 
-	// preflights reporting
-	go func() {
-		err = reporting.ReportAppInfo(opts.AppID, sequence, opts.SkipPreflights, opts.IsCLI)
+	if opts.IsAutomatic {
```
divolgin, line 306: this is always true I think

divolgin (approved):

pkg/updatechecker/updatechecker.go
sgalsaleh: not when you check for updates manually

sgalsaleh: https://github.com/replicatedhq/kots/pull/2333/files#diff-43223b6dda7aafc2d70ba222243d37325e88c939eac603a79f33c1042e1e9272R48

PR #2334, opened 2021-11-17
UI features are based on semver sorting order

divolgin:

Fixes #

```release-note

```

### DIFF
--- a/pkg/handlers/app.go
+++ b/pkg/handlers/app.go
@@ -167,18 +167,24 @@ func responseAppFromApp(a *apptypes.App) (*types.ResponseApp, error) {
 		return nil, errors.Wrap(err, "failed to get license")
 	}
 
-	isIdentityServiceSupportedForVersion, err := store.GetStore().IsIdentityServiceSupportedForVersion(a.ID, a.CurrentSequence)
+	downstreamVersions, err := store.GetStore().FindAppVersions(a.ID)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to check if identity service is supported for version %d", a.CurrentSequence)
+		return nil, errors.Wrap(err, "failed to find app versions")
+	}
+	latestVersion := downstreamVersions.AllVersions[0]
+
+	isIdentityServiceSupportedForVersion, err := store.GetStore().IsIdentityServiceSupportedForVersion(a.ID, latestVersion.ParentSequence)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to check if identity service is supported for version %d", latestVersion.ParentSequence)
 	}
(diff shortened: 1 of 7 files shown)


pkg/handlers/registry.go
```diff
@@ -169,24 +169,9 @@ func (h *Handler) UpdateAppRegistry(w http.ResponseWriter, r *http.Request) {
 			skipImagePush = true
 		}
 
-		downstreams, err := store.GetStore().ListDownstreamsForApp(foundApp.ID)
+		appVersions, err := store.GetStore().FindAppVersions(foundApp.ID)
```
sgalsaleh, line 172: we should stick to a consistent name for this variable. here it's called `downstreamVersions`: https://github.com/replicatedhq/kots/pull/2334/files#diff-a7b1ffb8e0d64772914b79d513e230636436763524feeee3e17d06925b2e15edR22

sgalsaleh: and i think the name should be `appVersions`

pkg/handlers/app.go
```diff
@@ -167,18 +167,24 @@ func responseAppFromApp(a *apptypes.App) (*types.ResponseApp, error) {
 		return nil, errors.Wrap(err, "failed to get license")
 	}
 
-	isIdentityServiceSupportedForVersion, err := store.GetStore().IsIdentityServiceSupportedForVersion(a.ID, a.CurrentSequence)
+	appVersions, err := store.GetStore().FindAppVersions(a.ID)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to check if identity service is supported for version %d", a.CurrentSequence)
+		return nil, errors.Wrap(err, "failed to get app versions")
+	}
+	topVersionSequence := appVersions.AllVersions[0].ParentSequence
```
sgalsaleh, line 174: `topVersionSequence` can be confusing. how about:
```
latestVersion := appVersions.AllVersions[0]
```
and then use `latestVersion.ParentSequence`?

pkg/handlers/app.go
```diff
@@ -167,18 +167,24 @@ func responseAppFromApp(a *apptypes.App) (*types.ResponseApp, error) {
 		return nil, errors.Wrap(err, "failed to get license")
 	}
 
-	isIdentityServiceSupportedForVersion, err := store.GetStore().IsIdentityServiceSupportedForVersion(a.ID, a.CurrentSequence)
+	appVersions, err := store.GetStore().FindAppVersions(a.ID)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to check if identity service is supported for version %d", a.CurrentSequence)
+		return nil, errors.Wrap(err, "failed to get app versions")
```
sgalsaleh, line 172:
```suggestion
		return nil, errors.Wrap(err, "failed to find app versions")
```

sgalsaleh (approved):

PR #2342, opened 2021-11-18
fix: use the right base sequence for semantic versions

sgalsaleh:


pkg/store/ocistore/version_store.go
```diff
@@ -46,44 +46,6 @@ func (s *OCIStore) appVersionConfigMapNameForApp(appID string) (string, error) {
 	return fmt.Sprintf("%s%s", AppVersionConfigmapPrefix, a.Slug), nil
 }
 
-func (s *OCIStore) getLatestAppVersion(appID string) (*versiontypes.AppVersion, error) {
```
divolgin, line 49: Is this replaced by an unimplemented method? :)

divolgin (approved):

PR #2346, opened 2021-11-18
handling renamed (removed prefix) chart directory

jeffreygolden:
#### What type of PR is this?

kind/bug

If helm charts have a common prefix such as a company name, the name gets stripped from the resulting chart directory name in /base.  This no longer matches the original chart name.

When writing kustomize files for these subcharts on the Native Helm feature, the subchart directories are not properly excluded when searching for private images.  This causes kustomization that targets subchart resources, resulting in Kustomize errors on deploy.

This PR uses the chart's Base name, after modification, and not the original chart name.

```release-note
Fixed a Native Helm bug that caused Kustomize "resource not found" errors when supplying multiple Helm charts with the same name prefix.
```

NONE

### DIFF
--- a/pkg/pull/pull.go
+++ b/pkg/pull/pull.go
@@ -424,11 +424,19 @@ func Pull(upstreamURI string, pullOptions PullOptions) (string, error) {
 	// for writing Common Midstream, every chart and subchart is in this map as Helm Midstreams will be processed later in the code
 	commonWriteMidstreamOptions.UseHelmInstall = map[string]bool{}
 	for _, v := range newHelmCharts {
-		commonWriteMidstreamOptions.UseHelmInstall[v.Spec.Chart.Name] = v.Spec.UseHelmInstall
+		chartBaseName := v.Spec.Chart.Name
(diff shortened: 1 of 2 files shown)


pkg/pull/pull.go
```diff
@@ -424,11 +424,19 @@ func Pull(upstreamURI string, pullOptions PullOptions) (string, error) {
 	// for writing Common Midstream, every chart and subchart is in this map as Helm Midstreams will be processed later in the code
 	commonWriteMidstreamOptions.UseHelmInstall = map[string]bool{}
 	for _, v := range newHelmCharts {
-		commonWriteMidstreamOptions.UseHelmInstall[v.Spec.Chart.Name] = v.Spec.UseHelmInstall
+		chartBaseName := v.Spec.Chart.Name
+		// the helmBase may have a chart name prefix removed - we must find the base name instead of the original chart name
```
laverya, line 428: I'd rather do the reverse of this, and just not remove the prefix in question
but this will work, I suppose
just feels like more complexity

laverya (approved):

PR #2358, opened 2021-11-20
feat(cli): generate-manifests now supports minimal rbac

DanStough:
#### What type of PR is this?
kind/enhancement

Adds flags to the `kots generate-manifests` command to let vendors/customers enable minimal rbac without manually editing the files.

Fixes SC-39602

```release-note
Added flags `--minimal-rbac` and `--additional-namespaces` to the `kots generate-manifests` command to enable minimal-rbac deployments through YAML.
```

Yes, TBD.

### DIFF
--- a/pkg/upstream/admin-console.go
+++ b/pkg/upstream/admin-console.go
@@ -17,6 +17,7 @@ import (
 )
 
 type UpstreamSettings struct {
+	Namespace              string
 	SharedPassword         string
 	SharedPasswordBcrypt   string
 	S3AccessKey            string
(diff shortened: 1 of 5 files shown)


pkg/upstream/admin-console.go
```diff
@@ -140,7 +149,7 @@ func generateNewAdminConsoleFiles(settings *UpstreamSettings) ([]types.UpstreamF
 	upstreamFiles := []types.UpstreamFile{}
 
 	deployOptions := kotsadmtypes.DeployOptions{
-		Namespace:              "default",
+		Namespace:              settings.Namespace,
```
jala-dx, line 152: Please double check if this value is ok to change.

jala-dx (approved):

pkg/upstream/admin-console.go
DanStough: Good callout, looks like the public version of this function is used elsewhere. Adding a check in the public method.

PR #2363, opened 2021-11-23
Migrate snapshots from objectstore to PVC

jala-dx:
#### What type of PR is this?

This PR chooses PVC over objectstore based on disableS3 flag.

The Pre-requisite for this PR: https://github.com/replicatedhq/kURL/pull/2366

How to test this:

1. Create a KURL cluster by setting disableS3 flag in the kotsadm addon (curl https://kurl.sh/e0fe387 | sudo bash)
2. Build the kotsadm image in codeserver and copy to ttl.sh (./hack/build-ttl.sh from kots repo)
3. In the kurl cluster, edit the statefulset to point the image to image: ttl.sh/jalaja/kotsadm:12h
4. velero plugin add ttl.sh/jalaja/local-volume-provider:12h
5. Go to kotsadm ui, choose snapshot settings, choose **Internal Storage**  from the Pull down, Choose it, Save it.
6. Take snapshots.

```

spec:
  contour:
    version: 1.18.0
  docker:
    version: 20.10.5
  ekco:
    version: 0.12.0
  kotsadm:
    disableS3: true
    version: 1.57.0
  kubernetes:
    version: 1.19.15
  longhorn:
    version: 1.1.2
  minio:
    version: 2020-01-25T02-50-51Z
  prometheus:
    version: 0.49.0-17.1.3
  registry:
    version: 2.7.1
  velero:
    version: 1.6.2
  weave:
    version: 2.6.5

```

### DIFF
--- a/cmd/kots/cli/velero.go
+++ b/cmd/kots/cli/velero.go
@@ -161,7 +161,7 @@ func VeleroConfigureInternalCmd() *cobra.Command {
 	}
 
 	cmd.Flags().Bool("skip-validation", false, "skip the validation of the internal store endpoint/bucket")
-	cmd.Flags().Bool("with-minio", true, "when set, kots will deploy minio for NFS snapshot locations")
+	cmd.Flags().Bool("with-minio", true, "when set, kots will deploy minio for INTERNAL snapshot locations")
 
 	return cmd
 }
--- a/cmd/kots/cli/velero.go
+++ b/cmd/kots/cli/velero.go
@@ -146,6 +146,7 @@ func VeleroConfigureInternalCmd() *cobra.Command {
 				RegistryOptions:   &registryOptions,
 				SkipValidation:    v.GetBool("skip-validation"),
 				ValidateUsingAPod: true,
+				IsMinioDisabled:   !v.GetBool("with-minio"),
 			}
 			_, err = snapshot.ConfigureStore(cmd.Context(), configureStoreOptions)
 			if err != nil {
@@ -160,6 +161,7 @@ func VeleroConfigureInternalCmd() *cobra.Command {
 	}
 
 	cmd.Flags().Bool("skip-validation", false, "skip the validation of the internal store endpoint/bucket")
+	cmd.Flags().Bool("with-minio", true, "when set, kots will deploy minio for NFS snapshot locations")
 
 	return cmd
 }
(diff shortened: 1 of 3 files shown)


web/src/components/snapshots/SnapshotStorageDestination.jsx
```diff
@@ -930,6 +930,12 @@ class SnapshotStorageDestination extends Component {
                 value: "hostpath",
                 label: "Host Path",
             });
+            if (snapshotSettings.isKurl == false) {
+              availableDestinations.push({
+                value: "internal",
+                label: "Internal Storage (Default)",
+              });
+            }
```
DanStough, line 938: 👍

jala-dx:
> We can't delete the previous behavior yet - there will be folks on old kurl install with newer versions of kots that don't migrate away from object storage. We'll need to detect what's going on in the cluster here.

modified the code to use disableS3 flag in kurl spec which basically creates a kostsadm-confg configmap and has the flag.

cmd/kots/cli/velero.go
```diff
@@ -160,6 +161,7 @@ func VeleroConfigureInternalCmd() *cobra.Command {
 	}
 
 	cmd.Flags().Bool("skip-validation", false, "skip the validation of the internal store endpoint/bucket")
+	cmd.Flags().Bool("with-minio", true, "when set, kots will deploy minio for NFS snapshot locations")
```
DanStough, line 164: Help message is copy-pasta from NFS. Should mention something about internal snapshots.

pkg/snapshot/store.go
```diff
@@ -267,7 +267,28 @@ func ConfigureStore(ctx context.Context, options ConfigureStoreOptions) (*types.
...
+		}
+		store.AWS = nil
+		store.Google = nil
+		store.Azure = nil
+		store.Other = nil
+		store.FileSystem = nil
 
+		store.Provider = "replicated.com/pvc"
+		store.Bucket = "velero-internal-snapshots"
+		store.Path = "/var/velero-local-volume-provider/velero-internal-snapshots/restic"
```
DanStough, line 291: `spec.config.storageSize` is a required field, so I think we either need to check the current volume size with client-go, or we could default to 50Gb since the plugin won't change the volume size if it already exists. The volume is ALWAYS created as part of the kurl spec so the impact is minor.

DanStough (changes requested):
Minor stuff but I think it should be addressed before merging.

DanStough (approved):

PR #2364, opened 2021-11-23
prevent kots downgrades

sgalsaleh:
#### What type of PR is this?

kind/bug

prevents kots downgrades since it can ruin a customer's installation

```release-note
Fixed an issue where downgrading to an older kots version was not failing early.
```

### DIFF
--- a/cmd/kots/cli/admin-console-upgrade.go
+++ b/cmd/kots/cli/admin-console-upgrade.go
@@ -1,21 +1,26 @@
 package cli
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/blang/semver"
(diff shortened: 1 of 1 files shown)


cmd/kots/cli/admin-console-upgrade.go
```diff
@@ -151,3 +160,43 @@ func AdminConsoleUpgradeCmd() *cobra.Command {
...
+			containers = d.Spec.Template.Spec.Containers
+		}
+	}
+
+	if len(containers) == 0 {
+		return "", false
+	}
+
+	parts := strings.Split(containers[0].Image, ":")
+	if len(parts) != 2 {
```
divolgin, line 186: There can be 3 parts if host name is followed by port number:  `parts[len(parts)-1]`

divolgin (approved):

PR #2386, opened 2021-12-09
centralize encryption key management

laverya:
#### What type of PR is this?

kind/chore

This PR centralizes encryption key management. Instead of one key per instance plus one per app, only the instance key will be created on new instances and used to encrypt things going forward. Existing keys will be preserved to decrypt things as required.

Fixes #

I'm going to have to do a LOT of testing here (snapshot restores, etc)

```release-note
NONE
```

NONE


laverya:
This PR replaces https://github.com/replicatedhq/kots/pull/2379

divolgin (approved):

PR #2395, opened 2021-12-09
explicitly do not decrypt passwords shown on the config page

laverya:
#### What type of PR is this?

followup to kind/chore in https://github.com/replicatedhq/kots/pull/2386

The config page should not show decrypted passwords if they were not freshly entered

Fixes #

```release-note
NONE
```
(this is how things work in the currently released kotsadm version)

### DIFF
--- a/pkg/crypto/aes_test.go
+++ b/pkg/crypto/aes_test.go
@@ -1,0 +1,68 @@
+package crypto
+
+import (
+	"encoding/base64"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func Test_General(t *testing.T) {
+	req := require.New(t)
+	// ensure that it works the first time
+	req.NoError(NewAESCipher())
+	cipherKey := ToString()
+
+	// ensure that values that are encrypted can be decrypted again
+	testValue := "this is a test value"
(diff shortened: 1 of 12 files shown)


divolgin (dismissed):

pkg/crypto/aes_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+	testDecrypted, err = Decrypt(testEncrypted)
+	req.NoError(err)
+	req.Equal(testValue, string(testDecrypted))
+
+	// ensure that string values of the cipherKey can be added
+	err = InitFromString(cipherKey)
+	req.NoError(err)
+
+	fmt.Printf("cipherKey: %s encrypted: %s\n", cipherKey, base64.StdEncoding.EncodeToString(testEncrypted))
```
divolgin, line 37: this need?

divolgin (dismissed):

pkg/crypto/aes_test.go
laverya: 🤦‍♂️ no, that was for me to get some values to hardcode below

divolgin (approved):

PR #2396, opened 2021-12-09
add version to deploy to upstream upgrade cmd

morningvera:
#### What type of PR is this?

kind/enhancement

This adds the "version to deploy" to the `upstream upgrade` command output. This works for the `--deploy` and `--deploy-label-version` flags but currently does not for the auto deploy case.

[36671](https://app.shortcut.com/replicated/story/36671/kots-cli-add-current-and-latest-version-details-and-ensure-this-information-is-available-as-json-output)

```release-note
Added the version that will deploy to the upstream upgrade CLI command output.
```

NONE

### DIFF
--- a/pkg/updatechecker/updatechecker.go
+++ b/pkg/updatechecker/updatechecker.go
@@ -158,6 +158,7 @@ type UpdateCheckResponse struct {
 	AvailableUpdates  int64
 	CurrentRelease    UpdateCheckRelease
 	AvailableReleases []UpdateCheckRelease
+	DeployingRelease  UpdateCheckRelease
 }
 
 type UpdateCheckRelease struct {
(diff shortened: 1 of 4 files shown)


pkg/updatechecker/updatechecker.go
```diff
@@ -338,6 +340,49 @@ func ensureDesiredVersionIsDeployed(opts CheckForUpdatesOpts, clusterID string)
 	return nil
 }
 
+func getVersionToDeploy(opts CheckForUpdatesOpts, clusterID string, availableReleases []UpdateCheckRelease) UpdateCheckRelease {
```
laverya, line 343: I wish we could just have the logic for deployLatestVersion etc handle this
but that all runs after this returns

laverya (approved):

PR #2405, opened 2021-12-14
Write helm midstream with HelmChart namespace

emosbaugh:
#### What type of PR is this?

kind/bug

When using useHelmInstall with namespace in kots.io/v1beta1.HelmChart, image pull secrets are rendered in the kotsadm console namespace. This will render in the overridden namespace.

Screen Shot 2021-12-13 at 5 51 46 PM

Fixes NONE

```release-note
Fixed an issue that would cause image pull secrets to be rendered in the Admin Console namespace rather than the `namespace` specified in the kots.io/v1beta1.HelmChart when using `useHelmInstall`.
```

NONE

### DIFF
--- a/pkg/tests/pull/cases/configcontext/testcase.yaml
+++ b/pkg/tests/pull/cases/configcontext/testcase.yaml
@@ -1,5 +1,6 @@
 Name: test helm pull
 PullOptions:
+  Namespace: app-namespace
   ExcludeAdminConsole: true
   IsAirgap: true
   Silent: true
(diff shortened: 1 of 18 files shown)


divolgin (dismissed):

divolgin (approved):

pkg/tests/pull/cases/configcontext/testcase.yaml
```diff
@@ -1,5 +1,6 @@
 Name: test helm pull
 PullOptions:
+  Namespace: app-namespace
```
emosbaugh, line 3: when kotsadm invokes kots pull it always passes a namespace

divolgin (approved):

PR #2414, opened 2021-12-23
Populated KnownImages value in the installation file

divolgin:
#### What type of PR is this?

kind/bug

When running garbage collection in private registry, Admin Console uses the `KnownImages` list to determine what images can be deleted.  Under some conditions, that list was not being populated, which would cause garbage collection to ignore corresponding images in the registry.

Fixes #

```release-note
Fixes a bug that causes application images to not be [deleted from private registry](https://kots.io/kotsadm/registries/kurl-registry/).
Fixes a bug that causes images included in support bundle's [`run` collector](https://troubleshoot.sh/docs/collect/run/#image-required) to not be deleted from private registry.
```

NONE

### DIFF
--- a/pkg/registry/images.go
+++ b/pkg/registry/images.go
@@ -253,7 +253,7 @@ func deleteUnusedImages(ctx context.Context, registry types.RegistrySettings, us
 		}
 	}
 
-	for _, i := range usedImages {
+	for _, usedImage := range usedImages {
 		registryOptions := dockerregistry.RegistryOptions{
 			Endpoint:  registry.Hostname,
 			Namespace: registry.Namespace,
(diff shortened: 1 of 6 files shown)


pkg/registry/images.go
```diff
@@ -387,12 +389,15 @@ func runGCCommand(ctx context.Context) error {
 		logger.Infof("garbage collect command stderr: %s", stderr.Bytes())
 
 		if err != nil {
-			return errors.Wrap(err, "failed to stream command output")
+			errs = append(errs, errors.Wrap(err, "failed to stream command output"))
+			continue
 		}
 
+		// terminate ater the first successful loop iteration
```
DanStough, line 396:
```suggestion
		// terminate after the first successful loop iteration
```

DanStough (approved):

DanStough (approved):

PR #2425, opened 2022-01-07
feat(cli): build universal M1 binary

DanStough:
#### What type of PR is this?
type::feature

M1 support for Mac clients. Bash script to install still needs to be updated accordingly because `/usr/local/bin` doesn't exist by default.

I checked the deprecation notices between the version we were using and the new version.

```release-note
Support universal binary for KOTS CLI MacOS M1.
```

NONE

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -401,7 +401,7 @@ jobs:
     - name: Run GoReleaser
       uses: goreleaser/goreleaser-action@v2
       with:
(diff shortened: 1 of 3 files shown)


laverya (dismissed):
I like this, but we should REALLY make sure goreleaser still works after this change

DanStough:
Test locally with `goreleaser build ...` and I get the following artifacts:
image

laverya (approved):

PR #2427, opened 2022-01-07
improve the helm chart deployment method changed logic

laverya:
now it will only error for a change if the chart was not previously excluded

This PR makes it so that an 'excluded' helm chart can have its deployment method changed without an error.

Fixes #

```release-note
* excluded helm charts can change between 'UseHelmInstall: true' and 'UseHelmInstall: false' without errors, bypassing the normal check.
```


pkg/pull/pull.go
```diff
@@ -292,6 +292,10 @@ func Pull(upstreamURI string, pullOptions PullOptions) (string, error) {
 		}
 
 		for _, prevChart := range prevHelmCharts {
+			if prevChart.Spec.Exclude.BoolVal || prevChart.Spec.Exclude.String() == "1" {
```
MikaelSmith, line 295: Feels like there should be Truthy/Falsey methods to avoid having to copy this pattern around.

laverya: yeah this isn't how I should actually implement it, in retrospect

I should just be using the .String method, for one

laverya: because .String already does the conversion of true/false into "1"/"0"

morningvera (dismissed):

MikaelSmith:
Anything blocking this?

divolgin (approved):

pkg/pull/pull.go
```diff
@@ -292,6 +292,10 @@ func Pull(upstreamURI string, pullOptions PullOptions) (string, error) {
 		}
 
 		for _, prevChart := range prevHelmCharts {
+			if prevChart.Spec.Exclude.String() == "1" {
```
MikaelSmith, line 295: What is prevChart here? The currently deployed version or the last rendered version?

laverya: I believe it's the last rendered version, but I am not sure right now - @divolgin ?

divolgin (approved):

pkg/pull/pull.go
divolgin: it's the previous rendered version, `sequence - 1`

MikaelSmith: Seems like it'd be better to compare to the actually deployed version, but I don't know how hard that is to do.

PR #2431, opened 2022-01-13
Support for okteto dev environment

marccampbell:
This adds support for using KOTS in the replicated okteto dev cluster. This is still super early and not fully supported and has some limitations.

What did I do?
- I added an okteto-pipeline.yml in the root. 
- I created a kustomize/overlays/okteto to go alongside the dev overlay. This allows us to merge this in and work in main, while some engineers continue to work using codeserver and skaffold.
- I rewrote the migrations from a job to a deployment with sleep. This matches our other okteto environments and has been working well

Limitations and next steps:
- Add support for kurl_proxy
- There's no cluster role or service account! The okteto env doesn't support these. We can discuss adding RBAC policies to allow, but for now, the app will have to be deployed to the same namespace as the kots service
- Work on (test and fix) deploying an app using this admin console
- Add an `okteto.yml` to each subproject
- Our in-cluster pod is currently running webpack-dev-server. Let's move this to nginx and compile at build time. The `okteto up` version should run a dev server only. This will make startup time quicker because we can absorb the cost at build time.

If you want to use this, I strongly recommend that you create a `$USERNAME-kots` namespace in okteto to try it in, and not use your current okteto namespace. You might want to delete, and not have to redeploy other services.


hack/dev/Dockerfile.okteto
```diff
@@ -0,0 +1,176 @@
+FROM golang:1.17 as builder
```
syntastical, line 1: I think we should make part of these a permanent image in a shared registry somewhere.  I don't think it would need to hold this back, but more of a thought on next steps.  I'm experimenting on some like that now, mixed with some other optimizations.

web/Dockerfile.okteto
```diff
@@ -0,0 +1,15 @@
+FROM node:10
+EXPOSE 3000 9229
+
+ADD ./package.json /src/package.json
+ADD ./yarn.lock /src/yarn.lock
+WORKDIR /src
+RUN yarn install --silent --frozen-lockfile
+
+ADD . /src/
+WORKDIR /src
```
syntastical, line 10: There is another WORKDIR above.  This one can be remove.  Also, you could put WORKDIR at the top and simplify your COPY/ADD by remove the "/src"

syntastical (approved):

web/Dockerfile.okteto
marccampbell: This is just a copy of the current Skaffold dockerfile. Thoughts on keeping iterating after merge?

PR #2432, opened 2022-01-13
descriptive errmsg when hostpath dne

stefanrepl:
#### What type of PR is this?

type::feature

Shows a more descriptive error message when a user chooses a host path that does not exist.  It is needed so our users/vendors can easily understand what they are doing wrong.

Fixes #

Attached screenshots show the two separate views a user might see this.  One is updating  storage and one creating storage.

```release-note
     Display descriptive error when provided host path does not exist.
```

NONE
Screen Shot 2022-01-13 at 9 46 04 AM
Screen Shot 2022-01-13 at 9 46 14 AM

### DIFF
--- a/pkg/handlers/snapshots.go
+++ b/pkg/handlers/snapshots.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"os"
 	"regexp"
 	"strconv"
 	"strings"
@@ -180,6 +181,12 @@ func (h *Handler) UpdateGlobalSnapshotSettings(w http.ResponseWriter, r *http.Re
 					JSON(w, http.StatusConflict, globalSnapshotSettingsResponse)
 					return
 				}
+				if _, ok := errors.Cause(err).(*kotssnapshot.HostPathNotFoundError); ok {
+					globalSnapshotSettingsResponse.Error = err.Error()
+					JSON(w, http.StatusBadRequest, globalSnapshotSettingsResponse)
+					return
+				}
+
 				logger.Error(err)
 				globalSnapshotSettingsResponse.Error = "failed to configure file system provider"
 				JSON(w, http.StatusInternalServerError, globalSnapshotSettingsResponse)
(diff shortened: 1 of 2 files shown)


pkg/snapshot/filesystem.go
```diff
@@ -52,6 +52,14 @@ type ResetFileSystemError struct {
 	Message string
 }
 
+type HostPathFileSystemError struct {
```
sgalsaleh, line 55: how about calling it
```suggestion
type HostPathNotFoundError struct {
```

pkg/handlers/snapshots.go
```diff
@@ -431,6 +444,11 @@ func configureMinioFileSystemProvider(ctx context.Context, clientset kubernetes.
 		ForceReset:       fileSystemOptions.ForceReset,
 		FileSystemConfig: fileSystemOptions.FileSystemConfig,
 	}
+
+	if _, err := os.Stat(*deployOptions.FileSystemConfig.HostPath); os.IsNotExist(err) {
```
sgalsaleh, line 448: would be a good idea to catch other errors as well, something like
```go
if _, err := os.Stat(*deployOptions.FileSystemConfig.HostPath); os.IsNotExist(err) {
    return &kotssnapshot.HostPathFileSystemError{Message: "provided host path does not exist"}
} else if err != nil {
    return errors.Wrap(err, "failed to os stat")
}
```

pkg/snapshot/filesystem.go
```diff
@@ -52,6 +52,14 @@ type ResetFileSystemError struct {
...
+	Message string
+}
+
+func (e HostPathFileSystemError) Error() string {
```
sgalsaleh, line 59: nit pick: let's keep the type definition and the `Error()` function implementation coupled for the `ResetFileSystemError` type. something like:
```go
type ResetFileSystemError struct {
	Message string
}

func (e ResetFileSystemError) Error() string {
	return e.Message
}

type HostPathNotFoundError struct {
	Message string
}

func (e HostPathNotFoundError) Error() string {
	return e.Message
}
```

pkg/handlers/snapshots.go
stefanrepl: ah ofcourse, addressed

pkg/snapshot/filesystem.go
stefanrepl: sounds good, changed

sgalsaleh (approved):

PR #2438, opened 2022-01-14
fix: try to kubectl apply single docs

DanStough:
#### What type of PR is this?
type::bug

If one of the resources in a multi-doc yaml is too big to apply, then KOTS was automatically `PATCH`ing all the resources. This caused load balancer services to regenerate, causing downtime.

Fixes [SC-41322](https://app.shortcut.com/replicated/story/41322/kots-uses-patch-on-all-child-manifests-if-parent-manifest-was-too-large)

```release-note
Fixes issues with Load Balancer Services (and potentially other objects) being regenerated by Kubernetes in special cases when  no changes were applied.
```

NONE


pkg/operator/applier/kubectl.go
```diff
@@ -154,6 +157,18 @@ func (c *Kubectl) ApplyCreateOrPatch(targetNamespace string, slug string, yamlDo
 
 	var combinedStdout, combinedStderr []byte
 	for name, b := range docs {
+
+		stdout, stderr, err := c.Apply(targetNamespace, slug, b, dryRun, wait, annotateSlug)
+		if err == nil {
+			combinedStdout = append(combinedStdout, stdout...)
+			combinedStderr = append(combinedStderr, stderr...)
+			continue
+		} else if !strings.Contains(string(stderr), "metadata.annotations: Too long") {
```
sgalsaleh, line 166: why not always fall back to create/patch if there is a failure, and not just when the annotation is too long? just to preserve the old behavior

DanStough: A little confused here, that would be exactly the way it was before, correct?

The problem with the existing behavior was that patching objects like a load balancer would cause them to change their nodeports and become momentarily disconnected. I think the failure case for this vendor is that they have one YAML file where a doc fails to apply, then EVERYTHING in the doc has to be patched or created regardless of size.

sgalsaleh (approved):

PR #2442, opened 2022-01-19
Update web to use nginx in Okteto when not in dev mode.

syntastical:
#### What type of PR is this?
type::chore

Updates Web's dev environment to use Nginx when not in dev mode

NONE

NONE


web/Dockerfile.okteto
```diff
@@ -1,15 +1,19 @@
-FROM node:10
-EXPOSE 3000 9229
+FROM node:10 as dev
```
marccampbell, line 1: ouch

web/src/components/apps/AppConfig.jsx
```diff
@@ -414,7 +414,7 @@ class AppConfig extends Component {
 
   isConfigReadOnly = (app) => {
     const { match } = this.props;
-    if (!match.params.sequence) return false;
+    if (!match.params.sequence) {return false;}
```
marccampbell, line 417: was this generated by some prettier?

web/webpack.config.dev.js
```diff
@@ -19,14 +20,14 @@ module.exports = {
   module: {
     rules: [
       {
-        test: /\.(js|jsx)$/,
+        test: /\.[jt]sx?$/,
```
marccampbell, line 23: intentional?  `jt`?

web/src/components/apps/AppConfig.jsx
syntastical: Definitely generated.  I think it was ESLint

web/webpack.config.dev.js
syntastical: Right now there is no typescript, but I was trying to standardize these so it's more versatile.  It would match js/ts/tsx/jsx

web/Dockerfile.okteto
syntastical: yeah, I wanted to upgrade the node version but it's going to require some retooling that could take some time.  I was slow for Vendor web but I think I know the steps.  I can try to get it done immediately after this though.

PR #2449, opened 2022-01-20
Grayson/version history dashboard udpates

GraysonNull:
#### What type of PR is this?
type::bug

Fixes an issue where the dashboard version card would not stop showing that preflight checks were running for the latest pending version

Yes
```release-note
Fixes an issue where the dashboard version card would not stop showing that preflight checks were running for the latest pending version
```

NONE

### DIFF
--- a/web/src/components/apps/Dashboard.jsx
+++ b/web/src/components/apps/Dashboard.jsx
@@ -13,7 +13,7 @@ import AutomaticUpdatesModal from "@src/components/modals/AutomaticUpdatesModal"
 import SnapshotDifferencesModal from "@src/components/modals/SnapshotDifferencesModal";
 import Modal from "react-modal";
 import { Repeater } from "../../utilities/repeater";
-import { Utilities } from "../../utilities/utilities";
+import { Utilities, isAwaitingResults } from "../../utilities/utilities";
 import { AirgapUploader } from "../../utilities/airgapUploader";
 
 import "../../scss/components/watches/Dashboard.scss";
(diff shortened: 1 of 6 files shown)


web/src/components/apps/Dashboard.jsx
```diff
@@ -232,6 +234,38 @@ class Dashboard extends Component {
...
+        headers: {
+          "Authorization": Utilities.getToken(),
+          "Content-Type": "application/json",
+        },
+        method: "GET",
+      });
+      if (res.ok && res.status == 200) {
+        const app = await res.json();
+        if (app?.downstreams?.length > 0 && app?.downstreams[0].pendingVersions?.length > 0) {
+          const isAwaitingResults = app.downstreams[0].pendingVersions[0].status === "pending_preflight" || app.downstreams[0].pendingVersions[0].status === "deploying" || app.downstreams[0].pendingVersions[0].status === "unknown";
```
sgalsaleh, line 252: if this logic exists in the version history too, we should move it to a shared function

GraysonNull: @sgalsaleh updated this.

web/src/components/apps/Dashboard.jsx
```diff
@@ -257,6 +290,7 @@ class Dashboard extends Component {
 
             if (this.props.updateCallback) {
               this.props.updateCallback();
+              this.state.fetchAppDownstreamJob.start(this.fetchAppDownstream, 2000);
```
sgalsaleh, line 293: this shouldn't depend on the `updateCallback` function, and should be outside of this if statement

sgalsaleh (approved):

sgalsaleh (approved):

PR #2493, opened 2022-01-31
Admin-console listen on all interfaces

Iristyle:
- When building a tools container for use in development / CI that
   contains KOTS, to able to access the admin-console portforward from
   the system hosting the container, the port needs be exposed on all
   interfaces. Otherwise, the port is inaccessible from the host.

   host -> container -> k8s cluster

   This allows for a workflow where kubectl / kots is not installed on
   a host.

#### What type of PR is this?

type::feature

Putting this up for discussion... didn't build / test it just yet.

Relates to https://github.com/kubernetes/kubernetes/issues/36152

```release-note
Admin-console port will listen on all adapters instead of just localhost. This necessary to support tool container workflows.
```

NONE

### DIFF
--- a/pkg/k8sutil/portforward.go
+++ b/pkg/k8sutil/portforward.go
@@ -27,8 +27,8 @@ import (
 )
 
 func IsPortAvailable(port int) bool {
-	// portforward explicitly listens on localhost
-	host := net.JoinHostPort("127.0.0.1", strconv.Itoa(port))
+	// portforward listens on all interfaces
+	host := net.JoinHostPort("0.0.0.0", strconv.Itoa(port))
 	server, err := net.Listen("tcp4", host)
 	if err != nil {
 		return false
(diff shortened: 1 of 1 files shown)


pkg/k8sutil/portforward.go
```diff
@@ -96,7 +96,7 @@ func PortForward(localPort int, remotePort int, namespace string, podName string
 	stopChan, readyChan := make(chan struct{}, 1), make(chan struct{}, 1)
 	out, errOut := new(bytes.Buffer), new(bytes.Buffer)
 
-	forwarder, err := portforward.New(dialer, []string{fmt.Sprintf("%d:%d", localPort, remotePort)}, stopChan, readyChan, out, errOut)
+	forwarder, err := portforward.NewOnAddresses(dialer, []string{"localhost", "0.0.0.0"} []string{fmt.Sprintf("%d:%d", localPort, remotePort)}, stopChan, readyChan, out, errOut)
```
Iristyle, line 99: Not sure it's necessary to also have `localhost` in the list.

sgalsaleh:
Thanks for your contribution! As Mikael mentioned, this has security implications, and listening on localhost only is a conscious decision. If you'd like to widely expose the service, I'd suggest creating a NodePort service. Thanks!

PR #2505, opened 2022-02-03
Show message to upgrade to the target kots version in the UI footer

sgalsaleh:
#### What type of PR is this?

type::feature

Show message to upgrade to the target kots version in the UI footer

```release-note
NONE
```

NONE

Image 2022-02-03 at 12 44 29 PM


pkg/handlers/app.go
```diff
@@ -184,6 +184,11 @@ func responseAppFromApp(a *apptypes.App) (*types.ResponseApp, error) {
 		return nil, errors.Wrap(err, "failed to check if rollback is supported")
 	}
 
+	targetKotsVersion, err := store.GetStore().GetTargetKotsVersionForVersion(a.ID, latestVersion.Sequence)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get target kots version")
```
divolgin, line 189: how about not failing for this?

divolgin (dismissed):

pkg/handlers/app.go
sgalsaleh: this will only fail if the app spec doesn't load for some reason, which will already be a problem for other fields. but i see your point. i can update

GraysonNull (dismissed):

divolgin (approved):

PR #2522, opened 2022-02-10
sc-39403 use the non-minio alternative for hostpath snapshots

jala-dx:
#### What type of PR is this?
type:: feature

#### Which issue(s) this PR fixes:
sc-39403

Fixes #

#### Special notes for your reviewer:
wip

#### Does this PR introduce a user-facing change?

```release-note
TODO

```

#### Does this PR require documentation?
May be YES (TODO)

PR #2523, opened 2022-02-11
hostpath snapshots use available minio image

laverya:
#### What type of PR is this?

type::feature

This PR unpins minio versions from kotsadm, allowing them to be updated independently - and allowing kURL to not ship a minio image with every kotsadm installation.

Fixes #

```release-note
Kotsadm will use the existing minio image within the cluster to run hostpath snapshots, instead of pinning a specific version.
```

### DIFF
--- a/pkg/snapshot/filesystem.go
+++ b/pkg/snapshot/filesystem.go
@@ -276,17 +276,22 @@ func ensureFileSystemMinioDeployment(ctx context.Context, clientset kubernetes.I
 }
 
 func fileSystemMinioDeploymentResource(clientset kubernetes.Interface, secretChecksum string, deployOptions FileSystemDeployOptions, registryOptions kotsadmtypes.KotsadmOptions) (*appsv1.Deployment, error) {
-	minioTag, err := image.GetTag(image.Minio)
+	existingImage, err := image.MinioImage(clientset)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to find minio image")
+	}
+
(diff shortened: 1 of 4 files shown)


pkg/snapshot/filesystem.go
```diff
@@ -276,17 +276,22 @@ func ensureFileSystemMinioDeployment(ctx context.Context, clientset kubernetes.I
 }
 
 func fileSystemMinioDeploymentResource(clientset kubernetes.Interface, secretChecksum string, deployOptions FileSystemDeployOptions, registryOptions kotsadmtypes.KotsadmOptions) (*appsv1.Deployment, error) {
```
jala-dx, line 278: Do we still want kots to deploy minio pods in the upcoming releases?

laverya: yes, to continue to support the people already using it

cmd/imagedeps/main_test.go
```diff
@@ -37,16 +37,6 @@ func TestFunctional(t *testing.T) {
 		fn          tagFinderFn
 		expectError bool
 	}{
-		{
```
jala-dx, line 40: not sure why this code is removed.

laverya: because I was trying to remove minio entirely
I'm going to back out of doing that

jala-dx (approved):

jala-dx:
merged with #2522

laverya:
This is why you use normal merges 😆

PR #2529, opened 2022-02-15
update okteto debug configurations

pavansokkenagaraj:
- fix build issues 
- update Makefile commands
- optimize build time
- update debug pod resources
-


hack/dev/Dockerfile.okteto
```diff
@@ -2,12 +2,17 @@ FROM golang:1.17 as builder
 
 EXPOSE 2345
 
+ENV GO111MODULE=on
```
marccampbell, line 5: this is no longer necessary since we are using Go 1.17, let's not set it?

hack/dev/Dockerfile.okteto
```diff
@@ -2,12 +2,17 @@ FROM golang:1.17 as builder
...
+
 RUN curl -k https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - \
   && echo "deb http://apt.postgresql.org/pub/repos/apt/ buster-pgdg main" > /etc/apt/sources.list.d/PostgreSQL.list \
   && apt-get update && apt-get install -y --no-install-recommends gnupg2 postgresql-client-10 python3-pip ca-certificates \
   && pip install s3cmd \
   && rm -rf /var/lib/apt/lists/*
 
+RUN apt-get update -y \
+  && apt-get install -y libdevmapper-dev golang-github-proglottis-gpgme-dev libbtrfs-dev
```
marccampbell, line 14: `&& rm -rf /var/lib/apt/lists/* ` ?

okteto.yml
```diff
@@ -4,7 +4,12 @@ securityContext:
   capabilities:
     add:
     - SYS_PTRACE
+resources:
```
marccampbell, line 7: why is this wanted in the dev environment pod?

pavansokkenagaraj: I had troubles debugging the app. so, I updated the resources. I will remove and lets see if we have debugging troubles in future.

hack/dev/Dockerfile.okteto
pavansokkenagaraj: Should have removed this. 
I lost my calm figuring why `go build` in debug mode was downloading modules again and agin.
Later found out that our `Makefile` imports `Makefile.build` and sets the below values
https://github.com/replicatedhq/kots/blob/8985152ebb9c40cc64ff53c9e1a85f66d63af05b/Makefile.build#L1-L3

Now, In okteto debug environment, only one time modules are downloaded and later they are not downloaded again as `CGO_ENABLED=1` in `make kotsadm-debug-build` is set

marccampbell (approved):

okteto.yml
pavansokkenagaraj: Testing the debug environment with default settings [cpu: 1, memory: 1Gi] fails during modules download.
attempt 1: The debug pod dies and deployment spins up the new pod
```
google.golang.org/protobuf/types/descriptorpb
google.golang.org/protobuf/types/known/anypb
github.com/golang/protobuf/ptypes/any
Connection lost to your development container, reconnecting...
 ✓  Persistent volume successfully attached
 ✓  Images successfully pulled
 ✓  Files synchronized
    Context:   replicated.okteto.dev
    Namespace: pavansokkenagaraj
    Name:      kotsadm
    Forward:   2353 -> 2345
 ```
 
 ```
➜  ~ k get podmetrics kotsadm-okteto-99cc857d-gz2pn -o yaml
apiVersion: metrics.k8s.io/v1beta1
containers:
- name: kotsadm
  usage:
    cpu: 1995794206n
    memory: 1732132Ki
kind: PodMetrics
metadata:
  creationTimestamp: "2022-02-15T01:40:00Z"
  name: kotsadm-okteto-99cc857d-gz2pn
  namespace: pavansokkenagaraj
  selfLink: /apis/metrics.k8s.io/v1beta1/namespaces/pavansokkenagaraj/pods/kotsadm-okteto-99cc857d-gz2pn
timestamp: "2022-02-15T01:39:40Z"
window: 30s
➜  ~
```

attempt 2: worked, but could see mod download being slow.

hack/dev/Dockerfile.okteto
```diff
@@ -8,6 +8,10 @@ RUN curl -k https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - \
   && pip install s3cmd \
   && rm -rf /var/lib/apt/lists/*
 
+RUN apt-get update -y \
+  && apt-get install -y libdevmapper-dev golang-github-proglottis-gpgme-dev libbtrfs-dev \
```
syntastical, line 12: What are these additional dependencies used for?

Makefile
```diff
@@ -59,14 +59,11 @@ kotsadm:
 # Debugging
 .PHONY: kotsadm-debug-build
 kotsadm-debug-build:
-	go build ${LDFLAGS} -gcflags="all=-N -l" \
- 		$(BUILDFLAGS) \
-		-o ./bin/kotsadm-api-debug \
-		./cmd/kotsadm
+	CGO_ENABLED=1 go build -v -o ./bin/kotsadm-api-debug ./cmd/kotsadm
```
syntastical, line 62: Seems odd that we're enabling CGO here and dropping the gcflags

Makefile
```diff
@@ -59,14 +59,11 @@ kotsadm:
...
 
 .PHONY: kotsadm-debug
 kotsadm-debug: kotsadm-debug-build
-	dlv --listen=:2345 --headless=true --api-version=2 exec -- ./bin/kotsadm-api-debug
+	LOG_LEVEL=$(LOG_LEVEL) dlv --listen=:2345 --headless=true --api-version=2  exec ./bin/kotsadm-api-debug api
```
syntastical, line 66: as far as I know this LOG_LEVEL environment variable is unused.

syntastical (changes requested):

Makefile
pavansokkenagaraj: It is used in 
https://github.com/replicatedhq/kots/blob/8985152ebb9c40cc64ff53c9e1a85f66d63af05b/cmd/kotsadm/cli/api.go#L25-L27

hack/dev/Dockerfile.okteto
pavansokkenagaraj: These packages are needed to build the kotsadm projects, since some of the modules are c/c++ compiled libs are dependent on these packages.

Makefile
syntastical: I'd have to test this, I wouldn't think that CGO would impact when/if a module is downloaded

pavansokkenagaraj: @syntastical this one is easily testable
run `CGO_ENABLED=0 go build -v -o ./bin/kotsadm-api-debug ./cmd/kotsadm`
and you can see the modules downloading
and again run `CGO_ENABLED=1 go build -v -o ./bin/kotsadm-api-debug ./cmd/kotsadm` 
you can see the modules not downloading
and again run `CGO_ENABLED=0 go build -v -o ./bin/kotsadm-api-debug ./cmd/kotsadm`
you will see the mods downloading again

syntastical: I tested with and without CGO and I didn't notice a need re-download.  I think we should remove this flag.
We should add back the follow ${LDFLAGS} ${GCFLAGS} $(BUILDFLAGS) to the build.  I think this eliminates the need to alter Dockerfile.okteto.  We didn't include these in the skaffold environment so I'm skeptical that they'd be required here.

syntastical (changes requested):

syntastical (approved):

syntastical (dismissed):

PR #2550, opened 2022-02-24
feat: cli will warn user if version does not match api

cbodonnell:
#### What type of PR is this?

type::feature

The KOTS CLI version may not match the API version after an upgrade of KOTS.  This PR introduces a check that is run after all KOTS CLI commands and will warn the user of a version mismatch.

Fixes [SC-43774](https://app.shortcut.com/replicated/story/43774/cli-to-message-user-when-it-does-not-match-api-version)

Related to [SC-43597](https://app.shortcut.com/replicated/story/43597/poc-auto-upgrading-kots)

```release-note
NONE
```

NONE

### DIFF
--- a/cmd/kots/cli/check-version.go
+++ b/cmd/kots/cli/check-version.go
@@ -1,0 +1,98 @@
+package cli
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+
+	"github.com/pkg/errors"
+	"github.com/replicatedhq/kots/pkg/buildversion"
+	"github.com/replicatedhq/kots/pkg/handlers"
+	"github.com/replicatedhq/kots/pkg/k8sutil"
+	"github.com/replicatedhq/kots/pkg/logger"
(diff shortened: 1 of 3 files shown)


cmd/kots/cli/check-version.go
```diff
@@ -0,0 +1,98 @@
...
+		select {
+		case err := <-errChan:
+			if err != nil {
+				log.Error(err)
+			}
+		case <-stopCh:
+		}
+	}()
+
+	getMetadataURL := fmt.Sprintf("http://localhost:%d/api/v1/metadata", localPort)
```
divolgin, line 58: Can we change this to hit /healthz instead?   Metadata specifically refers to application metadata.  And there's already a version field that needs a real value: https://github.com/replicatedhq/kots/blob/10b6d379ae2b7ade0e482e3ebbb9fe22709da00e/pkg/handlers/healthz.go#L33

cbodonnell: Perfect.  Will make this edit...

divolgin (approved):

PR #2551, opened 2022-02-24
Create failed updates as pending download versions that can be downloaded later

sgalsaleh:
#### What type of PR is this?

type::feature

Create failed updates as pending download versions that can be downloaded later instead of discarding them in favor of newer valid updates.

```release-note
Creates failed updates as pending download versions that can be downloaded later instead of discarding them in favor of newer valid updates.
```

TBD


pkg/upstream/types/types.go
```diff
@@ -34,8 +34,13 @@ type Upstream struct {
...
 
 type Update struct {
-	Cursor       string `json:"cursor"`
-	VersionLabel string `json:"versionLabel"`
+	ChannelID    string     `json:"channelID"`
+	ChannelName  string     `json:"channelName"`
+	Cursor       string     `json:"cursor"`
+	VersionLabel string     `json:"versionLabel"`
+	ReleaseNotes string     `json:"releaseNotes"`
+	ReleasedAt   *time.Time `json:"releasedAt"`
```
divolgin, line 42: `omitempty` for pointer types?

sgalsaleh: this struct isn't returned to the client anywhere, just internal api usage

divolgin (approved):

PR #2559, opened 2022-02-25
bug: added --no-port-forward flag for install command

cbodonnell:
#### What type of PR is this?

type::bug

The existing flag --port-forward defaults to true and was inconsistent with the rest of the boolean flags, which default to false.  This occasionally led to support requests where `--port-forward false` was not leading to the expected behavior.  This PR introduces a new flag `--no-port-forward` that defaults to false and deprecates the old `--port-forward` flag.

Fixes [SC-38396](https://app.shortcut.com/replicated/story/38396/kots-install-port-forward-flag-doesn-t-work-without-the)

Also modified the Github actions to use this new flag in place of `--port-forward=false`.

```release-note
Adds a new flag `--no-port-forward` to the CLI install command to disable automatic port-forwarding. The old flag `--port-forward` has been deprecated.
```

Yes: [Docs PR](https://github.com/replicatedhq/kots.io/pull/680)

### DIFF
--- a/cmd/kots/cli/install.go
+++ b/cmd/kots/cli/install.go
@@ -373,7 +373,13 @@ func InstallCmd() *cobra.Command {
 
 			m.ReportInstallFinish()
 
-			if v.GetBool("port-forward") && !deployOptions.ExcludeAdminConsole {
+			isPortForwarding := !v.GetBool("no-port-forward")
+			if isPortForwarding {
+				// if --no-port-forward not specififed, check deprecated method
+				isPortForwarding = v.GetBool("port-forward")
+			}
+
+			if isPortForwarding && !deployOptions.ExcludeAdminConsole {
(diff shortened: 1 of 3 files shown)


cmd/kots/cli/install.go
```diff
@@ -373,7 +373,13 @@ func InstallCmd() *cobra.Command {
 
 			m.ReportInstallFinish()
 
-			if v.GetBool("port-forward") && !deployOptions.ExcludeAdminConsole {
+			isPortForwarding := !v.GetBool("no-port-forward")
```
sgalsaleh, line 376: why not just
```suggestion
			isPortForwarding := !v.GetBool("no-port-forward") || v.GetBool("port-forward")
```

sgalsaleh (approved):

PR #2568, opened 2022-03-01
Allow gitops to be disabled when there is not a valid config

marccampbell:
#### What type of PR is this?

type::bug

Allows gitops to be disabled when it's enabled but has an invalid config. Before this change, you were required to fix the configuration and then disable.

no

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,21 +48,24 @@ mock:
 	mockgen -source=pkg/handlers/interface.go -destination=pkg/handlers/mock/mock.go
 
 .PHONY: build
+build: kots
 build:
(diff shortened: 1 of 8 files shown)


sgalsaleh (approved):

syntastical (approved):

Makefile
```diff
@@ -48,21 +48,24 @@ mock:
 	mockgen -source=pkg/handlers/interface.go -destination=pkg/handlers/mock/mock.go
 
 .PHONY: build
+build: kots
```
divolgin, line 51: this will slow down `make build run` in okteto

PR #2572, opened 2022-03-02
Grayson/preflight UI updates

GraysonNull:
UI for app blocking preflights

Screen Shot 2022-03-02 at 9 20 27 AM


sgalsaleh (approved):

markpundsack:
@GraysonNull What does it like when a required (strict) preflight succeeds?

GraysonNull:
There's no text specifically calling that out but it would have the green check next to it.
Screen Shot 2022-03-02 at 11 19 44 AM

markpundsack:
Oh, but the `Required` tag is still there, so I think that's sufficient.

PR #2585, opened 2022-03-03
Ability to choose kotsadm-only restore option from full snapshots in the UI

sgalsaleh:
#### What type of PR is this?

type::feature

Adds the ability to choose kotsadm-only restore option from full snapshots in the UI

```release-note
Adds the option to be able to only restore the KOTS admin console to the "Full Snapshots" page.
```

Probably

Image 2022-03-03 at 1 49 37 PM


GraysonNull (dismissed):
Code looks good. Only suggestion would be using the different icon that is in figma. Just to help differentiate them more and have the icon make sense with the option.

GraysonNull (approved):

PR #2587, opened 2022-03-04
parse installation params

cbodonnell:


pkg/kotsadm/updates.go
```diff
@@ -102,8 +104,18 @@ func UpdateToVersion(newVersion string) error {
...
 	}
 
+	installationParams, err := kotsutil.GetInstallationParams(kotsadmtypes.KotsadmConfigMap)
+	if err != nil {
+		return errors.Wrap(err, "failed to get installation params")
+	}
+
 	args := []string{
 		fmt.Sprintf("namespace=%s", ns),
+		fmt.Sprintf("ensure-rbac=%v", installationParams.EnsureRBAC),
```
divolgin, line 114: these won't be automatically added to command.  The script is pretty simplistic: https://github.com/replicatedhq/kots/blob/main/deploy/kots-upgrade.sh#L28

cbodonnell: I see.  Will see about modifying it to include them.

PR #2589, opened 2022-03-04
feat: add sequence-based auto-updates non-semver apps

DanStough:
Signed-off-by: Kira Boyle <kira@replicated.com>

#### What type of PR is this?
type::feature

The behavior of automatic-deployments for non-semver channels still followed the same sorting logic as those with semantic versioning enabled. This feature uses sequence-based updates for those apps. It also cleans up the modal dialog to make user selections clearer.

Screenshots of Automatica Updates Modal:

Semver Enabled:
image

Semver Disabled:
image

Fixes [SC-44111](https://app.shortcut.com/replicated/story/44111/automatic-upgrades-should-sequence-based-on-semvers-if-the-channel-is-semver-enabled-otherwise-they-should-sort-based-on)

```release-note
- Application release channels that do not have [semantic versioning](https://docs.replicated.com/vendor/releases-understanding#semantic-versioning) enabled can now perform automatic release deployments. The most recent release will be used when updating, regardless of its version tag.
```

Yes, TBD.


pkg/updatechecker/updatechecker.go
```diff
@@ -449,8 +449,9 @@ func deployVersionLabel(opts CheckForUpdatesOpts, clusterID string, versionLabel
 	return nil
 }
 
-func autoDeploy(opts CheckForUpdatesOpts, clusterID string, semverAutoDeploy apptypes.SemverAutoDeploy) error {
-	if semverAutoDeploy == "" || semverAutoDeploy == apptypes.SemverAutoDeployDisabled {
+func autoDeploy(opts CheckForUpdatesOpts, clusterID string, autoDeploy apptypes.AutoDeploy) error {
+	if autoDeploy == "" || autoDeploy == apptypes.AutoDeployDisabled {
+		fmt.Println("auto deploy not running")
```
divolgin, line 454: debug stuff?

divolgin (dismissed):

pkg/updatechecker/updatechecker.go
DanStough: yerp!

pkg/updatechecker/updatechecker.go
```diff
@@ -469,33 +469,40 @@ func autoDeploy(opts CheckForUpdatesOpts, clusterID string, semverAutoDeploy app
...
 	var versionToDeploy *downstreamtypes.DownstreamVersion
 
-Loop:
-	for _, v := range appVersions.AllVersions {
-		if v == nil || v.Semver == nil {
-			continue
-		}
+	// In the case semver is not required, we only need to check if the newest app version
+	// is greater than the current version.
+	if autoDeploy == apptypes.AutoDeploySequence && currentVersion.Sequence < appVersions.AllVersions[0].Sequence {
```
sgalsaleh, line 474: what if `autoDeploy` is set to `apptypes.AutoDeploySequence` but the current sequence is greater than or equal to the latest version sequence? then it's gonna execute the semantic versioning logic below instead, which i think is incorrect.

web/src/components/apps/AppVersionHistory.jsx
```diff
@@ -1455,8 +1488,9 @@ class AppVersionHistory extends Component {
               isOpen={this.state.showAutomaticUpdatesModal}
               onRequestClose={this.toggleAutomaticUpdatesModal}
               updateCheckerSpec={app?.updateCheckerSpec}
-              semverAutoDeploy={app?.semverAutoDeploy}
+              autoDeploy={app?.autoDeploy}
               appSlug={app?.slug}
+              semVerEnabled={appLicense?.isSemverRequired}
```
sgalsaleh, line 1493: this should be part of the application response object, we shouldn't be making an extra request to get the license and determine that.

sgalsaleh: in here i mean https://github.com/replicatedhq/kots/blob/a606fcfa00ba79894376f43ab96053c3b11a7c43/pkg/handlers/app.go#L165

web/src/components/modals/AutomaticUpdatesModal.jsx
```diff
@@ -142,15 +164,28 @@ export default class AutomaticUpdatesModal extends React.Component {
...
-  handleSemverAutoDeployOptionChange = selectedSemverAutoDeploy => {
+  handleAutoDeployOptionChange = selectedAutoDeploy => {
     this.setState({
-      selectedSemverAutoDeploy: { ...selectedSemverAutoDeploy },
+      selectedAutoDeploy: { ...selectedAutoDeploy },
     });
   }
 
+  handleSequenceAutoUpdatesChange = (sequenceAutoDeployEnabled) => {
+    console.log("sequenceAutoDeployEnabled", sequenceAutoDeployEnabled);
```
sgalsaleh, line 174: remove log

web/src/components/modals/AutomaticUpdatesModal.jsx
```diff
@@ -214,18 +249,42 @@ export default class AutomaticUpdatesModal extends React.Component {
...
-              classNamePrefix="replicated-select"
-              placeholder="Automatically deploy new versions"
-              options={SEMVER_AUTO_DEPLOY_OPTIONS}
-              isSearchable={false}
-              getOptionValue={(option) => option.label}
-              value={selectedSemverAutoDeploy}
-              onChange={this.handleSemverAutoDeployOptionChange}
-              isOptionSelected={(option) => { option.value === selectedSemverAutoDeploy }}
-              />
+            { semVerEnabled ?
```
sgalsaleh, line 252: how about having the variable names be consistent? `semVerEnabled` vs `isSemverRequired`

sgalsaleh: i would've thought they're two different things, but it seems like they're the same?

web/src/components/apps/AppVersionHistory.jsx
DanStough: I'll have to add it to the App Response, but I see there are other license fields in their too. I'll consolidate that.

pkg/updatechecker/updatechecker.go
DanStough: Yeah, great callout.

web/src/components/modals/AutomaticUpdatesModal.jsx
DanStough: Yes, they are the same. I will change to `isSemverRequired`

sgalsaleh (dismissed):

sgalsaleh (approved):

PR #2590, opened 2022-03-05
Don't allow kots upgrades in airgap and kurl installs

divolgin:
#### What type of PR is this?

type::feature

Kots cannot self upgrade in airgap and kurl clusters.  This PR disables that functionality.

Fixes #

```release-note
NONE
```

NONE


pkg/handlers/update.go
```diff
@@ -217,6 +217,22 @@ func (h *Handler) UpdateAdminConsole(w http.ResponseWriter, r *http.Request) {
 		Success: false,
 	}
 
+	isKurl, err := kotsadm.IsKurl()
+	if err != nil {
+		logger.Error(errors.Wrap(err, "failed to check kURL"))
+		updateAdminConsoleResponse.Error = err.Error()
```
sgalsaleh, line 223: let's not return the error stack trace to the user? how about a hardcoded message?
```suggestion
		updateAdminConsoleResponse.Error = "failed to check kURL"
```

pkg/handlers/update.go
```diff
@@ -217,6 +217,22 @@ func (h *Handler) UpdateAdminConsole(w http.ResponseWriter, r *http.Request) {
...
+		JSON(w, http.StatusInternalServerError, updateAdminConsoleResponse)
+		return
+	}
+
+	if isKurl || kotsadm.IsAirgap() {
+		err := errors.New("cannot automatically update Admin Console in kURL or airgap installs")
```
sgalsaleh, line 229:
```suggestion
		err := errors.New("cannot automatically update the admin console in kURL or airgapped installations")
```

web/src/Root.jsx
```diff
@@ -367,15 +367,15 @@ class Root extends PureComponent {
...
+                  <ProtectedRoute path="/access/:tab?" render={(props) => <Access {...props} appName={this.state.selectedAppName} isKurlEnabled={this.state.adminConsoleMetadata.isKurl} isGeoaxisSupported={this.isGeoaxisSupported()} />} />
                   <ProtectedRoute
                     path={["/snapshots/:tab?"]}
                     render={
                       props => (
                         <SnapshotsWrapper
                           {...props} 
                           appName={this.state.selectedAppName} 
-                          isKurlEnabled={this.state.isKurlEnabled} 
+                          isKurlEnabled={this.state.adminConsoleMetadata.isKurl}
```
sgalsaleh, line 378:
```suggestion
                          isKurlEnabled={this.state.adminConsoleMetadata?.isKurl} 
```

web/src/Root.jsx
```diff
@@ -367,15 +367,15 @@ class Root extends PureComponent {
                   <Route path="/unsupported" component={UnsupportedBrowser} />
                   <ProtectedRoute path="/cluster/manage" render={(props) => <ClusterNodes {...props} appName={this.state.selectedAppName} />} />
                   <ProtectedRoute path="/gitops" render={(props) => <GitOps {...props} appName={this.state.selectedAppName} />} />
-                  <ProtectedRoute path="/access/:tab?" render={(props) => <Access {...props} appName={this.state.selectedAppName} isKurlEnabled={this.state.isKurlEnabled} isGeoaxisSupported={this.isGeoaxisSupported()} />} />
+                  <ProtectedRoute path="/access/:tab?" render={(props) => <Access {...props} appName={this.state.selectedAppName} isKurlEnabled={this.state.adminConsoleMetadata.isKurl} isGeoaxisSupported={this.isGeoaxisSupported()} />} />
```
sgalsaleh, line 370:
```suggestion
                  <ProtectedRoute path="/access/:tab?" render={(props) => <Access {...props} appName={this.state.selectedAppName} isKurlEnabled={this.state.adminConsoleMetadata?.isKurl} isGeoaxisSupported={this.isGeoaxisSupported()} />} />
```

web/src/components/apps/AppDetailPage.jsx
```diff
@@ -378,6 +378,7 @@ class AppDetailPage extends Component {
                         makingCurrentRelease={this.state.makingCurrentRelease}
                         redeployVersion={this.redeployVersion}
                         redeployVersionErrMsg={this.state.redeployVersionErrMsg}
+                        adminConsoleMetadata={this.props.adminConsoleMetadata}
```
sgalsaleh, line 381: `adminConsoleMetadata` is not being passed from Root.jsx to AppDetailPage.jsx as a prop, so this will be undefined

sgalsaleh, line 381: not sure if `adminConsoleMetadata` is being passed from Root.jsx to AppDetailPage.jsx as part of `{...props}` , so this might be undefined

web/src/components/apps/AppVersionHistory.jsx
```diff
@@ -1085,6 +1085,7 @@ class AppVersionHistory extends Component {
         handleSelectReleasesToDiff={this.handleSelectReleasesToDiff}
         renderVersionDownloadStatus={this.renderVersionDownloadStatus}
         isDownloading={this.state.versionDownloadStatuses?.[version.sequence]?.downloadingVersion}
+        adminConsoleMetadata={this.props.adminConsoleMetadata}
```
sgalsaleh, line 1088: same here

pkg/handlers/update.go
sgalsaleh: actually, why is there another function that doesn't return any errors? https://github.com/replicatedhq/kots/pull/2590/files?w=1#diff-eaeff8d34bc12281033f61dfba86f59d58aa83762ef0af09fb275bd5347fd8bfR11

pkg/handlers/update.go
```diff
@@ -217,6 +217,22 @@ func (h *Handler) UpdateAdminConsole(w http.ResponseWriter, r *http.Request) {
 		Success: false,
 	}
 
+	isKurl, err := kotsadm.IsKurl()
```
sgalsaleh, line 220: hmm, seems like there is another function that doesn't return any errors? https://github.com/replicatedhq/kots/pull/2590/files?w=1#diff-eaeff8d34bc12281033f61dfba86f59d58aa83762ef0af09fb275bd5347fd8bfR11

web/src/components/apps/AppVersionHistory.jsx
divolgin: it does work

sgalsaleh (approved):

pkg/handlers/update.go
divolgin: i think error is good here.  if error is ignored, we will end up with a broken kots instance.

PR #2595, opened 2022-03-07
parse installation parameters and pass to upgrade script

cbodonnell:
#### What type of PR is this?

type::feature

This PR parses the CLI flags saved from the install command and passes them to the upgrade script.

Fixes [SC-44365](https://app.shortcut.com/replicated/story/44365/match-original-cli-flags-during-upgrade)

n/a

```release-note
NONE
```

NONE


divolgin:
I think this can break older installs where flags are missing from config map.  `with-minio` is defaulted to true, but on the first upgrade it will be flipped to false.  I think we need to check if there is minio present if it's missing from config map.  The other three boolean flags are less scary since they will not break the deployment, but could break the upgrade process.

cbodonnell:
> I think this can break older installs where flags are missing from config map. `with-minio` is defaulted to true, but on the first upgrade it will be flipped to false. I think we need to check if there is minio present if it's missing from config map. The other three boolean flags are less scary since they will not break the deployment, but could break the upgrade process.

Weren't we thinking that this feature would be limited to newer versions only?  Or would this be the only reason why that might be necessary?

divolgin:
There maybe a way to infer other flags...  But there is nothing blocking the upgrade right now.  If you are thinking about addressing the issue separately, let's just make sure we have a story to address it since I think this is a blocker even for a beta release.

divolgin (approved):

cbodonnell:
Ok, I'll have think about that some more. Seems we will need to check if the flags exist in the config map first. If so, then just parse. Otherwise infer from other sources. Still feels to me like this falls under the same story, so I'm going to wait on this until we have it working for old and new installations.

cbodonnell:
Merging this now since we will be limiting this feature to newer installations only (v1.66.0+) that store the CLI flags in the config map.

PR #2606, opened 2022-03-11
don't autodeploy if preflights are failing

kcboyle:
#### What type of PR is this?
type::bug

This PR fixes a bug where auto-deploys would deploy updates to the application automatically even when the preflight checks were failing

```release-note
- fixes a bug with auto deploys where deployments would happen regardless of preflight success or failure. Auto-deploys will now only deploy if the preflights succeed
```

NONE

### DIFF
--- a/pkg/updatechecker/updatechecker.go
+++ b/pkg/updatechecker/updatechecker.go
@@ -1,7 +1,10 @@
 package updatechecker
 
 import (
+	"encoding/json"
 	"fmt"
+	"github.com/replicatedhq/troubleshoot/pkg/preflight"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"sync"
 	"time"
 
@@ -14,7 +17,7 @@ import (
 	"github.com/replicatedhq/kots/pkg/logger"
 	kotspull "github.com/replicatedhq/kots/pkg/pull"
 	"github.com/replicatedhq/kots/pkg/reporting"
-	"github.com/replicatedhq/kots/pkg/store"
+	storepkg "github.com/replicatedhq/kots/pkg/store"
 	storetypes "github.com/replicatedhq/kots/pkg/store/types"
 	upstreamtypes "github.com/replicatedhq/kots/pkg/upstream/types"
 	"github.com/replicatedhq/kots/pkg/util"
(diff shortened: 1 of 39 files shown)


kcboyle:
Screen Shot 2022-03-11 at 1 57 18 PM
Screenshot of log message when the preflights fail

DanStough (approved):

pkg/updatechecker/updatechecker.go
```diff
@@ -513,13 +516,49 @@ func autoDeploy(opts CheckForUpdatesOpts, clusterID string, autoDeploy apptypes.
...
+	}
+
+	var parsedResults preflight.UploadPreflightResults
+	if err = json.Unmarshal([]byte(preflightResult.Result), &parsedResults); err != nil {
+		return errors.Wrap(err, "failed to parse preflight results")
+	}
+
+	if parsedResults.Errors != nil {
+		return errors.Wrap(errors.New(preflightResult.Result), "preflight errors")
+	}
```
pavansokkenagaraj, line 557: preflight result doesn't account for failed analyzer

preflightresult.results[].isFail=true, the deployments should be blocked.

+ when there is no preflight spec in the app, we may need to test the edge case

pavansokkenagaraj (changes requested):

kcboyle:
messed up the rebase on this one. opening up a [new PR](https://github.com/replicatedhq/kots/pull/2628) that is fixed

pkg/updatechecker/updatechecker.go
```diff
@@ -513,13 +517,53 @@ func autoDeploy(opts CheckForUpdatesOpts, clusterID string, autoDeploy apptypes.
...
+	if err != nil {
+		return errors.Wrap(err, "failed to poll for preflights results")
+	}
+	// refetch latest results
+	preflightResult, err := store.GetPreflightResults(appID, sequence)
+	if err != nil {
+		return errors.Wrap(err, "failed to fetch preflight results")
+	}
+
+	if preflightResult == nil || len(preflightResult.Result) == 0 {
```
sgalsaleh, line 551: this should not fail, some apps don't have preflights

PR #2624, opened 2022-03-19
Support float sequences instead of integers

sgalsaleh:
#### What type of PR is this?

type::feature

Support decimal sequences instead of integers. this is helpful in many scenarios, for example editing the config of the currently deployed version when there are pending upstream versions available in the version history, and we don't want the new version to be on top of the list and dismiss them.

- User changes will bump the decimal part. e.g. config changes, registry settings, license syncs, download/upload, etc…
- Upstream updates will bump the integer part
- If the limit is reached for the patch sequence number (`.999`), will fall back to the old behavior of getting the next available app sequence.

```release-note
TBD
```

NONE

Image 2022-03-19 at 12 32 48 PM

### DIFF
--- a/pkg/handlers/download.go
+++ b/pkg/handlers/download.go
@@ -220,13 +220,13 @@ func (h *Handler) DownloadAppVersion(w http.ResponseWriter, r *http.Request) {
 	version, err := store.GetStore().GetAppVersion(a.ID, sequence)
 	if err != nil {
 		if store.GetStore().IsNotFound(err) {
-			errMsg := fmt.Sprintf("version for sequence %f not found", sequence)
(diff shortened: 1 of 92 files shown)


pkg/handlers/download.go
```diff
@@ -189,7 +189,7 @@ func (h *Handler) DownloadAppVersion(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	sequence, err := strconv.Atoi(mux.Vars(r)["sequence"])
+	sequence, err := strconv.ParseFloat(mux.Vars(r)["sequence"], 64)
```
divolgin, line 192: is 1.1 and 1.10 the same thing here?

PR #2625, opened 2022-03-19
Add hostpreflight to kotsutil

marccampbell:
This adds troubleshoot/v1beta2, HostPreflight to the kotsutil LoadKotsKinds function.

KOTS doesn't require this type be loaded because it's generally removed before the manifests are passed to kots. kURL should be the handler for these kinds. But this function (LoadKotsKinds) is exported and used elsewhere, including some other replicatedhq projects that lint and extract custom resources for other purposes. 

This PR will not affect KOTS, but allows a single, unified location for kots kinds to be loaded for maintainability of other projects.


sgalsaleh (dismissed):

erquhart:
@sgalsaleh can you give this one more review

PR #2627, opened 2022-03-21
use redeploy endpoint when redeploying from dashboard version card

GraysonNull:
#### What type of PR is this?
type::bug

Fixes a bug that called the deploy endpoint instead of the redeploy endpoint when attempting to redeploy a version from the application dashboard version card.

```release-note
Fixes a bug that called the deploy endpoint instead of the redeploy endpoint when attempting to redeploy a version from the application dashboard version card.
```

NONE


pavansokkenagaraj (approved):
LGTM

PR #2628, opened 2022-03-21
don't autodeploy if preflights are failing

kcboyle:
#### What type of PR is this?
type::bug

This PR fixes a bug where auto-deploys would deploy updates to the application automatically even when the preflight checks were failing

```release-note
- fixes a bug with auto deploys where deployments would happen regardless of preflight success or failure. Auto-deploys will now only deploy if the preflights succeed
```

NONE


pkg/updatechecker/updatechecker.go
```diff
@@ -26,13 +30,14 @@ import (
 // jobs maps app ids to their cron jobs
 var jobs = make(map[string]*cron.Cron)
 var mtx sync.Mutex
+var store = storepkg.GetStore()
```
sgalsaleh, line 33: could there be a race condition where this package is initialized before the store package and it would get the wrong store or maybe nil?

pkg/updatechecker/updatechecker.go
```diff
@@ -513,13 +518,63 @@ func autoDeploy(opts CheckForUpdatesOpts, clusterID string, autoDeploy apptypes.
...
 		return errors.Wrapf(err, "failed to deploy sequence %d with version label %s", versionToDeploy.Sequence, versionToDeploy.VersionLabel)
 	}
 
 	return nil
 }
 
+func waitForPreflightsToFinish(appID string, sequence int64) error {
+	app, err := store.GetApp(appID)
+	if err != nil {
+		return errors.Wrap(err, "failed get fetch app to check for preflights")
```
pavansokkenagaraj, line 535:
```suggestion
		return errors.Wrap(err, "failed get app to check for preflights")
```

sgalsaleh (approved):

PR #2632, opened 2022-03-22
Check all prerequisites exist before installing airgap bundle

divolgin:
#### What type of PR is this?

Block airgap bundle upload if prerequisite versions are not installed

Screen Shot 2022-03-22 at 9 14 55 AM

type::feature

Fixes #

```release-note
* Adds support for required versions in airgap installs.
```

NONE

### DIFF
--- a/pkg/airgap/update.go
+++ b/pkg/airgap/update.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/blang/semver"
@@ -94,6 +95,18 @@ func UpdateAppFromPath(a *apptypes.App, airgapRoot string, airgapBundlePath stri
 	airgap, err := pull.FindAirgapMetaInDir(airgapRoot)
 	if err != nil {
 		return errors.Wrap(err, "failed to parse license from file")
+	}
+
+	missingPrereqs, err := GetMissingRequiredVersions(a.ID, airgap)
+	if err != nil {
+		return errors.Wrapf(err, "failed to check required versions")
+	}
+
+	if len(missingPrereqs) > 0 {
+		return util.ActionableError{
+			NoRetry: true,
+			Message: fmt.Sprintf("This airgap bundle requires the following releases to be installed first: %s", strings.Join(missingPrereqs, ", ")),
+		}
 	}
 
 	archiveDir, baseSequence, err := store.GetStore().GetAppVersionBaseArchive(a.ID, airgap.Spec.VersionLabel)
@@ -300,4 +313,29 @@ func canInstall(beforeKotsKinds *kotsutil.KotsKinds, afterKotsKinds *kotsutil.Ko
 	}
 
 	return nil
+}
+
+func GetMissingRequiredVersions(appID string, airgap *kotsv1beta1.Airgap) ([]string, error) {
+	appVersions, err := store.GetStore().GetAppVersions(appID)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get downstream versions")
+	}
+
+	installedVersions := make(map[string]bool)
+	for _, appVersion := range appVersions {
+		if appVersion.ChannelID != airgap.Spec.ChannelID {
+			continue
+		}
+		installedVersions[appVersion.UpdateCursor] = true
+	}
+
+	missingVersions := make([]string, 0)
+	for _, requiredRelease := range airgap.Spec.RequiredReleases {
+		if installedVersions[requiredRelease.UpdateCursor] {
+			break
+		}
+		missingVersions = append(missingVersions, requiredRelease.VersionLabel)
+	}
+
+	return missingVersions, nil
 }
--- a/pkg/airgap/update.go
+++ b/pkg/airgap/update.go
@@ -106,7 +106,7 @@ func UpdateAppFromPath(a *apptypes.App, airgapRoot string, airgapBundlePath stri
 	if len(missingPrereqs) > 0 {
 		return util.ActionableError{
 			NoRetry: true,
-			Message: fmt.Sprintf("This airgap bundle cannot be deployed because versions %s are required and must be uploaded first.", strings.Join(missingPrereqs, ", ")),
+			Message: fmt.Sprintf("This airgap bundle cannot be uploaded because versions %s are required and must be uploaded first.", strings.Join(missingPrereqs, ", ")),
 		}
 	}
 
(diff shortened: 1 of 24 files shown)


pkg/airgap/update.go
```diff
@@ -301,3 +314,28 @@ func canInstall(beforeKotsKinds *kotsutil.KotsKinds, afterKotsKinds *kotsutil.Ko
...
+		}
+		installedVersions[appVersion.UpdateCursor] = true
+	}
+
+	missingVersions := make([]string, 0)
+	for _, requiredRelease := range airgap.Spec.RequiredReleases {
+		if installedVersions[requiredRelease.UpdateCursor] {
+			break
+		}
+		missingVersions = append(missingVersions, requiredRelease.VersionLabel)
```
sgalsaleh, line 337: nit pick, Alex mentioned that these should be in ascending order, so i did this for online installs
```go
versionLabels = append([]string{v.VersionLabel}, versionLabels...)
```

pkg/airgap/update.go
```diff
@@ -315,26 +316,66 @@ func canInstall(beforeKotsKinds *kotsutil.KotsKinds, afterKotsKinds *kotsutil.Ko
...
+		return missingVersions, nil
+	}
+
 	for _, requiredRelease := range airgap.Spec.RequiredReleases {
-		if installedVersions[requiredRelease.UpdateCursor] {
+		laterReleaseInstalled := false
+		for _, appVersion := range installedVersions {
+			if license.Spec.IsSemverRequired {
+				// semvers can be compared across channels
+				if appVersion.Semver == nil {
```
sgalsaleh, line 344: semver and non-semver can live together. we probably shouldn't fail here, but rather just continue?

pkg/handlers/app.go
```diff
@@ -421,50 +425,97 @@ func (h *Handler) RemoveApp(w http.ResponseWriter, r *http.Request) {
...
+			logger.Error(errors.Wrap(err, response.Error))
+			JSON(w, http.StatusInternalServerError, response)
+			return
+		}
 
-	if kotsApp != nil {
-		response.IsCompatible = kotsutil.IsKotsVersionCompatibleWithApp(*kotsApp, request.IsInstall)
-	} else {
-		response.IsCompatible = true
+		if kotsApp != nil {
```
sgalsaleh, line 465: if kotsApp is nil, `CanInstall` should be explicitly set to `true` (same as old behavior). otherwise it will return an error in the if statement below.

pkg/airgap/update.go
```diff
@@ -96,6 +98,18 @@ func UpdateAppFromPath(a *apptypes.App, airgapRoot string, airgapBundlePath stri
...
 
+	missingPrereqs, err := GetMissingRequiredVersions(a, airgap)
+	if err != nil {
+		return errors.Wrapf(err, "failed to check required versions")
+	}
+
+	if len(missingPrereqs) > 0 {
+		return util.ActionableError{
+			NoRetry: true,
+			Message: fmt.Sprintf("This airgap bundle requires the following releases to be installed first: %s", strings.Join(missingPrereqs, ", ")),
```
sgalsaleh, line 109: @ajp-io any suggestion on the wording here? the one we're using for online installs doesn't necessarily work here.

pkg/handlers/app.go
```diff
@@ -421,50 +425,97 @@ func (h *Handler) RemoveApp(w http.ResponseWriter, r *http.Request) {
...
+		missingPrereqs, err := airgap.GetMissingRequiredVersions(a, decoded.(*kotsv1beta1.Airgap))
+		if err != nil {
+			response.Error = "failed to get release prerequisites"
+			logger.Error(errors.Wrap(err, response.Error))
+			JSON(w, http.StatusInternalServerError, response)
+			return
+		}
+
+		if len(missingPrereqs) > 0 {
+			response.Error = fmt.Sprintf("This airgap bundle requires the following releases to be installed first: %s", strings.Join(missingPrereqs, ", "))
```
sgalsaleh, line 512: same here about the wording, let'd discuss with Alex

pkg/airgap/update.go
sgalsaleh: @divolgin after discussing with @ajp-io:
```
This airgap bundle cannot be deployed because versions %s are required and must be uploaded first.
```

web/src/utilities/airgapUploader.js
```diff
@@ -160,13 +162,14 @@ export class AirgapUploader {
     }
   }
 
-  checkKotsVersionCompatibility = async appSpec => {
-    const res = await fetch(`${process.env.API_ENDPOINT}/app/iscompatible`, {
+  checkReleaseInstallable = async (appSpec, airgapSpec) => {
```
sgalsaleh, line 165: how about `isReleaseInstallable` or `canInstallRelease`?

sgalsaleh (approved):

pkg/airgap/update.go
```diff
@@ -96,6 +98,18 @@ func UpdateAppFromPath(a *apptypes.App, airgapRoot string, airgapBundlePath stri
...
 
+	missingPrereqs, err := GetMissingRequiredVersions(a, airgap)
+	if err != nil {
+		return errors.Wrapf(err, "failed to check required versions")
+	}
+
+	if len(missingPrereqs) > 0 {
+		return util.ActionableError{
+			NoRetry: true,
+			Message: fmt.Sprintf("This airgap bundle cannot be deployed because versions %s are required and must be uploaded first.", strings.Join(missingPrereqs, ", ")),
```
ajp-io, line 109: @sgalsaleh I think you meant uploaded here, right?
```suggestion
			Message: fmt.Sprintf("This airgap bundle cannot be uploaded because versions %s are required and must be uploaded first.", strings.Join(missingPrereqs, ", ")),
```

pkg/handlers/app.go
```diff
@@ -421,50 +425,97 @@ func (h *Handler) RemoveApp(w http.ResponseWriter, r *http.Request) {
...
+		missingPrereqs, err := airgap.GetMissingRequiredVersions(a, decoded.(*kotsv1beta1.Airgap))
+		if err != nil {
+			response.Error = "failed to get release prerequisites"
+			logger.Error(errors.Wrap(err, response.Error))
+			JSON(w, http.StatusInternalServerError, response)
+			return
+		}
+
+		if len(missingPrereqs) > 0 {
+			response.Error = fmt.Sprintf("This airgap bundle cannot be deployed because versions %s are required and must be uploaded first.", strings.Join(missingPrereqs, ", "))
```
ajp-io, line 512: Same here.
```suggestion
			response.Error = fmt.Sprintf("This airgap bundle cannot be uploaded because versions %s are required and must be uploaded first.", strings.Join(missingPrereqs, ", "))
```

PR #2634, opened 2022-03-22
handle css in grid of graphs better when scaling beyond three graphs

GraysonNull:
#### What type of PR is this?
type::bug

Fixes a style bug that causes the grid of metric graphs to be broken when there are more than three graphs.

Screen Shot 2022-03-22 at 4 45 55 PM

```release-note
Fixes a style bug that causes the grid of metric graphs to be broken when there are more than three graphs.
```

NONE


alicenstar (approved):
I would like to be able to see the changes reflected live in the UI (ie in Okteto), but until then, these changes look good.

PR #2640, opened 2022-03-23
make fetchSnapshotSettings wait for response

cbodonnell:
#### What type of PR is this?

type::bug

This PR fixes an issue where the UI was checking the snapshot settings every 2 seconds regardless of how long the request took to complete.  This introduced a race condition where later requests to check the settings sometimes returned earlier, causing some weird behavior with the UI.  This would sometimes cause the banner/modal to display indicating that velero was not configured, when it actually was.  This was confusing for customers and led to support issues as a result.

Fixes [SC-37857](https://app.shortcut.com/replicated/story/37857/snapshot-settings-shows-error-while-restic-and-velero-are-restarting)

Without this change, the promise would resolve immediately after the fetch request was initiated.  With this change, it will return the promise generated by the fetch api and will resolve once the request has completed.

```release-note
Fixes an issue where updating the snapshot storage location to NFS or Host Path would incorrectly display a dialog indicating that Velero was not installed and configured properly.
```

NONE

### DIFF
--- a/web/src/components/snapshots/SnapshotSettings.jsx
+++ b/web/src/components/snapshots/SnapshotSettings.jsx
@@ -32,7 +32,7 @@ class SnapshotSettings extends Component {
     checkForVeleroAndRestic: false
   };
 
-  fetchSnapshotSettings = async (isCheckForVelero) => {
+  fetchSnapshotSettings = (isCheckForVelero) => {
     this.setState({
       isLoadingSnapshotSettings: true,
       snapshotSettingsErr: false,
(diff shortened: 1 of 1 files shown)


web/src/components/snapshots/SnapshotSettings.jsx
```diff
@@ -41,7 +41,7 @@ class SnapshotSettings extends Component {
       minimalRBACKotsadmNamespace: "",
     });
 
-    fetch(`${process.env.API_ENDPOINT}/snapshots/settings`, {
+    return fetch(`${process.env.API_ENDPOINT}/snapshots/settings`, {
```
sgalsaleh, line 44:
```suggestion
    await fetch(`${process.env.API_ENDPOINT}/snapshots/settings`, {
```

divolgin (dismissed):

sgalsaleh (dismissed):

sgalsaleh (approved):

PR #2643, opened 2022-03-23
Upload support bundle to s3 when testim tests fail

emosbaugh:
#### What type of PR is this?

type::chore

This will upload a support bundle to s3 when testim tests fail to aid in debugging.

Fixes NONE

```release-note
NONE
```

NONE

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -358,6 +358,27 @@ jobs:
           fi
           exit $EXIT_CODE
 
+      - name: Generate support bundle on failure
+        if: failure()
+        env:
+          # aws replicated dev account e2e-kots-support-bundle user
+          AWS_ACCESS_KEY_ID: ${{ secrets.E2E_SUPPORT_BUNDLE_AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.E2E_SUPPORT_BUNDLE_AWS_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test.yaml
```diff
@@ -781,6 +886,27 @@ jobs:
...
           echo "All pods started"
 
+      - name: Generate support bundle on failure
+        if: failure()
+        env:
+          # aws replicated dev account e2e-kots-support-bundle user
+          AWS_ACCESS_KEY_ID: ${{ secrets.E2E_SUPPORT_BUNDLE_AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.E2E_SUPPORT_BUNDLE_AWS_SECRET_ACCESS_KEY }}
+          AWS_DEFAULT_REGION: us-east-1
+          KOTS_NAMESPACE: multi-namespace-yeti
```
emosbaugh, line 896: not sure what namespace should be here

sgalsaleh: seems right to me

sgalsaleh (dismissed):

sgalsaleh (approved):

.github/workflows/build-test.yaml
```diff
@@ -358,6 +358,27 @@ jobs:
           fi
           exit $EXIT_CODE
 
+      - name: Generate support bundle on failure
```
divolgin, line 361: The kubectl commands above are no longer needed with this change

PR #2644, opened 2022-03-23
show earliest required version as latest available on dashboard

GraysonNull:

Fixes #

```release-note

```

### DIFF
--- a/web/src/components/apps/DashboardVersionCard.jsx
+++ b/web/src/components/apps/DashboardVersionCard.jsx
@@ -2,7 +2,9 @@ import React from "react";
 import { Link, withRouter } from "react-router-dom";
 import ReactTooltip from "react-tooltip"
 
-import dayjs from "dayjs";
+import filter from "lodash/filter";
+import orderBy from "lodash/orderBy";
+import findIndex from "lodash/findIndex";
 import MarkdownRenderer from "@src/components/shared/MarkdownRenderer";
 import DownstreamWatchVersionDiff from "@src/components/watches/DownstreamWatchVersionDiff";
 import Modal from "react-modal";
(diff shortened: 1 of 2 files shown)


web/src/components/apps/DashboardVersionCard.jsx
```diff
@@ -907,16 +909,28 @@ class DashboardVersionCard extends React.Component {
       return null;
     }
 
+    let latestVersion = downstream?.latestVersion;
+    let newVersionsAfterUpgrade = null;
+    let versionsToSkip = downstream?.pendingVersions?.length - 1;
+    const requiredVersions = downstream?.pendingVersions?.length > 0 ? filter(downstream.pendingVersions, ["isRequired", true]) : [];
+    if (this.props.currentVersion?.deployedAt && requiredVersions.length > 0) { // If there is a version already deployed, and there is at least one required pending versions, set latestVersion to the earliest required version
```
sgalsaleh, line 916:
```suggestion
    if (this.props.currentVersion && requiredVersions.length > 0) { // If there is a version already deployed, and there is at least one required pending versions, set latestVersion to the earliest required version
```

web/src/components/apps/DashboardVersionCard.jsx
```diff
@@ -907,16 +909,28 @@ class DashboardVersionCard extends React.Component {
...
+      const sortedVersions = orderBy(requiredVersions, ["createdOn", "desc"]);
+      latestVersion = sortedVersions[0];
```
sgalsaleh, line 918: wouldn't this get the latest required version instead of earliest?

web/src/components/apps/DashboardVersionCard.jsx
```diff
@@ -907,16 +909,28 @@ class DashboardVersionCard extends React.Component {
       return null;
     }
 
+    let latestVersion = downstream?.latestVersion;
```
sgalsaleh, line 912: let's rename this to something like `nextAppVersion`? because reading the code otherwise is a bit confusing

web/src/components/apps/DashboardVersionCard.jsx
```diff
@@ -907,16 +909,28 @@ class DashboardVersionCard extends React.Component {
...
+    let newVersionsAfterUpgrade = null;
+    let versionsToSkip = downstream?.pendingVersions?.length - 1;
+    const requiredVersions = downstream?.pendingVersions?.length > 0 ? filter(downstream.pendingVersions, ["isRequired", true]) : [];
+    if (this.props.currentVersion?.deployedAt && requiredVersions.length > 0) { // If there is a version already deployed, and there is at least one required pending versions, set latestVersion to the earliest required version
+      const sortedVersions = orderBy(requiredVersions, ["createdOn", "desc"]);
+      latestVersion = sortedVersions[0];
+      const indexOfLatestVersion = findIndex(downstream?.pendingVersions, ["versionLabel", latestVersion.versionLabel]);
+      versionsToSkip = (downstream?.pendingVersions?.length - 1) - indexOfLatestVersion;
```
sgalsaleh, line 920: shouldn't this just be
```suggestion
      versionsToSkip = indexOfLatestVersion;
```

web/src/components/apps/DashboardVersionCard.jsx
GraysonNull: I'm not sure exactly why but using `asc` and `desc` return the same sort order for me. I'm going to update the method though to use `sortBy` which defaults to ascending order and is also grabbing the correct required version.

web/src/components/apps/DashboardVersionCard.jsx
GraysonNull: I don't think so. `indexOfLatestVersion` is just finding it's position in the array so for example it might be a the second position in the array but there could be 6 versions between it and the one that's currently deployed so showing "2 versions will be skipped...." would be incorrect. 

This number has been correct in all the scenarios i've tested. I'm trying to get loom working so I can get a demo video recorded to be able to see it all.

sgalsaleh: how would that even happen? pending versions is the list of versions above the currently deployed version, and the index of the "nextAppVersion" should be exactly the number of versions between the currently deployed and the next app version.

PR #2695, opened 2022-04-02
fix: ensure registry image pull secrets are preset in additional namespaces for minimal RBAC installations

sgalsaleh:
#### What type of PR is this?

type::bug

Ensures that the registry image pull secrets are preset in the additional namespaces specified by the application for minimal RBAC installations

```release-note
Fixes an issue where the registry image pull secrets were not applied in the additional namespaces specified by the application in minimal RBAC installations
```

NONE


marccampbell (dismissed):

marccampbell (approved):

pkg/operator/client/client.go
```diff
@@ -183,11 +173,14 @@ func (c *Client) deployManifests(deployArgs operatortypes.DeployAppArgs) (*deplo
 		if additionalNamespace == "*" {
 			continue
 		}
-
```
divolgin, line 186: Removed file has handling for wildcard namespace

```
				if watchedNamespace == "*" {
					deployImagePullSecret = true
				} else {
```

divolgin:
I think this breaks cluster level installs with `*` namespaces.  That removed implementation is still needed.

pkg/operator/client/deploy.go
```diff
@@ -246,6 +246,45 @@ func (c *Client) ensureNamespacePresent(name string) error {
 	return nil
 }
 
+func (c *Client) ensureImagePullSecretsPresent(namespace string, imagePullSecrets []string) error {
+	clientset, err := k8sutil.GetClientset()
```
divolgin, line 250: this should have  a mutex now that it's called from multiple go-routines

divolgin (dismissed):

PR #2706, opened 2022-04-05
fix: delete expired sessions from kotsStore

pavansokkenagaraj:
#### What type of PR is this?
type::bug

fix: delete expired sessions from kotsStore

Fixes # [SC-34580](https://app.shortcut.com/replicated/story/34580/kots-jwt-timeouts-are-too-long-non-oidc-login)

We are storing the sessions in k8s opaque secrets. Once we invalidate the expired sessions, we should delete the expired sessions as k8s secrets can hold 1 MB of data.

NONE

NONE

### DIFF
--- a/pkg/handlers/session.go
+++ b/pkg/handlers/session.go
@@ -75,6 +75,12 @@ func requireValidSession(kotsStore store.Store, w http.ResponseWriter, r *http.R
 	}
 
 	if time.Now().After(sess.ExpiresAt) {
+		if err := kotsStore.DeleteSession(sess.ID); err != nil {
+			err := errors.Wrapf(err, "session expired. failed to delete expired session %s", sess.ID)
+			response := ErrorResponse{Error: err.Error()}
+			JSON(w, http.StatusInternalServerError, response)
+			return nil, err
+		}
 		err := errors.New("session expired")
 		response := ErrorResponse{Error: err.Error()}
 		JSON(w, http.StatusUnauthorized, response)
--- a/pkg/handlers/session.go
+++ b/pkg/handlers/session.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/replicatedhq/kots/pkg/k8sutil"
+	"github.com/replicatedhq/kots/pkg/logger"
 	"github.com/replicatedhq/kots/pkg/session"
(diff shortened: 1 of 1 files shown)


pkg/handlers/session.go
```diff
@@ -75,6 +75,12 @@ func requireValidSession(kotsStore store.Store, w http.ResponseWriter, r *http.R
 	}
 
 	if time.Now().After(sess.ExpiresAt) {
+		if err := kotsStore.DeleteSession(sess.ID); err != nil {
+			err := errors.Wrapf(err, "session expired. failed to delete expired session %s", sess.ID)
+			response := ErrorResponse{Error: err.Error()}
+			JSON(w, http.StatusInternalServerError, response)
```
emosbaugh, line 81: should we just log here and continue?

emosbaugh (changes requested):

pkg/handlers/session.go
pavansokkenagaraj: that is what I had. But later changed it to Internal Server Error after looking at the logout flow.
but log and continue(returns 401) would be a better option

pavansokkenagaraj: @emosbaugh changed to logging in https://github.com/replicatedhq/kots/pull/2706/commits/bbbb57fe68071689624b9a5a3851a3fa450b9706#diff-8be07e05f305ae1ebec9b3760f46190803066daf92afb41de47e5adf9e182837R80

emosbaugh (dismissed):

laverya:
Won't this only delete a session from the store if someone attempts to use it after expiry?

Should there be a dailyish task to look through all the sessions in the store and clear old ones?

pavansokkenagaraj:
> Won't this only delete a session from the store if someone attempts to use it after expiry?
> 
> Should there be a dailyish task to look through all the sessions in the store and clear old ones?

Not sure about it. lemme bring it up tomorrow on standup

erquhart:
@pavansokkenagaraj remaining failure looks like we're printing the error message without parsing it correctly: https://github.com/replicatedhq/kots/pull/2696/files

cc/ @sgalsaleh this is the change license test, in case you have insights

pavansokkenagaraj:
Closing PR as the commits are merged as part of PR: https://github.com/replicatedhq/kots/pull/2711

PR #2707, opened 2022-04-05
Velero in okteto

syntastical:

Fixes #

```release-note

```

### DIFF
--- a/cmd/kots/cli/velero.go
+++ b/cmd/kots/cli/velero.go
@@ -81,11 +81,14 @@ func VeleroEnsurePermissionsCmd() *cobra.Command {
 				return errors.Wrap(err, "failed to get clientset")
 			}
 
-			if err := snapshot.EnsureVeleroPermissions(cmd.Context(), clientset, veleroNamespace, kotsadmNamespace); err != nil {
-				return err
+			skipRbac := v.GetBool("skip-rbac")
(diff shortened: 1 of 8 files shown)


cmd/kots/cli/velero.go
```diff
@@ -81,11 +81,14 @@ func VeleroEnsurePermissionsCmd() *cobra.Command {
 				return errors.Wrap(err, "failed to get clientset")
 			}
 
-			if err := snapshot.EnsureVeleroPermissions(cmd.Context(), clientset, veleroNamespace, kotsadmNamespace); err != nil {
-				return err
+			skipRbac := v.GetBool("skip-rbac")
```
sgalsaleh, line 84: can we have this consistent with the other commands? which i believe is called `ensure-rbac`

syntastical: Would they need then specificy `--ensure-rbac false` and the default would be true?

sgalsaleh: `--ensure-rbac=false`, and yes.

syntastical: Feels a little clunky, how about a compromise.  We use `skip-rbac-check` already for other commands, how do you feel about me changing `skip-rbac` to `skip-rbac-check`?

PR #2711, opened 2022-04-07
 fix: extend active user session by an hour if session expiryAt is less than one hour of session timeout

pavansokkenagaraj:
#### What type of PR is this?
type::bug

Currently kots UI session timeout is 12 hours.
If a user is logged in and is using the UI, he would be automatically logged out.
For user's convenience, extend the session expiry by an hour when current session is less than 11 hours (sessionTimeout: 12 hours - 1 hour) 

Fixes #

```release-note
fixes session timeout by extending the session by an hour if session expiry is at least less than one hour of session timeout
```

NO

### DIFF
--- a/pkg/handlers/session.go
+++ b/pkg/handlers/session.go
@@ -75,6 +75,12 @@ func requireValidSession(kotsStore store.Store, w http.ResponseWriter, r *http.R
 	}
 
 	if time.Now().After(sess.ExpiresAt) {
+		if err := kotsStore.DeleteSession(sess.ID); err != nil {
+			err := errors.Wrapf(err, "session expired. failed to delete expired session %s", sess.ID)
+			response := ErrorResponse{Error: err.Error()}
+			JSON(w, http.StatusInternalServerError, response)
+			return nil, err
+		}
 		err := errors.New("session expired")
 		response := ErrorResponse{Error: err.Error()}
 		JSON(w, http.StatusUnauthorized, response)
--- a/pkg/handlers/session.go
+++ b/pkg/handlers/session.go
@@ -85,9 +85,9 @@ func requireValidSession(kotsStore store.Store, w http.ResponseWriter, r *http.R
 		return nil, err
 	}
 
-	// extend the session by an hour, if user is still active after an hour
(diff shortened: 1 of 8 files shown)


pkg/handlers/session.go
```diff
@@ -75,12 +76,23 @@ func requireValidSession(kotsStore store.Store, w http.ResponseWriter, r *http.R
...
 		response := ErrorResponse{Error: err.Error()}
 		JSON(w, http.StatusUnauthorized, response)
 		return nil, err
 	}
 
+	// extend the session by an hour, if user is still active after an hour
+	if delta := time.Until(sess.ExpiresAt); delta.Hours() <= (SessionTimeout - 1*time.Hour).Hours() {
+		sess.ExpiresAt = sess.ExpiresAt.Add(time.Hour * 1)
+		if err := kotsStore.UpdateSessionExpiresAt(sess.ID, sess.ExpiresAt); err != nil {
+			logger.Error(errors.Wrapf(err, "failed to update session expiry %s", sess.ID))
```
pavansokkenagaraj, line 92: logging the error instead of returning it.
Since it would cause user inconvenience as their session is still valid and active, this may cause a false positive.

pkg/handlers/session.go
```diff
@@ -75,12 +76,23 @@ func requireValidSession(kotsStore store.Store, w http.ResponseWriter, r *http.R
...
+		}
+	}
```
laverya, line 94:
```suggestion
	// give the user the full session timeout if they have been active at least an hour
	if time.Now().Add(SessionTimeout - time.Hour).After(sess.ExpiresAt){
		sess.ExpiresAt = time.Now().Add(SessionTimeout)
		if err := kotsStore.UpdateSessionExpiresAt(sess.ID, sess.ExpiresAt); err != nil {
			logger.Error(errors.Wrapf(err, "failed to update session expiry %s", sess.ID))
		}
	}
```

Given 5 requests in a row from a user with 1 hour remaining, your code would update it to 2 then 3 then 4 then 5 then 6 hours remaining - just setting it to a constant value is both easier to read and more efficient.

laverya (approved):

PR #2714, opened 2022-04-07
remove now-unused ocistore backend

laverya:
As far as I can tell this is unused and not maintained - there are bugfixes in the kots store that aren't present here.

As such, we'd really need to rewrite it before using it - and I don't know it's worth it to keep while waiting for that.

Fixes #

```release-note

```


sgalsaleh (approved):

pkg/kotsadm/objects/kotsadm_objects.go
```diff
@@ -304,16 +301,7 @@ func KotsadmDeployment(deployOptions types.DeployOptions) (*appsv1.Deployment, e
...
-		env = append(env, corev1.EnvVar{
-			Name:  "STORAGE_BASEURI",
-			Value: deployOptions.StorageBaseURI,
-		})
-		env = append(env, corev1.EnvVar{
-			Name:  "STORAGE_BASEURI_PLAINHTTP",
-			Value: strconv.FormatBool(deployOptions.StorageBaseURIPlainHTTP),
-		})
-	} else {
+	if deployOptions.IncludeMinio {
```
sgalsaleh, line 304: kotsadm deployment should always have an object store, let's keep the old logic where this check wasn't there? because it will fail differently

laverya: 🤦 right, the deployment will always need one
it's the statefulset that won't

I'll fix

sgalsaleh (approved):

sgalsaleh (approved):

PR #2719, opened 2022-04-09
fix: propagate automated install errors from kotsadm to cli

cbodonnell:
#### What type of PR is this?

type::bug

Currently, there is nothing that indicates to the user that an automated installation has failed.  This PR introduces changes that will propagate errors encountered during kotsadm automated installations to the user.  Errors during automated installations will now also result in a non-zero exit code.

Fixes [SC-29399](https://app.shortcut.com/replicated/story/29399/admin-console-kots-install-cli-command-does-not-show-any-errors-when-an-automated-install-of-a-license-fails), half of [SC-29729](https://app.shortcut.com/replicated/story/29729/kots-install-license-file-flag-fails-silently), and half of [SC-45886](https://app.shortcut.com/replicated/story/45886/when-installing-from-the-cli-if-preflights-fail-or-have-a-warning-the-application-isn-t-deployed-but-there-is-no-message)

Refactored the automated install logic to remove the need for `continue` statements.  This also allows for propagation of resulting errors.

Currently, the task is left inside the `kotsadm-tasks` ConfigMap regardless of the result.  Might be worth clearing out at some point, but having this info in there seems useful.

```release-note
- Fixes an issue where automated installs using the CLI were not propagating license and preflight errors to the user.
```

NONE


cmd/kots/cli/install.go
```diff
@@ -373,6 +376,12 @@ func InstallCmd() *cobra.Command {
 
 			m.ReportInstallFinish()
 
+			if deployOptions.License != nil {
+				if err := validateAutomatedInstall(deployOptions, clientset, apiEndpoint); err != nil {
+					return errors.Wrap(err, "failed to validate automated install")
+				}
+			}
```
cbodonnell, line 383: Should do this before we report install finish since we may exit non-zero.

cmd/kots/cli/install.go
```diff
@@ -371,6 +377,24 @@ func InstallCmd() *cobra.Command {
...
+				}
+				log.Debug("automated install validated successfully")
+			}
+			if !deployOptions.SkipPreflights {
+				if err := ValidatePreflightStatus(deployOptions, authSlug, apiEndpoint); err != nil {
+					return errors.Wrap(err, "failed to validate preflight status")
+				}
+				log.Debug("preflights validated successfully")
+			}
+
```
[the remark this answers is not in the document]
kcboyle: Just saw that this morning. Fixed and hopefully will pass now. I was just hoping i didn't have to have 2 nested ifs.

cmd/kots/cli/install.go
```diff
@@ -745,3 +777,173 @@ func CheckRBAC() error {
...
+	newReq.Header.Add("Content-Type", "application/json")
+	newReq.Header.Add("Authorization", authSlug)
+
+	resp, err := http.DefaultClient.Do(newReq)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to execute request")
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 {
```
sgalsaleh, line 817:
```suggestion
	if resp.StatusCode != http.StatusOK {
```

kcboyle: updated. spaced using the "commit suggestion" button

kcboyle:
associated PR in the docs: https://github.com/replicatedhq/replicated-docs/pull/322

pkg/automation/automation.go
```diff
@@ -71,174 +78,191 @@ func AutomateInstall() error {
...
+				logger.Error(errors.Wrap(err, "failed to set error on install task status"))
 			}
 		}
+	}()
+
+	verifiedLicense, err := kotspull.VerifySignature(unverifiedLicense)
+	if err != nil {
+		return appSlug, errors.Wrap(err, "failed to verify license signature")
+	}
+	appSlug = verifiedLicense.Spec.AppSlug
```
sgalsaleh, line 146: not needed, let's remove this line. app slug shouldn't ever change for the upstream app

pkg/handlers/app.go
```diff
@@ -578,3 +579,18 @@ func (h *Handler) GetLatestDeployableVersion(w http.ResponseWriter, r *http.Requ
 
 	JSON(w, http.StatusOK, getLatestDeployableVersionResponse)
 }
+
+func (h *Handler) GetAutomatedInstallStatus(w http.ResponseWriter, r *http.Request) {
+	status, msg, err := store.GetStore().GetTaskStatus(fmt.Sprintf("automated-install-slug-%s", mux.Vars(r)["appSlug"]))
+	if err != nil {
+		logger.Error(errors.Wrapf(err, "failed to get install status for app %s", mux.Vars(r)["appSlug"]))
+		w.WriteHeader(500)
```
sgalsaleh, line 587:
```suggestion
		w.WriteHeader(http.StatusInternalServerError)
```

pkg/handlers/app.go
```diff
@@ -578,3 +579,18 @@ func (h *Handler) GetLatestDeployableVersion(w http.ResponseWriter, r *http.Requ
...
+		return
+	}
+	response := kotsstore.TaskStatus{
+		Status:  status,
+		Message: msg,
+	}
+
+	JSON(w, 200, response)
```
sgalsaleh, line 595:
```suggestion
	JSON(w, http.StatusOK, response)
```

cmd/kots/cli/install.go
```diff
@@ -371,6 +383,25 @@ func InstallCmd() *cobra.Command {
 				}
 			}()
 
+			authSlug, err := auth.GetOrCreateAuthSlug(clientset, deployOptions.Namespace)
+			if err != nil {
+				return errors.Wrap(err, "failed to get kotsadm auth slug")
+			}
+
+			if deployOptions.License != nil {
+				if err := ValidateAutomatedInstall(deployOptions, authSlug, apiEndpoint); err != nil {
```
sgalsaleh, line 392: the task status is not being reset at the end, so if the user retries a failed installation for example, there could be a race condition where the old status is read and the install command will fail instead of waiting for the new status.

pkg/handlers/app.go
kcboyle: will be repushed

pkg/handlers/app.go
kcboyle: will be repushed

pkg/automation/automation.go
kcboyle: will be repushed

sgalsaleh (approved):

PR #2722, opened 2022-04-13
update local-volume-provider to v0.3.1

laverya:

Fixes #

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ CURRENT_USER := $(shell id -u -n)
 MINIO_TAG ?= RELEASE.2022-01-08T03-11-54Z
 POSTGRES_ALPINE_TAG ?= 10.19-alpine
 DEX_TAG ?= v2.30.0
-LVP_VERSION := v0.3.1
+LVP_VERSION := v0.3.3
 
 BUILDFLAGS = -tags='netgo containers_image_ostree_stub exclude_graphdriver_devicemapper exclude_graphdriver_btrfs containers_image_openpgp' -installsuffix netgo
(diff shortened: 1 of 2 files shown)


emosbaugh:
Looks like we are using v0.3.0 in kurl https://github.com/replicatedhq/kURL/blob/04a733496c9c3e99793bc538ef4eb7d3cb9d25e2/addons/velero/1.7.1/Manifest#L6

laverya:
v0.3.1 is the same code as v0.3.0, just a newer build

Makefile
```diff
@@ -140,9 +140,6 @@ build-release:
 	mkdir -p bin/docker-archive/local-volume-provider
 	skopeo copy docker://replicated/local-volume-provider:${LVP_VERSION} docker-archive:bin/docker-archive/local-volume-provider/${LVP_VERSION}
 
-	mkdir -p bin/docker-archive/local-volume-fileserver
```
sgalsaleh, line 143: why is this removed? i think this is needed for velero to function (getting backup logs, etc...)

pkg/print/fs-lvp-velero-config.go
```diff
@@ -92,7 +92,7 @@ func printLvpFileSystemVeleroInstructions(c *LvpFileSystemVeleroConfig, log *log
 	  replicated.com/nfs: ObjectStore
 	  replicated.com/hostpath: ObjectStore
 	data:
-	  fileserverImage: %s/local-volume-fileserver:v0.1.0
```
sgalsaleh, line 95: so this is part of the `local-volume-provider` image now?

laverya: it is, yes

sgalsaleh (approved):

pkg/print/fs-lvp-velero-config.go
laverya: Or at least the v0.2.0+ versions of `local-volume-provider` no longer include that image, and the velero plugin in kurl doesn't use it

sgalsaleh (approved):

PR #2730, opened 2022-04-19
add default dynamic nfs client analyzer

cbodonnell:
#### What type of PR is this?

type::feature

This PR adds a default dynamic analyzer to KOTS that will run when snapshots are enabled for the license.  This will help to reduce the support burden for issues where customers cannot configure NFS snapshots because an NFS client package is not installed on the host(s).

**Fail:**
nfs-client-analyzer-fail

**Pass:**
nfs-client-analyzer-pass

Fixes [SC-45145](https://app.shortcut.com/replicated/story/45145/analyzer-missing-nfs-client-package)

This will only produce a failure if snapshots are enabled and the user has tried to configure NFS snapshots without a client package installed.  It could also potentially produce a failure if the user has resolved the issue, but the message we search for still exists in the event logs.

This was added as a default dynamic analyzer so that we can include it only for snapshot-enabled licenses and also target the desired namespace when searching the event logs.

```release-note
Support bundles generated in the "Troubleshoot" tab will now indicate if an NFS client package was missing when configuring snapshots.
```

NONE

### DIFF
--- a/pkg/supportbundle/spec.go
+++ b/pkg/supportbundle/spec.go
@@ -527,6 +527,36 @@ func getDefaultDynamicAnalyzers(app *apptypes.App) []*troubleshootv1beta2.Analyz
 	analyzers := make([]*troubleshootv1beta2.Analyze, 0)
 	analyzers = append(analyzers, makeAPIReplicaAnalyzer())
 
+	license, err := store.GetStore().GetLicenseForAppVersion(app.ID, app.CurrentSequence)
+	if err != nil {
+		logger.Errorf("Failed to get license for dynamic analyzers for app id %s sequence %d: %v", app.ID, app.CurrentSequence, err)
+	} else if license.Spec.IsSnapshotSupported {
+		analyzers = append(analyzers, &troubleshootv1beta2.Analyze{
+			TextAnalyze: &troubleshootv1beta2.TextAnalyze{
+				AnalyzeMeta: troubleshootv1beta2.AnalyzeMeta{
+					CheckName: "NFS Client Package",
+				},
+				FileName:     fmt.Sprintf("cluster-resources/events/%s.json", util.PodNamespace),
+				RegexPattern: "bad option; for several filesystems \\(e\\.g\\. nfs, cifs\\) you might need a \\/sbin\\/mount\\..+type.+ helper program\\.",
+				Outcomes: []*troubleshootv1beta2.Outcome{
+					{
+						Fail: &troubleshootv1beta2.SingleOutcome{
+							When:    "true",
+							Message: "NFS client errors were found. Refer to [documentation](https://docs.replicated.com/enterprise/snapshots-configuring-nfs) on how to configure NFS snapshots.",
+							URI:     "https://docs.replicated.com/enterprise/snapshots-configuring-nfs",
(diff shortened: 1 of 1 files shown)


pkg/supportbundle/spec.go
```diff
@@ -527,6 +527,36 @@ func getDefaultDynamicAnalyzers(app *apptypes.App) []*troubleshootv1beta2.Analyz
 	analyzers := make([]*troubleshootv1beta2.Analyze, 0)
 	analyzers = append(analyzers, makeAPIReplicaAnalyzer())
 
+	license, err := store.GetStore().GetLicenseForAppVersion(app.ID, app.CurrentSequence)
```
sgalsaleh, line 530:
```suggestion
	license, err := store.GetStore().GetLatestLicenseForApp(app.ID)
```

pkg/supportbundle/spec.go
```diff
@@ -527,6 +527,36 @@ func getDefaultDynamicAnalyzers(app *apptypes.App) []*troubleshootv1beta2.Analyz
...
+	if err != nil {
+		logger.Errorf("Failed to get license for dynamic analyzers for app id %s sequence %d: %v", app.ID, app.CurrentSequence, err)
```
sgalsaleh, line 532:
```suggestion
		logger.Errorf("Failed to get latest license for app %s: %v", app.Slug, err)
```

pkg/supportbundle/spec.go
```diff
@@ -527,6 +527,36 @@ func getDefaultDynamicAnalyzers(app *apptypes.App) []*troubleshootv1beta2.Analyz
...
+			TextAnalyze: &troubleshootv1beta2.TextAnalyze{
+				AnalyzeMeta: troubleshootv1beta2.AnalyzeMeta{
+					CheckName: "NFS Client Package",
+				},
+				FileName:     fmt.Sprintf("cluster-resources/events/%s.json", util.PodNamespace),
+				RegexPattern: "bad option; for several filesystems \\(e\\.g\\. nfs, cifs\\) you might need a \\/sbin\\/mount\\..+type.+ helper program\\.",
+				Outcomes: []*troubleshootv1beta2.Outcome{
+					{
+						Fail: &troubleshootv1beta2.SingleOutcome{
+							When:    "true",
```
sgalsaleh, line 544: i don't think this is needed

pkg/supportbundle/spec.go
```diff
@@ -527,6 +527,36 @@ func getDefaultDynamicAnalyzers(app *apptypes.App) []*troubleshootv1beta2.Analyz
...
+							Message: "NFS client errors were found. Refer to [documentation](https://docs.replicated.com/enterprise/snapshots-configuring-nfs) on how to configure NFS snapshots.",
+							URI:     "https://docs.replicated.com/enterprise/snapshots-configuring-nfs",
+						},
+					},
+					{
+						Pass: &troubleshootv1beta2.SingleOutcome{
+							When:    "false",
```
sgalsaleh, line 551: this means that a success outcome won't ever show up, no?

sgalsaleh (approved):

divolgin (approved):

PR #2733, opened 2022-04-21
ignore preflights and deploy application

stefanrepl:
#### What type of PR is this?
type::feature

Deploy application when a user chooses to ignore preflights and deploy.  We need this because the previous flow of choosing ignore preflights and continue was a confusing for some users as they expected their app to deploy when choosing this option.

Fixes #
https://app.shortcut.com/replicated/story/36852/skip-preflight-checks-on-initial-install-does-not-install-app

Ran against nightly regression tests.  Tested deploy functionality in okteto.

```release-note
     Automate deployment of application when preflights are ignored.
```

### DIFF
--- a/web/src/components/shared/modals/SkipPreflightsModal.jsx
+++ b/web/src/components/shared/modals/SkipPreflightsModal.jsx
@@ -2,7 +2,7 @@ import React from "react";
 import Modal from "react-modal";
 
 export default function SkipPreflightsModal(props) {
-  const { showSkipModal, hideSkipModal, onForceDeployClick, sendPreflightsReport, appsList } = props;
+  const { showSkipModal, hideSkipModal, deployKotsDownstream, onForceDeployClick } = props;
 
   return (
     <Modal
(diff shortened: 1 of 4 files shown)


web/src/components/shared/modals/SkipPreflightsModal.jsx
```diff
@@ -22,9 +22,9 @@ export default function SkipPreflightsModal(props) {
           <div className="u-marginTop--30 flex flex-column">
             <button type="button" className="btn blue primary" onClick={hideSkipModal}>Wait for Preflights to finish</button>
             {onForceDeployClick ?
-              <span className="u-fontSize--normal u-fontWeight--medium u-textDecoration--underline u-textColor--bodyCopy u-marginTop--15 u-cursor--pointer" onClick={() => onForceDeployClick(false)}>Ignore Preflights and deploy</span>
+              <span className="u-fontSize--normal u-fontWeight--medium u-textDecoration--underline u-textColor--bodyCopy u-marginTop--15 u-cursor--pointer" onClick={() => onForceDeployClick(true)}>Ignore Preflights and deploy</span>
```
sgalsaleh, line 25: why was this changed to `true`? it looks like the `onForceDeployClick` takes a `continueWithFailedPreflights` and in this case they didn't fail but rather skipped, right?

stefanrepl: correct, thanks, updated

sgalsaleh (approved):

PR #2739, opened 2022-04-21
add an API to allow changing password (if the current password is known) while logged into kotsadm

pavansokkenagaraj:
#### What type of PR is this?

type::feature

**add an API to allow changing password  (if the current password is known) while logged into kotsadm**

- add a new `PUT /api/v1/password/change` endpoint to change kots shared password
- request takes current and new password, checks if current password matches, current pwd != new pwd and new pwd length greater than 6.
- once the password is created, deletes all the sessions, so pwd change is enforced
- added the delete all sessions to kots cli cmd `kubectl kots reset-password`
- added a new pacakge `pkg/password` which is being used by cli and api for password change code.
- add a extra check on session validation: if password was updated after session was issued, session will be invalidated and try to delete and log out the user with appropriate message.
- add a new `passwordUpdatedAt` key/value to `kotsadm-password` to record when password was changed (addresses backward compatibility for existing secrets/newly created secrets by using secret.CreatedAt )

Fixes # [SC-44273](https://app.shortcut.com/replicated/story/44273/method-to-change-password-in-kots-admin-console-ui#activity-46631)

```release-note
* Fixes an issue where `kubectl kots reset-password` will force the password change by deleting all the user sessions.
* Add an API to allow changing password (if the current password is known) while logged into kotsadm
```

NONE
(since we don't document API changes, password reset documentation will be part of UI changes)


pkg/store/kotsstore/user_store.go
```diff
@@ -231,3 +232,64 @@ func (s *KOTSStore) flagSuccessfulLoginInDatabase() error {
 
 	return nil
 }
+
+// SetSharedPasswordBcrypt - set the shared password bcrypt hash in the kotsadm secret
+func (s *KOTSStore) SetSharedPasswordBcrypt(bcryptPassword []byte) error {
```
laverya, line 237: is there not already a function for this, called by the CLI?

Can we use that instead?

Further - do we ever actually use a password from sqlite?

cmd/kots/cli/reset-password.go
```diff
@@ -146,3 +151,29 @@ func setKotsadmPassword(password string, namespace string) error {
...
 }
+
+// delete all sessions from the kotsadm-sessions secret
+func deleteAllSessions(namespace string) error {
+	clientset, err := k8sutil.GetClientset()
+	if err != nil {
+		return errors.Wrap(err, "failed to create k8s client")
+	}
+
+	existingSecret, err := clientset.CoreV1().Secrets(namespace).Get(context.TODO(), "kotsadm-sessions", metav1.GetOptions{})
```
emosbaugh, line 162: why do we need to get the secret at all here?

pkg/handlers/password_change.go
```diff
@@ -0,0 +1,74 @@
...
+		return
+	}
+	if identityConfig.Spec.Enabled && identityConfig.Spec.DisablePasswordAuth {
+		err := errors.New("password authentication disabled")
+		JSON(w, http.StatusForbidden, NewErrorResponse(err))
+		return
+	}
+
+	var passwordChangeRequest PasswordChangeRequest
+	if err := json.NewDecoder(r.Body).Decode(&passwordChangeRequest); err != nil {
```
emosbaugh, line 42: can we move lines 41-51 before `identity.GetConfig` as an optimization?

pkg/policy/policies.go
```diff
@@ -57,6 +57,12 @@ var (
 	PrometheussettingsWrite = Must(NewPolicy(ActionWrite, "prometheussettings."))
 )
 
+// Password change
+
+var (
+	PasswordChange = Must(NewPolicy(ActionWrite, "passwordchange."))
```
emosbaugh, line 63: PasswordWrite or PasswordUpdate for consistency?

pkg/store/kotsstore/session_store.go
```diff
@@ -333,3 +333,21 @@ func (s *KOTSStore) DeleteExpiredSessions() error {
 
 	return nil
 }
+
+// DeleteAllSessions - delete all sessions from kotsadm-sessions secret
+func (s *KOTSStore) DeleteAllSessions() error {
+	sessionLock.Lock()
+	defer sessionLock.Unlock()
+
+	secret, err := s.getSessionSecret()
```
emosbaugh, line 342: why do we need to get the secret at all here?

pkg/handlers/password_change.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	if err := user.ValidatePasswordInput(passwordChangeRequest.CurrentPassword, passwordChangeRequest.NewPassword); err != nil {
+		JSON(w, http.StatusBadRequest, NewErrorResponse(err))
+		return
+	}
+
+	// change password
+	if err := user.ChangePassword(store.GetStore(), passwordChangeRequest.CurrentPassword, passwordChangeRequest.NewPassword); err != nil {
+		if err == user.ErrInvalidPassword {
```
emosbaugh, line 55: should you change this to the following in case a change is made to the called function?

```
if errors.Is(err, user.ErrInvalidPassword) {
```

cmd/kots/cli/reset-password.go
```diff
@@ -146,3 +151,29 @@ func setKotsadmPassword(password string, namespace string) error {
 
 	return nil
 }
+
+// delete all sessions from the kotsadm-sessions secret
+func deleteAllSessions(namespace string) error {
```
emosbaugh, line 156: could we also share this code?

emosbaugh (changes requested):
I do wonder if we could share some more code with the cli as well as andrew pointed out

cmd/kots/cli/reset-password.go
pavansokkenagaraj: never knew there was a `func DeleteCollection` , should have checked the interface.

cmd/kots/cli/reset-password.go
```diff
@@ -56,6 +56,11 @@ func ResetPasswordCmd() *cobra.Command {
 				return errors.Wrap(err, "failed to set new password")
 			}
 
+			// delete all sessions to force password change
+			if err := deleteAllSessions(namespace); err != nil {
```
trelore, line 60: From a security perspective (I also lack full context, so feel free to point that out):

What happens if `deleteAllSessions` errors here?
- The user gets a exit code `0`.
- The user gets a log statement saying it didn't delete all sessions (I can guarantee someone isn't going to read it if they get `0`).
- A compromised session may still exist.

Other possibly alternatives (in no particular order);
- Error here
  - Compromised sessions may still exist
  - Rerunning the command may produce the same error (though hopefully it's a transient error)
- Put in some kind of retry logic (definitely a route I'd avoid unless we have to/have a pattern for it)
  - This way transient errors are hopefully caught
- Some kind of db style transaction and roll back the password
  - Feels very ugly, and hard to implement if there's no precedent for it
- Make a manual delete sessions CLI command to ensure if this step fails, the user can try just the last bit (and error out here to highlight they have work to do)
  - Adds an edge case for users, but allows them to have more granular control over their password management.
 
In summary: Maybe line 61 needs to be an error to bring attention to the user that they _may_ still have something to do

cmd/kots/cli/reset-password.go
trelore: Out of curiousity, what happens if we delete the secret, does it expect a secret to exist to append to, or will the first session create the secret?

pkg/handlers/session.go
```diff
@@ -85,6 +85,20 @@ func requireValidSession(kotsStore store.Store, w http.ResponseWriter, r *http.R
...
+	if passwordUpdatedAt.After(sess.IssuedAt) {
+		if err := kotsStore.DeleteSession(sess.ID); err != nil {
+			logger.Error(errors.Wrapf(err, "password was updated after session created. failed to delete invalid session %s", sess.ID))
+		}
+		err := errors.New("password changed, please login again")
+		response := ErrorResponse{Error: err.Error()}
+		JSON(w, http.StatusUnauthorized, response)
+		return nil, err
+	}
+
```
laverya, line 101: isn't this redundant if the sessions are being deleted?

pavansokkenagaraj: if delete sessions fail, this would be a good fallback option 
 additionally, addresses @trelore [comment](https://github.com/replicatedhq/kots/pull/2739#discussion_r856173161)

cmd/kots/cli/reset-password.go
pavansokkenagaraj: @trelore if `kotsadm-secret` gets deleted, kots will fallback to environment password value `SHARED_PASSWORD_BCRYPT` 
when someone changes the secret, the new secret get's created

cmd/kots/cli/reset-password.go
pavansokkenagaraj: @trelore I have added a session check, if a session was created before the passwordUpdatedAt, we invalidate session,  try deleting again and log the user out.
it should address the above comment
https://github.com/replicatedhq/kots/pull/2739#discussion_r856450037

pkg/store/kotsstore/user_store.go
pavansokkenagaraj: I was bit hesitant to refactor existing code, added a new package `password` which is being used by cli as well as api

pkg/store/kotsstore/user_store.go
```diff
@@ -231,3 +231,35 @@ func (s *KOTSStore) flagSuccessfulLoginInDatabase() error {
...
+	var passwordUpdatedAt time.Time
+	updatedAtBytes, ok := passwordSecret.Data["passwordUpdatedAt"]
+	if ok {
+		updatedAt, err := time.Parse(time.RFC3339, string(updatedAtBytes))
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to parse passwordUpdatedAt")
+		}
+		passwordUpdatedAt = updatedAt
+	} else {
+		// backward compatibility, for older secret/ newly created secert with no passwordUpdatedAt value
```
laverya, line 260:
```suggestion
		// backward compatibility, for older secret/ newly created secret with no passwordUpdatedAt value
```

laverya (approved):

pkg/password/password_test.go
```diff
@@ -0,0 +1,309 @@
...
+			name: "expect no error when existing secret updated successfully",
+			args: args{
+				clientset:      newMockClientForExistingSecretUpdateSuccess(),
+				namespace:      "test",
+				bcryptPassword: []byte("password"),
+			},
+			wantErr: false,
+		},
+		{
+			name: "expect error when existing secret updated successfully and sessions delete err",
```
laverya, line 293:
```suggestion
			name: "expect no error when existing secret updated successfully and sessions delete err",
```

laverya (dismissed):

emosbaugh (approved):

pkg/password/password.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+
+		return errors.Wrap(err, "failed to compare current password")
+	}
+
+	return nil
+}
+
+// ChangePassword - will change the password in the kotsadm secret
+func ChangePassword(clientset *kubernetes.Clientset, namespace string, newPassword string) error {
```
sgalsaleh, line 55: needs a mutex

pkg/password/password.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	if newPassword == currentPassword {
+		return ErrNewPasswordShouldBeDifferent
+	}
+	return nil
+}
+
+// ValidateCurrentPassword - will compare the password with the stored password and return an error if they don't match
+func ValidateCurrentPassword(kotsStore store.Store, currentPassword string) error {
```
sgalsaleh, line 37: needs a mutex

pkg/password/password.go
```diff
@@ -0,0 +1,133 @@
...
+
+	if err := setSharedPasswordBcrypt(clientset, namespace, shaBytes); err != nil {
+		return errors.Wrap(err, "failed to set new shared password bcrypt")
+	}
+
+	return nil
+}
+
+// setSharedPasswordBcrypt - set the shared password bcrypt hash in the kotsadm secret
+func setSharedPasswordBcrypt(clientset kubernetes.Interface, namespace string, bcryptPassword []byte) error {
```
sgalsaleh, line 69: needs a mutex

pkg/password/password.go
laverya: I personally looked at this and thought the same thing at first, but this can be called from both the CLI and the server itself so IMO a mutex is unlikely to help prevent races

Can't hurt though

emosbaugh: lgtm although its possible to use resource version in k8s object as a mutex

somthing like validateAndSetSharedPassword

emosbaugh: lgtm as i dont think a mutex is really needed but maybe there is some way to use resource version mismatch as a mutex?

sgalsaleh (approved):

emosbaugh (approved):

pkg/password/password.go
pavansokkenagaraj: resource reconciliation would be one good option, but that one is complicated.

pavansokkenagaraj:
```bash
➜  curl --location --request PUT 'https://kotsadm-pavansokkenagaraj.replicated.okteto.dev/api/v1/password/change' \
--header 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzZXNzaW9uSWQiOiIyOElnSmlOQlhyQVlRcEVkdkVvNEM1M3dMcHQifQ.z-lJxZ0Y_jE9TJ_xdV-P-a4BA41PcwyejZmobUSwBFs' \
--header 'Content-Type: application/json' \
--data-raw '{
	"current_password": "passowrds",
    "new_password": "passwords"
}'
```
request
```json
{
    "current_password": "passwords",
    "new_password": "password"
}

```
response: 400 BadRequest
```json
{
    "error": "current password does not match",
    "success": false
}
```
request
```json
{
    "current_password": "password",
    "new_password": "test"
}
```
response: 400 BadRequest
```json
{
    "error": "new password must be at least 6 characters",
    "success": false
}
```
request
```json
{
    "current_password": "password",
    "new_password": "password"
}

```
response: 400 BadRequest
```json
{
    "error": "new password should be different from current password",
    "success": false
}
```
request
```json
{
    "current_password": "password",
    "new_password": "kotsadm"
}

```
response: 200 Ok
```json
{
    "success": true
}
```

PR #2740, opened 2022-04-22
downloadable file in kots config

stefanrepl:
#### What type of PR is this?
type::feature

This PR allows for a user to download their previously uploaded config file for a given app sequence.

Fixes #

```release-note
     Add ability to download Config file types for a given application sequence.
```


pkg/handlers/config.go
```diff
@@ -64,6 +65,86 @@ type CurrentAppConfigResponse struct {
...
+	}
+
+	fileForDownload, err := os.Open(fmt.Sprintf("%s/%s", archiveDir, filename))
+	if err != nil {
+		logger.Error(err)
+		downloadAppConfigResponse.Error = "failed to open temp config file"
+		JSON(w, http.StatusInternalServerError, downloadAppConfigResponse)
+	}
+
+	fileStat, _ := fileForDownload.Stat()
```
divolgin, line 140: this error should be handled.  `fileStat` will be nil  on error and possibly cause a panic.

pkg/handlers/config.go
```diff
@@ -433,6 +514,26 @@ func shouldCreateNewAppVersion(appID string, sequence int64) (bool, error) {
 	return true, nil
 }
 
+func getAppConfigValue(downloadApp *apptypes.App, sequence int64, filename string, archiveDir string) (string, error) {
```
divolgin, line 517: how about naming this functions something more specific since it only works for file values, like `getAppConfigValueForFile`?

pkg/handlers/config.go
```diff
@@ -433,6 +514,26 @@ func shouldCreateNewAppVersion(appID string, sequence int64) (bool, error) {
...
+	err := store.GetStore().GetAppVersionArchive(downloadApp.ID, sequence, archiveDir)
+	if err != nil {
+		return "", err
```
divolgin, line 520: errors.Wrap

pkg/handlers/config.go
```diff
@@ -433,6 +514,26 @@ func shouldCreateNewAppVersion(appID string, sequence int64) (bool, error) {
...
+	}
+
+	kotsKinds, err := kotsutil.LoadKotsKindsFromPath(archiveDir)
+	if err != nil {
+		return "", err
```
divolgin, line 525: errors.Wrap

pkg/handlers/config.go
```diff
@@ -64,6 +65,86 @@ type CurrentAppConfigResponse struct {
...
+
+	decoded, err := base64.StdEncoding.DecodeString(configValue)
+	if err != nil {
+		logger.Error(err)
+		downloadAppConfigResponse.Error = "failed to decode config value"
+		JSON(w, http.StatusInternalServerError, downloadAppConfigResponse)
+	}
+
+	// create file in temp directory to use for io.Copy
+	err = os.WriteFile(fmt.Sprintf("%s/%s", archiveDir, filename), decoded, 0666)
```
divolgin, line 126: This is not needed.  `reader := strings. NewReader(decoded)` can be used without intermediate files

pkg/handlers/config.go
```diff
@@ -64,6 +65,86 @@ type CurrentAppConfigResponse struct {
...
+		logger.Error(err)
+		downloadAppConfigResponse.Error = "failed to open temp config file"
+		JSON(w, http.StatusInternalServerError, downloadAppConfigResponse)
+	}
+
+	fileStat, _ := fileForDownload.Stat()
+	fileSize := strconv.FormatInt(fileStat.Size(), 10)
+	w.Header().Set("Content-Type", "application/octet-stream")
+	w.Header().Set("Content-Length", fileSize)
+	io.Copy(w, fileForDownload)
```
divolgin, line 144: this error should be handled and logged.

pkg/handlers/config.go
```diff
@@ -64,6 +65,86 @@ type CurrentAppConfigResponse struct {
...
+
+	configValue, err := getAppConfigValue(foundApp, int64(sequence), filename, archiveDir)
+	if err != nil {
+		logger.Error(err)
+		downloadAppConfigResponse.Error = "failed to get app config"
+		JSON(w, http.StatusInternalServerError, downloadAppConfigResponse)
+		return
+	}
+
+	decoded, err := base64.StdEncoding.DecodeString(configValue)
```
divolgin, line 118: also consider using `Decode` instead of `DecodeString` since files can be binary and that would be cleaner and also would match `Content-Type` header better set at the end of this function.

web/src/components/AppConfigRenderer.jsx
```diff
@@ -35,6 +35,8 @@ export class AppConfigRenderer extends React.Component {
           handleChange={this.props.handleChange || (() => { return; })}
           getData={this.props.getData || (() => { return; })}
           readonly={readonly}
+          configSequence={this.props.configSequence}
```
rianfowler, line 38: Can you elaborate on what `configSequence` is and how it's used?

rianfowler (commented):
Added some questions / suggestions.

pkg/handlers/config.go
stefanrepl: addressed

pkg/handlers/config.go
stefanrepl: addressed

pkg/handlers/config.go
stefanrepl: addressed

pkg/handlers/config.go
stefanrepl: ahh great suggestion, thanks! addressed

pkg/handlers/config.go
stefanrepl: addressed

pkg/handlers/config.go
stefanrepl: addressed

web/src/components/config_render/ConfigFileInput.jsx
```diff
@@ -21,6 +22,32 @@ export default class ConfigFileInput extends React.Component {
...
+      fetch(url, {
+        method: "GET",
+        headers: {
+          "Content-Type": "application/octet-stream",
+          "Authorization": Utilities.getToken(),
+        }
+      }).then((response) => {
+        return response.blob()
+      }).then((blob) => {
+        const url2 = window.URL.createObjectURL(
```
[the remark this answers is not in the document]
stefanrepl: url2 was temp url that is used for downloading the file.  I renamed the url to reflect its use better

web/src/components/AppConfigRenderer.jsx
GraysonNull: configSequence is the identifier for a given config. When a customer goes to download a file we need to send the sequence number so we know which config to pull the file out of. If you look in the URL of a config page you'll see the sequence in there so we pull it out of the url and send it along through props to where the api request is made.

rianfowler: gotcha thanks

rianfowler (dismissed):

pkg/handlers/config.go
```diff
@@ -64,6 +65,77 @@ type CurrentAppConfigResponse struct {
...
+
+	decoded := make([]byte, 0)
+	_, err = base64.StdEncoding.Decode(decoded, configValue)
+	if err != nil {
+		logger.Error(err)
+		downloadAppConfigResponse.Error = "failed to decode config value"
+		JSON(w, http.StatusInternalServerError, downloadAppConfigResponse)
+	}
+
+	reader := strings.NewReader(string(decoded))
```
divolgin, line 126: `bytes.NewReader(decoded)` to match new type

pkg/handlers/config.go
```diff
@@ -64,6 +66,77 @@ type CurrentAppConfigResponse struct {
...
+	w.Header().Set("Content-Type", "application/octet-stream")
+	_, err = io.Copy(w, reader)
+	if err != nil {
+		logger.Error(err)
+		downloadAppConfigResponse.Error = "failed to copy file to client"
+		JSON(w, http.StatusBadRequest, downloadAppConfigResponse)
+		return
+	}
+
+	JSON(w, http.StatusOK, DownloadAppConfigResponse{Success: true})
```
divolgin, line 137: this should not be called here, and io.Copy should not be used before headers are written.  this function should call these directly
```
	w.WriteHeader(http.StatusOK)
	w.Write(decoded)
```

pkg/handlers/config.go
```diff
@@ -64,6 +64,69 @@ type CurrentAppConfigResponse struct {
...
+
+	configValue, err := getAppConfigValueForFile(foundApp, int64(sequence), filename, archiveDir)
+	if err != nil {
+		logger.Error(err)
+		downloadAppConfigResponse.Error = "failed to get app config"
+		JSON(w, http.StatusInternalServerError, downloadAppConfigResponse)
+		return
+	}
+
+	decoded := make([]byte, base64.StdEncoding.DecodedLen(len(string(configValue))))
```
divolgin, line 117: `len` works on any "array" type: `len(configValue)`

stefanrepl: ofcourse

web/src/components/config_render/ConfigFileInput.jsx
```diff
@@ -21,6 +22,37 @@ export default class ConfigFileInput extends React.Component {
     }
   }
 
+  handleDownloadFile = async (fileName) => {
+    const url = `${process.env.API_ENDPOINT}/app/${this.props.appSlug}/config/${this.props.configSequence}/${fileName}/download`;
+      fetch(url, {
```
sgalsaleh, line 27: indentation here is wrong

pkg/handlers/config.go
```diff
@@ -64,6 +64,69 @@ type CurrentAppConfigResponse struct {
 	ConfigGroups []kotsv1beta1.ConfigGroup `json:"configGroups"`
 }
 
+type DownloadAppConfigResponse struct {
+	Success bool   `json:"success"`
+	Error   string `json:"error,omitempty"`
+}
+
+func (h *Handler) DownloadFileFromConfig(w http.ResponseWriter, r *http.Request) {
```
sgalsaleh, line 72: since this was renamed, i think you'll need to run `make mock` too.

pkg/handlers/config.go
```diff
@@ -64,6 +64,69 @@ type CurrentAppConfigResponse struct {
 	ConfigGroups []kotsv1beta1.ConfigGroup `json:"configGroups"`
 }
 
+type DownloadAppConfigResponse struct {
```
sgalsaleh, line 67: let's rename the response to match the function name? `DownloadFileFromConfigResponse`

pkg/handlers/config.go
stefanrepl: yup, done

sgalsaleh (approved):

PR #2741, opened 2022-04-22
Images garbage collection should be limited to the application's registry namespace

sgalsaleh:
#### What type of PR is this?

type::bug

Limits images garbage collection to the application's registry namespace

Fixes [SC-46748](https://app.shortcut.com/replicated/story/46748/images-garbage-collection-should-be-limited-to-the-application-s-registry-namespace)

```release-note
Fixes an issue that causes images garbage collection in embedded clusters to remove images outside of the application's dedicated registry namespace.
```

NONE

### DIFF
--- a/pkg/registry/images.go
+++ b/pkg/registry/images.go
@@ -214,17 +214,18 @@ func deleteUnusedImages(ctx context.Context, registry types.RegistrySettings, us
 
 	digestsInRegistry := map[string]string{}
 	for _, r := range searchResult {
-		imageName := path.Join(registry.Hostname, r.Name)
+		parts := strings.Split(r.Name, "/")
 
-		parts := strings.Split(imageName, "/")
-		if len(parts) < 2 {
-			continue
+		registryNamespace := ""
+		if len(parts) > 1 {
+			// e.g.: my/namespace/imagename => my/namespace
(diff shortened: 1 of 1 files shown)


pkg/registry/images.go
```diff
@@ -215,6 +215,17 @@ func deleteUnusedImages(ctx context.Context, registry types.RegistrySettings, us
 	digestsInRegistry := map[string]string{}
 	for _, r := range searchResult {
 		imageName := path.Join(registry.Hostname, r.Name)
+
+		parts := strings.Split(imageName, "/")
```
emosbaugh, line 219: Why not use `reference.Path(ref.DockerReference())`?

emosbaugh (approved):
This looks correct so i will approve but why not use docker libs they provide?

pkg/registry/images.go
sgalsaleh: cuz it's not parsed as a reference yet at that point. i guess i could, but we do use this pattern in our codebase

emosbaugh:
It is parsed on the next line

Sent via Superhuman iOS ***@***.***>

emosbaugh (approved):

emosbaugh (approved):

PR #2744, opened 2022-04-25
fix broken monitoring graph legend, fix broken date templating to show localized date instead of LLL

GraysonNull:
#### What type of PR is this?
type::bug

- Fixes an issue where the legend on the dashboard graphs were blank
- Fixes an issue when hovering on a graph the tooltip showed "LLL" instead of a formatted date

Fixes https://github.com/replicated-collab/mandiant-replicated/issues/64, Fixes https://github.com/replicated-collab/julia-kots/issues/48

Screen Shot 2022-04-25 at 10 36 20 AM

Screen Shot 2022-04-25 at 10 28 09 AM

```release-note
- Fixes an issue where the legend on the dashboard graphs were blank
- Fixes an issue when hovering on a graph the tooltip showed "LLL" instead of a formatted date
```

NONE

### DIFF
--- a/web/webpack.config.js
+++ b/web/webpack.config.js
@@ -34,7 +34,6 @@ module.exports = function (env) {
       publicPath: "/",
       filename: "[name].[fullhash].js"
     },
-
     resolve: {
       extensions: [".js", ".mjs", ".jsx", ".css", ".scss", ".png", ".jpg", ".svg", ".ico"],
       fallback: {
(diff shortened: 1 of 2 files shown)


web/webpack.config.js
```diff
@@ -48,7 +47,8 @@ module.exports = function (env) {
         tty: require.resolve("tty-browserify")
       },
       alias: {
-        "@src": path.resolve(__dirname, "src")
+        "@src": path.resolve(__dirname, "src"),
+        "handlebars" : "handlebars/dist/handlebars.js"
```
rianfowler, line 51: My understanding is that this is forcing `from "handlebars"` to resolve to the module's `/dist/handlebars.js`

rianfowler (approved):
Verified this is working for me. 
Screen Shot 2022-04-25 at 12 49 49

web/webpack.config.js
GraysonNull: I believe so yeah, was the only thing that could get it to work...for the life of my everything else I was trying RE importing it in the component either wouldn't resolve it or would have compiler errors.

PR #2746, opened 2022-04-25
add a template function that escapes arbitary strings for inclusion in yaml

laverya:

The strings produced by template functions like [ConfigOptionData](https://docs.replicated.com/reference/template-functions-config-context#configoptiondata) will sometimes include characters that the YAML spec disallows. If the value is to be included within a yaml file (for instance, a configmap or secret), it needs to have these characters escaped.

This PR provides a function to turn this unescaped series of characters into a properly quoted and escaped yaml string. A demonstration of the input and output strings can be seen [here](https://goplay.tools/snippet/wfDddBBYJIm).

Fixes #

```release-note
add template function `YamlEscape` to escape a string for inclusion within a yaml file
```

This PR will need an addition on https://docs.replicated.com/reference/template-functions-static-context.


jala-dx (dismissed):

pkg/template/static_context.go
```diff
@@ -588,3 +590,20 @@ func (ctx StaticCtx) noProxy() string {
 func (ctx StaticCtx) kotsVersion() string {
 	return strings.TrimPrefix(buildversion.Version(), "v")
 }
+
+func (ctx StaticCtx) yamlEscape(plain string) string {
```
jala-dx, line 594: Is there a reason you have in mind to use yamlEscape template function instead of using !!binary on the base64encoded data ?

laverya: Because !! binary is different than base64

jala-dx (dismissed):

pkg/template/static_context.go
laverya: or at least I thought it was different than base64 - https://yaml.org/type/binary.html

now I feel silly!

laverya: it seems that that doesn't work great with template functions, so this is still worthwhile 😄

garcialuis (approved):

pkg/template/static_context.go
jala-dx: soo andrew that answer is what i came up with too :slightly_smiling_face: this works
```
trustStore: !!binary |
          abc=
```
but not `when abc= is replaced with repl{{ ConfigOptionData "custom_trust_store" }}` hence something like this `repl{{ ConfigOptionData "custom_trust_store" | binary }}` would have been ideal, instead of yamlEscape :slightly_smiling_face: but the above is still worthwhile

PR #2749, opened 2022-04-26
Re-download newer versions if an older one was updated

divolgin:
#### What type of PR is this?

type::bug

This ensures that application upgrades are re-downloaded and new configuration is applied when an older version is updated with new configuration.

Fixes #

To create this dataset, 45 was deployed, 46 was downloaded, then 45 was updated.

```
# kotsadm=# SELECT sequence, update_cursor, version_label, is_required FROM app_version;
 sequence | update_cursor | version_label | is_required 
----------+---------------+---------------+-------------
        0 | 45            | 0.3.1         | f
        1 | 46            | 0.4.2         | f
        2 | 45            | 0.3.1         | f
        3 | 45            | 0.3.1         | f
(4 rows)
```

Before the change in the PR, the function would return the _latest known_ cursor, which is 46, which would cause replicated.app API to return no updates.

With this change, the function now returns the _currently deployed_ cursor, which is 45, which allows replicated.app return 46 as the update.

```release-note
* Fixes an issue where a newer version may not have a Deploy button after configuration for currently deployed version is updated.
```

NONE

### DIFF
--- a/pkg/store/kotsstore/version_store.go
+++ b/pkg/store/kotsstore/version_store.go
@@ -924,8 +924,8 @@ func (s *KOTSStore) getNextAppSequence(db queryable, appID string) (int64, error
 
 func (s *KOTSStore) GetCurrentUpdateCursor(appID string, channelID string) (string, string, bool, error) {
 	db := persistence.MustGetDBSession()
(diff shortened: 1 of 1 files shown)


sgalsaleh (dismissed):

jala-dx (approved):

jala-dx:
since it is already reviewed by salah, i just gave 👍

PR #2755, opened 2022-04-26
feature: add change password modal to footer

alicenstar:
#### What type of PR is this?

type::feature

This PR adds a 'Change Password' button to the footer in KOTS, which opens a modal so the user can change the login password. In this modal, the user must enter their current password as well as their new password (and confirm it). There is a mix of client and server side validation that runs, and if an error occurs it will show the error message next to the Change Password button. Each input field also has a show/hide toggle so the user can see what they are typing.

Fixes [SC-44273](https://app.shortcut.com/replicated/story/44273/method-to-change-password-in-kots-admin-console-ui)

None

```release-note
Adds the ability to change your password. Adds a 'Change Password' button to the page footer.
```

Yes

#### Demo

https://www.loom.com/share/fc10d8422cf84603ad380031fa695abb


web/src/components/modals/ChangePasswordModal/ChangePasswordModal.jsx
```diff
@@ -0,0 +1,106 @@
...
+            throw new Error(`Unexpected status code ${response.status}: ${res.error}`);
+          }
+          throw new Error(`Unexpected status code ${response.status}`);
+        }
+  
+        const loginInfo = await response.json();
+        if (loginInfo?.method === "identity-service") {
+          setIdentityServiceEnabled(true);
+        }
+        setIdentityServiceEnabled(false);
```
rianfowler, line 42: the way i read this, if they they're using an identity-service, this sets IdentityServiceEnabled to true and then sets it to false.

web/src/components/modals/ChangePasswordModal/ChangePasswordModal.jsx
```diff
@@ -0,0 +1,106 @@
+import React from "react";
+import Modal from "react-modal";
+import Warning from "../../shared/Warning";
+import ChangePasswordForm from "./ChangePasswordForm";
+import { Link } from "react-router-dom";
+
+import "@src/scss/components/modals/ChangePasswordModal/ChangePasswordModal.scss";
+
+const ChangePasswordModal = ({ closeModal, isOpen }) => {
+  const [isSuccessful, setIsSuccessful] = React.useState(false);
```
rianfowler, line 10: maybe something like `passwordChangeSuccessful` or `passwordChangeComplete`

rianfowler (dismissed):
Looks good to me. Had one question and an a suggestion

web/src/components/modals/ChangePasswordModal/ChangePasswordModal.jsx
alicenstar: Ah you're right, I need to return. Thanks for catching that. :)

GraysonNull (dismissed):

rianfowler (approved):

JustinN-Replicated:
Just wanted to add some screenshots
image
image

PR #2760, opened 2022-04-27
Okteto build performance

syntastical:

Fixes #

```release-note

```


syntastical:
**Pros**
1. Cache stays on the buildkit server
2. Containers are much smaller (except web apps that are still expected to use `okteto up` for dynamic serving)
3. Pipeline deploy is faster

    ```
    completely cached build >1m
    docker layer invalidation 2m
    simple code change 3m
    ```
4. We can deploy single apps, including schema changes
    ```
    okteto build kotsadm 
    okteto deploy 
    ```
5. Okteto up works for all supported services from the base of the project (no need to switch directories)
6. Go projects no longer okteto up, they just get rebuilt and deployed
7. Less files to manage

**Cons**

1. First build will still take forever 
2. Will the buildkit server get overwhelmed with everyone using  it for every compile?
3. Will the buildkit servers caching ability get overwelmed? (could be a problem today)
4. New workflows for developers to learn

README.md
```diff
@@ -79,4 +79,60 @@ Your permissions will be the same between both namespaces, and you will be able
...
+
+This V2 work flow attempts to solve these issues by:
+
+1. Build application only on the buildkit servers so that the cache lives in one place and image sizes stay lean.  This excluded applications that have live reloading (web).
+2. Only use development containers where needed.  (web apps, schema hero, etc)
+3. Update the Okteto manifest to the new schema which allows for separating build and deploy specs, allowing us to run `okteto deploy` and only deploy the manifest. 
+
+#### Reference
+
+| Blah | Syntax | Description                                                                                                |
```
alicenstar, line 108: Could we change 'Blah' to 'Terminology/Term' or something like that?

README.md
```diff
@@ -79,4 +79,60 @@ Your permissions will be the same between both namespaces, and you will be able
...
+| Deploy | `okteto deploy -f okteto-v2.yml` | Deploys the kubernetes manifests. If there were builds before this command, the new images will be used in the deployment. | 
+| Development mode | `okteto up -f okteto-v2.yml` | Prompts the use for what container to put into development mode.  Only web and migrations should appear at this time. |
+
+#### Warning
+
+Because this new workflow is experimental, we still have the old workflow in the project.  If you are using the new workflow, and fail to provide the `-f` flag with the v2 manifest, you will be invoking the old workflow.
+
+#### Example workflow: kotsadm change
+
+1. `okteto pipeline deploy -f okteto-v2.yml`
```
emosbaugh, line 121: i understand this is experimental but once ga could some of these be aliased and combined via make commands?

marccampbell: IMO, it's ok that the user knows it's okteto under the hood.  If we wrap these, it will work 99% of the time, but it might be a lot more context loading to understand how to fix when it breaks.  I like the approach documented here where you learn the `okteto` CLI commands

README.md
syntastical: dang it I knew I would forget to come back to that.

alicenstar (approved):

PR #2761, opened 2022-04-27
return nil instead of default time when kotsadm password updated is nil

pavansokkenagaraj:
#### What type of PR is this?
type::bug

return nil instead of default time when kotsadm password updated is nil

during snapshot restore, if a backup is old and doesn't contain passwordUpdatedAt
when user restores the old snapshot, user will be logged out automatically.
instead of using passwordSecret.CreateTimeStamp, use nil when there is no valid updated timestamp

Fixes #

```release-note
NONE
```

NONE

### DIFF
--- a/pkg/store/kotsstore/user_store.go
+++ b/pkg/store/kotsstore/user_store.go
@@ -248,7 +248,7 @@ func (s *KOTSStore) GetPasswordUpdatedAt() (*time.Time, error) {
 		return nil, errors.Wrap(err, "failed to get password secret")
 	}
 
-	passwordUpdatedAt := new(time.Time)
+	var passwordUpdatedAt *time.Time
 	updatedAtBytes, ok := passwordSecret.Data["passwordUpdatedAt"]
 	if ok {
 		updatedAt, err := time.Parse(time.RFC3339, string(updatedAtBytes))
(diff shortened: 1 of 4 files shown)


pkg/store/kotsstore/user_store.go
```diff
@@ -243,23 +243,23 @@ func (s *KOTSStore) GetPasswordUpdatedAt() (*time.Time, error) {
...
 		if kuberneteserrors.IsNotFound(err) {
 			//  similar to fallback case when password secret is not found and uses the default password from environment variable
-			return &time.Time{}, nil
+			return nil, nil
 		}
 		return nil, errors.Wrap(err, "failed to get password secret")
 	}
 
-	var passwordUpdatedAt time.Time
+	passwordUpdatedAt := new(time.Time)
```
sgalsaleh, line 251:
```suggestion
	var passwordUpdatedAt *time.Time
```

pkg/store/kotsstore/user_store.go
```diff
@@ -243,23 +243,23 @@ func (s *KOTSStore) GetPasswordUpdatedAt() (*time.Time, error) {
...
 		updatedAt, err := time.Parse(time.RFC3339, string(updatedAtBytes))
 		if err != nil {
 			return nil, errors.Wrap(err, "failed to parse passwordUpdatedAt")
 		}
-		passwordUpdatedAt = updatedAt
+		passwordUpdatedAt = &updatedAt
 	} else {
 		// backward compatibility, for older secret/ newly created secret with no passwordUpdatedAt value
-		passwordUpdatedAt = passwordSecret.CreationTimestamp.Time
+		passwordUpdatedAt = nil
```
sgalsaleh, line 261: this else statement is not needed when my above suggestion is done

sgalsaleh (approved):

PR #2763, opened 2022-04-28
refactor AutomateAirgapInstall to use task logic

cbodonnell:
#### What type of PR is this?

type::bug

#### What this PR does / why we need it:

This PR fixes an issue where automated airgapped installs were not working.  Automated airgapped installs used different logic to install the license, so the status was not being reported using a task and the `kots install` command was unable to verify the installation.  This PR refactors the automated airgap install to use the same logic as the online install.

Additional messaging to the user was also added to better indicate the automated installation progress.

#### Which issue(s) this PR fixes:

Fixes [SC-45886](https://app.shortcut.com/replicated/story/45886/when-installing-from-the-cli-if-preflights-fail-or-have-a-warning-the-application-isn-t-deployed-but-there-is-no-message) for automated airgapped installations

#### Special notes for your reviewer:

Screenshots of various outcomes below...

Successful install:
license-with-config-success

Additional app configuration required:
additional-config-required

License installation failed:
license-failed

Preflight checks failed:
preflights-failed

#### Does this PR introduce a user-facing change?

```release-note
NONE (see linked PR below for the release note)
```

See release note in PR #2719 

#### Does this PR require documentation?

NONE

PR #2769, opened 2022-04-29
add a test to validate helm install weights are respected

laverya:

This adds a test to exercise https://github.com/replicatedhq/kots/pull/2743.

Fixes #

```release-note

```

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -1434,9 +1434,94 @@ jobs:
           echo "::notice ::support bundle uploaded to aws replicated-dev account s3://kots-e2e-build-test-support-bundles/${BUNDLE}"
 
 
+  validate-helm-install-order:
+    runs-on: ubuntu-20.04
+    needs: [ can-run-ci, build-kots, build-kurl-proxy, build-migrations, push-minio, push-postgres ]
+    strategy:
+      fail-fast: false
+      matrix:
+        k8s_version: [ v1.20.14-k3s2,v1.21.8-k3s2,v1.22.5-k3s2,v1.23.3-k3s1 ]
+    steps:
+      - uses: replicatedhq/action-k3s@main
+        id: k3s
+        with:
+          version: ${{ matrix.k8s_version }}
+
+      - name: download kots binary
+        uses: actions/download-artifact@v2
+        with:
+          name: kots
+          path: bin/
+
+      - run: chmod +x bin/kots
+
+      - name: run the test
+        env:
+          APP_SLUG: helm-install-order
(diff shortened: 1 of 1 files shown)


jala-dx (approved):

.github/workflows/build-test.yaml
```diff
@@ -1448,9 +1448,94 @@ jobs:
...
+          echo ${{ secrets.HELM_INSTALL_ORDER_LICENSE }} | base64 -d > license.yaml
+          ./bin/kots \
+            install $APP_SLUG/automated \
+            --license-file license.yaml \
+            --no-port-forward \
+            --namespace $APP_SLUG \
+            --shared-password password \
+            --kotsadm-registry ttl.sh \
+            --kotsadm-namespace automated-${{ github.run_id }} \
+            --skip-preflights \
```
sgalsaleh, line 1486: how about removing the preflights from the app itself instead?

.github/workflows/build-test.yaml
```diff
@@ -1448,9 +1448,94 @@ jobs:
           echo "::notice ::support bundle uploaded to aws replicated-dev account s3://kots-e2e-build-test-support-bundles/${BUNDLE}"
 
 
+  validate-helm-install-order:
+    runs-on: ubuntu-20.04
+    needs: [ can-run-ci, build-kots, build-kurl-proxy, build-migrations, push-minio, push-postgres ]
+    strategy:
+      fail-fast: false
+      matrix:
+        k8s_version: [ v1.23.3-k3s1 ]
```
sgalsaleh, line 1457: since this test does not use testim, let's run it on all 4 supported k8s/k3s versions (like the other tests that do not use testim)

.github/workflows/build-test.yaml
laverya: see https://github.com/replicatedhq/kots-test-apps/tree/main/helm-install-order - I think it's stuck on the default preflights?

.github/workflows/build-test.yaml
laverya: I'm fine with running it on all of them, I just didn't think it needed to be there. I'll make the change

.github/workflows/build-test.yaml
sgalsaleh: huh, there shouldn't be such a thing as default preflights. ah i think this is a bug that's fixed in another pr that is not yet merged: https://github.com/replicatedhq/kots/pull/2763

sgalsaleh (dismissed):

sgalsaleh (approved):

PR #2785, opened 2022-05-03
Kiraboyle/sc 46647/initial gitops commit does not translate all labeled secrets to sealed secrets

kcboyle:
- adds missing unit tests for secrets logic
- adds support for multidoc secrets that would improperly ignore multiple secrets and drop non secrets if there was a valid secret in that file

Fixes (shortcut)
https://app.shortcut.com/replicated/story/46647/initial-gitops-commit-does-not-use-sealed-secrets

## Steps to reproduce

In a clean environment:
- install a kubernetes cluster (ie kurl)
- create a namespace
- create a secret (example attached) and apply to the cluster
    - must have the labels:
        - `kots.io/buildphase: secret`
        - `kots.io/secrettype: sealedsecrets`
- install the app as normal (the app/config should not matter) 
- create a repo for enabling gitops
- In your app, go to the gitops tab and setup gitops as normal ([docs](https://docs.replicated.com/enterprise/gitops-single-app-workflow))
- In the repo, you should be able to see the history of the files and see that we are committing "Secrets" and not just "SealedSecrets"

What should be happening and is fixed by this PR
- when we enable gitops, we should be translating Secret objects into SealedSecret objects

What is happening without this PR:
- an inconsistent number of Secret objects are being translated when we initially setup gitops, so those plain text secrets are being committed
- when you make an update to the app, those remaining secrets are all translated as expected, but the plain text secrets still exist in the git history

```release-note
- when enabling gitops, the initial commit will now properly translate all labeled secrets to sealedsecrets
```

NONE

### DIFF
--- a/pkg/secrets/sealed_secrets.go
+++ b/pkg/secrets/sealed_secrets.go
@@ -5,17 +5,16 @@ import (
 	"crypto/rsa"
 	"crypto/x509"
 	"encoding/pem"
-	"io/ioutil"
-	"os"
-
 	sealedsecretsv1alpha1 "github.com/bitnami-labs/sealed-secrets/pkg/apis/sealed-secrets/v1alpha1"
 	sealedsecretsscheme "github.com/bitnami-labs/sealed-secrets/pkg/client/clientset/versioned/scheme"
 	"github.com/pkg/errors"
 	"github.com/replicatedhq/kots/pkg/util"
+	"io/ioutil"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	jsonserializer "k8s.io/apimachinery/pkg/runtime/serializer/json"
 	"k8s.io/client-go/kubernetes/scheme"
+	"os"
 )
 
 func replaceSecretsWithSealedSecrets(archivePath string, config map[string][]byte) error {
@@ -39,7 +38,6 @@ func replaceSecretsWithSealedSecrets(archivePath string, config map[string][]byt
 	}
 
 	sealedsecretsscheme.AddToScheme(scheme.Scheme)
-	codecFactory := serializer.NewCodecFactory(scheme.Scheme)
 
 	decode := scheme.Codecs.UniversalDeserializer().Decode
 	for _, secretPath := range secretPaths {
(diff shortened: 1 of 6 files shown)


pkg/secrets/sealed_secrets.go
```diff
@@ -44,44 +46,72 @@ func replaceSecretsWithSealedSecrets(archivePath string, config map[string][]byt
 			return errors.Wrap(err, "failed to read file")
 		}
 
-		decoded, _, err := decode(contents, nil, nil)
-		if err != nil {
-			return nil
-		}
+		multiDocYaml := bytes.Split(contents, []byte("---\n"))
+		var secrets []byte
+		var nonSecrets []byte
```
sgalsaleh, line 51: i'd recommend changing this to this, and then run `bytes.Join(nonSecrets, []byte("\n---\n"))` at the end. same for the secrets slice
```suggestion
		var nonSecrets [][]byte
```

pkg/secrets/sealed_secrets.go
```diff
@@ -44,44 +46,72 @@ func replaceSecretsWithSealedSecrets(archivePath string, config map[string][]byt
...
+				continue
+			}
 
-		secret, ok := decoded.(*v1.Secret)
-		if !ok {
-			return nil
-		}
+			decoded, _, err := decode(object, nil, nil)
+			if err != nil {
+				nonSecrets = append(nonSecrets, []byte("---\n")...)
```
sgalsaleh, line 60: i'd recommend following my suggestion above, since not all documents end with a new line. or you can use `\n---\n` instead

pkg/secrets/sealed_secrets.go
```diff
@@ -44,44 +46,72 @@ func replaceSecretsWithSealedSecrets(archivePath string, config map[string][]byt
 			return errors.Wrap(err, "failed to read file")
 		}
 
-		decoded, _, err := decode(contents, nil, nil)
-		if err != nil {
-			return nil
-		}
+		multiDocYaml := bytes.Split(contents, []byte("---\n"))
```
sgalsaleh, line 49:
```suggestion
		multiDocYaml := bytes.Split(contents, []byte("\n---\n"))
```

pkg/secrets/sealed_secrets.go
```diff
@@ -44,44 +46,79 @@ func replaceSecretsWithSealedSecrets(archivePath string, config map[string][]byt
...
+		var secrets [][]byte
+		var nonSecrets [][]byte
 
-		secret, ok := decoded.(*v1.Secret)
-		if !ok {
-			return nil
-		}
+		for i := 0; i < len(multiDocYaml); i++ {
```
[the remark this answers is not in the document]
kcboyle: changed and committed

pkg/secrets/secrets.go
```diff
@@ -45,17 +40,17 @@ func ReplaceSecretsInPath(archiveDir string) error {
...
-func getSecretsInPath(archiveDir string) ([]string, error) {
-	secretPaths := []string{}
+func findPathsWithSecrets(archiveDir string) ([]string, error) {
+	var paths []string
 	decode := scheme.Codecs.UniversalDeserializer().Decode
 
 	err := filepath.Walk(archiveDir, func(path string, info os.FileInfo, err error) error {
 		if err != nil {
-			return err
+			return errors.Wrap(err, "could not walk through the archive directory")
```
sgalsaleh, line 53: this message should be moved to line 84 https://github.com/replicatedhq/kots/pull/2785/files#diff-29941b86f5d6066d562f03783c8316b6c4f13dfddcdc57c320fea3d53a6635b4R83
```suggestion
			return err
```

kcboyle: good call. fixed and pushed

pkg/secrets/sealed_secrets.go
kcboyle: resolved with latest push

sgalsaleh (approved):

PR #2786, opened 2022-05-03
Exclude the api version when identifying native k8s manifests to be deleted on deploy

sgalsaleh:
#### What type of PR is this?

type::bug

Excludes the api version when identify manifests to be deleted on deploy. kubernetes does not consider the api version when identifying manifests to be deleted or updated, just the kind, name, and namespace.

## Steps to reproduce

1- add an ingress object with api version `networking.k8s.io/v1` to your app
2- install and deploy the app
3- replace with an ingress object with api version `extensions/v1beta1` in your app
4- deploy that version => ingress object is recreated when it should've been updated (since kubernetes supports that)

```release-note
* Fixes an issue where changing the api version for a native Kubernetes object caused that object to be deleted and recreated instead of just being updated.
```

NONE

### DIFF
--- a/pkg/operator/client/gvkn.go
+++ b/pkg/operator/client/gvkn.go
@@ -31,7 +31,18 @@ func GetGVKWithNameAndNs(content []byte, baseNS string) (string, OverlySimpleGVK
 		namespace = o.Metadata.Namespace
 	}
 
-	return fmt.Sprintf("%s-%s-%s-%s", o.APIVersion, o.Kind, o.Metadata.Name, namespace), o
+	// TODO: this is a hack, find a better way to do this.
+	// kubernetes does not consider the api version when identifying manifests,
+	// and it automatically converts the schema when the api version changes for native k8s objects.
+	// kubernetes doesn't/can't handle automatically converting the schema when the api version changes for CRDs though,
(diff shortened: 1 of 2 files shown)


pkg/operator/client/gvkn.go
```diff
@@ -31,7 +31,8 @@ func GetGVKWithNameAndNs(content []byte, baseNS string) (string, OverlySimpleGVK
 		namespace = o.Metadata.Namespace
 	}
 
-	return fmt.Sprintf("%s-%s-%s-%s", o.APIVersion, o.Kind, o.Metadata.Name, namespace), o
+	// kubernetes does not consider the api version when identifying manifests
+	return fmt.Sprintf("%s-%s-%s", o.Kind, o.Metadata.Name, namespace), o
```
divolgin, line 35: It's unlikely but possible that two specs with different `APIVersion` will trample each other in the map where this is used a key.  We may need a distinction between what we use in our internal map key and what we use to identify objects that need to be deleted.

divolgin (dismissed):

pkg/operator/client/gvkn.go
jala-dx: @sgalsaleh this fix is bit risky in mind. we need to discuss with @marccampbell for a solution.

divolgin (approved):

PR #2793, opened 2022-05-04
Don't re-deploy the app when the api restarts

sgalsaleh:
#### What type of PR is this?

type::bug

Fixes an issue where the app was always being re-deployed when the api restarts which can cause unexpected consequences or even downtime.

This PR also changes the way we set and read the status of the deployment.

## Steps to reproduce
- install and deploy an app
- restart the kotsadm pod => app is redeployed

```release-note
Fixes an issue where the application was being re-deployed when the admin console restarts.
```

NONE


pkg/operator/client/client.go
```diff
@@ -143,10 +144,11 @@ func (c *Client) DeployApp(deployArgs operatortypes.DeployAppArgs) {
 	var helmResult *commandResult
 	var deployError, helmError error
 	defer func() {
-		err := c.setResults(deployArgs, &deployRes.dryRunResult, &deployRes.applyResult, helmResult)
+		results, err := c.setResults(deployArgs, &deployRes.dryRunResult, &deployRes.applyResult, helmResult)
 		if err != nil {
-			log.Printf("failed to set result: %v", err)
+			finalError = errors.Wrap(err, "failed to set results")
 		}
+		deployed = !results.IsError
```
divolgin, line 151: results can be nil here

pkg/operator/client/client.go
```diff
@@ -296,9 +300,9 @@ func (c *Client) deployHelmCharts(deployArgs operatortypes.DeployAppArgs) (*comm
 	return installResult, nil
 }
 
-func (c *Client) setResults(deployArgs operatortypes.DeployAppArgs, dryRunResult *commandResult, applyResult *commandResult, helmResult *commandResult) error {
+func (c *Client) setResults(deployArgs operatortypes.DeployAppArgs, dryRunResult *commandResult, applyResult *commandResult, helmResult *commandResult) (*DeployResults, error) {
```
divolgin, line 303: How about changing return type to `DeployResults`, so it's never nil

sgalsaleh: maybe, i feel `nil` is better than having empty results because it will have `IsError` set to false by default.

divolgin (approved):

PR #2795, opened 2022-05-04
Ensure secrets when only Helm is used with no other specs

divolgin:
#### What type of PR is this?

type::bug

When application specifies `additionalNamespaces` but does not included any non-helm resources, secrets will not be applied to additional namespaces.

Fixes #

```release-note
* FIxes a bug that would cause image pull secrets to not be created in additional namespaces when only Helm charts are used by the application.
```

NONE

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -1,6 +1,7 @@
 package operator
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
 	"encoding/json"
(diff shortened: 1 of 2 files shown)


pkg/operator/operator.go
```diff
@@ -638,3 +650,34 @@ func RedeployAppVersion(appID string, sequence int64) error {
...
+		obj, gvk, err := decode([]byte(secretSpec), nil, nil)
+		if err != nil {
+			continue
+		}
+
+		if gvk.Group != "" || gvk.Version != "v1" || gvk.Kind != "Secret" {
+			continue
+		}
+		secret := obj.(*corev1.Secret)
+		uniqueSecrets[secret.Name] = secret
```
sgalsaleh, line 668: what about namespace?

divolgin: they are applied to all namespaces.  we just need the list of secrets without namespaces.

sgalsaleh (approved):

pkg/operator/operator.go
divolgin: here https://github.com/replicatedhq/kots/blob/847a03667df7c19b59398399f34d7ad67dd8c03c/pkg/operator/client/deploy.go#L267-L268

PR #2796, opened 2022-05-05
always show cron input. auto switch to custom when typing in the box

GraysonNull:
#### What type of PR is this?
type::feature

Makes the ability to do a custom snapshot schedule more discoverable.

Here is a video of the change: https://www.loom.com/share/a66a2cca6ffa45bcb76858ef9cfb214e

```release-note
When automatic snapshots are enabled an input showing the CRON schedule is always visible and editable to allow for more granular. 
```

NONE


miaawong (approved):
looks good to me! 
would be cleaner to create an 'handleCronChange' function, it is best practice to pull into a function instead of running code inline.

miaawong (approved):
approved!

PR #2798, opened 2022-05-05
Prevent multiple support bundles being generated, show in progress support bundle

alicenstar:
#### What type of PR is this?

type::feature

Currently, users can generate multiple support bundles at once, and if you navigate away from the `/generate` page, you will lose the ability to track the progress of your support bundle.

This PR adds a check when visiting the `/generate` page, so that if there is an existing bundle in progress, the user will not be shown the button to generate another support bundles. If there is a support bundle in progress, this page will show the progress for the latest bundle.

Fixes [#47061](https://app.shortcut.com/replicated/story/47061/support-bundle-generation-from-kots-ui-can-t-be-done-in-the-background)

There is some follow up work that will need to be done, such as:

- Add the ability to cancel the generation of a support bundle (in the event that it gets "stuck" and never completes) in the API.
- Possibly add a cancel button in the UI
- Show a better message on the Troubleshoot tab for a bundle that is in progress. Currently it shows "Unable to surface insights for this bundle" if a bundle is in progress, which makes it sound like an error has occurred.

## Steps to reproduce

Go to the Troubleshoot tab, click 'Generate a support bundle'. Click 'Analyze <App>' and then navigate to a different page. Return to the generate a support bundle page. You should see the progress of your support bundle, and not see a button to generate a new one.

```release-note
Prevents user from creating more than one support bundle at a time, and allows the user to see the progress of the current support bundle if you return to the `/troubleshoot/generate` route.
```

No?

#### Behavior Before

https://www.loom.com/share/d8274e1f44784c59867e39f4e913be77

#### Behavior After

https://www.loom.com/share/6ad80612eea34cd49eb4f07e52cd6d95

### DIFF
--- a/web/src/components/troubleshoot/GenerateSupportBundle.jsx
+++ b/web/src/components/troubleshoot/GenerateSupportBundle.jsx
@@ -231,6 +231,8 @@ class GenerateSupportBundle extends React.Component {
   pollForBundleAnalysisProgress = async () => {
     const { newBundleSlug } = this.state;
     if (!newBundleSlug) {
+      // component may start polling before bundle slug is set
+      // this is to prevent an api call if the slug is not set
       return;
     }
     fetch(`${process.env.API_ENDPOINT}/troubleshoot/supportbundle/${newBundleSlug}`, {
--- a/web/src/components/troubleshoot/GenerateSupportBundle.jsx
+++ b/web/src/components/troubleshoot/GenerateSupportBundle.jsx
@@ -13,8 +13,6 @@ import { Repeater } from "../../utilities/repeater";
 
 import "../../scss/components/troubleshoot/GenerateSupportBundle.scss";
 
-const NEW_CLUSTER = "Create a new downstream cluster";
-
 class GenerateSupportBundle extends React.Component {
   constructor(props) {
     super(props);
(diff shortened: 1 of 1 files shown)


web/src/components/troubleshoot/GenerateSupportBundle.jsx
```diff
@@ -81,6 +81,69 @@ class GenerateSupportBundle extends React.Component {
...
+            errorMsg: `Unexpected status code: ${res.status}`,
+            displayErrorModal: true,
+            networkErr: false
+          });
+          return;
+        }
+        const response = await res.json();
+        const bundleRunning = response.supportBundles.find((bundle) => bundle.status === "running");
+        if (bundleRunning) {
+          console.log(bundleRunning);
```
rianfowler, line 112:
```suggestion
```
console.log

web/src/components/troubleshoot/GenerateSupportBundle.jsx
```diff
@@ -159,6 +231,9 @@ class GenerateSupportBundle extends React.Component {
 
   pollForBundleAnalysisProgress = async () => {
     const { newBundleSlug } = this.state;
+    if (!newBundleSlug) {
```
rianfowler, line 234: maybe add a comment explaining why this would return

rianfowler (dismissed):
Verified on Okteto.

rianfowler (approved):

PR #2805, opened 2022-05-07
Surface Minio FS errors better in UI and support bundles

divolgin:
#### What type of PR is this?

type::bug

When NFS backend configuration for snapshots fails, the user is presented with a generic error message. The specific error is also no surfaced by support bundle analyzers.

This will also clean up orphaned `kotsadm-fs-minio-check` pods once configuration is successful.

Fixes #

- This probably doesn't cover every possible failure mode.
- UI could use some work for cases where error messages are excessively long.
- Screenshots in comments.

## Steps to reproduce

1. Navigate to Snapshots settings page
2. Trigger error in one or more of the following ways:
   - enter invalid IP address/host name (invalid will cause error to surface sooner; i/o timeout takes longer and may not happen fast enough)
   - remove/rename `mount.nfs` binary on the host
   - remove write permissions from the NFS directory
3. Click "Update" and wait
4. Collect support bundle.

```release-note
* Failed `kotsadm-fs-minio-check` pods will now be cleaned up once NFS backend for snapshots has been configured successfully.
* Adds collector and analyzer for cases when NFS configuration fails because the `mount.nfs` binary is missing on the host.
```

NONE

### DIFF
--- a/pkg/snapshot/filesystem.go
+++ b/pkg/snapshot/filesystem.go
@@ -9,6 +9,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/google/uuid"
@@ -19,6 +20,7 @@ import (
 	kotsadmtypes "github.com/replicatedhq/kots/pkg/kotsadm/types"
 	kotsadmversion "github.com/replicatedhq/kots/pkg/kotsadm/version"
 	"github.com/replicatedhq/kots/pkg/kotsutil"
+	"github.com/replicatedhq/kots/pkg/logger"
 	kotss3 "github.com/replicatedhq/kots/pkg/s3"
 	types "github.com/replicatedhq/kots/pkg/snapshot/types"
 	"github.com/replicatedhq/kots/pkg/util"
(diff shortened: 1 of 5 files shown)


divolgin:
Some screenshots:
Screen Shot 2022-05-06 at 12 40 41 PM
Screen Shot 2022-05-06 at 6 02 51 PM
Screen Shot 2022-05-06 at 5 59 33 PM
Screen Shot 2022-05-06 at 5 42 03 PM
Screen Shot 2022-05-06 at 3 39 06 PM
Screen Shot 2022-05-06 at 2 39 51 PM

pkg/snapshot/filesystem.go
```diff
@@ -41,6 +45,12 @@ const (
 	FileSystemMinioServicePort                                                = 9000
 )
 
+const (
+	fdMinioCheckTage  = "kotsadm-fs-minio-check"
```
sgalsaleh, line 49:
```suggestion
	fsMinioCheckTag  = "kotsadm-fs-minio-check"
```

pkg/snapshot/filesystem.go
```diff
@@ -548,7 +558,40 @@ func shouldResetFileSystemMount(ctx context.Context, clientset kubernetes.Interf
 	}
 
 	if err := k8sutil.WaitForPod(ctx, clientset, deployOptions.Namespace, checkPod.Name, time.Minute*2); err != nil {
-		finalErr = errors.Wrap(err, "failed to wait for file system minio check pod to complete")
+		genericErr := errors.Wrap(err, "failed to wait for file system minio check pod to complete")
```
sgalsaleh, line 561: what about the reset pod and the keys SHA pod? i see this PR will only return errors for the check pod

divolgin: I think those will still show a generic error and logs will be in the support bundle....  This is the only one I tested.

sgalsaleh (approved):

PR #2825, opened 2022-05-13
improvement: remove deploy loop from the operator

sgalsaleh:
removes the deploy loop from the operator so that everything happens adhoc

```release-note
NONE
```

NONE

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -76,11 +76,8 @@ func DeployVersion(appID string, sequence int64) error {
 
 	logger.Info("deploying app version", zap.String("appId", appID), zap.Int64("sequence", sequence))
 
-	db := persistence.MustGetDBSession()
-	query := `update app_downstream set current_sequence = $1 where app_id = $2`
-	_, err = db.Exec(query, sequence, appID)
-	if err != nil {
-		return errors.Wrap(err, "failed to update app downstream current sequence")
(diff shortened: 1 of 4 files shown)


pkg/version/version.go
```diff
@@ -73,31 +74,16 @@ func DeployVersion(appID string, sequence int64) error {
...
-	query = `update app_downstream_version set status = $3, applied_at = $4 where sequence = $1 and app_id = $2`
-	_, err = tx.Exec(query, sequence, appID, storetypes.VersionDeploying, time.Now())
-	if err != nil {
-		return errors.Wrap(err, "failed to update app downstream version status")
-	}
-
-	if err := tx.Commit(); err != nil {
-		return errors.Wrap(err, "failed to commit")
-	}
+	go operator.DeployApp(appID, sequence)
```
divolgin, line 86: I think this requires `DeployApp` to be thread safe

divolgin: Could also keep the loop and make it read from a channel so only one deployment happens at a time.

Also if kotsadm restarts before deployment happens, someone will have to click the button again.

sgalsaleh: `DeployApp` should be thread safe since it uses a mutex? but i do agree with your other point. i'll close this PR then for now.

sgalsaleh: also, the loop already does one deployment at a time.

divolgin (dismissed):

divolgin (approved):

PR #2838, opened 2022-05-24
Replace k3s replaces with the offical k8s

syntastical:
Remove the dependency on k8s.io/kubernetes
Use the k8s deps instead of k3s

Fixes #47238

None

## Steps to reproduce
None

```release-note
NONE
```


Makefile.build
```diff
@@ -1,4 +1,5 @@
 export GOPROXY=https://proxy.golang.org
+export CGO_ENABLED=0
```
marccampbell, line 2: What's this needed for?

pkg/k8sutil/cluster.go
```diff
@@ -1,70 +1,57 @@
 package k8sutil
 
 import (
+	"context"
+	"fmt"
+	v1 "k8s.io/api/core/v1"
```
marccampbell, line 6: move this into the section below the space, and rename the import to `corev1` for consistency

pkg/k8sutil/cluster.go
```diff
@@ -1,70 +1,57 @@
...
-	kubeadmconfig "k8s.io/kubernetes/cmd/kubeadm/app/util/config"
 )
 
 // GenerateBootstrapToken will generate a node join token for kubeadm.
 // ttl defines the time to live for this token.
 func GenerateBootstrapToken(client kubernetes.Interface, ttl time.Duration) (string, error) {
 	token, err := bootstraputil.GenerateBootstrapToken()
 	if err != nil {
-		return "", errors.Wrap(err, "generate kubeadm token")
+		return "", errors.Wrap(err, "failed to generate bootstrap token")
```
marccampbell, line 21: why renaming this text? will this impact error messages that exist in our docs, community, analyzers, etc?  Unless there's a product reason, feels unecessary and risky

pkg/k8sutil/cluster.go
```diff
@@ -1,70 +1,57 @@
...
 	}
+	substrs := bootstraputil.BootstrapTokenRegexp.FindStringSubmatch(token)
+	tokenID := substrs[0]
+	tokenSecret := substrs[1]
 
-	bts, err := bootstraptokenv1.NewBootstrapTokenString(token)
-	if err != nil {
-		return "", errors.Wrap(err, "new kubeadm token string")
+	data := map[string][]byte{
+		bootstrapapi.BootstrapTokenIDKey:     []byte(tokenID),
```
marccampbell, line 28: Will this continue to work with older K8s versions? How has this implementation been validated and tested?

marccampbell (commented):
The go.mod changes look mostly good to me. I'm not sure I understand how the kubeadm bootstrapping changes were tested and confirmed across supported versions. Can you explain that in the PR?

Makefile.build
syntastical: We use CGO_ENABLED=0 in goreleaser so this change will match the configuration that we release with.  This will bring down the build time a little bit.

pkg/k8sutil/cluster.go
syntastical: I can revert this.

pkg/k8sutil/cluster.go
syntastical: It should, this is basically the same thing that the kubernetes framework was doing so we would get the same result either way.  This way just allows us to drop a direct dependency.

syntastical:
@marccampbell  in regards to the testing.  I tested in Okteto and locally but I want to run the regression tests on this for sure.  That will give us more confidences about kube versions and configurations.  But right now the regression stuff triggers off main.

syntastical:
@marccampbell Fixed the issues you mentioned in the code.

syntastical:
These changes were tested locally and in Okteto.

syntastical:
@marccampbell here is the passing regression test.  There was one failure but it was completely unrelated to the change.  everything else passed.

PR #2844, opened 2022-05-26
Add support to do existing install on k8s 1.24.0

jala-dx:
On k8s 1.24.0 node labels are changed to `control-plane ` and `master` label is removed. In order to support existing kots installs on 1.24.0, added logic to accommodate  this new label `control-plane`

Fixes # SC-48706

None

No

```release-note
TODO

```

yes

### DIFF
--- a/pkg/kurl/join_cert.go
+++ b/pkg/kurl/join_cert.go
@@ -10,9 +10,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/blang/semver"
 	"github.com/pkg/errors"
-	"github.com/replicatedhq/kots/pkg/k8sutil"
 	"github.com/replicatedhq/kots/pkg/logger"
 	"github.com/replicatedhq/kots/pkg/util"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


garcialuis (approved):

emosbaugh (changes requested):
What about https://github.com/replicatedhq/kots/blob/main/pkg/kurl/join_cert.go#L183-L192 ?

jala-dx:
> What about https://github.com/replicatedhq/kots/blob/main/pkg/kurl/join_cert.go#L183-L192 ?

Sigh just missed we do `existing` install usecase here, assumed that labels are already set by kurl. Will fix this.

pkg/kurl/join_cert.go
```diff
@@ -180,12 +205,10 @@ func getPodSpec(clientset kubernetes.Interface, namespace string) (*corev1.Pod,
...
 		Spec: corev1.PodSpec{
 			SecurityContext: &securityContext,
-			NodeSelector: map[string]string{
-				"node-role.kubernetes.io/master": "",
-			},
+			NodeSelector:    nodeSelector,
 			Tolerations: []corev1.Toleration{
 				{
-					Key:      "node-role.kubernetes.io/master",
+					Key:      tolerationKey,
```
emosbaugh, line 211: lets just always add both tolerations rather than detecting version

pkg/kurl/join_cert.go
```diff
@@ -180,12 +205,10 @@ func getPodSpec(clientset kubernetes.Interface, namespace string) (*corev1.Pod,
 		},
 		Spec: corev1.PodSpec{
 			SecurityContext: &securityContext,
-			NodeSelector: map[string]string{
-				"node-role.kubernetes.io/master": "",
-			},
+			NodeSelector:    nodeSelector,
```
emosbaugh, line 208: lets change this to nodeAffinity and add both rather than detecting version

https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity

> If you specify multiple nodeSelectorTerms associated with nodeAffinity types, then the Pod can be scheduled onto a node if one of the specified nodeSelectorTerms can be satisfied.

```
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: node-role.kubernetes.io/control-plane
            operator: Exists
        - matchExpressions:
          - key: node-role.kubernetes.io/master
            operator: Exists
```

emosbaugh (changes requested):

emosbaugh (approved):

pkg/kurl/join_cert.go
jala-dx: Tested this

PR #2856, opened 2022-06-06
check if base dir exists before using it

trelore:

Checks if a base directory exists before using it as a base in kustomize

Fixes #44092

## Steps to reproduce

Create an app with a chart dependency on if a bool is checked or not
Check and uncheck the dependency
Deploy the app
The app used to break as the dependency is no longer there, but the midstream tries to use the chart.

```release-note
Fixes issue where a helm chart is excluded after previously being included
```

Fixes issue where a helm chart is excluded after previously being included.


pkg/operator/helm.go
```diff
@@ -87,7 +87,8 @@ func renderChartsArchive(deployedVersionArchive string, downstreamName string, k
 				}
 			}
 
-			if info.Name() == "kustomization.yaml" {
+			_, err = os.Stat(sourceChartsDir)
```
emosbaugh, line 90: charts dir is `base/charts`. inside this dir is `base/charts/[chart name]` for each chart. are you sure you dont want to check each chart individually? if not why iterate at all and just check this after line 70?

https://github.com/replicatedhq/kots/pull/2856/files#diff-685325e8ac6019af11c7f04174b7ae09e382a86ae9d0e59563788a5a936354e5R70

trelore: Ah, I'll have another look, thanks for catching this

trelore: @emosbaugh can you take another look, I added the relative path to the source charts dir. I think this is what you meant initially.

emosbaugh: lgtm

emosbaugh:
i am rerunning the tests

emosbaugh:
i dont know why the tests are failing. i can help you look tomorrow.

emosbaugh (approved):

pkg/operator/helm.go
```diff
@@ -88,16 +88,19 @@ func renderChartsArchive(deployedVersionArchive string, downstreamName string, k
 			}
 
 			if info.Name() == "kustomization.yaml" {
-				archiveChartOutput, err := exec.Command(kustomizeBinPath, "build", filepath.Dir(path)).Output()
-				if err != nil {
-					if ee, ok := err.(*exec.ExitError); ok {
-						err = fmt.Errorf("kustomize %s: %q", path, string(ee.Stderr))
+				_, err = os.Stat(filepath.Join(sourceChartsDir, relPath))
+				if !os.IsNotExist(err) {
```
sgalsaleh, line 92: why not
```go
if err == nil {
```

sgalsaleh: and then fail if there is an error that is not `IsNotExist`

PR #2864, opened 2022-06-07
Set user agent on outbound http requests

marccampbell:
Consistently sets the user-agent to the kots version string in outgoing http requests.

Tracked in SC: https://app.shortcut.com/replicated/story/48588/kots-instances-reporting-kots-version-go-http-client-2-0

None

None


laverya (approved):
looks good to me, my only suggestion would be making a library function that returns a request with this header already set instead of adding it individually everywhere, but I'll make a different ticket for that

laverya (approved):

laverya (approved):

PR #2878, opened 2022-06-13
[Still testing] Fix to generate image pull secrets for each subchart

jala-dx:
Fixes a bug that is caused when top level `templates` folder is not present in an helm chart that also has subcharts and top level charts.
Generates uniquely named imagePull secrets for each of the top level charts and also subcharts if the chart manfiests has a private image. 
Refer to the ticket for the detailed analysis. 

Fixes # SC-49488

## Steps to reproduce

Deploy a Helm chart with subcharts that does not have a top level `templates` folder 
Notice that the base rendering looks incorrect that all the subcharts are consolidated and appears like a single chart

```release-note

Fixes a bug that is caused when top level `templates` folder is not present in an helm chart that also has subcharts and top level charts.

```

### DIFF
--- a/pkg/base/helm_test.go
+++ b/pkg/base/helm_test.go
@@ -82,7 +82,7 @@ func Test_RenderHelm(t *testing.T) {
 			name: "helm v3 namespace insertion",
 			args: args{
 				upstream: &upstreamtypes.Upstream{
-					Name: "namespace-test",
+					Name: "test-chart",
 					Files: []upstreamtypes.UpstreamFile{
 						{
 							Path:    "Chart.yaml",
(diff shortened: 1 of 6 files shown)


pkg/base/helm_test.go
```diff
@@ -725,6 +725,201 @@ func Test_removeCommonPrefix(t *testing.T) {
...
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := RenderHelm(tt.args.upstream, tt.args.renderOptions)
```
laverya, line 911: Why not add this to the existing renderHelm unit test?

emosbaugh (dismissed):

laverya (approved):

PR #2888, opened 2022-06-14
Fix e2e dockerfile environment

emosbaugh:


e2e/README.md
```diff
@@ -56,10 +56,10 @@ $ make e2e \
     SKIP_TEARDOWN=1
 ...
     To set kubecontext run:
-      export KUBECONFIG=/tmp/kots-e2e3629427925/.kubeconfig
+      export KUBECONFIG="$(k3d kubeconfig merge kots-e2e3629427925)"
```
syntastical, line 59: Haven't tested this yet but my understanding is that `k3d kubeconfig merge` will write a file, don't we want to use `k3d kubeconfig` get here instead?

e2e/README.md
```diff
@@ -56,10 +56,10 @@ $ make e2e \
...
     To delete cluster run:
       k3d cluster delete kots-e2e3629427925
-$ export KUBECONFIG=/tmp/kots-e2e3629427925/.kubeconfig
+$ export KUBECONFIG="$(k3d kubeconfig merge kots-e2e3629427925)"
```
syntastical, line 62: same as above

e2e/cluster/k3d.go
```diff
@@ -54,7 +54,7 @@ func (c *K3d) Teardown() {
 }
 
 func (c *K3d) PrintDebugInfo() {
-	GinkgoWriter.Printf("To set kubecontext run:\n  export KUBECONFIG=%s\n", c.GetKubeconfig())
+	GinkgoWriter.Printf("To set kubecontext run:\n  export KUBECONFIG=\"$(k3d kubeconfig merge %s)\"\n", c.GetClusterName())
```
syntastical, line 57: same as readme question

e2e/Makefile
```diff
@@ -23,10 +23,11 @@ test: TESTIM_BRANCH ?= master
 test: SKIP_TEARDOWN ?= 0
 test:
 	export TESTIM_ACCESS_TOKEN
-	docker run --rm -it --net host \
+	docker run --rm -i --net host \
```
syntastical, line 26: Just curious, was the `t` causing a problem or just some clean up?

syntastical (approved):

e2e/README.md
emosbaugh: Can you set the env var to a string and not a filepath?

PR #2930, opened 2022-06-29
Use chart path instead of name as a map key

divolgin:

Don't use chart name as the key in an in-memory structure because charts names are not unique.

Fixes #

```release-note
* Fixes a bug that caused kustomize to fail if a Helm chart and one of its sub-charts had the same name
```

NONE


pkg/pull/pull.go
```diff
@@ -440,9 +440,12 @@ func Pull(upstreamURI string, pullOptions PullOptions) (string, error) {
 		chartBaseName := v.Spec.Chart.Name
 		// the helmBase may have a chart name prefix removed - we must find the base name instead of the original chart name
 		for _, helmBase := range helmBases {
-			chartName := strings.Split(helmBase.Path, "/")[len(strings.Split(helmBase.Path, "/"))-1]
-			if strings.HasSuffix(v.Spec.Chart.Name, chartName) {
+			baseDir := filepath.Dir(helmBase.Path)
+			chartName := filepath.Base(helmBase.Path)
```
sgalsaleh, line 444: i think the base path comes from `metadata.name` or the new `releaseName` field, not the chart name. so comparing this to `v.Spec.Chart.Name` is not accurate i believe

sgalsaleh: there's a `v.GetDirName()` function that should return the base directory name that was used

pkg/rewrite/rewrite.go
```diff
@@ -237,8 +236,10 @@ func Rewrite(rewriteOptions RewriteOptions) error {
 		chartBaseName := v.Spec.Chart.Name
 		// the helmBase may have a chart name prefix removed - we must find the base name instead of the original chart name
 		for _, helmBase := range helmBases {
-			chartName := strings.Split(helmBase.Path, "/")[len(strings.Split(helmBase.Path, "/"))-1]
-			if strings.HasSuffix(chartBaseName, chartName) {
+			baseDir := filepath.Dir(helmBase.Path)
+			chartName := filepath.Base(helmBase.Path)
+			// v is a top level chart, so only top level charts directory name should match
```
sgalsaleh, line 241: > so only top level charts directory name should match

how is that enforced here? by `baseDir == "charts"`? isn't there a `charts` directory for each subchart as well?

divolgin: yes, and baseDir is the entire path. it won't match anything nested deeper.

pkg/pull/pull.go
```diff
@@ -440,9 +440,12 @@ func Pull(upstreamURI string, pullOptions PullOptions) (string, error) {
 		chartBaseName := v.Spec.Chart.Name
 		// the helmBase may have a chart name prefix removed - we must find the base name instead of the original chart name
 		for _, helmBase := range helmBases {
-			chartName := strings.Split(helmBase.Path, "/")[len(strings.Split(helmBase.Path, "/"))-1]
-			if strings.HasSuffix(v.Spec.Chart.Name, chartName) {
+			baseDir := filepath.Dir(helmBase.Path)
```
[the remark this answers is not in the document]
divolgin: that leave a trailing slash in directory name, which i deemed ugly :)

pkg/image/builder.go
```diff
@@ -102,16 +102,22 @@ func GetPrivateImages(upstreamDir string, checkedImages map[string]ImageInfo, al
...
-				return filepath.SkipDir
-			}
-
 			if info.IsDir() {
+				chartName, err := filepath.Rel(upstreamDir, path)
+				if err != nil {
+					logger.Debugf("Failed to remove prefix from %s: %v", path, err)
+					return nil
+				}
+				chartName = filepath.Join(parentHelmChartPath, chartName)
```
sgalsaleh, line 111: shouldn't `filepath.Rel(upstreamDir, path)` already have the parent helm chart path?

pkg/pull/pull.go
```diff
@@ -437,12 +437,15 @@ func Pull(upstreamURI string, pullOptions PullOptions) (string, error) {
...
-		chartBaseName := v.Spec.Chart.Name
+		chartBaseName := v.GetDirName()
 		// the helmBase may have a chart name prefix removed - we must find the base name instead of the original chart name
 		for _, helmBase := range helmBases {
-			chartName := strings.Split(helmBase.Path, "/")[len(strings.Split(helmBase.Path, "/"))-1]
-			if strings.HasSuffix(v.Spec.Chart.Name, chartName) {
+			baseDir := filepath.Dir(helmBase.Path)
+			chartName := filepath.Base(helmBase.Path)
+			// v is a top level chart, so only top level charts directory name should match
+			if baseDir == "charts" && strings.HasSuffix(chartBaseName, chartName) {
```
sgalsaleh, line 446: why not
```suggestion
			if baseDir == "charts" && chartBaseName == chartName {
```

pkg/rewrite/rewrite.go
```diff
@@ -234,11 +233,13 @@ func Rewrite(rewriteOptions RewriteOptions) error {
...
-		chartBaseName := v.Spec.Chart.Name
+		chartBaseName := v.GetDirName()
 		// the helmBase may have a chart name prefix removed - we must find the base name instead of the original chart name
 		for _, helmBase := range helmBases {
-			chartName := strings.Split(helmBase.Path, "/")[len(strings.Split(helmBase.Path, "/"))-1]
-			if strings.HasSuffix(chartBaseName, chartName) {
+			baseDir := filepath.Dir(helmBase.Path)
+			chartName := filepath.Base(helmBase.Path)
+			// v is a top level chart, so only top level charts directory name should match
+			if baseDir == "charts" && strings.HasSuffix(chartBaseName, chartName) {
```
sgalsaleh, line 242: ditto

pkg/image/builder.go
```diff
@@ -91,7 +91,7 @@ func ProcessImages(srcRegistry, destRegistry registry.RegistryOptions, appSlug s
 	return newImages, nil
 }
 
-func GetPrivateImages(upstreamDir string, checkedImages map[string]ImageInfo, allPrivate bool, dockerHubRegistry registry.RegistryOptions, useHelmInstall map[string]bool) ([]string, []k8sdoc.K8sDoc, error) {
+func GetPrivateImages(upstreamDir string, checkedImages map[string]ImageInfo, allPrivate bool, dockerHubRegistry registry.RegistryOptions, parentHelmChartPath string, useHelmInstall map[string]bool) ([]string, []k8sdoc.K8sDoc, error) {
```
sgalsaleh, line 94: this function is not just for helm charts though, a bit confusing that there is a parameter named `parentHelmChartPath`

sgalsaleh: maybe pass base? or convert this function to be something like
```
func (b *Base) GetPrivateImages(
```

pkg/image/builder.go
divolgin: each helm chart is treated as a separate base, so starting root will be nested inside one or more charts, and the map contains values relative to top level.

pkg/pull/pull.go
divolgin: This makes this loop pointless.  I don't actually know what it does.  How about just removing it?

pkg/rewrite/rewrite.go
divolgin: ditto :)

pkg/image/builder.go
sgalsaleh: ah that's why i asked if a double nested subchart would work here

pkg/pull/pull.go
sgalsaleh: yes, i agree. we should remove it

sgalsaleh: it does seem pointless indeed. no objection on removing it

pkg/image/builder.go
divolgin: `maybeParentHelmChartPath`?  it's actually a bit of an invariant.  without parent chart, it's empty string, which still works with the `useHelmInstall` map

sgalsaleh: i don't have any bright ideas. no strong objection to keep it as is. maybe aggregate all of these params in a new options struct or something

divolgin: the name makes sense to me given existing context

sgalsaleh (approved):

pkg/tests/pull/cases/subcharts/wantResults/overlays/midstream/charts/fluent-bit/kustomization.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+bases:
+- ../../../../base/charts/fluent-bit
+commonAnnotations:
+  kots.io/app-slug: my-app
+images:
```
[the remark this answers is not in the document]
divolgin: yes, this is current behavior.  not introduced in this PR

pkg/tests/pull/cases/subcharts/wantResults/overlays/midstream/kustomization.yaml
```diff
@@ -34,6 +34,14 @@ images:
   newName: fake-docker-proxy/busybox
 - name: library/busybox
   newName: fake-docker-proxy/busybox
+- name: cr.fluentbit.io/fluent/fluent-bit
```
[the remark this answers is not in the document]
divolgin: yep, same issue as the one in the parent chart.

PR #2931, opened 2022-06-29
Input CLEANUP_INTERVAL_HOURS for regression-test-cleanup workflow

emosbaugh:

If too many tests run in 24 hours we run into resource quotas in aws. Make this configurable.


e3b0c442 (approved):
Looks good to me

PR #2932, opened 2022-06-30
Set default CLEANUP_INTERVAL_HOURS for scheduled workflow

emosbaugh:

Scheduled workflow is failing https://github.com/replicatedhq/kots/runs/7130980544?check_suite_focus=true


e3b0c442 (approved):
Looks good to me

PR #2937, opened 2022-07-01
push-images command should not require username and password

divolgin:

`kots admin-console push-images` command should work with anonymous (or role bases) registries, which means users should be able to omit registry credentials.  Right now the command fails with:

```
Error: failed get registry login from secret: failed to get secret: secrets "kotsadm-private-registry" not found
```

Fixes #

## Steps to reproduce

Run `kots admin-console push-images kotsadm.tar.gz ttl.sh/somens` 

```release-note
* Fixes a bug that caused the [push-images](/reference/kots-cli-admin-console-push-images) command to fail when`--registry-password` and `--registry-username` are not specified for use with anonymous registries.
```

NONE

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -1268,6 +1268,35 @@ jobs:
           aws-secret-access-key: '${{ secrets.E2E_SUPPORT_BUNDLE_AWS_SECRET_ACCESS_KEY }}'
 
 
+  validate-kots-push-images-anonymous:
+    runs-on: ubuntu-20.04
+    needs: [ can-run-ci, build-kots, build-kurl-proxy, build-migrations, push-minio, push-postgres ]
+    strategy:
+      fail-fast: false
+      matrix:
+        k8s_version: [ v1.23.3-k3s1 ]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - uses: replicatedhq/action-k3s@main
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-test.yaml
```diff
@@ -1268,6 +1268,33 @@ jobs:
...
+
+      - name: download kots binary
+        uses: actions/download-artifact@v2
+        with:
+          name: kots
+          path: bin/
+
+      - run: chmod +x bin/kots
+
+      - name: run kots pull
```
sgalsaleh, line 1292:
```suggestion
      - name: run kots admin-console push-images
```

.github/workflows/build-test.yaml
```diff
@@ -1290,5 +1317,7 @@ jobs:
       - validate-yamlescape
       - validate-no-redeploy-on-restart
       - validate-kubernetes-installer-preflight
+      # cli-only tests
```
sgalsaleh, line 1320: wanna move the `validate-kots-pull` test to this section as well?

divolgin: kots pull does deploy stuff... this one is a one off right now because it does not run on a k3s matrix.  there is no kubernetes at all.

sgalsaleh: ah ok. makes sense.

sgalsaleh (dismissed):

sgalsaleh (approved):

PR #2955, opened 2022-07-11
fix: use parsed url for dashboard link href

cbodonnell:

This PR addresses an issue where users are unable to specify a different hostname in the url for the deployed application.  For example, the below spec would create a link with the url: `https://<admin-console-hostname>/some/path` instead of `https://www.example.com/some/path`:

```
  descriptor:
    links:
      - description: "Open the Application"
        url: https://www.example.com/some/path
```

In some cases, the user might have a different hostname for the deployed application, so the hostname in the url should reflect what was provided in the spec.

Additionally, if the user provided an invalid url, it could cause an uncaught error on the client-side and render the admin console unusable.

Fixes [SC-48585](https://app.shortcut.com/replicated/story/48585/open-msv-dashboard-button-appears-and-it-redirects-to-the-host-ip-or-fqdn-that-we-specified-in-the-config-ui-but-when-we)

Is there a reason why we use `window.location.hostname` [here](https://github.com/replicatedhq/kots/blob/v1.75.0/web/src/components/apps/AppStatus.jsx#L48) instead of the actual hostname provided in the spec?

#### Steps to reproduce

Add the following link to the k8s application spec and the resulting url will be different:
```
  descriptor:
    links:
      - description: "Open the Application"
        url: https://www.example.com/some/path
```

```release-note
Fixes an issue where dashboard links were being re-written to use the admin console hostname instead of the hostname provided in the application manifest.
```

NONE

### DIFF
--- a/web/src/components/apps/AppStatus.jsx
+++ b/web/src/components/apps/AppStatus.jsx
@@ -40,15 +40,15 @@ export default class AppStatus extends React.Component {
   };
 
   createDashboardActionLink = (uri) => {
-    const parsedUrl = new URL(uri);
-    let port;
-    if (!parsedUrl.port) {
-      port = "";
-    } else {
-      port = ":" + parsedUrl.port;
+    try {
+      const parsedUrl = new URL(uri);
+      if (parsedUrl.hostname === "localhost") {
+        parsedUrl.hostname = window.location.hostname;
+      }
+      return parsedUrl.href;
+    } catch (error) {
+      return "";
     }
-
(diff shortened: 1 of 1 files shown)


divolgin:
> Is there a reason why we use window.location.hostname [here](https://github.com/replicatedhq/kots/blob/v1.75.0/web/src/components/apps/AppStatus.jsx#L48) instead of the actual hostname provided in the spec?

Yes.  Normally these links point to the installed application.  Because it's an on-prem instance.

divolgin (approved):

PR #2958, opened 2022-07-13
Add workflow for cleaning up a single workspace [sc-51778]

e3b0c442:
Adds a reusable workflow for cleaning up a single e2e test workspace. This will eventually be called by a workflow which finds the workspaces ready to be cleaned up (and can also be called as one-off where needed).

### DIFF
--- a/.github/workflows/e2e-workspace-cleanup.yaml
+++ b/.github/workflows/e2e-workspace-cleanup.yaml
@@ -1,0 +1,102 @@
+name: e2e-workspace-cleanup
+
+on:
+  workflow_call:
+    inputs:
+      workspace:
+        type: string
+        description: Terraform workspace to turn down
+        required: true
+    secrets:
+      E2E_TESTIM_AWS_ACCESS_KEY_ID:
+        required: true
+      E2E_TESTIM_AWS_SECRET_ACCESS_KEY:
+        required: true
+      E2E_GH_PAT:
+        required: true
+  workflow_dispatch:
+    inputs:
+      workspace:
+        type: string
+        description: Terraform workspace to turn down
+        required: true
+
+concurrency: cleanup-${{ github.event.inputs.workspace }}
+
+env:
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-workspace-cleanup.yaml
```diff
@@ -0,0 +1,102 @@
...
+          repository: replicatedhq/kots-regression-automation
+          token: ${{ secrets.E2E_GH_PAT }}
+          path: automation
+          ref: main
+      - name: Setup Terraform
+        uses: hashicorp/setup-terraform@v2
+      - name: Initialize Terraform
+        env:
+          TF_WORKSPACE: default
+        run: terraform init -reconfigure -backend-config="key=jumpbox.tfstate"
```
syntastical, line 96: key is already hard coded in the terraform so you shouldn't need it here

.github/workflows/e2e-workspace-cleanup.yaml
```diff
@@ -0,0 +1,102 @@
...
+        with:
+          repository: replicatedhq/kots-regression-automation
+          token: ${{ secrets.E2E_GH_PAT }}
+          path: automation
+          ref: main
+      - name: Setup Terraform
+        uses: hashicorp/setup-terraform@v2
+      - name: Initialize Terraform
+        env:
+          TF_WORKSPACE: default
```
syntastical, line 95: This is hard coded in a couple places, is this just for testing?

syntastical (changes requested):

.github/workflows/e2e-workspace-cleanup.yaml
e3b0c442: Removed, thanks

.github/workflows/e2e-workspace-cleanup.yaml
e3b0c442: This overrides line 33 which is set by the parameter passed to the workflow

PR #2961, opened 2022-07-13
include credentials for pulling helm chart in ListApps response

stefanrepl:
This PR adds Credentials to ListApps response for helm applications.  These credentials are the username and password used when someone does a helm install of a given chart.  This is needed so we can provide proper `helm registry login` commands in the frontend.

Fixes #
https://app.shortcut.com/replicated/story/52569/inject-username-into-values-yaml-and-serve-in-api

```release-note

```

See credentials object now being returned for a given app in /apps response
Screen Shot 2022-07-13 at 3 18 27 PM


pkg/handlers/app.go
```diff
@@ -251,38 +268,44 @@ func (h *Handler) ListApps(w http.ResponseWriter, r *http.Request) {
...
-				allow, err := rbac.CheckAccess(r.Context(), defaultRoles, "read", fmt.Sprintf("app.%s", a.Slug), sess.Roles)
-				if err != nil {
-					logger.Error(errors.Wrapf(err, "failed to check access for app %s", a.Slug))
-					w.WriteHeader(http.StatusInternalServerError)
-					return
-				} else if !allow {
-					continue
-				}
-			}
+	defaultRoles := rbac.DefaultRoles() // TODO (ethan): this should be set in the handler
```
stefanrepl, line 288: The diff from here down is solely caused by moving this code out of an `else` block since the above `if` now returns on its own.

divolgin (approved):

PR #2968, opened 2022-07-14
Replace regresssion cleanup workflow with e2e cleanup workflow, remove legacy cleanup bits [sc-51778]

e3b0c442:
* Adds the new parallelized e2e-infra-cleanup workflow which replaces the regression-test-cleanup workflow
* Removes regression-test-cleanup workflow and action as they are no longer necessary
* Address a bug in `e2e-workspace-cleanup` where a `repository_dispatch` event would not have the workspace properly populated for deletion due to differences in how the inputs are passed.


.github/workflows/e2e-infra-cleanup.yaml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Setup Terraform
+        uses: hashicorp/setup-terraform@v2
+        with:
+          terraform_wrapper: false
+      - name: Initialize terraform
+        run: terraform init -reconfigure
+      - name: Get automation workspaces
+        id: get-workspaces
+        run: |
+          WORKSPACES=($(terraform workspace list | grep "nickm"))
```
syntastical, line 54: Remove the grep?

.github/workflows/e2e-infra-cleanup.yaml
```diff
@@ -0,0 +1,78 @@
...
+          repository: replicatedhq/kots-regression-automation
+          token: ${{ secrets.E2E_GH_PAT }}
+          path: automation
+          ref: main
+      - name: Setup Terraform
+        uses: hashicorp/setup-terraform@v2
+        with:
+          terraform_wrapper: false
+      - name: Initialize terraform
+        run: terraform init -reconfigure
```
syntastical, line 50: You shouldn't need to reconfigure here

.github/workflows/e2e-infra-cleanup.yaml
```diff
@@ -0,0 +1,78 @@
...
+        required: true
+  schedule:
+    - cron: "0 * * * *"
+
+concurrency: e2e-infra-cleanup
+
+env:
+  AWS_DEFAULT_REGION: us-east-1
+  AWS_ACCESS_KEY_ID: ${{ secrets.E2E_TESTIM_AWS_ACCESS_KEY_ID }}
+  AWS_SECRET_ACCESS_KEY: ${{ secrets.E2E_TESTIM_AWS_SECRET_ACCESS_KEY }}
```
syntastical, line 25: Are these populated for cron?

syntastical (changes requested):

.github/workflows/e2e-infra-cleanup.yaml
e3b0c442: nope, but the value's wrong. Nice catch.

.github/workflows/e2e-infra-cleanup.yaml
e3b0c442: Removed.

.github/workflows/e2e-infra-cleanup.yaml
e3b0c442: The secrets context should be available for all runs. Unfortunately not a way to test it without being on the main branch. We'll need to monitor the automated cleanup and address if it's failing.

syntastical (approved):

PR #2976, opened 2022-07-18
Fix nil pointer panics and update cache using watches

divolgin:

- Fix nil pointer panics when cache is not initialized
- Don't reload cache when the `/apps` endpoint is called
- Update cache using watches

Fixes #

```release-note
NONE
```

NONE

### DIFF
--- a/pkg/helm/cache.go
+++ b/pkg/helm/cache.go
@@ -1,0 +1,209 @@
+package helm
+
+import (
+	"context"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"sync"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/replicatedhq/kots/pkg/k8sutil"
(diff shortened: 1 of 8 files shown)


pkg/helm/cache.go
```diff
@@ -0,0 +1,209 @@
...
+				continue
+			}
+
+			AddHelmRelease(releaseInfo.Release.Name, releaseInfo)
+		}
+
+		go func(namespace string) {
+			err := watchSecrets(ctx, namespace, secretsSelector)
+			if err != nil {
+				logger.Errorf("Faied to watch secrets in ns %s and pplication cache will not be updated: %v", err)
```
stefanrepl, line 74:
```suggestion
				logger.Errorf("Faied to watch secrets in ns %s application cache will not be updated: %v", err)
```

stefanrepl (approved):

PR #2987, opened 2022-07-19
fix: non-semantic version automatic deployments do not work when version labels are not valid semantic versions

sgalsaleh:

Enabling automatic deployments when semantic versioning is not enabled does not automatically deploy new versions if the version label is not a valid semver. This PR will allow the automatic deployment to go through even if the version label is not a valid semantic version.

This PR also addresses another issue where any new version would be automatically deployed even if it's not a new upstream release (e.g. a config change, license change, etc...).

Fixes #

```release-note
Fixes an issue where non-semantic version automatic deployments do not work when version labels are not valid [SemVer](https://semver.org/).
```

NONE

### DIFF
--- a/pkg/updatechecker/updatechecker.go
+++ b/pkg/updatechecker/updatechecker.go
@@ -472,21 +472,23 @@ func autoDeploy(opts CheckForUpdatesOpts, clusterID string, autoDeploy apptypes.
 	}
 
 	currentVersion := appVersions.CurrentVersion
-	if currentVersion == nil || currentVersion.Semver == nil {
+	if currentVersion == nil {
 		return nil
 	}
 
 	var versionToDeploy *downstreamtypes.DownstreamVersion
 
-	// In the case semver is not required, we only need to check if the newest app version
-	// is greater than the current version.
 	if autoDeploy == apptypes.AutoDeploySequence {
-		if currentVersion.Sequence < appVersions.AllVersions[0].Sequence {
+		// semver is not required/enabled, we only need to check if the newest app version is newer than the current version.
+		// use cursor instead of sequence in order to only deploy newer upstream versions, and not versions created by config changes, license changes, etc...
+		currentCursor := currentVersion.Cursor
+		latestCursor := appVersions.AllVersions[0].Cursor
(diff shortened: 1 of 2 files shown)


pkg/updatechecker/updatechecker.go
```diff
@@ -472,21 +472,23 @@ func autoDeploy(opts CheckForUpdatesOpts, clusterID string, autoDeploy apptypes.
...
 
-	// In the case semver is not required, we only need to check if the newest app version
-	// is greater than the current version.
 	if autoDeploy == apptypes.AutoDeploySequence {
-		if currentVersion.Sequence < appVersions.AllVersions[0].Sequence {
+		// semver is not required/enabled, we only need to check if the newest app version is newer than the current version.
+		// use cursor instead of sequence in order to only deploy newer upstream versions, and not versions created by config changes, license changes, etc...
+		currentCursor := currentVersion.Cursor
+		latestCursor := appVersions.AllVersions[0].Cursor
+		if currentCursor != nil && latestCursor != nil && (*currentCursor).Before(*latestCursor) {
```
divolgin, line 486: Should this be determined by that weird sorting algorithm we have?

sgalsaleh: that algorithm will sort using sequences and not cursors, which is what causes the issue of automatically deploying config changes, license changes, etc...

this code block is only relevant for when semantic versioning is disabled any ways

sgalsaleh: also, `appVersions.AllVersions` is already sorted using that algorithm

divolgin (dismissed):

pkg/updatechecker/updatechecker.go
sgalsaleh: also, `appVersions.AllVersions[0]` is the top version after sorting, so it for sure is higher than or equal to the current version in terms of order. then the only question becomes whether or not it has a higher cursor if it's a variation of the same upstream version.

dexhorthy:
I'm Dexter Horthy and I approve this PR

divolgin (approved):

PR #3017, opened 2022-08-01
dont show add application in dropdown when  in helm mode

stefanrepl:
This PR hides "Add New Application" from dropdown when running in helm managed mode.

Fixes #
https://app.shortcut.com/replicated/story/51186/add-new-application-button-should-not-be-present-in-helm-mode

See attached screenshot from a KOTS instance running in helm managed mode , "Add Application" is not present

```release-note

```

NONE
Screen Shot 2022-08-01 at 2 15 30 PM

### DIFF
--- a/web/src/Root.jsx
+++ b/web/src/Root.jsx
@@ -82,6 +82,7 @@ class Root extends PureComponent {
     connectionTerminated: false,
     snapshotInProgressApps: [],
     errLoggingOut: "",
+    isHelmManaged: false,
   };
 
   /**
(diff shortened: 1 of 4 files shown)


web/src/Root.jsx
```diff
@@ -340,6 +362,7 @@ class Root extends PureComponent {
       connectionTerminated,
       errLoggingOut,
     } = this.state;
+    this.checkIsHelmManaged();
```
miaawong, line 365: If we only need to check this when the component first mounts (when the user first gets on the page) we can run this in componentDidMount. 
Otherwise, the app will continuously run checkIsHelmManaged. 

```
componentDidMount = async () => {
    this.onRootMounted();
    this.interval = setInterval(async () => await this.ping(), 10000);
    this.checkIsHelmManaged()
  };
```

web/src/Root.jsx
```diff
@@ -136,6 +137,27 @@ class Root extends PureComponent {
...
+  checkIsHelmManaged = async () => {
+    try {
+      const res = await fetch(`${process.env.API_ENDPOINT}/is-helm-managed`, {
+        headers: {
+          Authorization: Utilities.getToken(),
+          "Content-Type": "application/json",
+        },
+        method: "GET",
+      });
+      if (res.ok && res.status == 200) {
```
miaawong, line 149: a nit and you don't have to update but would recommend using `===` in JS because it helps with testing for strict equality

miaawong (commented):
Looks great! Let me know if you have any questions!

miaawong (approved):

PR #3025, opened 2022-08-03
feat: add support for alias field in helm chart dependencies

cbodonnell:

Currently there is an issue where helm charts with aliased dependencies will not produce the expected base file structure.  This PR modifies how the translation is performed between the base and upstream directories so that aliased dependencies are properly accounted for.

Screen Shot 2022-08-03 at 11 41 27 AM

Fixes [SC-49462](https://app.shortcut.com/replicated/story/49462/add-support-for-alias-in-chart-depenencies)

Needed to update the helm subchart-alias test since now the `Chart.yaml` will be present in directories created for aliased dependencies.

There was also some odd behavior with finding the subchart additional files.  Previously it was using `strings.Contains(upstreamPath, basePath)`, which was matching the test cases' upstream `./helm-charts/subchart-1/Chart.yaml` file.  This seemed like a coincidence since this directory is a file repository for this dependency, but the path just happens to contain the base path `charts/subchart-1/Chart.yaml` substring, so the test was passing.  This test was updated so that the files in the `helm-charts` directory would not be expected to appear in the base files.

#### Steps to reproduce

Install a helm chart with aliased dependencies.

```release-note
Adds support for the `alias` field in Helm chart dependencies.
```

I don't think so since it doesn't look like we ever explicitly called out that the `alias` field is not supported.


pkg/base/helm.go
```diff
@@ -243,67 +249,179 @@ func removeCommonPrefix(baseFiles []BaseFile) []BaseFile {
...
+	basePath := ""
+
+	// iterate over the charts in the dependency chain to build the base path
+	for _, chart := range charts {
+		// check for alias if not at the top level
+		if chart != "" {
+			upstreamPath = filepath.Join(upstreamPath, "charts", chart)
+			for _, dep := range deps.Dependencies {
+				// apply any aliases defined in the chart dependencies
+				if dep.Name == chart && dep.Alias != "" {
```
sgalsaleh, line 351: same chart can have multiple aliases

pkg/base/helm.go
```diff
@@ -243,57 +254,168 @@ func removeCommonPrefix(baseFiles []BaseFile) []BaseFile {
...
+	// iterate over the subcharts and find any that are aliased
+	for i := 1; i < len(charts); i++ {
+		parts := []string{}
+		chart := charts[i]
+		parts = append(parts, chart)
+		parentChart := charts[i-1]
+
+		// build the full parent chart path and check if the chart is aliased by its parent
+		parentChartYaml := filepath.Join(parentChart, "Chart.yaml")
+		for j := 0; j < i-1; j++ {
```
sgalsaleh, line 326: i think this would break if you have a sub-sub-sub-chart like:
```
charts/mariadb/charts/common/charts/anothercharthere
```
i'd suggest tracking the full path in a variable outside of the loop

pkg/base/helm.go
```diff
@@ -243,57 +254,168 @@ func removeCommonPrefix(baseFiles []BaseFile) []BaseFile {
...
-		basePathMap[basePath] = true
+func helmChartUpstreamPathToBasePaths(upstreamPath string, upstreamFileMap map[string][]byte) ([]string, error) {
+	charts := pathToCharts(upstreamPath)
+	basePathParts := make([][]string, len(charts))
+	basePathParts[0] = []string{""}
+	// iterate over the subcharts and find any that are aliased
+	for i := 1; i < len(charts); i++ {
+		parts := []string{}
+		chart := charts[i]
+		parts = append(parts, chart)
```
sgalsaleh, line 321: this would break if i only have aliases for this particular chart. i think this should be done [here](https://github.com/replicatedhq/kots/pull/3025/files?diff=split&w=0#diff-cc6bf3b28c8d864ab3c297aa234a72fc8a7e035ae541171f91de94971a10c40aR336-R338) if the chart exists but doesn't have an alias

pkg/base/helm.go
```diff
@@ -243,57 +254,168 @@ func removeCommonPrefix(baseFiles []BaseFile) []BaseFile {
...
+//  // "" => [""]
+//  // "charts/mariadb" => ["", "mariadb"]
+//  // "charts/mariadb/charts/common" => ["", "mariadb", "common"]
+func pathToCharts(path string) []string {
+	re := regexp.MustCompile(`\/?charts\/`)
+	return re.Split(path, -1)
+}
+
+// takes a slice of string slices representing the aliased base paths for each level of a chart
+// and returns a slice of strings representing all of the possible base paths
```
sgalsaleh, line 361: could you please add an example in the comment here of input => output?

pkg/base/helm.go
```diff
@@ -243,57 +254,168 @@ func removeCommonPrefix(baseFiles []BaseFile) []BaseFile {
...
+func flattenBasePathParts(basePathParts [][]string) []string {
```
sgalsaleh, line 362: if i understand the purpose of this function correctly, each index of `basePathParts` contains a list of aliases/names for the subcharts. wouldn't it be simpler to flatten those with a recursive function? i don't quite grasp the implementation of this to be honest.

cbodonnell: same as above

pkg/base/helm.go
cbodonnell: was able to do away with the need for this method entirely

pkg/base/helm.go
cbodonnell: Done. This also got me thinking that the whole `flattenBasePathParts` workflow is not necessary since I am already iterating through all of the possible base branches for the upstream path here.

pkg/base/helm.go
cbodonnell: If there is an upstream path for a chart/subchart, then it's dependency, but it is not aliased (or was aliased manually by the name of the subdirectory).  Dependencies can be managed either by including them in the `charts/` directory or in the `dependencies` field (which *is* where they might be aliased).  This is why I add the actual chart name here by default and then add any aliases for it later.

This is what I gathered from reading this: https://helm.sh/docs/topics/charts/#chart-dependencies

cbodonnell: Addressed this by using the dependencies list from the parent chart to determine the aliases (if any).  If the subchart is not present in the parent chart's dependencies list, then the chart name is used.

pkg/base/helm.go
```diff
@@ -326,19 +326,30 @@ func helmChartUpstreamPathToBasePaths(upstreamPath string, upstreamFileMap map[s
...
 				return nil, errors.Wrapf(err, "failed to unmarshal %s", parentChartYaml)
 			}
+			foundSubchartInDeps := false
 			for _, dep := range deps.Dependencies {
-				if dep.Name == subchart && dep.Alias != "" {
-					parts = append(parts, dep.Alias)
+				if dep.Name != subchart {
+					continue
 				}
+				foundSubchartInDeps = true
```
sgalsaleh, line 334: i think this needs to be something like this
```golang
if dep.Alias != "" {
	aliases = append(aliases, dep.Alias)
	foundSubchartInDeps = dep.Alias == subchart
} else {
	aliases = append(aliases, dep.Name)
	foundSubchartInDeps = dep.Name == subchart
}
```

sgalsaleh: the subchart is not necessarily found just because the name matches. it either has to match a listed alias, or a name _when_ the alias is not specified.

pkg/base/helm.go
```diff
@@ -243,57 +254,147 @@ func removeCommonPrefix(baseFiles []BaseFile) []BaseFile {
...
+				}
+				foundSubchartInDeps = true
+				if dep.Alias != "" {
+					aliases = append(aliases, dep.Alias)
+				} else {
+					aliases = append(aliases, dep.Name)
+				}
+			}
+			if !foundSubchartInDeps {
+				// if the subchart is not listed in the dependencies, then it is not aliased by the parent
```
sgalsaleh, line 342: how about?
```suggestion
				// subchart is not managed via dependencies
```

pkg/base/helm.go
```diff
@@ -243,57 +254,147 @@ func removeCommonPrefix(baseFiles []BaseFile) []BaseFile {
...
+			continue // only care about Chart.yaml files
+		}
+		upstreamPath := strings.TrimSuffix(upstreamFile.Path, "Chart.yaml")
+		upstreamPath = strings.TrimSuffix(upstreamPath, string(os.PathSeparator))
+		if _, ok := renderedUpstreamPaths[upstreamPath]; ok {
+			continue // already rendered this upstream path
+		}
+		basePaths := upstreamToBasePathsMap[upstreamPath]
+		for _, basePath := range basePaths {
+			logger.Infof("adding missing dependency %s to base path %s\n", upstreamFile.Path, basePath)
```
sgalsaleh, line 397: let's not log this? since it will just be noise and is not relevant to the user. or log it as debug?

sgalsaleh (approved):

pkg/base/helm.go
sgalsaleh: because this will always happen

sgalsaleh (approved):

sgalsaleh (approved):

PR #3034, opened 2022-08-04
implement deploy logs view for helm managed mode

stefanrepl:
This PR implements deploy logs API response for KOTS when running in helm managed mode

Fixes #
https://app.shortcut.com/replicated/story/54248/404-returned-when-trying-to-view-deploy-logs

```release-note
     Show deploy logs of helm charts when running in [Helm managed mode (Alpha)](/vendor/helm-install).
```

Loom: https://www.loom.com/share/9ea4867e71fc4087b572d9f7d53b7c20

### DIFF
--- a/pkg/handlers/downstream.go
+++ b/pkg/handlers/downstream.go
@@ -5,8 +5,11 @@ import (
 	"strconv"
 
 	"github.com/gorilla/mux"
+	"github.com/replicatedhq/kots/pkg/api/downstream/types"
+	"github.com/replicatedhq/kots/pkg/helm"
 	"github.com/replicatedhq/kots/pkg/logger"
 	"github.com/replicatedhq/kots/pkg/store"
+	"github.com/replicatedhq/kots/pkg/util"
 )
 
 type GetDownstreamOutputResponse struct {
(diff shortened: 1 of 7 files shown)


divolgin:
Presumably this is received in the watch loop with secret info.  Is there a revision number?  The output should match the correct revision, but this will be shown all revisions of the app.  The API handler receives sequence number.  It should be used the retrieve matching output.

pkg/handlers/downstream.go
```diff
@@ -26,24 +29,40 @@ func (h *Handler) GetDownstreamOutput(w http.ResponseWriter, r *http.Request) {
...
 
-	a, err := store.GetStore().GetAppFromSlug(appSlug)
-	if err != nil {
-		logger.Error(err)
-		w.WriteHeader(http.StatusInternalServerError)
-		return
-	}
+	if util.IsHelmManaged() {
+		app := helm.GetHelmApp(appSlug)
+		releaseSecret, err := helm.GetChartSecret(app.Release.Name, app.Release.Namespace, mux.Vars(r)["sequence"])
```
divolgin, line 41: if app == nil, return NotFound

pkg/helm/charts.go
```diff
@@ -59,6 +59,40 @@ func (v InstalledReleases) Less(i, j int) bool {
...
+		"version": version,
+	}
+	listOpts := metav1.ListOptions{
+		LabelSelector: labels.SelectorFromSet(selectorLabels).String(),
+	}
+
+	secrets, err := clientSet.CoreV1().Secrets(namespace).List(context.TODO(), listOpts)
+	if err != nil {
+		if kuberneteserrors.IsNotFound(err) {
+			return &helmrelease.Release{}, nil
```
divolgin, line 80: `return nil, nil` or just remove this check.  the caller should check that the app is not found.

The problem is that `Release` contains pointers that will be nil here and will cause panics.  `Info` is a pointer and `releaseSecret.Info.Description` will panic.

pkg/handlers/downstream.go
stefanrepl: done

pkg/helm/charts.go
stefanrepl: done, returning nil, nil

pkg/handlers/downstream.go
```diff
@@ -26,24 +29,44 @@ func (h *Handler) GetDownstreamOutput(w http.ResponseWriter, r *http.Request) {
...
+			w.WriteHeader(http.StatusNotFound)
+			return
+		}
+		releaseSecret, err := helm.GetChartSecret(app.Release.Name, app.Release.Namespace, mux.Vars(r)["sequence"])
+		if err != nil {
+			logger.Error(err)
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+		if releaseSecret.Info.Status != "failed" {
```
divolgin, line 51: this needs to check for releaseSecret  != nil now

stefanrepl: ah of course, thanks

web/src/components/snapshots/SnapshotDetails.jsx
```diff
@@ -86,23 +86,23 @@ class SnapshotDetails extends Component {
       tooltip: {
         custom: function ({ series, seriesIndex, dataPointIndex, w }) {
           return (
-            '<div class="arrow_box">' +
-            '<p class="u-textColor--primary u-fontSize--normal u-fontWeight--medium">' +
+            "<div class=\"arrow_box\">" +
```
stefanrepl, line 89: my IDE's JS linter did these quoting changes, but seems to make sense

divolgin (approved):

PR #3036, opened 2022-08-05
Use defaul collectors and analyzers, if support bundle spec cannot be downloaded

divolgin:

For applications that ship without support bundle spec, support bundle is now collected using the default collectors and analyzers.

Fixes #

```release-note
* Fixes an issue that caused an error when clicking the Analyze application button on the [Troubleshoot page](/enterprise/troubleshooting-an-app#create-a-support-bundle-using-the-admin-console) when application does not include a support bundle spec in [Helm managed mode (Alpha)](/vendor/helm-install).
```

NONE

### DIFF
--- a/pkg/supportbundle/supportbundle.go
+++ b/pkg/supportbundle/supportbundle.go
@@ -212,13 +212,20 @@ func getKotsKindsForHelmApp(app *apptypes.HelmApp) (*kotsutil.KotsKinds, error) 
 	specURL := strings.TrimSuffix(app.ChartPath, fmt.Sprintf("/%s", app.Release.Chart.Name()))
 	upstreamSupportBundle, upstreamRedactors, err := getSupportBundleSpecFromOCI(license.Spec.LicenseID, specURL)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to download support bundle spec from %s", specURL)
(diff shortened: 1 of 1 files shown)


pkg/supportbundle/supportbundle.go
```diff
@@ -212,13 +212,20 @@ func getKotsKindsForHelmApp(app *apptypes.HelmApp) (*kotsutil.KotsKinds, error)
...
+		// don't return; use default collectors/analyzers when spec cannot be downloaded
+		logger.Infof("failed to download support bundle spec from %s: %v", specURL, err)
 	}
 
 	kotsKinds := kotsutil.EmptyKotsKinds()
 	kotsKinds.License = license
-	kotsKinds.SupportBundle = upstreamSupportBundle
-	kotsKinds.Redactor = upstreamRedactors
+
+	if upstreamSupportBundle != nil {
```
stefanrepl, line 222: do we need to set a default if upstream bundle is nil?

divolgin: `EmptyKotsKinds` has some initialization logic.  Didn't want to override whatever it could do with nil values on error.

stefanrepl (approved):

PR #3040, opened 2022-08-08
Store Helm config values in ConfigValues kind

divolgin:

Values entered on the config screen should be stored in ConfigValues, not in Config.

Fixes #

```release-note
NONE
```

NONE

### DIFF
--- a/pkg/helm/config_values.go
+++ b/pkg/helm/config_values.go
@@ -1,0 +1,139 @@
+package helm
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+
+	"github.com/pkg/errors"
+	kotsv1beta1 "github.com/replicatedhq/kots/kotskinds/apis/kots/v1beta1"
+	kotsscheme "github.com/replicatedhq/kots/kotskinds/client/kotsclientset/scheme"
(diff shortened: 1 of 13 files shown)


pkg/reporting/app.go
```diff
@@ -66,7 +69,7 @@ func initFromHelm() error {
 	}
 
 	for _, secret := range secrets.Items {
-		helmRelease, err := helm.HelmReleaseFromSecretData(secret.Data["release"])
```
divolgin, line 69: reporting should not be importing packages like helm

pkg/helm/config_values.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	if gvk.Group != "kots.io" || gvk.Version != "v1beta1" || gvk.Kind != "ConfigValues" {
+		return configValues, errors.Errorf("%q is not a valid ConfigValues GVK", gvk.String())
+	}
+
+	return decoded.(*kotsv1beta1.ConfigValues), nil
+}
+
+// TODO: this function is not thread safe
```
sgalsaleh, line 65: why not just add a mutex then?

sgalsaleh (approved):

PR #3044, opened 2022-08-08
Add apps versions hooks

rianfowler:
- adds hooks for metadata, versions, and apps calls. These hooks are not used in components yet so this should have no impact on existing functionality. 

n/a

- this passes current unit tests 

## Steps to reproduce
n/a

n/a

none


web/src/features/AppVersionHistory/api/getVersions.jsx
```diff
@@ -0,0 +1,197 @@
...
+
+  // if kots managed
+  return _getVersionsSelectorForKotsManaged;
+}
+
+function useVersions({
+  currentPage,
+  refetchInterval,
+  pageSize,
+  _getVersions = getVersions,
```
miaawong, line 156: out of curiosity.. what does adding the underscore do?

miaawong (approved):
lgtm! nice job!

web/src/features/AppVersionHistory/api/getVersions.jsx
rianfowler: It's just a convention for renaming imports injected as arguments. So rather than trying to mock an import during unit tests, I can pass it in directly as an argument. 

in other languages `_variable` conveys "this is private, don't use it" to other developers.

PR #3052, opened 2022-08-10
Support images names formatted as docker.io/image:tag

divolgin:

1. Images hosted in docker hub can be formatted in different ways:
```
busybox:1.32
library/busybox:1.32
docker.io/library/busybox:1.32
```
This PR adds support for `docker.io/busybox:1.32`, which is also a valid way to format the same image ref.

2. Removes duplicate entries from the `images` in the kustomization.yaml files
3. Fixes references similar to ECR's 111122222333.dkr.ecr.us-east-1.amazonaws.com/frontend:v1.0.1, which were being prefixed with `docker.io` (but caused no harm)

Fixes #

```release-note
* Fixes an issue that caused images formatted as `docker.io/image:tag` to not be [rewritten](https://docs.replicated.com/vendor/operator-referencing-images) when [upgrading applications](https://docs.replicated.com/enterprise/updating-apps) in airgapped environments.
```

NONE

### DIFF
--- a/pkg/base/replicated_test.go
+++ b/pkg/base/replicated_test.go
@@ -6,14 +6,15 @@ import (
 	"reflect"
 	"testing"
 
+	envsubst "github.com/drone/envsubst/v2"
 	"github.com/replicatedhq/kots/pkg/logger"
 	"github.com/replicatedhq/kots/pkg/template"
 	upstreamtypes "github.com/replicatedhq/kots/pkg/upstream/types"
+	"github.com/replicatedhq/kots/pkg/util"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
-
 	"k8s.io/client-go/kubernetes/scheme"
 )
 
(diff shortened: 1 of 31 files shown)


sgalsaleh (dismissed):

pkg/base/replicated_test.go
```diff
@@ -946,14 +953,14 @@ metadata:
     helm.sh/chart: postgresql-10.13.8
     app.kubernetes.io/instance: postgresql
     app.kubernetes.io/managed-by: Helm
-  namespace: 
+  namespace: ${POD_NAMESPACE}
```
divolgin, line 956: This and `TestGetenv` allow running tests in okteto.  This change is unrelated to the fix in the PR.

pkg/k8sutil/kustomization.go
```diff
@@ -38,12 +38,20 @@ func (s kustPatches) Less(i, j int) bool {
 }
 
 func WriteKustomizationToFile(kustomization kustomizetypes.Kustomization, file string) error {
-	// we remove newTags from here... because...
 	cleanedImages := []kustomizetypes.Image{}
+
+	// Remove tags and deduplicate image list.
+	// Tags are removed because we don't want kustomize to change tags, only image names
```
sgalsaleh, line 44: should this be true for digests as well?

sgalsaleh (approved):

pkg/k8sutil/kustomization.go
divolgin: The comment refers to this line `image.NewTag = ""` (which isn't new), so I guess digests would be there too.

divolgin: Actually, no.  Kustomize image has a separate field for `Digest`.

sgalsaleh: yeah, you shouldn't be able to change a digest anyways now that i think about it.

PR #3054, opened 2022-08-10
refactor fetchRenderedApplicationTree to use promise.all

rianfowler:
- prerequisite to upgrading to react 18. this component has a logic error that react no longer ignores 

NONE- this is just a refactor 

NONE


web/src/components/watches/DownstreamWatchVersionDiff.jsx
```diff
@@ -47,12 +47,6 @@ class DownstreamWatchVersionDiff extends React.Component {
         } else {
           this.setState({ secondApplicationTree: files });
         }
-        if (
-          this.state.firstApplicationTree.files &&
-          this.state.secondApplicationTree.files
-        ) {
-          this.setState({ loadingFileTrees: false });
-        }
```
rianfowler, line 55: My guess is that this is checking to see if all of the data is loaded before removing the loading state. `setState` is async though so I'm not sure how this ever worked.

miaawong (approved):

PR #3064, opened 2022-08-16
Include config values in deployment values file

divolgin:

In Helm managed installs, this allows tracking config values for each release revision.  This information can be used to show correct config values on the config screen.

Fixes #

This change is not backwards compatible with existing installs and charts.

```release-note
* Adds support for displaying application configuration values the [config page](/vendor/config-screen-about#admin-console-config-tab) for all revisions revision in [Helm managed mode (Alpha)](/vendor/helm-install).
```

NONE


pkg/helm/charts.go
```diff
@@ -306,73 +323,120 @@ func GetKotsKinds(helmApp *apptypes.HelmApp) (kotsutil.KotsKinds, error) {
...
-	secrets, err := clientSet.CoreV1().Secrets(helmApp.Namespace).List(context.TODO(), listOpts)
+	secrets, err := clientSet.CoreV1().Secrets(util.PodNamespace).List(context.TODO(), listOpts)
 	if err != nil {
-		return errors.Wrapf(err, "failed to find release secret with options %v", selectorLabels)
+		return kotsKinds, errors.Wrap(err, "failed to list secrets")
 	}
 
 	if len(secrets.Items) != 1 {
-		return errors.Wrapf(err, "expected 1 secret but found %d with options %v", len(secrets.Items), selectorLabels)
+		return kotsKinds, errors.Errorf("expected to mathc 1 secret, but found %d", len(secrets.Items))
```
stefanrepl, line 357:
```suggestion
		return kotsKinds, errors.Errorf("expected to match 1 secret, but found %d", len(secrets.Items))
```

pkg/helm/charts.go
```diff
@@ -306,73 +323,120 @@ func GetKotsKinds(helmApp *apptypes.HelmApp) (kotsutil.KotsKinds, error) {
...
+	}
+
+	return kotsKinds, nil
+}
+
+func getKotsKindsFromConfigSecret(secret *corev1.Secret) (kotsutil.KotsKinds, error) {
+	kotsKinds := kotsutil.EmptyKotsKinds()
+
+	licenseData := secret.Data["license"]
+	if len(licenseData) != 0 {
```
stefanrepl, line 415: Shouldn't there be an error returned if `len(licenseData) == 0`?  I would assume `len(licenseData) == 0` would mean the chart isnt managed by us/this process and we likely should not continue.

divolgin: That would probably be ideal.  But right now license is added dynamically by kots.  If release is rolled back, license data will be wiped out.  Maybe we could throw a license download in here if it's missing, and then it would be a food reason to fail here.

stefanrepl: ah gotchya, thanks

pkg/helm/cache.go
```diff
@@ -163,23 +156,6 @@ func helmAppFromSecret(secret *corev1.Secret) (*apptypes.HelmApp, error) {
 	_, helmApp.IsConfigurable = configSecret.Data["config"] // TODO: also check if there are any config items
 	helmApp.ChartPath = string(configSecret.Data["chartPath"])
 
-	if configSecret.Data["license"] == nil {
```
divolgin, line 166: `helmAppFromSecret` no longer mutates anything.  License is download is now triggered explicitly when license is needed.

stefanrepl (dismissed):

sgalsaleh (approved):

PR #3070, opened 2022-08-17
fix failing fetch

rianfowler:
- fixes minor bug related to fetching app for dashboard. I believe this is what is causing the tests to fail in the react 18 upgrade. 

- none

none

### DIFF
--- a/web/src/components/apps/Dashboard.jsx
+++ b/web/src/components/apps/Dashboard.jsx
@@ -717,9 +717,15 @@ class Dashboard extends Component {
                           startASnapshot={this.startASnapshot}
                           startSnapshotOptions={this.state.startSnapshotOptions}
                           startSnapshotErr={this.state.startSnapshotErr}
-                          startSnapshotErrorMsg={this.state.startSnapshotErrorMsg}
-                          snapshotInProgressApps={this.props.snapshotInProgressApps}
-                          selectedSnapshotOption={this.state.selectedSnapshotOption}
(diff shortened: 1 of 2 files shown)


web/src/components/apps/Dashboard.jsx
```diff
@@ -176,6 +176,14 @@ class Dashboard extends Component {
 
   getAppDashboard = () => {
     return new Promise((resolve, reject) => {
+      // this function is in a repeating callback that terminates when
+      // the promise is resolved
+
+      // TODO: use react-query to refetch this instead of the custom repeater
+      if (!this.props.app) {
```
rianfowler, line 183: The dashboard is hanging on the react 18 upgrade branch. The data is loading but the loader just stays. I'm not able to recreate it locally so I'm trying out a couple of small tweaks. 

This gives a warning locally because the fetch fails. The fetch fails because it's called before the app state is set. This function is called in a repeater so exiting when the app data is not available with return is fine.

web/src/components/apps/Dashboard.jsx
```diff
@@ -620,253 +628,264 @@ class Dashboard extends Component {
...
     }
 
-    if (!app) {
-      return (
-        <div className="flex-column flex1 alignItems--center justifyContent--center">
-          <Loader size="60" />
-        </div>
-      );
-    }
-
```
rianfowler, line 630: Moved this down into the main render

miaawong (approved):

PR #3077, opened 2022-08-18
refactor Dashboard into SFC

rianfowler:
- converts dashboard into functional component 
- this is needed to use api hooks (incoming in another PR)

none

## Steps to reproduce
none
none this is just a refactor

none


web/src/components/apps/Dashboard.jsx
```diff
@@ -26,65 +26,106 @@ const COMMON_ERRORS = {
...
+  const setWatchState = (app) => {
+    setAppName(app.name);
+    setIconUri(app.iconUri);
+    setCurrentVersion(app?.downstream.currentVersion);
+    setDownstream(app.downstream);
+    setLinks(app.downstream?.links);
+  };
 
-  getAppLicense = async (app) => {
+  const getAppLicense = async (app) => {
```
rianfowler, line 128: assigning all of this functions to const on every render will be inefficient but it should be fine for now.

web/src/components/apps/Dashboard.jsx
```diff
@@ -94,58 +135,33 @@ class Dashboard extends Component {
...
-          });
+          setAppLicense(body.license);
+          setGettingAppLicenseErrMsg("");
         } else if (body.error) {
-          this.setState({
-            appLicense: {},
-            gettingAppLicenseErrMsg: body.error,
-          });
+          setAppLicense({});
+          setGettingAppLicenseErrMsg(body.error);
```
rianfowler, line 151: This is a common theme- basically every `setState` had to be replaced with a hook.

web/src/components/apps/Dashboard.jsx
```diff
@@ -164,33 +180,47 @@ class Dashboard extends Component {
...
+    if (app) {
+      setWatchState(app);
+      getAppLicense(app);
+    }
+    return () => {
+      updateChecker.stop();
+      getAppDashboardJob.stop();
+      fetchAppDownstreamJob.stop();
+    };
+  }, []);
```
rianfowler, line 204: this replaces `componentDidMount` and `componentWillUnmount`.  Original [here](https://github.com/replicatedhq/kots/blob/0276eb1d5cf21073dbc77225bc319fb93ae6fdd4/web/src/components/apps/Dashboard.jsx#L126-L145). 

```
  componentDidMount() {
    const { app } = this.props;

    if (app?.isAirgap && !this.state.airgapUploader) {
      this.getAirgapConfig();
    }

    this.state.updateChecker.start(this.updateStatus, 1000);
    this.state.getAppDashboardJob.start(this.getAppDashboard, 2000);
    if (app) {
      this.setWatchState(app);
      this.getAppLicense(app);
    }
  }

  componentWillUnmount() {
    this.state.updateChecker.stop();
    this.state.getAppDashboardJob.stop();
    this.state.fetchAppDownstreamJob.stop();
  }
```

web/src/components/apps/Dashboard.jsx
```diff
@@ -164,33 +180,47 @@ class Dashboard extends Component {
...
       // this function is in a repeating callback that terminates when
       // the promise is resolved
 
       // TODO: use react-query to refetch this instead of the custom repeater
-      if (!this.props.app) {
+      if (!app) {
         return;
       }
 
-      if (this.props.cluster?.id == "" && this.props.isHelmManaged === true) {
```
rianfowler, line 187: replacing `this.props` had to be done one by one so I could check for anything like `this.props.cluster` and `this.state.cluster`

web/src/components/apps/Dashboard.jsx
```diff
@@ -26,65 +26,106 @@ const COMMON_ERRORS = {
...
+  makeCurrentVersion,
+  redeployVersion,
+  isHelmManaged,
+  ping,
+  snapshotInProgressApps,
+  isVeleroInstalled,
+  isBundleUploading,
+  onUploadNewVersion,
+}) => {
+  const [appName, setAppName] = useState("");
```
alicenstar, line 50: In the future, if it's easier, when I've refactored class components that have a lot of state, I use `useReducer` like this:

```
const [state, setState] = useReducer((prevState, nextState) => ({
	...prevState,
	...nextState
}, {
	* your initial state here *
});
```

as it allows me to easily swap out `this.setState(...)` for `setState(...)` instead of having to manually declare all of these. What you have obviously works too, just wanted to share in case you were unaware.

alicenstar (approved):
Great work refactoring!

PR #3088, opened 2022-08-22
make repeater for initial load

rianfowler:
- refactors dashboard to use a `Repeater` to control its initial data loading instead of a react lifecycle method 
- this fix is needed for the react 18 upgrade 

none - refactor 

none -refactor


web/src/components/apps/AppDetailPage.jsx
```diff
@@ -55,13 +56,6 @@ class AppDetailPage extends Component {
       this.props;
     const { app, loadingApp } = this.state;
 
-    // Used for a fresh reload
-    if (history.location.pathname === "/apps") {
-      this.checkForFirstApp();
-      // updates state but does not cause infinite loop because app navigates away from /apps
-      return;
-    }
```
rianfowler, line 63: This doesn't work in react 18. It's inside of a lifecycle method that is called differently after the upgrade. I'm putting it into a repeater job on mount for now. We will refactor this initial loading stuff soon and completely remove that concern from this component.

PR #3114, opened 2022-08-25
When redeploying or updating config, pin helm command to the correct chart version

divolgin:

When a Helm release is redeployed or when a release config changes, the `helm upgrade` command must be pinned to the correct release version, otherwise the latest release (if available) will be pulled down and installed.

Fixes #

## Steps to reproduce

1. Install a chart version
2. Ship a new chart release, but don't "Check for updates"
3. Try redeploying the installed version or updating the config

```release-note
* Fixes an issue that caused inadvertent application upgrades while redeploying or updating the config of the currently installed release in [Helm managed mode (Alpha)](/vendor/helm-install)
```

NONE

### DIFF
--- a/pkg/helm/charts.go
+++ b/pkg/helm/charts.go
@@ -1,14 +1,12 @@
 package helm
 
 import (
-	"bufio"
 	"bytes"
 	"compress/gzip"
 	"context"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"sort"
 	"strconv"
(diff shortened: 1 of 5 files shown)


web/src/components/shared/modals/HelmDeployModal.jsx
```diff
@@ -12,15 +12,12 @@ function makeDeployCommand({
   if (revision) {
     return `helm rollback ${appSlug} ${revision}`;
   }
+
   if (showDownloadValues) {
-    if (!version) {
```
divolgin, line 16: no check because version must always be present

rianfowler (approved):

pkg/helm/charts.go
```diff
@@ -122,88 +120,104 @@ func ListChartVersions(releaseName string, namespace string) ([]InstalledRelease
...
 
 	sort.Sort(sort.Reverse(releases))
 
 	return releases, nil
 }
 
-func HelmReleaseFromSecretData(data []byte) (*helmrelease.Release, error) {
-	base64Reader := base64.NewDecoder(base64.StdEncoding, bytes.NewReader(data))
-	gzreader, err := gzip.NewReader(base64Reader)
+func GetChartVersion(releaseName string, revision int64, namespace string) (*InstalledRelease, error) {
```
stefanrepl, line 137: a nit pick but could we rename to `GetChartByVersion`?

stefanrepl (approved):

PR #3120, opened 2022-08-26
Live-render config in Helm mode

divolgin:

Live config render functionality in Helm managed mode

Fixes #

```release-note
* Fixed an issue that could cause templates to be shown on the [config page](/vendor/config-screen-about#admin-console-config-tab) instead of actual rendered values in [Helm managed mode (Alpha)](/vendor/helm-install)
```

NONE

### DIFF
--- a/pkg/handlers/config.go
+++ b/pkg/handlers/config.go
@@ -220,6 +220,8 @@ func (h *Handler) LiveAppConfig(w http.ResponseWriter, r *http.Request) {
 		Success: false,
 	}
 
+	appSlug := mux.Vars(r)["appSlug"]
+
 	liveAppConfigRequest := LiveAppConfigRequest{}
 	if err := json.NewDecoder(r.Body).Decode(&liveAppConfigRequest); err != nil {
 		logger.Error(err)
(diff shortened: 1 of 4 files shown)


pkg/handlers/config.go
```diff
@@ -228,88 +230,103 @@ func (h *Handler) LiveAppConfig(w http.ResponseWriter, r *http.Request) {
...
+	appInfo := template.ApplicationInfo{Slug: app.GetSlug()}
+	renderedConfig, err := kotsconfig.TemplateConfigObjects(kotsKinds.Config, configValues, appLicense, &kotsKinds.KotsApplication, localRegistry, &versionInfo, &appInfo, kotsKinds.IdentityConfig, app.GetNamespace(), false)
 	if err != nil {
 		logger.Error(err)
 		liveAppConfigResponse.Error = "failed to render templates"
 		JSON(w, http.StatusInternalServerError, liveAppConfigResponse)
 		return
 	}
 
-	if renderedConfig != nil {
```
stefanrepl, line 308: lets keep a nil check, `kotsconfig.TemplateConfigObjects` can return a nil err and nil Config object.

stefanrepl (approved):

PR #3142, opened 2022-09-01
start status informers when operator starts

cbodonnell:

This PR fixes an issue where status informers were not being re-applied when the kotsadm pod restarted.  This is because status informers were only applied after an application deployment, so when the pod restarted, the application status would remain as it was prior to the restart and would only update once a version was deployed/redeployed.  This PR changes this so that status informers for the currently deployed release (if there is one) will be applied when the operator starts up.

Fixes [SC-55879](https://app.shortcut.com/replicated/story/55879/status-informers-do-not-get-re-applied-after-kotsadm-pod-restarts)

Refactored the `Start(...)` method for the operator to accept the client and store interfaces as arguments so we can more easily unit test it.

#### Steps to reproduce

1. Deploy an application with status informers
2. Restart the kotsadm pod
3. Delete some application pods and observe that the statuses update properly

```release-note
Fixes an issue where application [status informers](/vendor/admin-console-display-app-status) would not update if the Admin Console pod was restarted.
```

NONE

### DIFF
--- a/pkg/apiserver/server.go
+++ b/pkg/apiserver/server.go
@@ -18,6 +18,7 @@ import (
 	"github.com/replicatedhq/kots/pkg/informers"
 	"github.com/replicatedhq/kots/pkg/k8sutil"
 	"github.com/replicatedhq/kots/pkg/operator"
+	"github.com/replicatedhq/kots/pkg/operator/client"
 	"github.com/replicatedhq/kots/pkg/persistence"
 	"github.com/replicatedhq/kots/pkg/policy"
 	"github.com/replicatedhq/kots/pkg/rbac"
@@ -87,7 +88,13 @@ func Start(params *APIServerParams) {
 	}
 
 	if !util.IsHelmManaged() {
-		if err := operator.Start(params.AutocreateClusterToken); err != nil {
+		client := &client.Client{
+			TargetNamespace:   util.AppNamespace(),
+			ExistingInformers: map[string]bool{},
+			HookStopChans:     []chan struct{}{},
+		}
+		store := store.GetStore()
+		if err := operator.Start(params.AutocreateClusterToken, client, store); err != nil {
 			log.Println("error starting the operator")
 			panic(err)
 		}
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -46,27 +46,26 @@ import (
 )
 
 var OperatorClient client.ClientInterface
+var operatorStore store.Store
 var clusterID string
 var deployMtxs = map[string]*sync.Mutex{} // key is app id
 
-func Start(clusterToken string) error {
+func Start(clusterToken string, operatorClient client.ClientInterface, store store.Store) error {
 	logger.Debug("starting the operator")
 
-	OperatorClient = &client.Client{
-		TargetNamespace:   util.AppNamespace(),
-		ExistingInformers: map[string]bool{},
-		HookStopChans:     []chan struct{}{},
-	}
+	OperatorClient = operatorClient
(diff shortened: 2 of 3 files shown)


pkg/operator/operator.go
```diff
@@ -46,27 +46,26 @@ import (
 )
 
 var OperatorClient client.ClientInterface
+var operatorStore store.Store
```
divolgin, line 49: Why make it a global variable instead of calling GetStore whenever it's needed?

cbodonnell: Looking for a way to have the operator use the store that is passed in when initialized, which would help a lot with unit testing.  How about if we convert the operator to a struct that implements the `Start` and `Shutdown` method.  This way a client and store can be passed in when creating and there won't be a need for any of these globals?  Something like:
```
o := operator.New(client, store, etc...)
o.Start()
defer o.Shutdown()
```

divolgin (approved):

pkg/operator/operator.go
```diff
@@ -46,27 +46,26 @@ import (
...
 
-	id, err := store.GetStore().GetClusterIDFromDeployToken(clusterToken)
+	operatorStore = store
+	id, err := operatorStore.GetClusterIDFromDeployToken(clusterToken)
 	if err != nil {
 		return errors.Wrap(err, "failed to get cluster id from deploy token")
 	}
 	clusterID = id
 
+	startStatusInformers()
```
sgalsaleh, line 68: is there a reason this is not in a go routine?

pkg/apiserver/server.go
```diff
@@ -87,7 +88,13 @@ func Start(params *APIServerParams) {
 	}
 
 	if !util.IsHelmManaged() {
-		if err := operator.Start(params.AutocreateClusterToken); err != nil {
+		client := &client.Client{
```
sgalsaleh, line 91: i don't see any values passed in this context, why can't this be initialized inside the operator?

pkg/apiserver/server.go
```diff
@@ -87,7 +88,13 @@ func Start(params *APIServerParams) {
...
+			TargetNamespace:   util.AppNamespace(),
+			ExistingInformers: map[string]bool{},
+			HookStopChans:     []chan struct{}{},
+		}
+		store := store.GetStore()
```
sgalsaleh, line 96: why can't this be [here](https://github.com/replicatedhq/kots/pull/3142/files#diff-99c0290799daf9abc6240df64063e20bfaf67b371577b67ac7eec6f4725622ffR49) and written as:
```go
var operatorStore = store.GetStore()
```

pkg/operator/operator.go
```diff
@@ -395,8 +394,65 @@ func applyStatusInformers(a *apptypes.App, sequence int64, kotsKinds *kotsutil.K
 	return nil
 }
 
+func startStatusInformers() {
```
sgalsaleh, line 397: personally i'd call this `resumeStatusInformers`

pkg/operator/operator.go
```diff
@@ -395,8 +394,65 @@ func applyStatusInformers(a *apptypes.App, sequence int64, kotsKinds *kotsutil.K
...
+		return
+	}
+	for _, app := range apps {
+		if err := startStatusInformersForApp(app); err != nil {
+			logger.Error(errors.Wrapf(err, "failed to start status informers for app %s in cluster %s", app.ID, clusterID))
+		}
+	}
+}
+
+func startStatusInformersForApp(app *apptypes.App) error {
```
sgalsaleh, line 410: ditto

PR #3146, opened 2022-09-06
refactor operator to improve unit testing

cbodonnell:

This PR refactors the operator so that it can be unit tested more easily.  It creates an `Init(...)` method that initializes the operator with a client and store interface.  This will allow for mocks to be used when unit testing.

n/a

Added a global reference to the initialized operator within the package that is accessed using the `MustGetOperator()` method.  This method will panic if the operator has not yet been initialized.  This was necessary because the deploy method is used elsewhere in the code and we need a way to access the currently initialized operator.

```release-note
NONE
```

NONE

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -45,38 +45,60 @@ import (
 	"k8s.io/client-go/kubernetes/scheme"
 )
 
-var OperatorClient client.ClientInterface
-var operatorStore store.Store
-var clusterID string
-var deployMtxs = map[string]*sync.Mutex{} // key is app id
+var (
+	operator *Operator
+)
 
-func Start(clusterToken string, operatorClient client.ClientInterface, store store.Store) error {
+type Operator struct {
+	client       client.ClientInterface
+	store        store.Store
+	clusterToken string
+	clusterID    string
+	deployMtxs   map[string]*sync.Mutex // key is app id
+}
+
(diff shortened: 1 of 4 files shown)


pkg/operator/operator.go
```diff
@@ -45,38 +45,60 @@ import (
...
+
+type Operator struct {
+	client       client.ClientInterface
+	store        store.Store
+	clusterToken string
+	clusterID    string
+	deployMtxs   map[string]*sync.Mutex // key is app id
+}
+
+func Init(client client.ClientInterface, store store.Store, clusterToken string) *Operator {
```
sgalsaleh, line 60: why even use an `Init` function? i'd suggest just having the caller initialize a new operator object (e.g. `Operator{`) and set the necessary exported fields, other fields can remain unexported.

pkg/operator/operator.go
```diff
@@ -45,38 +45,60 @@ import (
...
+	operator = &Operator{
+		client:       client,
+		store:        store,
+		clusterToken: clusterToken,
+		deployMtxs:   map[string]*sync.Mutex{},
+	}
+	return operator
+}
+
+func MustGetOperator() *Operator {
```
sgalsaleh, line 70: this won't be needed with my suggestion above

sgalsaleh (approved):

PR #3148, opened 2022-09-06
Redirect to apps when user already authenticated

miaawong:
#### Which issue(s) this PR fixes: If a user visits ${KOTS_URL}/secure-console when already logged in, it should redirect them to /apps. 

Fixes # https://app.shortcut.com/replicated/story/47115/add-redirect-if-user-visits-kots-login-when-already-authenticated

```release-note
Redirects user to /apps if they are already logged in. 
```

NONE

### DIFF
--- a/web/src/components/SecureAdminConsole.jsx
+++ b/web/src/components/SecureAdminConsole.jsx
@@ -217,13 +217,17 @@ class SecureAdminConsole extends React.Component {
   async componentDidMount() {
     window.addEventListener("keydown", this.submitForm);
 
-    const token = window.localStorage.getItem("token");
-    if (Utilities.isLoggedIn()) {
+    const localStorageToken = window.localStorage.getItem("token");
+    const cookieToken = Utilities.getCookie("token");
+    const cookieSessionRole = Utilities.getCookie("session_role");
+    const localStorageSessionRole = window.localStorage.getItem("session_role");
+    if (Utilities.isLoggedIn() || cookieToken) {
       // this is a redirect from identity service login
       // strip quotes from token (golang adds them when the cookie value has spaces, commas, etc..)
       const loginData = {
-        token: token.replace(/"/g, ""),
-        sessionRoles: window.localStorage.getItem("session_roles"),
+        token:
+          localStorageToken.replace(/"/g, "") || cookieToken.replace(/"/g, ""),
+        sessionRoles: localStorageSessionRole || cookieSessionRole,
       };
       await this.completeLogin(loginData);
     }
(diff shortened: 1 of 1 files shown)


web/src/components/SecureAdminConsole.jsx
```diff
@@ -214,23 +214,18 @@ class SecureAdminConsole extends React.Component {
...
   };
 
-  componentDidMount() {
+  async componentDidMount() {
     window.addEventListener("keydown", this.submitForm);
 
-    const token = Utilities.getCookie("token");
-    if (token) {
+    const token = window.localStorage.getItem("token");
+    if (Utilities.isLoggedIn()) {
```
miaawong: Utilities has a method to check if there is an existing token

web/src/components/SecureAdminConsole.jsx
```diff
@@ -214,23 +214,22 @@ class SecureAdminConsole extends React.Component {
...
 
-    const token = Utilities.getCookie("token");
-    if (token) {
+    const localStorageToken = window.localStorage.getItem("token");
+    const cookieToken = Utilities.getCookie("token");
+    const cookieSessionRole = Utilities.getCookie("session_role");
+    const localStorageSessionRole = window.localStorage.getItem("session_role");
+    if (Utilities.isLoggedIn() || cookieToken) {
       // this is a redirect from identity service login
       // strip quotes from token (golang adds them when the cookie value has spaces, commas, etc..)
```
miaawong, line 226: in the completeLogin function, we set token and session_role to localStorage. That's how we know if a user is logged in. 
However previously, on componentDidMount we retrieve token by cookies. (I believe that's why it would always prompt user to login) token is never set in cookies. Therefore, token was empty. 

However, I'm not comfortable removing cookies from this code, as the comment above mention it is a redirect from identity service login. Let me know what you think. :)

web/src/components/SecureAdminConsole.jsx
```diff
@@ -214,23 +214,22 @@ class SecureAdminConsole extends React.Component {
...
+          localStorageToken.replace(/"/g, "") || cookieToken.replace(/"/g, ""),
+        sessionRoles: localStorageSessionRole || cookieSessionRole,
       };
-      this.completeLogin(loginData).then((loggedIn) => {
-        if (loggedIn) {
-          Utilities.removeCookie("token");
-          Utilities.removeCookie("session_roles");
-        }
-      });
+      await this.completeLogin(loginData);
```
miaawong, line 232: adding await allows fetching and redirect to happen sooner. Otherwise, you will get a flash of the login page before being redirected.

alicenstar (approved):

PR #3152, opened 2022-09-07
auto update checks for helm charts

stefanrepl:
This PR implements checking for updates for helm charts when running in helm managed mode.

Fixes #
https://app.shortcut.com/replicated/story/54318/helm-mode-automatic-update-checks-don-t-happen

This also includes some minor FE changes.  Namely, default to default not custom, and set certain text in the modal based on what mode KOTS is running in.

```release-note
     Adds ability to automatically check for new chart versions that are available when running in [Helm managed mode (Alpha)](/vendor/helm-install).
```

### DIFF
--- a/pkg/app/types/helm_app.go
+++ b/pkg/app/types/helm_app.go
@@ -15,7 +15,8 @@ type HelmApp struct {
 	IsConfigurable    bool
 	ChartPath         string
 	CreationTimestamp time.Time
-
+	UpdateCheckerSpec string
+	UpdateStatus      string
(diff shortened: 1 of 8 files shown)


divolgin (changes requested):
It doesn't look like it persists the schedule.  How about migrating `update_checker_spec` values from the app table into a config map and using that for both helm and classic installs?

stefanrepl:
> It doesn't look like it persists the schedule. How about migrating `update_checker_spec` values from the app table into a config map and using that for both helm and classic installs?

ok

PR #3155, opened 2022-09-08
auto update checks for helm charts

stefanrepl:
This PR implements checking for updates for helm charts when running in helm managed mode.

Fixes #
https://app.shortcut.com/replicated/story/54318/helm-mode-automatic-update-checks-don-t-happen

This also includes some minor FE changes.  Namely, default to default not custom, and set certain text in the modal based on what mode KOTS is running in.

```release-note
     Adds ability to automatically check for new chart versions that are available when running in [Helm managed mode (Alpha)](/vendor/helm-install).
```

### DIFF
--- a/pkg/handlers/update_checker_spec.go
+++ b/pkg/handlers/update_checker_spec.go
@@ -3,9 +3,13 @@ package handlers
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"net/http"
 
+	"github.com/replicatedhq/kots/pkg/helm"
+	"github.com/replicatedhq/kots/pkg/kotsadm/types"
 	"github.com/replicatedhq/kots/pkg/kotsutil"
+	"github.com/replicatedhq/kots/pkg/util"
 
 	"github.com/gorilla/mux"
 	apptypes "github.com/replicatedhq/kots/pkg/app/types"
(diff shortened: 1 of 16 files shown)


pkg/handlers/update_checker_spec.go
```diff
@@ -35,6 +44,91 @@ func (h *Handler) ConfigureAutomaticUpdates(w http.ResponseWriter, r *http.Reque
...
+
+	selectorLabels := map[string]string{
+		"metadata.name": "kotsadm-confg",
+	}
+	listOpts := metav1.ListOptions{
+		FieldSelector: labels.SelectorFromSet(selectorLabels).String(),
+	}
+
+	configMaps, err := clientSet.CoreV1().ConfigMaps(util.PodNamespace).List(context.TODO(), listOpts)
+	if err != nil && !kuberneteserrors.IsNotFound(err) {
```
divolgin, line 63: List doesn't return "not found" errors.  Use `Get` instead of `List` like here https://github.com/replicatedhq/kots/blob/main/pkg/kotsadm/main.go#L1084

pkg/handlers/update_checker_spec.go
```diff
@@ -35,6 +44,91 @@ func (h *Handler) ConfigureAutomaticUpdates(w http.ResponseWriter, r *http.Reque
...
+			if err != nil {
+				logger.Error(err)
+				updateCheckerSpecResponse.Error = "failed to parse cron spec"
+				JSON(w, 500, updateCheckerSpecResponse)
+				return
+			}
+		}
+
+		cm.Data[fmt.Sprintf("%s-spec", release.GetID())] = cronSpec
+		_, err = clientSet.CoreV1().ConfigMaps(util.PodNamespace).Update(context.TODO(), cm, metav1.UpdateOptions{})
```
divolgin, line 113: If config map doesn't exist, this will fail.  This functionality can remain in store package and then all app types can use it.  There are some store functions that use configmaps and secrets already: https://github.com/replicatedhq/kots/blob/6502604a16cac170434a56e19257fbe33c73c97d/pkg/store/kotsstore/session_store.go#L198

pkg/handlers/update_checker_spec.go
```diff
@@ -85,9 +179,11 @@ func (h *Handler) ConfigureAutomaticUpdates(w http.ResponseWriter, r *http.Reque
 		}
 	}
 
-	if err := store.GetStore().SetUpdateCheckerSpec(foundApp.ID, cronSpec); err != nil {
+	cm.Data[fmt.Sprintf("%s-spec", foundApp.ID)] = cronSpec
```
divolgin, line 182: key name should be less generic since this is config map is not a specifically for update checks.  How about `"update-schedule-%s"`

pkg/updatechecker/updatechecker.go
```diff
@@ -183,18 +219,94 @@ type UpdateCheckRelease struct {
...
+	}
+
+	selectorLabels := map[string]string{
+		"metadata.name": "kotsadm-confg",
+	}
+	listOpts := metav1.ListOptions{
+		FieldSelector: labels.SelectorFromSet(selectorLabels).String(),
+	}
+
+	configMaps, err := clientSet.CoreV1().ConfigMaps(util.PodNamespace).List(context.TODO(), listOpts)
```
divolgin, line 234: Same, use `Get` instead of `List` and maybe move into store package

pkg/updatechecker/updatechecker.go
```diff
@@ -292,9 +404,10 @@ func CheckForUpdates(opts CheckForUpdatesOpts) (*UpdateCheckResponse, error) {
 	}
 
 	// this is to avoid a race condition where the UI polls the task status before it is set by the goroutine
-	status := fmt.Sprintf("%d Updates available...", ucr.AvailableUpdates)
-	if err := store.SetTaskStatus("update-download", status, "running"); err != nil {
```
divolgin, line 296: These store functions should still be used.  Maybe change key name so it doesn't collide for different apps.

pkg/updatechecker/updatechecker.go
```diff
@@ -183,18 +219,94 @@ type UpdateCheckRelease struct {
...
+			})
+		}
+
+		ucr := UpdateCheckResponse{
+			AvailableUpdates:  int64(len(updates)),
+			AvailableReleases: updates,
+			DeployingRelease:  nil,
+		}
+
+		cm.Data[fmt.Sprintf("%s-status", opts.AppID)] = "running"
```
divolgin, line 296: There are store functions for task status already

pkg/handlers/update_checker_spec.go
stefanrepl: ah thanks, done

pkg/handlers/update_checker_spec.go
stefanrepl: done

pkg/updatechecker/updatechecker.go
stefanrepl: looking into moving into store

divolgin (approved):

PR #3169, opened 2022-09-14
fix: Headless/unattended install hangs in embedded clusters with recent kubernetes versions

sgalsaleh:

Fixes an issue where headless/unattended installs hang in embedded clusters with recent Kubernetes versions.

Fixes [SC-55766](https://app.shortcut.com/replicated/story/55766/headless-unattended-install-does-not-work-in-kurl-clusters)

```release-note
Fixes an issue where headless/unattended installs hang in embedded clusters with recent Kubernetes versions.
```

NONE

### DIFF
--- a/pkg/k8sutil/portforward.go
+++ b/pkg/k8sutil/portforward.go
@@ -63,6 +63,32 @@ func IsPortAvailable(clientset kubernetes.Interface, port int) (bool, error) {
 }
 
 func FindFreePort(clientset kubernetes.Interface) (int, error) {
+	portCh := make(chan int)
+	errCh := make(chan error)
+
+	go func() {
+		port, err := findFreePort(clientset)
+		if err != nil {
+			errCh <- err
+			return
+		}
+		portCh <- port
+	}()
+
+	select {
+	case err := <-errCh:
+		if err != nil {
+			return 0, err
+		}
+	case port := <-portCh:
+		return port, nil
+	case <-time.After(30 * time.Second):
+	}
+
+	return 0, errors.New("Timed out trying to find a free port")
+}
+
+func findFreePort(clientset kubernetes.Interface) (int, error) {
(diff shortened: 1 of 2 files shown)


pkg/k8sutil/portforward.go
```diff
@@ -26,39 +27,90 @@ import (
...
+		return 0, errors.Wrap(err, "failed to get free port")
+	}
+
+	isPortAvailable, err := IsPortAvailable(clientset, port)
+	if err != nil {
+		return 0, errors.Wrap(err, "failed to check if port is available")
+	}
+
+	if !isPortAvailable {
+		return FindFreePort(clientset)
```
divolgin, line 77: Wanna put a limit on this recursion?  If `isPortAvailable` stays false, this will spin forever.

sgalsaleh: yeah, i knew you were gonna say that... okay :)

pkg/k8sutil/portforward.go
```diff
@@ -26,39 +27,90 @@ import (
...
 	host := net.JoinHostPort("127.0.0.1", strconv.Itoa(port))
 	server, err := net.Listen("tcp4", host)
 	if err != nil {
-		return false
+		return false, nil
 	}
 
 	server.Close()
-	return true
+	return true, nil
```
divolgin, line 62: Should we still do a net.Listen check at this point?  This function does not account for ports being in use by system processes that run outside of cluster.

sgalsaleh: it accounts for ports being used where the kots commands are being run though, and that's still useful?

divolgin: yeah, nm, the code i had in mind is already in this function

divolgin (dismissed):

divolgin (approved):

PR #3175, opened 2022-09-15
Use values from currently deployed version to display on config for pending version

divolgin:

When viewing config for a pending update, all previously configured values are missing and have to be re-entered manually.  With this change we use values for the currently deployed version to pre-populate the config screen.

This also adds caching for pending Helm releases to make Config page and live config faster.

Fixes #

```release-note
* Fixes an issue that caused values to be missing on the [config page](/vendor/config-screen-about#admin-console-config-tab) for pending updates in [Helm managed mode (Alpha)](/vendor/helm-install)
```

NONE

### DIFF
--- a/pkg/helm/updates.go
+++ b/pkg/helm/updates.go
@@ -4,6 +4,9 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
 	"sort"
 	"strings"
 	"sync"
@@ -13,6 +16,8 @@ import (
 	imagetypes "github.com/containers/image/v5/types"
 	"github.com/pkg/errors"
 	apptypes "github.com/replicatedhq/kots/pkg/app/types"
+	"github.com/replicatedhq/kots/pkg/logger"
+	"go.uber.org/zap"
 	helmgetter "helm.sh/helm/v3/pkg/getter"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/helm/updates.go
```diff
@@ -159,8 +162,13 @@ func removeDuplicates(tags []string) []string {
 	return u
 }
 
-// TODO: Add caching
 func PullChartVersion(helmApp *apptypes.HelmApp, licenseID string, version string) (*bytes.Buffer, error) {
+	data, _ := getUpdateChartFromCache(helmApp, version)
```
stefanrepl, line 166: why ignore error?

pkg/helm/updates.go
```diff
@@ -171,10 +179,58 @@ func PullChartVersion(helmApp *apptypes.HelmApp, licenseID string, version strin
...
+	return bytes.NewBuffer(b), nil
+}
+
+func saveUpdateChartInCache(helmApp *apptypes.HelmApp, version string, data *bytes.Buffer) (*bytes.Buffer, error) {
+	if updateCacheDir == "" {
+		dirName, err := ioutil.TempDir("", "chart-updates-")
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to create temp dir")
+		}
+		updateCacheDir = dirName
```
stefanrepl, line 215: so `updateCacheDir` will be lost if kots gets rebooted..then kots will create a new cache dir causing a memory leak..can we add a cleanup defer statement for the cache dir when kots is rebooted?

pkg/helm/updates.go
divolgin: cache is not critical.  if it's not there, try to download the release

stefanrepl: might as well log the error then no?  seems like a trivial TODO

pkg/helm/updates.go
divolgin: it's in the ephemeral storage.  it will be deleted kotsadm is restarted

stefanrepl (approved):

PR #3198, opened 2022-09-20
Add GVK for LintConfig

garcialuis:
Adds GVK for LintConfig

[SC-59102](https://app.shortcut.com/replicated/story/59102/kots-lintconfig)

```release-note
NONE
```

NONE


pavansokkenagaraj (dismissed):
let's check with kots team on backward compatibility and to check how a new CRD is handled for old kots version

cbodonnell (dismissed):

cbodonnell:
> let's check with kots team on backward compatibility and to check how a new CRD is handled for old kots version

this should be fine since it is `apiVersion: kots.io/v1beta1` and will have the proper exclude annotation injected upstream.  it would also be filtered out by [this method](https://github.com/replicatedhq/kots/blob/main/pkg/base/base.go#L279-L303).

cbodonnell:
Looks like we need to rebase this branch since our tests have been updated since it was created.

PR #3199, opened 2022-09-20
Sync license before checking for updates in Helm mode

divolgin:

When checking for updates in Helm managed mode, the license will be synced first.

Fixes #

```release-note
* Fixes an issue that caused the license to not be synced when checking for application updates in [Helm managed mode (Alpha)](/vendor/helm-install).
```

### DIFF
--- a/pkg/handlers/license.go
+++ b/pkg/handlers/license.go
@@ -22,6 +22,7 @@ import (
 	"github.com/replicatedhq/kots/pkg/online"
 	installationtypes "github.com/replicatedhq/kots/pkg/online/types"
 	"github.com/replicatedhq/kots/pkg/registry"
+	"github.com/replicatedhq/kots/pkg/replicatedapp"
 	"github.com/replicatedhq/kots/pkg/store"
 	"github.com/replicatedhq/kots/pkg/updatechecker"
 	"github.com/replicatedhq/kots/pkg/util"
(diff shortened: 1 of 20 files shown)


rianfowler (approved):
FE 👍🏽

pkg/handlers/license.go
```diff
@@ -130,42 +131,21 @@ func (h *Handler) SyncLicense(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
-		currentLicense, err := helm.GetChartLicenseFromSecretOrDownload(helmApp)
+		isSynced, err = helm.SyncLicense(helmApp)
 		if err != nil {
-			syncLicenseResponse.Error = "failed to get license from secret"
+			syncLicenseResponse.Error = "failed to get sync helm license"
```
stefanrepl, line 136:
```suggestion
			syncLicenseResponse.Error = "failed to sync helm license"
```

pkg/helm/license.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	err = SaveChartLicenseInSecret(helmApp, latestLicenseData.LicenseBytes)
+	if err != nil {
+		return isSynced, errors.Wrap(err, "failed to update helm license")
+	}
+
+	if currentLicense == nil {
+		isSynced = true
+	} else if currentLicense.Spec.LicenseSequence != latestLicenseData.License.Spec.LicenseSequence {
```
stefanrepl, line 41: should this be `<` isntead of `!=`?  Id imagine when we sync we would always expect the new sequence to be greater.

pkg/updatechecker/updatechecker.go
```diff
@@ -210,16 +210,18 @@ func CheckForUpdates(opts CheckForUpdatesOpts) (ucr *UpdateCheckResponse, finalE
 	}
 
 	finishedChan := make(chan error)
-	if opts.Wait {
-		defer close(finishedChan)
-	}
+	defer func() {
+		if opts.Wait || finalError != nil {
+			defer close(finishedChan)
```
stefanrepl, line 215: shouldnt this just be `close(finishedChan)` since its already within a defer statement?  I dont think the extra defer here is necessary, unless i am missing somehting.

pkg/helm/license.go
divolgin: yes, it's always increasing

pkg/updatechecker/updatechecker.go
divolgin: this spins up a go routine if `Wait` is true.  The channel needs to stay open until that terminates.

stefanrepl (approved):

PR #3200, opened 2022-09-20
Rianfowler/sc 58965/add typechecking to AppConfig

rianfowler:
- refactor: adds typechecking to config page 

https://app.shortcut.com/replicated/story/58965/adding-typechecking-to-top-level-routes-routes-in-the-navbar

NONE - refactor that adds type checking 

NONE


web/src/types/index.ts
```diff
@@ -0,0 +1,44 @@
+export type App = {
```
rianfowler, line 1: I moved the app type into here. There are a few references that needed to be updated.

web/src/features/AppConfig/components/AppConfig.tsx
```diff
@@ -590,31 +648,23 @@ class AppConfig extends Component {
           </SideNavWrapper>
           <div className="ConfigArea--wrapper">
             <UseIsHelmManaged>
-              {({ data = {} }) => {
-                const { isHelmManaged } = data;
-
-                const {
-                  mutate: saveConfig,
-                  isLoading: isSaving,
-                  isError: saveError,
```
rianfowler, line 599: had to remove some unused variables.

web/src/features/AppConfig/components/AppConfig.tsx
```diff
@@ -590,31 +648,23 @@ class AppConfig extends Component {
...
-                } = useSaveConfig({
-                  appSlug: this.getSlug(),
-                });
+              {({ data = {} }: { data: { isHelmManaged?: boolean } }) => {
+                const { isHelmManaged: isHelmManagedFromHook } = data;
 
-                const {
-                  download,
-                  clearError: clearDownloadError,
-                  error: downloadError,
```
rianfowler, line 607: this never returns an error. i think it did previously but i updated the fetch and didn't have typechecking in place to complain. added a todo in the `useDownloadValues` hook to fix this

web/src/features/AppConfig/components/AppConfig.tsx
```diff
@@ -670,26 +720,26 @@ class AppConfig extends Component {
...
                           downloadClicked={download}
-                          downloadError={downloadError}
-                          isDownloading={isDownloading}
                           hideHelmDeployModal={() => {
                             this.setState({ showHelmDeployModal: false });
                             clearDownloadError();
                           }}
                           registryUsername={
-                            this.props?.app?.credentials?.username
+                            this.props?.app?.credentials?.username || ""
```
rianfowler, line 728: Could probably remove the optional chaining here but I think we should continue typing App out some more and then figure out a way to initialize it with sane defaults.

web/src/features/AppConfig/components/AppConfig.tsx
```diff
@@ -781,4 +831,8 @@ class AppConfig extends Component {
   }
 }
 
-export default withRouter(AppConfig);
+// TODO: fix this type
+// eslint-disable-next-line
+const AppConfigWithRouter: any = withRouter(AppConfig);
+
+export default AppConfigWithRouter;
```
rianfowler, line 838: This is a mess.  I need to dig into this to figure out how to type components decorated with react router

web/webpack.config.js
```diff
@@ -49,6 +49,7 @@ module.exports = function (env) {
         "@src": path.resolve(__dirname, "src"),
         "@features": path.resolve(__dirname, "src/features"),
         "@stores": path.resolve(__dirname, "src/stores"),
+        "@types": path.resolve(__dirname, "src/types/index"),
```
rianfowler, line 52: will need to update these aliases if we decide to break the types file up. for now i think it'll be fine to shove everything into index.ts

web/src/features/AppConfig/components/AppConfig.tsx
miaawong: don't know if this is helpful but from researching, it seems like you can use RouteComponentProps https://www.reactandtypescript.dev/examples/react-router#withrouter

web/src/types/index.ts
miaawong: I like it! Makes it more reusable.

miaawong (approved):

PR #3212, opened 2022-09-22
Rianfowler/sc 58965/adding typechecking to dashboard component

rianfowler:
- adds typechecks to Dashboard component 

https://app.shortcut.com/replicated/story/58965/adding-typechecking-to-top-level-routes-routes-in-the-navbar

NONE- refactor 

None


web/src/components/apps/Dashboard.tsx
```diff
@@ -26,47 +26,136 @@ const COMMON_ERRORS = {
   "no such host": "No such host",
 };
 
-class Dashboard extends Component {
-  state = {
```
rianfowler, line 30: moved this into the constructor. took me hours to figure out why some of my typechecks were giving me weird errors. it was because the initial state was getting set outside of a constructor + parts of the state were getting set in a timeout. I don't think TS could see that the state was changing inside of that timeout so it was always just falling back to the types set here.

web/src/components/apps/Dashboard.tsx
```diff
@@ -174,17 +263,17 @@ class Dashboard extends Component {
...
-  getAppDashboard = () => {
+  getAppDashboard = (): Promise<void> => {
     return new Promise((resolve, reject) => {
       // this function is in a repeating callback that terminates when
       // the promise is resolved
 
       // TODO: use react-query to refetch this instead of the custom repeater
       if (!this.props.app) {
-        return;
+        resolve();
```
rianfowler, line 273: this was a little confusing but hopefully it passes the tests. if you define `resolve` inside of a promise you have to use it instead of return.

web/src/components/apps/Dashboard.tsx
```diff
@@ -289,12 +378,12 @@ class Dashboard extends Component {
         method: "GET",
       });
       if (res.ok && res.status == 200) {
-        const app = await res.json();
-        if (!isAwaitingResults(app.downstream.pendingVersions)) {
+        const appResponse = await res.json();
```
rianfowler, line 381: had to rename this since `app` was declared in the outer scope

web/src/components/apps/Dashboard.tsx
```diff
@@ -311,11 +400,13 @@ class Dashboard extends Component {
       }
     } catch (err) {
       console.log(err);
+      const errorMessage =
+        err instanceof Error
```
rianfowler, line 404: `catch (err)` can always be `any` so we have to do some type narrowing checks here. I check for `instanceof Error` before trying to access `err.message`. not sure if this is the best way to do it but it seems to work.

web/src/components/apps/Dashboard.tsx
```diff
@@ -386,23 +477,27 @@ class Dashboard extends Component {
     const params = {
       appId: this.props.app?.id,
     };
-    this.state.airgapUploader.upload(
+
+    // TODO: remove after adding type to airgap uploader
+    // eslint-disable-next-line
+    // @ts-ignore
+    this.state.airgapUploader?.upload(
```
rianfowler, line 484: We really have to stop putting callbacks on the state.

web/src/components/apps/Dashboard.tsx
```diff
@@ -445,14 +540,14 @@ class Dashboard extends Component {
     this.setState({ viewAirgapUploadError: !this.state.viewAirgapUploadError });
   };
 
-  toggleViewAirgapUpdateError = (err) => {
+  toggleViewAirgapUpdateError = (err?: string) => {
     this.setState({
       viewAirgapUpdateError: !this.state.viewAirgapUpdateError,
-      airgapUpdateError: !this.state.viewAirgapUpdateError ? err : "",
+      airgapUpdateError: !this.state.viewAirgapUpdateError && err ? err : "",
```
rianfowler, line 546: `err` is optional so we should check if it exists before trying to use it.

web/src/components/apps/Dashboard.tsx
```diff
@@ -634,6 +733,9 @@ class Dashboard extends Component {
     const appResourcesByState = this.getAppResourcesByState();
     const hasStatusInformers = this.checkStatusInformers();
 
+    // TODO: remove the type casting
```
rianfowler, line 736:
```suggestion
```

rianfowler: just going to leave this in for now. i used this to try to figure out an issue. will refactor when we convert to SFC and use hooks.

web/src/components/apps/Dashboard.tsx
```diff
@@ -785,7 +886,7 @@ class Dashboard extends Component {
             {this.state.viewAirgapUpdateError && (
               <Modal
                 isOpen={this.state.viewAirgapUpdateError}
-                onRequestClose={this.toggleViewAirgapUpdateError}
+                onRequestClose={() => this.toggleViewAirgapUpdateError()}
```
rianfowler, line 889: `toggleViewAirgapUpdateError` sometimes takes a string but never takes a mouseEvent. wrapping this function so it's clear that we're not trying to use what `onRequestClose` calls with

web/src/components/apps/Dashboard.tsx
```diff
@@ -830,8 +931,8 @@ class Dashboard extends Component {
                           {Utilities.toTitleCase(state)}
                         </p>
                         {appResourcesByState?.statesMap[state]?.map(
-                          (resource, i) => (
-                            <div key={`${resource?.name}-${i}`}>
+                          (resource, j) => (
```
rianfowler, line 934: another variable shadowing fix

web/src/types/index.ts
```diff
@@ -1,28 +1,75 @@
...
   username: string;
   password: string;
 };
 
-export type DownStream = {
+export type ResourceStates = {
+  kind: string;
+  name: string;
+  namespace: string;
+  // from https://github.com/replicatedhq/kots/blob/84b7e4e0e9275bb200a36be69691c4944eb8cf8f/pkg/appstate/types/types.go#L10-L14
```
rianfowler, line 51: this is the only time i've looked at the go code to get an enum. there aren't many available but anytime you see a hardcoded string like "ready" being used it's a good idea to check.

web/src/types/index.ts
```diff
@@ -36,6 +83,11 @@ export type KotsParams = {
   slug: string;
 };
 
+export type DashboardActionLink = {
```
rianfowler, line 86: some of these names are not great but i figure we'll have a proper chat about kots domain language once we get all of the stuff typed out

web/src/components/apps/Dashboard.tsx
jodillion: I've used something like this before, can we use the type Error and define that in our commonly used types interface file?
interface Error {
    name: string;
    message: string;
    stack?: string;
}

web/src/types/index.ts
```diff
@@ -1,28 +1,75 @@
 export type App = {
```
jodillion, line 1: okay I could be wrong here but I THINK that if you use interfaces instead of types in this file then you don't need to export them. then you also don't need to import them anywhere and can use them everywhere within your code

web/src/components/apps/Dashboard.tsx
```diff
@@ -26,47 +26,136 @@ const COMMON_ERRORS = {
...
+import {
+  App,
+  AppLicense,
+  Downstream,
+  DashboardResponse,
+  DashboardActionLink,
+  KotsParams,
+  ResourceStates,
+  Version,
+} from "@types";
```
jodillion, line 39: see comment at the bottom about importing from types and using interfaces instead!

web/src/components/apps/Dashboard.tsx
```diff
@@ -26,47 +26,136 @@ const COMMON_ERRORS = {
...
+  startSnapshotErrorMsg: string;
+  startSnapshotOptions: SnapshotOption[];
+  updateChecker: Repeater;
+  uploadProgress: number;
+  uploadResuming: boolean;
+  uploadSize: number;
+  uploadingAirgapFile: boolean;
+  viewAirgapUpdateError: boolean;
+  viewAirgapUploadError: boolean;
+};
```
jodillion, line 103: when we convert this to using functional components instead of class, this will save on a bunch of lines of extra code as we can define the types inline with useState :)

miaawong (approved):

web/src/components/apps/Dashboard.tsx
rianfowler: That makes sense. I added "Add a shared Error type" to our shortcut [doc](https://app.shortcut.com/replicated/write/IkRvYyI6I3V1aWQgIjYzMjRkZDFjLTg0OWUtNDE1NS04YWMyLTU2MmU1ZjI3MWVkMyI=).

web/src/types/index.ts
rianfowler: That makes sense. I've been following the examples in bullet proof react and [they're just exporting the types](https://github.com/alan2207/bulletproof-react/blob/master/src/types/index.ts). 

I think having global types is cleaner and we should do that long term. I think we have to go through a re-naming process for some of this stuff first.

Like we have a `Dashboard` component and `DashboardResponse` type. But we also have an `App` type that maybe should be called `AppResponse`. I'm not sure what kind of naming conventions we should use yet.

PR #3214, opened 2022-09-22
Miawong/sc 58965/adding typechecking to license

miaawong:

Fixes # https://app.shortcut.com/replicated/story/58965/adding-typechecking-to-top-level-routes-routes-in-the-navbar

```release-note
NONE - refactor
```

NONE


rianfowler (dismissed):
awesome work

web/src/types/index.ts
```diff
@@ -42,3 +43,9 @@ export type Version = {
   sequence: number;
   versionLabel?: string;
 };
+export type Entitlement = {
+  title: string;
+  value: string;
+  label: string;
+  valueType: "Text" | "Boolean" | "Integer" | "String";
```
jodillion, line 50: just curious, why are these as strings and capitalized?

miaawong: @jodillion I'm using a string literal type to only allow specific string values for valueType. These are the four options valueTypes can be. They are capitalized since that's how the data is return from the api.

rianfowler: sometimes we can see the enums in the go code: https://github.com/replicatedhq/kots/blob/84b7e4e0e9275bb200a36be69691c4944eb8cf8f/pkg/appstate/types/types.go#L10-L14

other times we have to constrain it based on usage in the FE. I would like us to get an OpenAPI spec for the kots apis so we can derive types like this from that instead of having to hardcode response types in the FE

rianfowler (approved):

PR #3222, opened 2022-09-24
Recalculate config_pending when new version is deployed

divolgin:

- When new version is configured with all required config items, all other pending versions should have the same values applied by default, making configuration step before deployment optional.

- Fixes the download values.yaml button on the Version History and Dashboard for pending Helm releases.

Fixes #

```release-note
NONE
```

NONE


rianfowler (approved):
FE 👍🏽

stefanrepl (approved):

PR #3232, opened 2022-09-27
Show automatic update check config that was previously saved

divolgin:

- Show previously saved values when `Configure automatic updates` is opened.
- Revert non-Helm apps to use postgres for update check schedule.

Fixes #

```release-note
* Fixes an issue that caused custom automatic update schedule to not be shown after it's saved on the [Configure automatic updates](https://docs.replicated.com/enterprise/updating-apps#configure-automatic-updates).
```

NONE

### DIFF
--- a/pkg/handlers/update_checker_spec.go
+++ b/pkg/handlers/update_checker_spec.go
@@ -2,60 +2,51 @@ package handlers
 
 import (
 	"encoding/json"
-	"errors"
-	"fmt"
 	"net/http"
 
-	"github.com/replicatedhq/kots/pkg/helm"
-	"github.com/replicatedhq/kots/pkg/kotsadm/types"
-	"github.com/replicatedhq/kots/pkg/kotsutil"
-	"github.com/replicatedhq/kots/pkg/util"
-
 	"github.com/gorilla/mux"
+	"github.com/pkg/errors"
 	apptypes "github.com/replicatedhq/kots/pkg/app/types"
+	"github.com/replicatedhq/kots/pkg/helm"
+	"github.com/replicatedhq/kots/pkg/kotsutil"
 	"github.com/replicatedhq/kots/pkg/logger"
 	"github.com/replicatedhq/kots/pkg/store"
 	"github.com/replicatedhq/kots/pkg/updatechecker"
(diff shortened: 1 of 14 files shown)


pkg/handlers/update_checker_spec.go
```diff
@@ -146,36 +131,68 @@ func (h *Handler) ConfigureAutomaticUpdates(w http.ResponseWriter, r *http.Reque
...
+		logger.Error(errors.Wrap(err, updateCheckerSpecResponse.Error))
+		JSON(w, http.StatusInternalServerError, updateCheckerSpecResponse)
 		return
 	}
 
-	JSON(w, 204, "")
+	JSON(w, http.StatusNoContent, "")
+}
+
+func (h *Handler) GetAutomaticUpdatesConfig(w http.ResponseWriter, r *http.Request) {
```
sgalsaleh, line 166: this used to be part of the application object returned to the ui when listing apps, is there a reason it's a separate handler now?

web/src/components/modals/AutomaticUpdatesModal.jsx
```diff
@@ -97,10 +101,10 @@ export default class AutomaticUpdatesModal extends React.Component {
       selectedSchedule,
       selectedAutoDeploy,
     };
+  }
 
-    if (this.state.updateCheckerSpec === "") {
```
sgalsaleh, line 101: why is this not needed anymore? is it never empty?

sgalsaleh (approved):

web/src/components/modals/AutomaticUpdatesModal.jsx
divolgin: yeah, there is a check above to find "custom" or "default" now

sgalsaleh: but that's for the selected schedule, not the actual cron spec being used, right?

sgalsaleh: the actual cron spec should also displayed as `@default` i believe when it's empty. the "custom" or "default" are for the drop down options

PR #3238, opened 2022-09-28
Show license expired message on Version History page

divolgin:

In Helm managed mode, when checking for updates, the API will return the "license expired" message.

Screen Shot 2022-09-28 at 9 55 07 AM

Fixes #

```release-note
* Fixes an issue where checking for updates from the [version history](/enterprise/updating-apps#update-an-application-in-the-admin-console) page would not show the "License is expired" error when the check fails due to an expired license in [Helm managed mode (Alpha)](/vendor/helm-install).
```

NONE

### DIFF
--- a/pkg/updatechecker/updatechecker.go
+++ b/pkg/updatechecker/updatechecker.go
@@ -15,6 +15,7 @@ import (
 	"github.com/replicatedhq/kots/pkg/kotsadmconfig"
 	license "github.com/replicatedhq/kots/pkg/kotsadmlicense"
 	upstream "github.com/replicatedhq/kots/pkg/kotsadmupstream"
+	kotslicense "github.com/replicatedhq/kots/pkg/license"
 	"github.com/replicatedhq/kots/pkg/logger"
 	"github.com/replicatedhq/kots/pkg/preflight"
 	kotspull "github.com/replicatedhq/kots/pkg/pull"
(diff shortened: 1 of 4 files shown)


web/src/components/apps/AppVersionHistory.jsx
```diff
@@ -1002,7 +1002,7 @@ class AppVersionHistory extends Component {
       });
 
       const res = await fetch(
-        `${process.env.API_ENDPOINT}/app/${app?.slug}/cluster/${clusterId}/sequence/${version?.sequence}/downstreamoutput`,
+        `${process.env.API_ENDPOINT}/app/${app.slug}/cluster/${clusterId}/sequence/${version.sequence}/downstreamoutput`,
```
divolgin, line 1005: Slug and version are required here.  If they are missing, UI state is not managed right.

web/src/components/apps/DashboardVersionCard.jsx
```diff
@@ -481,20 +483,22 @@ class DashboardVersionCard extends React.Component {
...
-                    currentVersion?.status === "failed"
-                  )
-                }
-                data-tip="View deploy logs"
-              >
-                <Icon icon="view-logs" size={22} className="clickable" />
-              </span>
-              <ReactTooltip effect="solid" className="replicated-tooltip" />
-            </div>
+            {app ? (
```
divolgin, line 486: Can't click without app.  Was causing random failures in tests.

pkg/updatechecker/updatechecker.go
```diff
@@ -246,6 +247,23 @@ func checkForHelmAppUpdates(opts CheckForUpdatesOpts, finishedChan chan<- error)
 
 	availableUpdateTags, err := helm.CheckForUpdates(helmApp, license, &currentVersion)
 	if err != nil {
+		expiredErr := util.ActionableError{
+			NoRetry: true,
+			Message: "License is expired",
+		}
+		isExpired := false
+
+		newLicense, _ := helm.GetChartLicenseFromSecretOrDownload(helmApp)
```
sgalsaleh, line 256: why not check for errors?

sgalsaleh (approved):

pkg/updatechecker/updatechecker.go
divolgin: if we can't sync and local copy is expired, what else is there to do....

PR #3239, opened 2022-09-28
store branding archive in database

cbodonnell:

This PR changes how font files are included in a release.  Previously, the data for these files was placed directly in the kots app spec as base64 string data and was stored in a configmap, but because of the [configmap size limit](https://kubernetes.io/docs/concepts/configuration/configmap/#:~:text=A%20ConfigMap%20is%20not%20designed,separate%20database%20or%20file%20service.), we needed an alternate way to store these files.  Branding files will instead be included as standalone files in an application release and will be extracted and stored in the database for each application version.

Initial installs will fetch an initial branding archive from a new replicated-app `/branding` endpoint or from the airgap bundle if it's an airgap installation.

Additionally, this PR changes how custom css is included in an application release.  Previously, it was specified directly in the kots app spec, but now it will included as a separate files (or files) and referenced in the app spec.

Example kots application spec with branding and css files:
```
apiVersion: kots.io/v1beta1
kind: Application
metadata:
  name: my-app
spec:
  title: My App
  branding:
    css:
      - "css/font.css"
      - "css/color.css"
    fonts:
      - fontFamily: MyFont
        sources:
          - "fonts/my-font.woff"
          - "fonts/my-font.woff2"
      - fontFamily: MyOtherFont
        sources:
          - "fonts/my-other-font.ttf"
```

Fixes [SC-58829](https://app.shortcut.com/replicated/story/58829/kots-will-fail-to-fetch-the-latest-version-if-the-application-metadata-kots-app-spec-is-too-large)

This PR expects that the new replicated-app `/branding` endpoint will be available.  If it is not, the metadata will fail to fetch and instead the default will be used.

#### Steps to reproduce

Include font files and/or CSS files directly in an application release and reference them via file path in the kots application spec.

```release-note
Fixes an issue where custom admin console branding (Alpha) would cause the admin console to fail to create a new application version if the data in the KOTS application manifest exceeded 1 MiB.
```

Yes, the documentation for this Alpha feature will need to be updated to reflect the new method for including branding files in an application release.


pkg/replicatedapp/api.go
```diff
@@ -87,20 +92,29 @@ func getLicenseFromAPI(url string, licenseID string) (*LicenseData, error) {
...
+		return nil, errors.Wrapf(err, "failed to get metadata from %s", host)
 	}
 
-	if len(metadata) == 0 {
-		metadata = []byte(DefaultMetadata)
+	if len(manifest) == 0 {
+		manifest = []byte(DefaultMetadata)
+	}
+
+	branding, err := getApplicationMetadataFromHost(host, "branding", upstream, versionLabel)
```
cbodonnell, line 106: Should we unmarshal the manifest here and check if the branding fields are empty or not before making this request?  Could prevent unnecessary requests for branding.

divolgin: i think this should be ok

divolgin (approved):

PR #3253, opened 2022-10-03
Use kURL kotsadm add-on package for release regression tests

emosbaugh:

Use kURL kotsadm add-on package for release regression tests

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -343,8 +343,10 @@ jobs:
               - 'deploy/kurl/**'
   validate-kurl-addon:
     runs-on: ubuntu-20.04
-    if: ${{ github.ref_type == 'branch' && needs.kurl-addon-changes-filter.outputs.ok-to-test == 'true' }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -482,11 +474,12 @@ jobs:
 
   regression-test:
     if: github.ref_type == 'branch'
-    needs: [regression-test-setup, generate-tag, build-go-api, release-go-api-tagged]
-    uses: replicatedhq/kots/.github/workflows/regression.yaml@main
+    needs: [ regression-test-setup, generate-tag, build-go-api, release-go-api-tagged, validate-kurl-addon ]
```
sgalsaleh, line 477: took me a minute to realize that `validate-kurl-addon` actually generates an addon version that the regression tests will use. the word `validate` doesn't indicate that.

sgalsaleh (approved):

.github/workflows/release.yaml
sgalsaleh: might be a good idea to move the `generate-addon` job to a separate action/job

PR #3254, opened 2022-10-04
Rianfowler/sc 50084/fix multi app ux 2

rianfowler:
- converts AppDetailsPage into a stateless functional component 
- necessary to use new app hooks 

https://app.shortcut.com/replicated/story/50084/fix-multi-app-ux

none- refactor

none


web/src/components/context/withTheme.jsx
```diff
@@ -10,3 +10,13 @@ export default function withTheme(Component) {
     );
   };
 }
+
+function useTheme() {
```
rianfowler, line 14: new custom hook for getting the theme

web/src/components/apps/AppDetailPage.tsx
```diff
@@ -85,69 +90,120 @@ class AppDetailPage extends Component<Props, State> {
...
       requiredKotsUpdateMessage: "",
       selectedWatchName: "",
-    };
-  }
+    }
+  );
+
+  const history = useHistory();
+  const params = useParams<KotsParams>();
+  const [currentApp, setCurrentApp] = useState(
```
rianfowler, line 98: i have a hook for the current app but i'm not using it here yet. I need to wire up the `useApps` stuff so the `currentApp` will be accurate. will do that in the next pr

web/src/Root.tsx
```diff
@@ -66,7 +66,9 @@ const ProtectedRoute = ({
 };
 
 const ThemeContext = React.createContext({
-  setThemeState: () => {},
+  setThemeState: (themeState?: ThemeState) => {
+    console.log("setThemeState used before being set", themeState);
```
alicenstar, line 70: Was this meant to be here? It's fine if so, just wanted to double check.

web/src/components/apps/AppDetailPage.tsx
```diff
@@ -202,442 +258,387 @@ class AppDetailPage extends Component<Props, State> {
...
+                  exact
+                  path="/app/:slug/registry-settings"
+                  render={() => (
+                    <AppRegistrySettings
+                      app={app}
+                      updateCallback={refetchData}
+                    />
+                  )}
+                />
+                {true && (
```
alicenstar, line 562: Is this true supposed to be hardcoded?

web/src/Root.tsx
rianfowler: Yeah this is just a placeholder for now. Going to move the themeset stuff into it's own provider hook and i wanted to make sure i remember to add some error handling.

web/src/components/apps/AppDetailPage.tsx
rianfowler: no- i added another pr to fix this.

PR #3256, opened 2022-10-04
fix: regenerate primary node join token

cbodonnell:

This PR fixes an issue where users were unable to generate a primary node join command via the "Cluster Management" tab in the admin console when running in a kURL cluster with the EKCO add-on.  The command to generate the new token was not able to complete and would timeout because the kubeconfig used was pointing to a local address on the host, which was not resolvable within the container.  This PR changes this so that the kubeconfig used for this command utilizes credentials from the associated service account for the running pod ([reference](https://kubernetes.io/docs/tasks/run-application/access-api-from-pod/)).

This PR also improves error handling on the front-end for this component so that if the process fails on the backend, the user is presented with a message that the command failed to generate.

Fixes [SC-50133](https://app.shortcut.com/replicated/story/50133/the-button-for-adding-a-primary-node-doesn-t-work-when-the-original-token-expires)

#### Steps to reproduce

- set up a kurl cluster with `curl https://kurl.sh/243cf3c | sudo bash -s ha`
- run `kubectl edit cm kurl-config -n kube-system` and change `upload_certs_expiration` to the day before (this is just for faster repro)
- install any application
- go the "Cluster Management" tab and click "Add a node"
- click "Primary Node" => will eventually show "Unexpected end of JSON input"

edit the deployment with a TTL image containing these changes and repeat to validate the fix

```release-note
Fixes an issue where the button for adding a primary node was not working after the original join token had expired.
```

### DIFF
--- a/pkg/kurl/join_cert.go
+++ b/pkg/kurl/join_cert.go
@@ -10,6 +10,8 @@ import (
 	"strings"
 	"time"
 
+	_ "embed"
+
 	"github.com/pkg/errors"
 	"github.com/replicatedhq/kots/pkg/logger"
 	"github.com/replicatedhq/kots/pkg/util"
@@ -18,7 +20,16 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+//go:embed scripts/join-cert-gen.sh
+var joinCertGenScript string
+
 func createCertAndKey(ctx context.Context, client kubernetes.Interface, namespace string) (string, error) {
+	configMap := getConfigMap(namespace)
+	_, err := client.CoreV1().ConfigMaps(namespace).Create(ctx, configMap, metav1.CreateOptions{})
+	if err != nil {
+		return "", errors.Wrap(err, "failed to create configmap")
+	}
+
 	pod, err := getPodSpec(client, namespace)
 	if err != nil {
 		return "", errors.Wrap(err, "failed to create pod spec")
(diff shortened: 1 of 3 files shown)


pkg/kurl/join_cert.go
```diff
@@ -127,6 +142,18 @@ func parseCertGenOutput(logs []byte) (string, error) {
 	return "", fmt.Errorf("key not found in %d bytes of output", len(logs))
 }
 
+func getConfigMap(namespace string) *corev1.ConfigMap {
```
sgalsaleh, line 145: nit pick: since this is part of the general `kurl` package, name conflict might happen in the future. would be nice to name it something like `joinCertGenConfigMap` or something

sgalsaleh (dismissed):

pkg/kurl/join_cert.go
cbodonnell: agreed

pkg/kurl/join_cert.go
```diff
@@ -127,6 +142,18 @@ func parseCertGenOutput(logs []byte) (string, error) {
 	return "", fmt.Errorf("key not found in %d bytes of output", len(logs))
 }
 
+func getJoinCertGenConfigMap(namespace string) *corev1.ConfigMap {
```
sgalsaleh, line 145:
```suggestion
func getJoinCertGenConfigMapSpec(namespace string) *corev1.ConfigMap {
```

pkg/kurl/join_cert.go
```diff
@@ -18,7 +20,16 @@ import (
...
+var joinCertGenScript string
+
 func createCertAndKey(ctx context.Context, client kubernetes.Interface, namespace string) (string, error) {
+	configMap := getJoinCertGenConfigMap(namespace)
+	_, err := client.CoreV1().ConfigMaps(namespace).Create(ctx, configMap, metav1.CreateOptions{})
+	if err != nil {
+		return "", errors.Wrap(err, "failed to create configmap")
+	}
+
 	pod, err := getPodSpec(client, namespace)
```
sgalsaleh, line 33: we can rename this too :)

sgalsaleh (approved):

PR #3264, opened 2022-10-05
Rianfowler/sc 50084/fix multi app ux 3 

rianfowler:
- removes app data fetching logic from `AppDetailPage`
- replaces it with the `useApps` and `useSelectedApps` hooks. These hooks create a global state for fetched data so that all components can render cached data before fetching 
- this only partially fixes the multi app issue. Additional work is needed for the version history and license card on the dashboard 

https://app.shortcut.com/replicated/story/50084/fix-multi-app-ux

None- will add release note to final fix. 

none


web/src/components/apps/AppDetailPage.tsx
```diff
@@ -30,26 +30,24 @@ import TroubleshootContainer from "../troubleshoot/TroubleshootContainer";
...
   onActiveInitSession: (session: string) => void;
   ping: () => void;
-  refetchAppsList: () => void;
   refetchAppMetadata: () => void;
   rootDidInitialAppFetch: boolean;
   snapshotInProgressApps: string[];
 };
 
 type State = {
-  app: App | null;
```
rianfowler, line 52: app is now stored in the app hook state which is using `react-query`

web/src/components/apps/AppDetailPage.tsx
```diff
@@ -95,9 +92,14 @@ function AppDetailPage(props: Props) {
...
-    props.appsList?.find((w) => w.slug === params.slug)
-  );
+  const { selectedApp } = useSelectedApp();
+  const {
+    data: appsData,
+    refetch: refetchApps,
+    isLoading: appsIsLoading,
+  } = useApps();
+
+  const { apps: appsList } = appsData || {};
```
rianfowler, line 104: for now the custom hooks just return exactly what `react-query` returns. that means that each component that uses the hook has to handle the response possibly being `null`

web/src/components/apps/AppDetailPage.tsx
```diff
@@ -95,72 +92,58 @@ function AppDetailPage(props: Props) {
...
 
-      const res = await fetch(`${process.env.API_ENDPOINT}/app/${slug}`, {
-        headers: {
-          Authorization: Utilities.getToken(),
-          "Content-Type": "application/json",
-        },
-        method: "GET",
+  // loading state stuff that was in the old getApp() implementation
+  useEffect(() => {
```
rianfowler, line 107: removed `getApp()` and replaced all of the state setting it does with this effect hook. The status code is handled in the fetch function so the handling logic here has been altered to just show the error message that comes from the fetch function. 

The old implementation would look at the status code and report it in the error message

PR #3268, opened 2022-10-05
Release regression tests not running because of kurl add-on changes

emosbaugh:

The regression tests are being skipped on all recently merged PRs because of the dependency on validate-kurl-addon job that runs only when changes are made to the deploy/kurl/kotsadm/template directory.

This pull request adds a generate-kurl-addon job as a dependency of the regression tests that is run on all merges to main.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -6,6 +6,7 @@ on:
       - "v*.*.*"
     branches:
       - main
+      - emosbaugh/fix-release-workflow
 
 jobs:
   generate-tag:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -6,6 +6,7 @@ on:
       - "v*.*.*"
     branches:
       - main
+      - emosbaugh/fix-release-workflow
```
emosbaugh, line 9: REMOVE THIS

sgalsaleh (approved):

PR #3280, opened 2022-10-07
render and run preflights

stefanrepl:
This PR adds a `plugin` command  that serves as an entry point for our helm plugin.  The preflight subcommand is capable of pulling a remote preflight spec from registry and will template out the preflight spec utilizing config values set if a config spec is also included.

Fixes #
https://app.shortcut.com/replicated/story/58512/make-preflight-checks-work-for-config-options-and-anything-templated

Related to plugin PR: https://github.com/replicatedhq/kots-helm/pull/15

```release-note
     Adds `plugin` command for kots that is utilized by our helm plugin.
```

### DIFF
--- a/cmd/kots/cli/plugin.go
+++ b/cmd/kots/cli/plugin.go
@@ -1,0 +1,25 @@
+package cli
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+func PluginCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:           "plugin",
+		Short:         "Run commands for helm plugin",
+		Long:          `Run commands for helm plugin, this serves as an entry point for subcommands such as preflight`,
+		SilenceUsage:  false,
+		SilenceErrors: false,
+		PreRun: func(cmd *cobra.Command, args []string) {
+			viper.BindPFlags(cmd.Flags())
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return nil
+		},
+	}
+
+	cmd.AddCommand(PreflightCmd())
+	return cmd
+}
--- a/cmd/kots/cli/preflight.go
+++ b/cmd/kots/cli/preflight.go
@@ -1,0 +1,129 @@
+package cli
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/url"
+
+	"github.com/replicatedhq/troubleshoot/pkg/preflight"
+
+	"github.com/pkg/errors"
+	"github.com/replicatedhq/kots/pkg/kotsutil"
+	"github.com/replicatedhq/kots/pkg/logger"
+	"github.com/replicatedhq/kots/pkg/template"
+	"github.com/replicatedhq/troubleshoot/pkg/oci"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+func PreflightCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:           "preflight",
+		Short:         "Run preflights",
+		Long:          `Run preflights without a running KOTS installation`,
+		SilenceUsage:  false,
+		SilenceErrors: false,
+		PreRun: func(cmd *cobra.Command, args []string) {
+			viper.BindPFlags(cmd.Flags())
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			log := logger.NewCLILogger(cmd.OutOrStdout())
+			v := viper.GetViper()
+			v.BindPFlags(cmd.Flags())
+
+			preflightFile, err := cmd.Flags().GetString("preflight-spec")
+			if err != nil {
+				log.Errorf("failed to get preflight flag: %v\n", err)
+				return err
+			}
+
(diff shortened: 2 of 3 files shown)


divolgin:
Should we make this new command hidden for now? There's a [`Hidden`](https://github.com/spf13/cobra/blob/v1.5.0/command.go#L199) flag in Cobra commands.

cmd/kots/cli/plugin.go
```diff
@@ -0,0 +1,201 @@
...
+			viper.BindPFlags(cmd.Flags())
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 1 {
+				cmd.Help()
+				os.Exit(1)
+			}
+			log := logger.NewCLILogger(cmd.OutOrStdout())
+
+			if args[0] == "preflight" {
```
divolgin, line 37: These should be defied as subcommands they they will show up as subcommands in help output and not as flags as dined below (because these are not flags)

Example of subcommands for `admin-console` [here](https://github.com/replicatedhq/kots/blob/main/cmd/kots/cli/admin-console.go#L100-L103)

cmd/kots/cli/plugin.go
```diff
@@ -0,0 +1,201 @@
...
+					return err
+				}
+			} else if args[0] == "config" {
+				// TODO: logic for initial install for pulling config yaml
+			}
+			return nil
+		},
+	}
+
+	cmd.Flags().String("preflight", "", "the filename or url of the Preflight spec")
```
divolgin, line 62: These subcommands,  not flags.

cmd/kots/cli/plugin.go
```diff
@@ -0,0 +1,201 @@
...
+
+func renderAndRunPreflight(preflightFile, configFile string) error {
+	isUrl := true
+	parsed, err := url.ParseRequestURI(preflightFile)
+	if err != nil {
+		isUrl = false
+	}
+
+	var prefBytes []byte
+	if isUrl && parsed != nil && parsed.Scheme == "oci" {
```
divolgin, line 76: All of this should be in the troubleshoot repo.  It's already implemented there.  It might be sufficient to export [this function](https://github.com/replicatedhq/troubleshoot/blob/374a65522dda8342a9893d96d6df756f224e4b74/cmd/preflight/cli/run.go#L34). Though it might be better to move it into a separate package and remove cobra references from it.  BUT we could also import [these](https://github.com/replicatedhq/troubleshoot/blob/3a5816cd18ddb991e3681a8d6d76d35e247ce680/cmd/preflight/cli/root.go#L40-L49) flags from troubleshoot.  And make this integration magical just like when we import kubectl flags in our code instead of redefining them.

cmd/kots/cli/plugin.go
```diff
@@ -0,0 +1,201 @@
...
+			URI:     result.URI,
+		}
+		analyzeResults = append(analyzeResults, res)
+	}
+
+	showStdoutResultsHuman(preflightSpec.Name, analyzeResults)
+	return nil
+}
+
+func showStdoutResultsHuman(preflightName string, analyzeResults []*analyzerunner.AnalyzeResult) error {
```
divolgin, line 162: Output should be the normal preflight output as implemented in the troubleshoot project.

pkg/preflight/execute.go
```diff
@@ -18,7 +18,7 @@ import (
 
 // execute will execute the preflights using spec in preflightSpec.
 // This spec should be rendered, no template functions remaining
-func execute(appID string, sequence int64, preflightSpec *troubleshootv1beta2.Preflight, ignorePermissionErrors bool) (*troubleshootpreflight.UploadPreflightResults, error) {
+func Execute(appID string, sequence int64, preflightSpec *troubleshootv1beta2.Preflight, ignorePermissionErrors bool, isHeadless bool) (*troubleshootpreflight.UploadPreflightResults, error) {
```
divolgin, line 21: I don't think this function should be used here, so there should be no need for `isHeadless` flag.  When preflights are running from CLI, the entire thing should just be delegated to the preflight package in the troubleshoot project.  CLI only needs to template the specs out.  Probably best way to do this is to pass templating context to troubleshoot.  I'm open to suggestions here though.

stefanrepl: ya definitely makes sense, thanks

cmd/kots/cli/plugin.go
stefanrepl: they were flags for preflight spec and config spec..definitely will pick better naming since we are going to have preflight and config subcommands

cmd/kots/cli/plugin.go
stefanrepl: ya im going to move the code in troubleshoot into the preflight package in that repo, export whats necessary, and remove cobra references from the funcs and pass them in instead

stefanrepl: PR for this: https://github.com/replicatedhq/troubleshoot/pull/769

cmd/kots/cli/plugin.go
stefanrepl: done

PR #3284, opened 2022-10-07
allow custom graphs to be templated

cbodonnell:

This PR adds support for templating to all properties of the `spec.graphs` field in the kots.io/v1beta1 Application custom resource.

This PR also updates the PR test for label enforcement.

Fixes [SC-59379](https://app.shortcut.com/replicated/story/59379/fmt-sprintf-usage-broken-for-custom-graph-queries) and [SC-51989](https://app.shortcut.com/replicated/story/51989/using-template-functions-in-kots-application-spec-graphs-manifests-fail-to-render-and-display-prometheus-graphs)

Little bit of refactoring for easier unit testing of these changes.

#### Steps to reproduce

Add custom graphs with template functions to the KOTS application spec:

```
apiVersion: kots.io/v1beta1
   kind: Application
   metadata:
     name: my-application
   spec:
     graphs:
       - title: CPU Usage
         query: 'sum(rate(container_cpu_usage_seconds_total{namespace="{{repl Namespace}}",container!="POD",pod!=""}[5m])) by (pod)'
         legend: '{{ pod }}'
```

```release-note
Adds support for template functions to the `spec.graphs` field of the [Application custom resource](/reference/custom-resource-application).
```

https://github.com/replicatedhq/replicated-docs/pull/689

### DIFF
--- a/pkg/handlers/dashboard.go
+++ b/pkg/handlers/dashboard.go
@@ -35,28 +35,30 @@ func (h *Handler) GetAppDashboard(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	a, err := store.GetStore().GetAppFromSlug(appSlug)
+	store := store.GetStore()
+
+	a, err := store.GetAppFromSlug(appSlug)
 	if err != nil {
 		logger.Error(err)
 		w.WriteHeader(500)
 		return
 	}
 
-	appStatus, err = store.GetStore().GetAppStatus(a.ID)
+	appStatus, err = store.GetAppStatus(a.ID)
 	if err != nil {
 		logger.Error(err)
 		w.WriteHeader(500)
 		return
 	}
 
-	parentSequence, err := store.GetStore().GetCurrentParentSequence(a.ID, clusterID)
+	parentSequence, err := store.GetCurrentParentSequence(a.ID, clusterID)
 	if err != nil {
 		logger.Error(err)
 		w.WriteHeader(500)
 		return
 	}
 
-	prometheusAddress, err := store.GetStore().GetPrometheusAddress()
+	prometheusAddress, err := store.GetPrometheusAddress()
 	if err != nil {
 		logger.Error(err)
 		w.WriteHeader(500)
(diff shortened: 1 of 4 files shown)


pkg/handlers/dashboard.go
```diff
@@ -35,28 +35,30 @@ func (h *Handler) GetAppDashboard(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	a, err := store.GetStore().GetAppFromSlug(appSlug)
+	store := store.GetStore()
```
sgalsaleh, line 38: i'd suggest avoiding variable shadowing. if you really wanna have this in a separate step, maybe change the variable name to something else

pkg/version/metrics.go
```diff
@@ -79,36 +82,45 @@ var (
...
 
-		if len(a.Spec.Graphs) > 0 {
-			graphs = a.Spec.Graphs
-		}
+	builder, err := render.NewBuilder(kotsKinds, registrySettings, app.Slug, sequence, app.IsAirgap, util.PodNamespace)
+	if err != nil {
+		return graphs, errors.Wrap(err, "failed to get template builder")
 	}
 
+	if len(kotsKinds.KotsApplication.Spec.Graphs) > 0 {
```
sgalsaleh, line 116: the application spec should already exist in the database in the app_version table, there is no need to retrieve the archive from s3 to get it

sgalsaleh: tip, you can use `GetDownstreamVersion` and then call `AddDetails` and the version would have kots kinds loaded from the db.

pkg/version/metrics.go
```diff
@@ -79,36 +82,45 @@ var (
...
+		graphs = renderGraphs(builder, kotsKinds.KotsApplication.Spec.Graphs)
```
sgalsaleh, line 117: why not just render the whole application spec using the `RenderFile` function we already have? it takes the same parameters as the `builder` you have here to initialize. after rendering it, you can use `LoadKotsAppFromContents` to parse it.

pkg/version/metrics.go
cbodonnell: we need more than just the application spec since we are creating a builder instance right above this.  is there a more efficient way to get the information necessary without going to s3?

sgalsaleh: ah you're right, kots kinds are needed to render things i guess.

pkg/version/metrics.go
cbodonnell: done

pkg/handlers/dashboard.go
cbodonnell: reverted this change and will just pass `store.GetStore()` to the `GetGraphs(...)` function

sgalsaleh (approved):

sgalsaleh (approved):

PR #3286, opened 2022-10-10
collect namespace from flag or kubeconfig context

alexanderjophus:

Fixes kots CLI to use kubeconfig namespace for finding kotsadm when no namespace is passed in via flag.

Fixes # [SC-56078](https://app.shortcut.com/replicated/story/56078/kots-cli-should-respect-the-namespace-from-the-current-context)

May change some backwards compatibility, in that if a user was depending on it not being passed via flags and instead using the default namespace. Now that user may experience a different behaviour if their kubeconfig is set to another namespace.

Another goal of this PR is to align all the `--namespace` functionality, so that the user has a consistent experience. They all get it from flags, kubeconfig context, or set it to `default` in that order, then validate the chosen namespace. 

## Steps to reproduce
- Run kubectl config set-context --current --namespace=NAMESPACE and replace NAMESPACE with the name of the namespace KOTS is in.
- Run kubectl kots admin-console and see that it fails, saying you need to provide the namespace.

```release-note
Allows kots cli commands to use the kubeconfig namespace by default if a flag is not provided.
```

None?

### DIFF
--- a/cmd/kots/cli/namespace.go
+++ b/cmd/kots/cli/namespace.go
@@ -4,8 +4,34 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/client-go/tools/clientcmd"
 )
+
+func getNamespaceOrDefault(namespace string) (string, error) {
+	if namespace == "" {
+		clientCfg, err := clientcmd.NewDefaultClientConfigLoadingRules().Load()
+		if err != nil {
+			return "", errors.Wrap(err, "failed to load kubeconfig")
+		}
+		cctx := clientCfg.Contexts[clientCfg.CurrentContext]
+		if cctx != nil {
+			namespace = cctx.Namespace
+		}
+	}
+
+	if namespace == "" {
+		namespace = metav1.NamespaceDefault
+	}
+
+	err := validateNamespace(namespace)
+	if err != nil {
+		return "", errors.Wrapf(err, "invalid namespace %s", namespace)
+	}
+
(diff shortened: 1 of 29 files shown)


cmd/kots/cli/namespace.go
```diff
@@ -1,12 +1,39 @@
...
+		}
+	}
+
+	if namespace == "" {
+		namespace = metav1.NamespaceDefault
+	}
+
+	err := validateNamespace(namespace)
+	if err != nil {
+		return "", errors.Wrap(err, fmt.Sprintf("invalid namespace %s", namespace))
```
cbodonnell, line 31: I think there's also an `errors.Wrapf(...)` function that does this, so we wouldn't even have to import `"fmt"` here.

cmd/kots/cli/set-config.go
```diff
@@ -46,7 +46,10 @@ func SetConfigCmd() *cobra.Command {
...
 			log := logger.NewCLILogger(cmd.OutOrStdout())
 			appSlug := args[0]
-			namespace := v.GetString("namespace")
+			namespace, err := getNamespaceOrDefault(v.GetString("namespace"))
+			if err != nil {
+				return errors.Wrap(err, "failed to get namespace")
+			}
 
 			if err := validateNamespace(namespace); err != nil {
 				return errors.Wrap(err, "failed to validate namespace")
```
cbodonnell, line 55: Can we also get rid of this call to `validateNamespace` now?

cbodonnell (commented):
Can we also add tests for this somewhere?  Thinking that something in the `build-test` workflow we have may be appropriate for this: https://github.com/replicatedhq/kots/blob/main/.github/workflows/build-test.yaml

cbodonnell (approved):

PR #3287, opened 2022-10-10
Fix OIDC login

rianfowler:
- this removes requests that trigger logout when the user is not logged in. I believe this is interfering with identity provider logins. 

<
Fixes #

```release-note
- fixes issue where apis that require auth token were called while client was in an logged out state 
```

none


web/src/Root.tsx
```diff
@@ -306,6 +306,9 @@ const Root = () => {
   };
 
   const ping = async (tries = 0) => {
+    if (!Utilities.isLoggedIn()) {
```
rianfowler, line 309: temp fix. block the ping request if there is no token

`isLoggedIn` is not actually checking a login state it's trying to fetch a token. if the token is not there we assume this request won't work. since ping is executed in an interval we don't want to trigger a logout from the interval running while the user is logged out.

web/src/Root.tsx
```diff
@@ -345,11 +348,10 @@ const Root = () => {
...
 
   const onRootMounted = () => {
-    fetchKotsAppMetadata();
-    ping();
-    checkIsHelmManaged();
-
     if (Utilities.isLoggedIn()) {
+      fetchKotsAppMetadata();
+      ping();
+      checkIsHelmManaged();
```
rianfowler, line 354: I believe all of these will still be called even if they're bypassed here.

metadata is refetched by `AppDetails`
ping is run in an interval
isHelmManaged is called after successful login in the `SecureAdmin` component

PR #3290, opened 2022-10-11
Adds typechecking to AutomaticUpdatesModal

miaawong:

Fixes # [sc-59400]

```release-note
NONE
```

NONE

### DIFF
--- a/web/src/components/modals/AutomaticUpdatesModal.tsx
+++ b/web/src/components/modals/AutomaticUpdatesModal.tsx
@@ -82,26 +82,28 @@ type Schedule = {
 
 type Props = {
   appSlug: string;
-  onAutomaticUpdatesConfigured: () => void;
-  isOpen: boolean;
-  onRequestClose: () => void;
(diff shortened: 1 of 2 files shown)


web/src/components/modals/AutomaticUpdatesModal.tsx
```diff
@@ -312,10 +339,9 @@ export default class AutomaticUpdatesModal extends React.Component {
                 isSearchable={false}
                 getOptionValue={(schedule) => schedule.label}
                 value={selectedSchedule}
+                // TODO: upgrade react-select and fix this
+                // @ts-ignore
                 onChange={this.handleScheduleChange}
-                isOptionSelected={(option) => {
```
miaawong, line 316: this was never true, selectedSchedule is an object not string. 
I tried fixing this by setting `option.value === selectedSchedule.value`, however it messed up the hover states on select element. I removed it and nothing changed.

rianfowler (approved):

PR #3304, opened 2022-10-13
diamonwiggins/troubleshoot-collect-interface

diamonwiggins:
Integrates latest troubleshoot collect pkg containing breaking changes. Collector Struct is now an Interface that each collector implements.

Fixes #

```release-note

```


go.mod
```diff
@@ -1,31 +1,31 @@
 module github.com/replicatedhq/kots
 
-go 1.17
+go 1.19
```
stefanrepl, line 3: with this update to 1.19 in the mod file, how are you running `go mod tidy` or equivalent commands?  when I run `go mod tidy` locally I now get the following error: `go: go.mod file indicates go 1.19, but maximum version supported by tidy is 1.18`

diamonwiggins: The commands work for me. I'm using Go 1.19. I was unsure if I needed to do this at all, but it seems like some dependency train started after adding the new troubleshoot that led me here? If I can go back to 1.17 I will, but not sure.

stefanrepl: ah gotchya, I likely just need to update, thanks

stefanrepl (approved):

stefanrepl (dismissed):

divolgin (approved):

stefanrepl (approved):

PR #3306, opened 2022-10-13
Miawong/add slow loader state to license api 

miaawong:
#### What this PR does / why we need it: This allows us to add a loading spinner if the data takes longer than .5 seconds to return. 
- Created a custom hook that run the` /license` query hook and return slowLoading state if the response takes longer than .5 seconds. 
- Created a query hook that queries `/license` endpoint
- Added axios to kots/web 
- Hook is called `useLicenseWithIntercept`
- Hook is not implemented yet and is not being used anywhere. 

Fixes # [sc-50084]

```release-note
NONE
```

NONE

### DIFF
--- a/web/src/components/apps/Dashboard.tsx
+++ b/web/src/components/apps/Dashboard.tsx
@@ -23,7 +23,7 @@ import { Paragraph } from "../../styles/common";
 const COMMON_ERRORS = {
   "HTTP 401": "Registry credentials are invalid",
   "invalid username/password": "Registry credentials are invalid",
-  "no such host": "No such host"
+  "no such host": "No such host",
 };
 
 // Types
@@ -35,7 +35,7 @@ import {
   DashboardActionLink,
   KotsParams,
   ResourceStates,
-  Version
+  Version,
 } from "@types";
 import { RouteComponentProps } from "react-router-dom";
 //import LicenseTester from "./LicenseTester";
@@ -126,7 +126,7 @@ class Dashboard extends Component<Props, State> {
       dashboard: {
         appStatus: null,
         metrics: [],
-        prometheusAddress: ""
+        prometheusAddress: "",
       },
       currentVersion: null,
       displayErrorModal: false,
@@ -150,7 +150,7 @@ class Dashboard extends Component<Props, State> {
       startSnapshotOptions: [
         { option: "partial", name: "Start a Partial snapshot" },
         { option: "full", name: "Start a Full snapshot" },
-        { option: "learn", name: "Learn about the difference" }
+        { option: "learn", name: "Learn about the difference" },
       ],
       updateChecker: new Repeater(),
       uploadingAirgapFile: false,
@@ -159,7 +159,7 @@ class Dashboard extends Component<Props, State> {
       uploadSize: 0,
       viewAirgapUpdateError: false,
       viewAirgapUploadError: false,
-      slowLoader: false
+      slowLoader: false,
     };
   }
 
@@ -169,7 +169,7 @@ class Dashboard extends Component<Props, State> {
       iconUri: app.iconUri,
       currentVersion: app.downstream?.currentVersion,
       downstream: app.downstream,
-      links: app.downstream?.links
+      links: app.downstream?.links,
     });
   };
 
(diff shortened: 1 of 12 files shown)


web/src/components/apps/Dashboard.tsx
```diff
@@ -38,6 +38,7 @@ import {
   Version,
 } from "@types";
 import { RouteComponentProps } from "react-router-dom";
+//import LicenseTester from "./LicenseTester";
```
miaawong, line 42: can be deleted, wanted to make it easy to test

web/src/components/apps/Dashboard.tsx
```diff
@@ -749,11 +751,24 @@ class Dashboard extends Component<Props, State> {
...
     return (
       <>
-        {!app && (
-          <div className="flex-column flex1 alignItems--center justifyContent--center">
-            <Loader size="60" />
-          </div>
-        )}
+        {!app ||
+          (this.state.slowLoader && (
+            <div
```
miaawong, line 756: added slowLoader state, but this is always false for now. 
changed some of the styles so it only shows loader on the dashboard not the multi app sidebar

web/src/components/apps/Dashboard.tsx
```diff
@@ -851,7 +867,15 @@ class Dashboard extends Component<Props, State> {
                       gettingAppLicenseErrMsg={
                         this.state.gettingAppLicenseErrMsg
                       }
-                    />
+                    >
+                      {/* leaving this here as an example: please delete later */}
+                      {/* <LicenseTester
```
miaawong, line 872: component to test the useLicenseWithIntercept hook, please delete when it's unnecessary for you.

web/src/components/apps/LicenseTester.jsx
```diff
@@ -0,0 +1,32 @@
...
+  const { data } = licenseQuery;
+  const { license } = data || {};
+  const { entitlements } = license || [];
+
+  useEffect(() => {
+    setLoader(isSlowLoading);
+  }, [isSlowLoading, licenseQuery?.data]);
+
+  return (
+    <div>
```
miaawong, line 18: test component, can be deleted when it seems unnecessary. just wanted to leave an example here

web/webpack.config.js
```diff
@@ -119,6 +119,12 @@ module.exports = function (env) {
               }
             }
           ]
+        }, 
+        {
+          test: /\.m?js/,
+          resolve: {
+              fullySpecified: false
+          }
```
miaawong, line 127: had to add this to get axios working
doesn't really make sense to me.

web/src/components/apps/LicenseTester.jsx
rianfowler: thank you

web/src/features/App/api/useLicense.ts
```diff
@@ -0,0 +1,66 @@
...
+  //   });
+  // };
+
+  let licenseQuery = useLicense(appSlug);
+
+  useEffect(() => {
+    axios.interceptors.request.use(
+      (x) => {
+        // not sure if this check is enough to make sure that it
+        // only happens on /license endpoint
```
rianfowler, line 24: so interceptors are globally applied? that's good to know

web/webpack.config.js
rianfowler: axios must use .mjs files https://webpack.js.org/configuration/module/#resolvefullyspecified

rianfowler (approved):

web/src/features/App/api/useLicense.ts
```diff
@@ -0,0 +1,66 @@
...
+          // TODO: handle unauthorized
+          console.log("an error occurred");
+        }
+        //TODO: handle other errors
+        console.log("Error => axios.interceptors.request", error);
+      }
+    );
+  }, []);
+  // return everything that comes with react query
+  return { licenseQuery, isSlowLoading };
```
rianfowler, line 63:
```suggestion
  return { ...licenseQuery, isSlowLoading };
```
This is how I think we should do it that way it's the same api for all of our api queries.

rianfowler (approved):

PR #3307, opened 2022-10-13
use fetch state instead of loading state to trigger redirect and remo…

rianfowler:
- fixes bug where redirect would not occur when skipping preflight checks 
- fixes issue causing console.warning in dev build 

Fixes #
https://app.shortcut.com/replicated/story/60500/kots-occasionally-does-not-redirect-to-dashboard-after-uploading-license-and-skipping-preflight-checks
[sc-60500]

```release-note
- fix an issue where kots would occasionally not redirect to first available app preflight checks were skipped 
```

none

### DIFF
--- a/web/src/components/apps/AppDetailPage.tsx
+++ b/web/src/components/apps/AppDetailPage.tsx
@@ -97,7 +97,7 @@ function AppDetailPage(props: Props) {
     data: appsData,
     error: appsError,
     isError: appsIsError,
-    isLoading: appsIsLoading,
+    isFetching: appIsFetching,
     refetch: refetchApps,
   } = useApps({ refetchInterval: appsRefetchInterval });
 
@@ -120,7 +120,7 @@ function AppDetailPage(props: Props) {
 
   // loading state stuff that was in the old getApp() implementation
   useEffect(() => {
-    if (appsIsLoading) {
+    if (appIsFetching && !selectedApp) {
       setState({
         loadingApp: true,
       });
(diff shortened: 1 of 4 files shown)


web/src/components/apps/AppDetailPage.tsx
```diff
@@ -97,7 +97,7 @@ function AppDetailPage(props: Props) {
     data: appsData,
     error: appsError,
     isError: appsIsError,
-    isLoading: appsIsLoading,
+    isFetching: appIsFetching,
```
rianfowler, line 100: `isLoading` will always be false after the initial load. we want to use `isFetching` instead

web/src/components/shared/ConfigureGraphs.tsx
```diff
@@ -8,6 +8,14 @@ const ConfigureGraphs = ({
...
   onPromValueChange,
   placeholder,
+}: {
+  toggleConfigureGraphs?: () => void;
+  updatePromValue?: () => void;
+  promValue?: string;
+  savingPromValue?: boolean;
+  savingPromError?: string;
+  onPromValueChange?: (e: React.ChangeEvent<HTMLInputElement>) => void;
+  placeholder?: string;
```
rianfowler, line 18: this was causing a warning in the console when passing in an {} into an onClick handler. I saw the problem with typescript getting the implicit type and making the value required so I had to add explicit type definition to make that go away.

miaawong (approved):

PR #3313, opened 2022-10-14
run status informers for charts when in helm managed mode

stefanrepl:
Run status informers for charts when running in helm managed mode

This PR implements running status informer checks for helm charts when running in helm managed mode

Fixes #
https://app.shortcut.com/replicated/story/59995/support-status-informers-in-helm-managed-mode

```release-note
     Added ability to run status informers for helm charts when running in [Helm managed mode (Alpha)](/vendor/helm-install).
```

### DIFF
--- a/pkg/handlers/dashboard.go
+++ b/pkg/handlers/dashboard.go
@@ -7,6 +7,7 @@ import (
 	"github.com/gorilla/mux"
 	"github.com/pkg/errors"
 	appstatetypes "github.com/replicatedhq/kots/pkg/appstate/types"
+	"github.com/replicatedhq/kots/pkg/helm"
 	"github.com/replicatedhq/kots/pkg/logger"
 	"github.com/replicatedhq/kots/pkg/store"
 	"github.com/replicatedhq/kots/pkg/util"
@@ -24,14 +25,26 @@ func (h *Handler) GetAppDashboard(w http.ResponseWriter, r *http.Request) {
 	clusterID := mux.Vars(r)["clusterId"]
 	appStatus := new(appstatetypes.AppStatus)
 	if util.IsHelmManaged() {
-		appStatus.State = "ready"
-		getAppDashboardResponse := GetAppDashboardResponse{
-			AppStatus:         appStatus,
-			Metrics:           nil,
-			PrometheusAddress: "",
+		release := helm.GetHelmApp(appSlug)
+		if release == nil {
+			logger.Errorf("release for app: %s does not exist\n", appSlug)
+			w.WriteHeader(http.StatusNotFound)
+			return
 		}
 
-		JSON(w, 200, getAppDashboardResponse)
+		// we have received informer data
+		if len(release.Status.ResourceStates) > 0 {
+			getAppDashboardResponse := GetAppDashboardResponse{
+				AppStatus:         &release.Status,
+				Metrics:           nil,
+				PrometheusAddress: "",
+			}
+
+			JSON(w, 200, getAppDashboardResponse)
+			return
+		}
+
+		w.WriteHeader(http.StatusNotFound)
 		return
 	}
 
--- a/pkg/handlers/metadata.go
+++ b/pkg/handlers/metadata.go
@@ -105,7 +105,9 @@ func GetMetadataHandler(getK8sInfoFn MetadataK8sFn, kotsStore store.Store) http.
 		}
 		application := obj.(*kotsv1beta1.Application)
 		metadataResponse.IconURI = application.Spec.Icon
-		metadataResponse.Branding = getBrandingResponse(kotsStore, appID)
+		if !util.IsHelmManaged() {
+			metadataResponse.Branding = getBrandingResponse(kotsStore, appID)
+		}
 		metadataResponse.Name = application.Spec.Title
 		metadataResponse.UpstreamURI = brandingConfigMap.Data[upstreamUriKey]
 		metadataResponse.ConsoleFeatureFlags = application.Spec.ConsoleFeatureFlags
--- a/pkg/helm/cache.go
+++ b/pkg/helm/cache.go
@@ -68,6 +68,7 @@ func Init(ctx context.Context) error {
 			continue
 		}
 
+		initMonitor(clientSet, namespace)
 		for _, s := range secrets.Items {
 			if s.Labels == nil || s.Labels["status"] != helmrelease.StatusDeployed.String() {
 				continue
@@ -86,6 +87,7 @@ func Init(ctx context.Context) error {
 			}
 
 			AddHelmApp(releaseInfo.Release.Name, releaseInfo)
+			resumeHelmStatusInformers(releaseInfo.Release.Name)
 		}
 
 		go func(namespace string) {
@@ -237,13 +239,13 @@ func watchSecrets(ctx context.Context, namespace string, labelSelector string) e
 				}
 
 				if err := recalculateCachedUpdates(helmApp); err != nil {
-					logger.Errorf("failed to recalculate updates for helm release info from secret %s in namespace %s: %s", secret.Name, namespace)
+					logger.Errorf("failed to recalculate updates for helm release info from secret %s in namespace %s: %s", secret.Name, namespace, err)
 					// Continue here.  Release has been installed and should show up up in Admin Console.
 				}
 
 				logger.Debugf("adding secret %s to cache", secret.Name)
 				AddHelmApp(helmApp.Release.Name, helmApp)
-
+				resumeHelmStatusInformers(helmApp.Release.Name)
 			case watch.Deleted:
 				secret, ok := e.Object.(*corev1.Secret)
 				if !ok {
@@ -265,7 +267,6 @@ func watchSecrets(ctx context.Context, namespace string, labelSelector string) e
 				deleteUpdateCacheForChart(helmApp.ChartPath)
 
 				RemoveHelmApp(helmApp.Release.Name)
-
 			default:
 				secret, ok := e.Object.(*corev1.Secret)
 				if !ok {
--- a/pkg/helm/informers.go
+++ b/pkg/helm/informers.go
@@ -41,6 +41,12 @@ func initMonitor(clientset kubernetes.Interface, targetNamespace string) {
 	go runAppStateMonitor(nsMon)
 }
 
+func getMonitor(namespace string) *appstate.Monitor {
+	monitorMux.Lock()
+	defer monitorMux.Unlock()
+	return monitorMap[namespace]
+}
+
 func resumeHelmStatusInformers(appName string) {
 	helmApp := GetHelmApp(appName)
 	app := &apptypes.App{
@@ -119,7 +125,6 @@ func applyStatusInformers(a *apptypes.App, sequence int64, kotsKinds *kotsutil.K
 		}
 		release.Status = appStatus
 		release.Status.State = appstatetypes.GetState(defaultReadyState)
-		AddHelmApp(release.Release.Name, release)
 	}
 
 	return nil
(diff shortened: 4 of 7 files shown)


pkg/apiserver/server.go
```diff
@@ -90,20 +90,18 @@ func Start(params *APIServerParams) {
...
-			panic(err)
-		}
-		defer op.Shutdown()
+	client := &client.Client{
+		TargetNamespace:   util.AppNamespace(),
+		ExistingInformers: map[string]bool{},
+		HookStopChans:     []chan struct{}{},
+	}
+	opStore := store.GetStore()
+	op := operator.Init(client, opStore, params.AutocreateClusterToken)
```
divolgin, line 99: opStore cannot be used in helm mode.  It will attempt to connect to postgres, which works in dev env, but won't work when deployed.

stefanrepl: perhaps I should only initialize it when not running in helm managed mode?  the store doesnt get used for helm managed in this PR

pkg/handlers/dashboard.go
```diff
@@ -24,15 +30,82 @@ func (h *Handler) GetAppDashboard(w http.ResponseWriter, r *http.Request) {
...
+				AppStatus:         &release.Status,
+				Metrics:           nil,
+				PrometheusAddress: "",
+			}
+
+			JSON(w, 200, getAppDashboardResponse)
+			return
+		} else {
+			// we must apply the informers and await result
+			app := &apptypes.App{
```
divolgin, line 52: This whole block that waits for informer results should not be in an API handler.  If we are starting up the operator in the `Start()` function above, it should be doing this work asynchronously.

pkg/operator/client/client.go
```diff
@@ -440,6 +441,15 @@ func (c *Client) ApplyAppInformers(args operatortypes.AppInformersArgs) {
 }
 
 func (c *Client) setAppStatus(newAppStatus appstatetypes.AppStatus) error {
+	if util.IsHelmManaged() {
+		// set status in the map
+		app := helm.GetHelmApp(newAppStatus.AppID)
+		app.Status = newAppStatus
+		app.Status.State = appstatetypes.GetState(newAppStatus.ResourceStates)
+		helm.AddHelmApp(newAppStatus.AppID, app)
```
divolgin, line 449: nit pick: this is not needed

pkg/operator/operator.go
```diff
@@ -406,6 +409,21 @@ func (o *Operator) applyStatusInformers(a *apptypes.App, sequence int64, kotsKin
...
+			release := helm.GetHelmApp(a.ID)
+			appStatus := appstatetypes.AppStatus{
+				AppID:          a.ID,
+				ResourceStates: defaultReadyState,
+				Sequence:       release.GetCurrentSequence(),
+				UpdatedAt:      time.Now(),
+			}
+			release.Status = appStatus
+			release.Status.State = appstatetypes.GetState(defaultReadyState)
+			helm.AddHelmApp(release.GetID(), release)
```
divolgin, line 423: nit pick: this is not needed

pkg/apiserver/server.go
divolgin: You have to call `operator.Init` with valid parameters.  That code does not check if store is nil.  Either it needs to be updated to not use store or operator should be turned into an interface and two different implementations, one for helm and one for normal installs.

pkg/operator/operator.go
```diff
@@ -416,6 +434,53 @@ func (o *Operator) applyStatusInformers(a *apptypes.App, sequence int64, kotsKin
 	return nil
 }
 
+func (o *Operator) ResumeHelmStatusInformers() {
+	helmApps := helm.GetCache()
```
divolgin, line 438: Can't expose cache like this.  It must be behind a muted.

This can work instead
```
		for _, releaseName := range helm.GetCachedHelmApps() {
			release := helm.GetHelmApp(releaseName)
...
```

pkg/operator/operator.go
```diff
@@ -416,6 +434,53 @@ func (o *Operator) applyStatusInformers(a *apptypes.App, sequence int64, kotsKin
...
+	for {
+		helmApps = helm.GetCache()
```
divolgin, line 440: Can this be cleaned up somehow.  There is already a channel that triggers this function.  Maybe change initialization order?

pkg/apiserver/server.go
stefanrepl: will work to go the interface route

divolgin: wouldn't this nil panic?
https://github.com/replicatedhq/kots/pull/3313/commits/d6bb59047d261a17679af19d71525cbbb7a9f824#diff-99c0290799daf9abc6240df64063e20bfaf67b371577b67ac7eec6f4725622ffR427

`o.client` also uses store.

How about not try to fit this into existing operator code?

stefanrepl: it exits before that line is hit, it seems the interface route is the best option here

pkg/handlers/metadata.go
```diff
@@ -105,7 +105,9 @@ func GetMetadataHandler(getK8sInfoFn MetadataK8sFn, kotsStore store.Store) http.
 		}
 		application := obj.(*kotsv1beta1.Application)
 		metadataResponse.IconURI = application.Spec.Icon
-		metadataResponse.Branding = getBrandingResponse(kotsStore, appID)
+		if !util.IsHelmManaged() {
+			metadataResponse.Branding = getBrandingResponse(kotsStore, appID)
```
stefanrepl, line 109: Had to add this to avoid a panic when devving in okteto

pkg/helm/informers.go
```diff
@@ -0,0 +1,179 @@
...
+			return
+		}
+	}
+}
+
+func applyStatusInformers(a *apptypes.App, sequence int64, kotsKinds *kotsutil.KotsKinds, builder *template.Builder, namespace string) error {
+	renderedInformers := []appstatetypes.StatusInformerString{}
+
+	// deploy status informers
+	if len(kotsKinds.KotsApplication.Spec.StatusInformers) > 0 {
```
divolgin, line 81: this check is not needed

pkg/helm/informers.go
```diff
@@ -0,0 +1,179 @@
...
+			renderedInformers = append(renderedInformers, appstatetypes.StatusInformerString(renderedInformer))
+		}
+	}
+
+	if identitydeploy.IsEnabled(kotsKinds.Identity, kotsKinds.IdentityConfig) {
+		renderedInformers = append(renderedInformers, appstatetypes.StatusInformerString(fmt.Sprintf("deployment/%s", identitytypes.DeploymentName(a.Slug))))
+	}
+
+	if len(renderedInformers) > 0 {
+		fmt.Printf("applied informer")
```
divolgin, line 101: debugging stuff

pkg/helm/informers.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/replicatedhq/kots/pkg/kotsutil"
+	"github.com/replicatedhq/kots/pkg/logger"
+	registrytypes "github.com/replicatedhq/kots/pkg/registry/types"
+	"github.com/replicatedhq/kots/pkg/render"
+	"github.com/replicatedhq/kots/pkg/template"
+	"github.com/replicatedhq/kots/pkg/util"
+	"k8s.io/client-go/kubernetes"
+)
+
+var monitorMap map[string]*appstate.Monitor
```
divolgin, line 25: looks like this map is accessed from different go routines but is not protected by a mutex

pkg/helm/informers.go
```diff
@@ -0,0 +1,179 @@
...
+		settings := registrytypes.RegistrySettings{
+			Namespace: util.PodNamespace,
+		}
+		builder, err := render.NewBuilder(&kotsKinds, settings, app.Slug, helmApp.GetCurrentSequence(), app.IsAirgap, util.PodNamespace)
+		if err != nil {
+			logger.Errorf("failed to create new builder: %v\n", err)
+			return
+		}
+
+		// apply the informers for this app since they havent been applied yet
```
divolgin, line 69: Is this true?  This function is called every time a secret is added or removed, and it looks at every app in the cache every time

pkg/helm/cache.go
```diff
@@ -265,7 +267,7 @@ func watchSecrets(ctx context.Context, namespace string, labelSelector string) e
 				deleteUpdateCacheForChart(helmApp.ChartPath)
 
 				RemoveHelmApp(helmApp.Release.Name)
-
+				resumeHelmStatusInformers()
```
divolgin, line 270: Why resume when app is removed?

pkg/helm/informers.go
```diff
@@ -0,0 +1,179 @@
...
+
+	if monitorMap[targetNamespace] == nil {
+		monitorMap[targetNamespace] = appstate.NewMonitor(clientset, targetNamespace)
+	}
+
+	go runAppStateMonitor(monitorMap[targetNamespace])
+}
+
+func resumeHelmStatusInformers() {
+	for _, appName := range GetCachedHelmApps() {
```
divolgin, line 40: I would change this to take app name as parameter and remove the loop since this is called when a secret is added.  This this comment will be true I think

> // apply the informers for this app since they havent been applied yet

pkg/helm/informers.go
stefanrepl: removed

pkg/helm/informers.go
stefanrepl: yup, added mutex

pkg/helm/cache.go
stefanrepl: no need

pkg/helm/informers.go
```diff
@@ -0,0 +1,178 @@
...
+	for _, str := range informerStrings {
+		informer, err := str.Parse()
+		if err != nil {
+			logger.Infof("failed to parse informer %s: %s", str, err.Error())
+			continue // don't stop
+		}
+		informers = append(informers, informer)
+	}
+	if len(informers) > 0 {
+		if mon, ok := monitorMap[namespace]; ok {
```
divolgin, line 141: This needs a mutex.  I'd add

```
func getMonitor(namespace string) *appstate.Monitor {
    monitorMux.Lock()
    defer monitorMux.Unlock()
    return monitorMap[namespace]
}
```

pkg/helm/informers.go
```diff
@@ -0,0 +1,178 @@
...
+			mtx.Unlock()
+			if lastHash != nextHash {
+				b, _ := json.Marshal(appStatus)
+				logger.Infof("Updating app status %s", b)
+			}
+
+			app := GetHelmApp(appStatus.AppID)
+			app.Status = appStatus
+			app.Status.State = appstatetypes.GetState(appStatus.ResourceStates)
+			AddHelmApp(app.Release.Name, app)
```
divolgin, line 172: Let's not call this here.  Modifying the app is sufficient because it's a pointer.

pkg/helm/informers.go
```diff
@@ -0,0 +1,178 @@
...
+		release := GetHelmApp(a.ID)
+		appStatus := appstatetypes.AppStatus{
+			AppID:          a.ID,
+			ResourceStates: defaultReadyState,
+			Sequence:       release.GetCurrentSequence(),
+			UpdatedAt:      time.Now(),
+		}
+		release.Status = appStatus
+		release.Status.State = appstatetypes.GetState(defaultReadyState)
+		AddHelmApp(release.Release.Name, release)
```
divolgin, line 122: Let's not call this here. Modifying the app is sufficient because it's a pointer.

pkg/helm/informers.go
stefanrepl: done

divolgin (approved):

PR #3322, opened 2022-10-17
update default matchExpression for darwin

stefanrepl:
allow for running kots installs into clusters running on M1 Macs.

Fixes #
https://app.shortcut.com/replicated/story/61035/allow-deploying-kots-on-darwin

## Steps to reproduce

I tested by performing the following on a M1 Mac:

1) Create a k8s cluster using `kind`
2) built KOTS cli locally via `make kots-real`
3) ran a `kots install ...` of an application into the cluster running on the M1 machine
4) saw KOTS get properly scheduled onto the node (previously it wouldnt schedule and would get stuck in `Pending`), I then saw my test application deploy successfully.

```release-note
     Adds the ability to install applications into clusters running on M1 Macs.
```

### DIFF
--- a/pkg/kotsadm/objects/affinity_objects.go
+++ b/pkg/kotsadm/objects/affinity_objects.go
@@ -1,6 +1,8 @@
 package kotsadm
 
 import (
+	"runtime"
+
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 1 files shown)


divolgin (dismissed):

pkg/kotsadm/objects/affinity_objects.go
```diff
@@ -1,34 +1,66 @@
...
+				Operator: corev1.NodeSelectorOpIn,
+				Values: []string{
+					"linux",
+				},
+			},
+			{
+				Key:      "kubernetes.io/arch",
+				Operator: corev1.NodeSelectorOpIn,
+				Values: []string{
+					"arm64",
```
stefanrepl, line 26: nodes on clusters running on M1 machines will have the arm64 /arch label

pkg/kotsadm/objects/affinity_objects.go
```diff
@@ -1,34 +1,66 @@
...
+				Operator: corev1.NodeSelectorOpNotIn,
+				Values: []string{
+					"arm64",
+				},
+			},
+		},
+	}
+
+	var selectorTerm corev1.NodeSelectorTerm
+	switch runtime.GOOS {
```
divolgin, line 53: This is evaluated in CLI context, correct?  So deployment is done into a Linux cluster from a Mac client, this will get the `darwinSelectorTerm`?

stefanrepl: that is correct, and that is what i believed to be the ask here

PR #3330, opened 2022-10-18
Use release timestamps from chart instead of current time

divolgin:

Uses timestamps from charts info rather than hardcoded time.Now values.

Fixes #

```release-note
* Fixes an issue that causes Released timestamp to be the same for all releases on the [version history](/enterprise/updating-apps#update-an-application-in-the-admin-console) page in [Helm managed mode (Alpha)](/vendor/helm-install).
```

### DIFF
--- a/pkg/helm/updates.go
+++ b/pkg/helm/updates.go
@@ -25,6 +25,7 @@ import (
 	storetypes "github.com/replicatedhq/kots/pkg/store/types"
 	"github.com/replicatedhq/kots/pkg/util"
 	"go.uber.org/zap"
+	"gopkg.in/yaml.v3"
 	helmgetter "helm.sh/helm/v3/pkg/getter"
 	corev1 "k8s.io/api/core/v1"
 	kuberneteserrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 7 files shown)


pkg/helm/updates.go
```diff
@@ -252,8 +271,40 @@ func GetReplicatedSecretFromUpstreamChartVersion(helmApp *apptypes.HelmApp, lice
...
+	b := valuesData.Bytes()
+	values := struct {
+		Replicated *ReplicatedMeta `yaml:"replicated,omitempty"`
+	}{}
+	err = yaml.Unmarshal(b, &values)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to decode values.yaml")
+	}
+
+	return values.Replicated, nil
```
cbodonnell, line 294: Is it possible that `values.Replicated` is `nil` here?  If it's possible, it looks like it may panic when passing it to `helm.SetCachedUpdateMetadata(...)` [here](https://github.com/replicatedhq/kots/pull/3330/files#diff-5312bb96f485723d97db2688be857c2730d95bb339425202e587e69e4d8e4f66R467).

cbodonnell (dismissed):

pkg/helm/updates.go
divolgin: yep, updated!

cbodonnell (approved):

PR #3335, opened 2022-10-19
Use rqlite instead of Postgres

sgalsaleh:

Replaces Postgres with rqlite because:

- rqlite supports data replication at the application layer which removes the need for infrastructure-distributed storage.
- rqlite is more lightweight

Fixes [SC-60222](https://app.shortcut.com/replicated/story/60222/deploy-rqlite-instead-of-postgres-and-migrate-data-from-postgres-to-rqlite)

**New Installs:**

- New installs will deploy and use an instance of rqlite instead of Postgres.

**Upgrades and migration notes:** 

- Postgres's schema is always updated to the most recent version (e.g. currently the schema in KOTS 1.88) before running the migration. This makes the migration process much simpler since there would only be a single permutation for the migration to cover. It also ensures a smoother migration process from older versions of KOTS that have outdated schemas.

- The migration process will only be triggered if Postgres is detected AND the migration did not run previously. This ensures that the migration does not run again even if Postgres still exists in the cluster for some reason after the migration (e.g. KOTS failed to clean up Postgres resources).

- A Postgres instance is detected if the `kotsadm-postgres` secret exists in the cluster with a connection string / URI.

- Postgres resources will be cleaned up and removed from the cluster after a successful upgrade.

- Two migration tests were added:
        - One that tests upgrading from an older version of KOTS (v1.57.0) and ensures the migration runs successfully.
        - A more thorough integration test that validates the integrity of both the schema and the data before and after the migration.

- Timestamps in Postgres are converted to Unix timestamps (integers) when stored in or migrated to rqlite.

- Byte arrays in Postgres are converted to Base64 encoded strings when stored in or migrated to rqlite.

**Other notes:**

- Adds a `kots enable-ha` CLI command that will enable HA mode by scaling up the `kotsadm-rqlite` StatefulSet to `3` replicas and distribute them across different nodes. The command will print an error message if the criteria to enable HA are not met. The criteria right now is having at least `3` nodes in the cluster.

## Steps to reproduce

- Regular KOTS Install. OR:
- Regular KOTS upgrade from an older version.

```release-note

```

https://github.com/replicatedhq/replicated-docs/pull/713


pkg/ha/ha.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+const (
+	KUBERNETES_HA_MIN_NODE_COUNT = 3
+	RQLITE_HA_NODE_COUNT         = 3
+	REASON_NOT_ENOUGH_NODES      = "not enough nodes to run in HA mode"
+)
+
+func CanRunHA(ctx context.Context, clientset kubernetes.Interface) (bool, string, error) {
+	nodes, err := clientset.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
```
divolgin, line 23: Should we add selectors to match out node affinities? https://github.com/replicatedhq/kots/blob/10b6d379ae2b7ade0e482e3ebbb9fe22709da00e/pkg/kotsadm/objects/affinity_objects.go#L8

pkg/app/app.go
```diff
@@ -1,53 +1,66 @@
...
-	query := `update app set last_update_check_at = $1 where id = $2`
-	_, err := db.Exec(query, t, appID)
+	query := `update app set last_update_check_at = ? where id = ?`
+	wr, err := db.WriteOneParameterized(gorqlite.ParameterizedStatement{
+		Query:     query,
+		Arguments: []interface{}{t.Unix(), appID},
+	})
 	if err != nil {
-		return errors.Wrap(err, "failed to update last_update_check_at")
+		return fmt.Errorf("failed to update last_update_check_at: %v: %v", err, wr.Err)
```
divolgin, line 21: How does RQLite API work:   If `err != nil` then `wr.Err` will also be set?  And will they not contain the same message?

How about
```
errors.Wrapf(err, "failed to update last_update_check_at: %v", wr.Err)
```

pkg/k8sutil/kotsadm.go
```diff
@@ -203,7 +204,54 @@ func WaitForKotsadm(clientset *kubernetes.Clientset, namespace string, timeoutWa
...
 
 		if time.Since(start) > timeoutWaitingForWeb {
-			return "", &kotsadmtypes.ErrorTimeout{Message: "timeout waiting for kotsadm pod"}
+			return "", &types.ErrorTimeout{Message: "timeout waiting for kotsadm pod"}
+		}
+	}
+}
+
+func RestartKotsadm(ctx context.Context, clientset *kubernetes.Clientset, namespace string, timeout time.Duration) error {
+	pods, err := clientset.CoreV1().Pods(namespace).List(context.TODO(), metav1.ListOptions{LabelSelector: "app=kotsadm"})
```
divolgin, line 213: This use `Delete` here with label selector instead of `List` and then there is `WaitForDeploymentReady` (or `calculateDeploymentState`, not sure which is better).

divolgin: And I just saw this this is existing function that was moved, so this is not new.

pkg/store/kotsstore/app_store.go
```diff
@@ -15,6 +14,7 @@ import (
 	"github.com/replicatedhq/kots/pkg/logger"
 	"github.com/replicatedhq/kots/pkg/persistence"
 	troubleshootanalyze "github.com/replicatedhq/troubleshoot/pkg/analyze"
+	"github.com/rqlite/gorqlite"
```
divolgin, line 17: After the fact and not asking for changes, but you could alias `sql "github.com/rqlite/gorqlite"` to have fewer changes in the code :)

pkg/kotsadm/objects/rqlite_objects.go
```diff
@@ -0,0 +1,277 @@
...
+					TargetPort: intstr.FromString("rqlite"),
+				},
+			},
+		},
+	}
+
+	return service
+}
+
+func getRqliteVolumes(deployOptions types.DeployOptions) []corev1.Volume {
```
cbodonnell, line 229: `deployOptions` is not used in this function or the one below this.

pkg/app/app.go
sgalsaleh: This one is odd and that's the best format I was able to come up with. Basically, if there is an error, both `err` and `wr.Err` will be set, but they'll contain different messages, like:

- `err` would be `there were 2 statement errors`
- `wr.Err` would be the actual error, e.g. `syntax error, unexpected )` or something like that.

i tried doing it the way you're suggesting, but the error message ended up weird, like:
```
failed to update: syntax error, unexpected ), there were 2 statement errors
```

but the way i currently have it, it would look a bit better, like:
```
failed to update: there were 2 statement errors: syntax error, unexpected )
```

pkg/k8sutil/kotsadm.go
sgalsaleh: actually, i'll fix it

pkg/store/kotsstore/app_store.go
sgalsaleh: ha, i'll keep that in mind for the next migration!

pkg/k8sutil/kotsadm.go
sgalsaleh: also, it doesn't look like the `Delete` function accepts label selectors

divolgin (approved):

cbodonnell (approved):

PR #3336, opened 2022-10-19
fix airgap uploader

rianfowler:
- fixes airgap uploader 

Fixes #

Uploaded an invalid airgap bundle to test fix: 

none

none

### DIFF
--- a/web/src/features/Dashboard/components/Dashboard.tsx
+++ b/web/src/features/Dashboard/components/Dashboard.tsx
@@ -850,7 +850,7 @@ const Dashboard = (props: Props) => {
                 </p>
                 <div className="ExpandedError--wrapper u-marginTop--10 u-marginBottom--10">
                   <p className="u-fontSize--normal u-textColor--error">
-                    {airgapUploaderror}
+                    {state.airgapUploadError}
                   </p>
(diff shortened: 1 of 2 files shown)


web/src/features/Dashboard/components/Dashboard.tsx
```diff
@@ -170,6 +168,7 @@ const Dashboard = (props: Props) => {
   const history = useHistory();
   const match = useRouteMatch();
   const { app, isBundleUploading, isVeleroInstalled } = props;
+  const airgapUploader = useRef<AirgapUploader | null>(null);
```
rianfowler, line 171: rather than storing the uploader on the state, we need to store it in a ref. async functions work differently in functional components and keep a reference to the context they're assigned in so the old `state.airgapUploader` references in async functions would never reference the latest version of state. 

https://css-tricks.com/dealing-with-stale-props-and-states-in-reacts-functional-components/

rianfowler: There's probably a better way to do this but this is a quick fix to get the regression tests passing.

erquhart (approved):

PR #3339, opened 2022-10-20
remove unneccessary withRouter calls

rianfowler:
- `withRouter` is deprecated in react-router 6.0. 
- this is the first removal step- removing `withRouter` calls in components that don't use any of the props that `withRouter` decorates on its children (e.g. `props.match` and `props.history`

- the branch tests should adequately verify this since any component that tries to use router params without it being available will crash or not route correctly

none- refactor 

none

### DIFF
--- a/web/src/components/AirgapUploadProgress.jsx
+++ b/web/src/components/AirgapUploadProgress.jsx
@@ -1,5 +1,4 @@
 import React from "react";
-import { withRouter } from "react-router-dom";
 import Loader from "./shared/Loader";
 import {
   formatByteSize,
--- a/web/src/components/AirgapUploadProgress.jsx
+++ b/web/src/components/AirgapUploadProgress.jsx
@@ -338,4 +338,4 @@ AirgapUploadProgress.defaultProps = {
   progress: 0,
 };
 
-export default withRouter(AirgapUploadProgress);
+export default AirgapUploadProgress;
--- a/web/src/components/apps/AppDetailPage.tsx
+++ b/web/src/components/apps/AppDetailPage.tsx
@@ -520,7 +520,7 @@ function AppDetailPage(props: Props) {
                   render={() => (
                     <TroubleshootContainer
                       app={selectedApp}
-                      appName={appName}
+                      appName={appName || ""}
                     />
                   )}
                 />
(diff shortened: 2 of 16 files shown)


web/src/components/AirgapUploadProgress.jsx
```diff
@@ -338,4 +338,4 @@ AirgapUploadProgress.defaultProps = {
   progress: 0,
 };
 
-export default withRouter(AirgapUploadProgress);
+export default AirgapUploadProgress;
```
rianfowler, line 340: most of the changes are removals of `withRouter` calls. This decorator adds several react-router related props to the component. Several components use this decorated but not the props.

web/src/components/clusters/GitOps.jsx
```diff
@@ -9,10 +9,6 @@ import "../../scss/components/watches/WatchedApps.scss";
 import "../../scss/components/watches/WatchCard.scss";
 
 export class GitOps extends React.Component {
-  static propTypes = {
-    history: PropTypes.object.isRequired,
-  };
-
```
rianfowler, line 15: Don't need this propType. Shouldn't use propTypes anymore since we're using TS

web/src/components/apps/AppDetailPage.tsx
```diff
@@ -520,7 +520,7 @@ function AppDetailPage(props: Props) {
                   render={() => (
                     <TroubleshootContainer
                       app={selectedApp}
-                      appName={appName}
+                      appName={appName || ""}
```
rianfowler, line 523: think the typechecks script is not returning an error on failure. will investigate in another pr. This was a type error i saw while scrolling through the component

web/src/components/troubleshoot/TroubleshootContainer.tsx
```diff
@@ -9,13 +9,12 @@ import EditRedactor from "../redactors/EditRedactor";
...
 // Types
 import { App } from "@types";
-import { RouteComponentProps } from "react-router-dom";
 
 type Props = {
-  app: App;
+  app: App | null;
   appName: string;
 };
-class TroubleshootContainer extends Component<Props & RouteComponentProps> {
```
rianfowler, line 18: no longer need to have the react router props type definition on this component.

miaawong (approved):

PR #3340, opened 2022-10-21
remove withRouter from BackupRestore

rianfowler:
- removes `withRouter` from BackupRestore
- `withRouter` is deprecated in react-router 6. This is necessary to upgrade the dependency 

none- refactor 

none

### DIFF
--- a/web/src/components/BackupRestore.jsx
+++ b/web/src/components/BackupRestore.jsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { withRouter, Link } from "react-router-dom";
+import { Link } from "react-router-dom";
 import { KotsPageTitle } from "@components/Head";
 import isEmpty from "lodash/isEmpty";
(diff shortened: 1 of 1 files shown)


web/src/components/BackupRestore.jsx
```diff
@@ -239,7 +239,6 @@ class BackupRestore extends React.Component {
         renderNotVeleroMessage={this.renderNotVeleroMessage}
         hideCheckVeleroButton={this.state.hideCheckVeleroButton}
         isLicenseUpload={true}
-        history={this.props.history}
```
rianfowler, line 242: This is the only `withRouter` prop used in this component. `ConfigureSnapshots` doesn't use it so we can safely delete it.

rianfowler (commented):
simple refactor

miaawong (approved):

PR #3343, opened 2022-10-21
Remove withRouter from UploadAirgapBundle

rianfowler:
- removes `withRouter` from `UploadAirgapBundle` and replaces with hooks from react-router
- `withRouter` is deprecated in react-router 6.0

none-refactor 

none


web/src/components/UploadAirgapBundle.tsx
```diff
@@ -0,0 +1,738 @@
...
+  preparingOnlineInstall: boolean;
+  supportBundleCommand?: string | string[];
+  showSupportBundleCommand: boolean;
+  simultaneousUploads?: number;
+  onlineInstallErrorMessage: string;
+  uploadProgress: number;
+  uploadSize: number;
+  uploadResuming: boolean;
+  viewOnlineInstallErrorMessage: boolean;
+};
```
rianfowler, line 67: Added type definitions

web/src/components/UploadAirgapBundle.tsx
```diff
@@ -0,0 +1,738 @@
...
+      uploadSize: 0,
+      uploadResuming: false,
+      viewOnlineInstallErrorMessage: false,
+    }
+  );
+
+  const emptyHostnameErrMessage = 'Please enter a value for "Hostname" field';
+  const match = useRouteMatch<KotsParams>();
+  const history = useHistory();
+  const appSlug = match.params.slug;
```
rianfowler, line 93: At first I tried to use `selectedApp` here but I think at this point in the flow there is no `selectedApp` because we're still uploading the airgap bundle. The `selectedApp` hook does look at the match.params like this but it returns an App object only if `/apps` has returned an app that matches the slug.

web/src/components/UploadAirgapBundle.tsx
```diff
@@ -0,0 +1,738 @@
...
+          >
+            Ok, got it!
+          </button>
+        </div>
+      </Modal>
+    </div>
+  );
+};
+
+export default UploadAirgapBundle;
```
rianfowler, line 738: no more `withRouter`

web/src/components/UploadAirgapBundle.tsx
```diff
@@ -0,0 +1,738 @@
...
+      viewOnlineInstallErrorMessage: false,
+    }
+  );
+
+  const emptyHostnameErrMessage = 'Please enter a value for "Hostname" field';
+  const match = useRouteMatch<KotsParams>();
+  const history = useHistory();
+  const appSlug = match.params.slug;
+  // TODO: refactor the /resume fetching so this isn't necessary
+  const onlineInstallErrorMessage = useRef<string>("");
```
rianfowler, line 93: This is being access from inside of an interval so need to use a ref to avoid stale state. We need to refactor it so that it fires a chain of promises instead of using an interval to inspect a value set by an async function.

PR #3346, opened 2022-10-24
remove withRouter from Access component

rianfowler:
- refactors Access component to use react router hook instead of `withRouter`
- `withRouter` is deprecated in react-router 6

none-refactor

none

### DIFF
--- a/web/src/Root.tsx
+++ b/web/src/Root.tsx
@@ -389,11 +389,6 @@ const Root = () => {
     return !!find(apps, (app) => app.isIdentityServiceSupported);
   };
 
-  const isGeoaxisSupported = () => {
-    const apps = state.appsList;
-    return !!find(apps, (app) => app.isGeoaxisSupported);
-  };
-
   const isSnapshotsSupported = () => {
     const apps = state.appsList;
     return !!find(apps, (app) => app.allowSnapshots);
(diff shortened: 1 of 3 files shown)


web/src/Root.tsx
```diff
@@ -595,11 +595,9 @@ const Root = () => {
...
                 path="/access/:tab?"
-                render={(props) => (
+                render={() => (
                   <Access
-                    {...props}
-                    appName={state.selectedAppName}
-                    isKurlEnabled={state.adminConsoleMetadata?.isKurl}
+                    isKurlEnabled={state.adminConsoleMetadata?.isKurl || false}
                     isGeoaxisSupported={isGeoaxisSupported()}
                   />
```
rianfowler, line 602: cleaned up some of the copy pasta here. Access doesn't use the props passed from ProtectedRoute

web/src/components/identity/Access.tsx
```diff
@@ -44,14 +43,13 @@ class Access extends Component {
...
+        />
       </div>
-    );
-  }
-}
+    </div>
+  );
+};
 
-export default withTheme(withRouter(Access));
```
rianfowler, line 57: I don't see any props from `withTheme` being used here or passed down to other components. LMK if removing this is an error

web/src/components/identity/Access.tsx
```diff
@@ -1,34 +1,33 @@
...
-  componentDidMount() {
-    const { history } = this.props;
+type Props = {
+  isKurlEnabled: boolean;
+  isGeoaxisSupported: boolean;
+};
 
+const Access = (props: Props) => {
+  const history = useHistory();
+  // TODO: move this into a redirect route or update links to default to /identity-providers
```
rianfowler, line 18: don't want to deal with this until we clean up the routes and put them into their own files

miaawong (approved):

PR #3347, opened 2022-10-24
fix: installing/updating/configuring applications that have many kots kinds images hangs or takes a really long time

sgalsaleh:

Fixes an issue where installing, updating, or configuring applications that have many images defined in KOTS custom resources (e.g. collectors, preflights, analyzers, etc..) hangs or takes a really long time.

Fixes [SC-61040](https://app.shortcut.com/replicated/story/61040/installing-updating-configuring-applications-that-have-many-images-used-in-kots-kinds-collectors-preflights-etc-hangs-or)

Fixes #

```release-note
Fixes an issue where installing, updating, or configuring applications that have many images defined in KOTS custom resources (e.g. collectors, preflights, analyzers, etc..) hangs or takes a really long time.
```

NONE

### DIFF
--- a/pkg/image/builder.go
+++ b/pkg/image/builder.go
@@ -85,6 +85,9 @@ func GetPrivateImages(upstreamDir string, kotsKindsImages []string, checkedImage
 	objectsWithImages := make([]k8sdoc.K8sDoc, 0) // all objects where images are referenced from
 
 	for _, image := range kotsKindsImages {
+		if _, ok := checkedImages[image]; ok {
+			continue
+		}
 		if allPrivate {
 			checkedImages[image] = types.ImageInfo{
 				IsPrivate: true,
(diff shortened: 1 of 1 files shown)


pkg/image/builder.go
```diff
@@ -85,6 +85,9 @@ func GetPrivateImages(upstreamDir string, kotsKindsImages []string, checkedImage
 	objectsWithImages := make([]k8sdoc.K8sDoc, 0) // all objects where images are referenced from
 
 	for _, image := range kotsKindsImages {
+		if _, ok := checkedImages[image]; ok {
```
divolgin, line 88: How about moving this to the `else` part, so if `allPrivate` is set at a later point, all the flags get flipped.  Would happen if `proxyPublicImages` is set after initial install.

divolgin (approved):

PR #3348, opened 2022-10-25
Don't keep polling preflight status when preflights are not running

divolgin:

If preflights fail to start due to some error, the UI should not be showing the progress bar as if preflights are running.  This change makes this UI progress with the message that preflights did not run.

Fixes #

```release-note
* Fixes an issue that could cause the preflight progress bar to be stuck close to 100% but never complete.
```

### DIFF
--- a/pkg/preflight/execute.go
+++ b/pkg/preflight/execute.go
@@ -9,6 +9,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/replicatedhq/kots/pkg/k8sutil"
 	"github.com/replicatedhq/kots/pkg/logger"
+	"github.com/replicatedhq/kots/pkg/preflight/types"
 	"github.com/replicatedhq/kots/pkg/store"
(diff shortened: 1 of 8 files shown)


pkg/preflight/execute.go
```diff
@@ -125,20 +165,6 @@ func execute(appID string, sequence int64, preflightSpec *troubleshootv1beta2.Pr
...
-	b, err := json.Marshal(uploadPreflightResults)
-	if err != nil {
-		return uploadPreflightResults, errors.Wrap(err, "failed to marshal results")
-	}
-
-	completeMx.Lock()
-	defer completeMx.Unlock()
-
-	isComplete = true
-	if err := store.GetStore().SetPreflightResults(appID, sequence, b); err != nil {
```
divolgin, line 138: This is the call that updates data and stops polling, but it was only being called when preflights run successfully.

rianfowler (dismissed):

cbodonnell (dismissed):

cbodonnell (approved):

PR #3352, opened 2022-10-26
Rianfowler/sc 59398/remove withRouter from EditRedactor

rianfowler:
- removes `withRouter` from EditRedactor 
- needed to upgrade react-router 

none- refactor 

none


web/src/components/redactors/EditRedactor.tsx
```diff
@@ -0,0 +1,433 @@
...
+  editingRedactor: boolean;
+  isLoadingRedactor: boolean;
+  redactorEnabled: boolean;
+  redactorErrMsg: string;
+  redactorName: string;
+  redactorYaml: string;
+};
+
+// TODO: upgrade AceEditor so we can use the type definitions
+let aceEditor: AceEditor | null;
```
rianfowler, line 34: this thing is 4 major versions behind. This was originally stored in the instance but storing here for now. I think this should be fine unless we have multiple editors visible at once. Not even sure why this is needed and maybe we can get rid of it after upgrading `AceEditor`

web/src/components/redactors/EditRedactor.tsx
```diff
@@ -0,0 +1,433 @@
+import React, { useEffect, useReducer } from "react";
+import { Link, useHistory, useRouteMatch } from "react-router-dom";
+import { KotsPageTitle } from "@components/Head";
+import AceEditor, { Marker } from "react-ace";
```
rianfowler, line 4: just took a guess that this `Marker` type would be available

web/src/components/redactors/EditRedactor.tsx
```diff
@@ -0,0 +1,433 @@
...
+      redactorEnabled: false,
+      redactorErrMsg: "",
+      redactorName: "",
+      redactorYaml: "",
+    }
+  );
+
+  const history = useHistory();
+  const match = useRouteMatch<KotsParams>();
+  const slug = useSelectedApp()?.selectedApp?.slug || "";
```
rianfowler, line 59: I've been thinking about how to standardize our hooks interfaces. i def like the idea of returning `react-query`'s api but `selectedApp` is fully custom hook and it might be better to follow the react-router approach and just return `selectedApp` and not `{ selectedApp }` 

Also thinking that on the `react-query` returns we could add something like `dataFlattened` and just add the stuff that components use from the responses in a flattened data structure instead of spreading the api structure all over the app.

web/src/components/redactors/EditRedactor.tsx
```diff
@@ -0,0 +1,433 @@
...
+    removals:
+      values:
+      - "removethis"`;
+      setState({
+        redactorEnabled: true,
+        redactorYaml: defaultYaml,
+        redactorName: "New redactor",
+      });
+    }
+  }, []);
```
rianfowler, line 271: replaces `componentDidMount`

```
  componentDidMount() {
    if (this.props.match.params.redactorSlug) {
      this.getRedactor(this.props.match.params.redactorSlug);
    } else {
      const defaultYaml = `kind: Redactor
apiVersion: troubleshoot.sh/v1beta2
metadata:
  name:
spec:
  redactors:
  - name: myredactor
    fileSelector:
      files:
      - "abc"
    removals:
      values:
      - "removethis"`;
      this.setState({
        redactorEnabled: true,
        redactorYaml: defaultYaml,
        redactorName: "New redactor",
      });
    }
  }
```

web/src/components/redactors/EditRedactor.tsx
```diff
@@ -0,0 +1,433 @@
...
+            theme="chrome"
+            className="flex1 flex"
+            value={state.redactorYaml}
+            height="100%"
+            width="100%"
+            markers={state.activeMarkers}
+            editorProps={{
+              $blockScrolling: Infinity,
+              // @ts-ignore
+              useSoftTabs: true,
```
rianfowler, line 384: ignoring until we upgrade this

web/src/components/redactors/EditRedactor.tsx
miaawong: curious how that would look like! I down to give it a try

miaawong (approved):

PR #3372, opened 2022-10-28
add unit test coverage for snapshot store

cbodonnell:

This PR refactors the code that updates the existing snapshot store into its own method so that it can be unit tested.  Coverage was added for the local-volume-provider store permutation.

n/a

In general, the pattern of instantiating a k8s client within a function makes it difficult to unit test that function since it requires a cluster context.  This PR instantiates the client outside of the new function so that a fake k8s client can be used for unit testing.

#### Steps to reproduce

`make test`

```release-note
NONE
```

NONE

### DIFF
--- a/pkg/snapshot/store.go
+++ b/pkg/snapshot/store.go
@@ -93,263 +93,26 @@ func (e *InvalidStoreDataError) Error() string {
 }
 
 func ConfigureStore(ctx context.Context, options ConfigureStoreOptions) (*types.Store, error) {
-	store, err := GetGlobalStore(ctx, options.KotsadmNamespace, nil)
+	existingStore, err := GetGlobalStore(ctx, options.KotsadmNamespace, nil)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get store")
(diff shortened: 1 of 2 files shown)


pkg/snapshot/store.go
```diff
@@ -275,13 +333,14 @@ func ConfigureStore(ctx context.Context, options ConfigureStoreOptions) (*types.
 		store.Internal.ObjectStoreClusterIP = string(secret.Data["object-store-cluster-ip"])
 		store.Internal.Region = "us-east-1"
 	} else if options.Internal && options.IsMinioDisabled {
+		// TODO: remove the need for cluster context so this code path can be unit tested
 		isKurl, err := kurl.IsKurl()
```
cbodonnell, line 337: `kurl.IsKurl(...)` requires cluster context, so I don't think this code path can be unit tested.  Wondering if we can have `kurl.IsKurl(...)` accept a k8s client and just fall back to the cluster context if `nil` is passed?

divolgin (approved):

PR #3377, opened 2022-10-31
change prometheus disk usage query

garcialuis:

This PR changes the prometheus query for disk usage so that it shows the usage per instance and device.
Prior to this change the disk usage graph would should a spike when a new device was mounted and was not as useful. 

Fixes #
[SC-50066](https://app.shortcut.com/replicated/story/50066/kots-disk-usage-reports-a-total-of-all-filesystems-mounted)

The disk usage graph now shows each device and what instance they are in:
Screen Shot 2022-10-31 at 10 06 59 AM

## Steps to reproduce

1. Use a cluster with Prometheus and KOTS
2. Add and mount a device to the instance
3. Inspect the disk usage graph, it should show the instance-device

Additional nodes can be joined to see metrics from different instances, the picture above is from a 2 node cluster.

```release-note
Updates prometheus query to show disk usage by instance and mount point.
```

NONE

### DIFF
--- a/pkg/version/metrics.go
+++ b/pkg/version/metrics.go
@@ -56,12 +56,12 @@ var (
 			Title: "Disk Usage",
 			Queries: []kotsv1beta1.MetricQuery{
 				{
-					Query:  `sum(node_filesystem_size_bytes{job=~"node-exporter|kubernetes-service-endpoints",fstype!="",instance!=""} - node_filesystem_avail_bytes{job=~"node-exporter|kubernetes-service-endpoints",fstype!="",instance!=""}) by (instance)`,
-					Legend: "Used: {{ instance }}",
+					Query:  `node_filesystem_size_bytes{fstype=~"ext4|xfs",instance!="",job=~"node-exporter|kubernetes-service-endpoints"} - node_filesystem_avail_bytes{fstype=~"ext4|xfs",instance!="",job=~"node-exporter|kubernetes-service-endpoints"}`,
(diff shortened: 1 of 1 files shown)


pkg/version/metrics.go
```diff
@@ -56,12 +56,12 @@ var (
 			Title: "Disk Usage",
 			Queries: []kotsv1beta1.MetricQuery{
 				{
-					Query:  `sum(node_filesystem_size_bytes{job=~"node-exporter|kubernetes-service-endpoints",fstype!="",instance!=""} - node_filesystem_avail_bytes{job=~"node-exporter|kubernetes-service-endpoints",fstype!="",instance!=""}) by (instance)`,
-					Legend: "Used: {{ instance }}",
+					Query:  `node_filesystem_size_bytes{fstype=~"ext4|xfs",instance!="",job=~"node-exporter|kubernetes-service-endpoints"} - node_filesystem_avail_bytes{fstype=~"ext4|xfs",instance!="",job=~"node-exporter|kubernetes-service-endpoints"}`,
```
divolgin, line 59: How about having these 4 on the list: BTRFS, ZFS, XFS, EXT4?

Another option is to have the exclude list based on this: https://github.com/replicatedhq/kURL/blob/738f82e486f8864b5dc85573db72a27dcbd3a7cc/addons/collectd/v5/collectd.conf#L24-L32

garcialuis: Thanks, I added the two that were missing: btrfs and zfs

divolgin (approved):

PR #3380, opened 2022-11-01
fix license expiration check in UI

cbodonnell:

This PR fixes an issue where customer licenses were not showing as expired in the UI, but when an end user tried to fetch updates, it would fail with a message indicating that the license was expired.  Additionally, it fixes an issue where the UTC expiration time of the license was being compared against the local time, so it made for an inconsistent comparison. This PR changes this so that the UTC time on the license is compared against the current UTC time.

When syncing license in KOTS (before):
kots-license-expiration-before

Checking for updates:
kots-ui-license-expired

After this PR, the UI will properly display if the license is expired after syncing:

kots-license-expiration-after

Fixes [SC-59063](https://app.shortcut.com/replicated/story/59063/can-t-check-for-updates-because-license-is-expired-but-neither-the-admin-console-nor-vendor-portal-says-it-is-expired-yet)

## Steps to reproduce

* Create or update a license in vendor portal with expiration set to today
* Observe that license does not show expired once license is synced
* Try to install or fetch updates with that license

```release-note
Fixes an issue where licenses were not showing as expired in the **License** tab, but would fail when checking for updates with the message "License is expired".
```

### DIFF
--- a/web/src/utilities/utilities.test.js
+++ b/web/src/utilities/utilities.test.js
@@ -1,0 +1,17 @@
+import { Utilities } from "./utilities";
+
+describe("Utilities", () => {
+  describe("checkIsDateExpired", () => {
+    it("should return true if date is expired", () => {
+      const date = new Date().setMinutes(new Date().getMinutes() - 1);
+      const timestamp = new Date(date).toISOString();
+      expect(Utilities.checkIsDateExpired(timestamp)).toBe(true);
+    });
+
+    it("should return false if date is not expired", () => {
+      const date = new Date().setHours(new Date().getHours() + 1);
+      const timestamp = new Date(date).toISOString();
+      expect(Utilities.checkIsDateExpired(timestamp)).toBe(false);
+    });
+  });
+});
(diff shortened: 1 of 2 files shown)


web/src/utilities/utilities.test.js
```diff
@@ -0,0 +1,17 @@
...
+      expect(Utilities.checkIsDateExpired(timestamp)).toBe(true);
+    });
+
+    it("should return false if date is not expired", () => {
+      const date = new Date().setHours(new Date().getHours() + 1);
+      const timestamp = new Date(date).toISOString();
+      expect(Utilities.checkIsDateExpired(timestamp)).toBe(false);
+    });
+  });
+});
```
cbodonnell, line 17: This test will fail on main branch:
```
    expect(received).toBe(expected) // Object.is equality

    Expected: true
    Received: false

       6 |       const date = new Date().setMinutes(new Date().getMinutes() - 1);
       7 |       const timestamp = new Date(date).toISOString();
    >  8 |       expect(Utilities.checkIsDateExpired(timestamp)).toBe(true);
         |                                                       ^
       9 |     });
      10 |
      11 |     it("should return false if date is not expired", () => {

      at Object.toBe (src/utilities/utilities.test.js:8:55)
```

rianfowler (approved):

PR #3383, opened 2022-11-02
add poll for apps in helm install component

rianfowler:
- redirect from helm install page if there is an app installed 

https://app.shortcut.com/replicated/story/60232/the-install-with-helm-screen-should-automatically-redirect-to-apps-page-when-a-chart-is-installed

with your kots install in helm mode, you can just navigate directly to /install-with-helm to test the redirect behavior. 

```release-note
- fixes an issue where kots would not redirect from `/install-with-helm` automatically after an application is installed 
```

none

### DIFF
--- a/web/src/features/AddNewApp/components/InstallWithHelm.tsx
+++ b/web/src/features/AddNewApp/components/InstallWithHelm.tsx
@@ -1,6 +1,17 @@
+import { useApps } from "@features/App";
 import React from "react";
+import { useHistory } from "react-router";
 
 function InstallWithHelm() {
+  // TODO: // move this into a routes component
+  // poll for apps data and redirect if app is installed
+  const { apps } = useApps({ refetchInterval: 2000 }).data || {};
+  const history = useHistory();
+
+  if (apps && apps?.length > 0) {
+    history.push("/apps");
+  }
+
   return (
     <div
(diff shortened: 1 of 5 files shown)


web/src/features/AddNewApp/components/InstallWithHelm.tsx
```diff
@@ -1,6 +1,17 @@
...
 function InstallWithHelm() {
+  // TODO: // move this into a routes component
+  // poll for apps data and redirect if app is installed
+  const { apps } = useApps({ refetchInterval: 2000 }).data || {};
+  const history = useHistory();
+
+  if (apps && apps?.length > 0) {
+    history.push("/apps");
+  }
+
```
rianfowler, line 14: pretty simple change here

web/src/features/App/api/useLicense.ts
```diff
@@ -26,7 +26,7 @@ function useLicenseWithIntercept() {
         return x;
       },
       (error) => {
-        if (error.response.status === 401) {
+        if (error?.response?.status=== 401) {
```
rianfowler, line 29: was getting a lot of console log noise from this from the cors errors. I think I fixed all of them.

rianfowler:
I'm updating the `helm-managed` suite by removing the step that manually navigates to the app page.

miaawong (approved):

PR #3386, opened 2022-11-03
Rianfowler/sc 59398/remove withRouter from UploadLicenseFile

rianfowler:
- removes `withRouter` from `UploadLicenseFile` and replaces with hooks from react-router
- `withRouter` is deprecated in react-router 6.0

none-refactor 

none


web/src/components/UploadLicenseFile.tsx
```diff
@@ -0,0 +1,683 @@
+import React, { useEffect, useReducer } from "react";
+import { useHistory } from "react-router";
+import { Link } from "react-router-dom";
+import { KotsPageTitle } from "@components/Head";
+// TODO: upgrade this dependency
+// @ts-ignore
+import Dropzone from "react-dropzone";
```
rianfowler, line 7: I tried adding the `@types` but I started getting a bunch of errors. I think we might be using this component the wrong way.

web/src/components/UploadLicenseFile.tsx
```diff
@@ -0,0 +1,683 @@
...
+                <CodeSnippet
+                  language="bash"
+                  canCopy={true}
+                  onCopyText={
+                    <span className="u-textColor--success">
+                      Command has been copied to your clipboard
+                    </span>
+                  }
+                >
+                  {typeof licenseExistErrData === "string"
```
rianfowler, line 663: added a type guard here. we will need to make sure we refactor this kind of thing out. shouldn't be assigning strings and objects to the same variable.

web/src/components/apps/AppLicense.tsx
```diff
@@ -157,15 +157,19 @@ const AppLicenseComponent = (props: Props) => {
     const airgapLicense = await yaml.safeLoad(contentStr);
     const { appLicense } = state;
 
-    if (airgapLicense.spec?.licenseID !== appLicense?.id) {
+    // TODO: FIX THIS
+    // @ts-ignore
+    if (airgapLicense?.spec?.licenseID !== appLicense?.id) {
```
rianfowler, line 162: dunno how this got in. going to standardize the makefiles / linting commands next week so hopefully that will block these kinds of things during PR checks

miaawong: what is the error?

web/src/components/UploadLicenseFile.tsx
miaawong: are you referring to the `js-yaml` package?

miaawong (approved):

web/src/components/apps/AppLicense.tsx
rianfowler: the type returned by `TextDecoder` is not the type that `yaml.safeLoad` expects. I'm not sure what any of this is so I'm avoiding changing it for now. 

```
 const content: any = await getFileContent(files[0]);
    const contentStr = new TextDecoder("utf-8").decode(content);
    const airgapLicense = await yaml.safeLoad(contentStr);
    const { appLicense } = state;

    if (airgapLicense.spec?.licenseID !== appLicense?.id) {
    // TODO: FIX THIS
    // @ts-ignore
    if (airgapLicense?.spec?.licenseID !== appLicense?.id) {
```

web/src/components/UploadLicenseFile.tsx
rianfowler: I was able to add the types for that one and get it to work but `Dropzone` is like ten versions behind.

PR #3407, opened 2022-11-10
Fix logout redirect 2

rianfowler:
- kots occasionally redirects from the snapshot restore page to the login page
- See [this](https://replicated.slack.com/archives/C02TQTV9G2V/p1667504115509199?thread_ts=1667502472.595329&cid=C02TQTV9G2V) thread

Fixes # https://app.shortcut.com/replicated/story/61561/kots-occasionally-redirects-from-restore-success-page

```release-note
- fixes an issue where kots would automatically redirect to the login page from the Snapshot Restore Complete page
```

NONE


web/src/components/snapshots/AppSnapshotRestore.jsx
```diff
@@ -333,8 +326,7 @@ class AppSnapshotRestore extends Component {
     }
 
     if (restoreCompleted && hasNoErrorsOrWarnings) {
-      this.logOutUser();
-      this.props.history.push("/restore-completed");
+      Utilities.logoutUser(null, { snapshotRestore: true });
```
rianfowler, line 329: Previously this was using it's own logout command that did basically the same thing as the utility. I got rid of that and swapped in the utility.

web/src/utilities/utilities.js
```diff
@@ -700,8 +700,11 @@ export const Utilities = {
       window.localStorage.removeItem("session_roles");
     }
 
-    if (window.location.pathname !== "/secure-console") {
-      window.location = "/secure-console";
+    const redirectPath = options?.snapshotRestore
+      ? "/restore-complete"
+      : "/secure-console";
```
rianfowler, line 705: Kots is not set up to go anywhere but the login page when the user is not authed. This is a hacky way to support another route. Basically if the user is logged out it will redirect to the `/restore-complete` page when `options.snapshotRestore` is passed in.

Future state I think this should be in a top level routes definition file similarly to how BPR does it. The top level routes file uses any state necessary to make routing decisions: 

https://github.com/alan2207/bulletproof-react/blob/master/src/routes/index.tsx

We have similar functionality but we need to isolated it like this.

rianfowler:
Need to merge this testim branch in at the same time: https://app.testim.io/#/project/wpYAooUimFDgQxY73r17/branch/fix-logout-redirect-2/test/3KsnsQNZ2U4cpbxQ/step/UPXyiw0G2Z5dLWIu?result-id=abzOMOKobxOsXMWP&shared-step-instance=false&path=UPXyiw0G2Z5dLWIu

web/src/utilities/utilities.js
```diff
@@ -700,8 +700,11 @@ export const Utilities = {
       window.localStorage.removeItem("session_roles");
     }
 
-    if (window.location.pathname !== "/secure-console") {
-      window.location = "/secure-console";
+    const redirectPath = options?.snapshotRestore
+      ? "/restore-completed"
+      : "/secure-console";
+    if (window.location.pathname !== redirectPath) {
+      window.location = redirectPath;
```
rianfowler, line 707: Kots is not set up to go anywhere but the login page when the user is not authed. This is a hacky way to support another route. Basically if the user is logged out it will redirect to the /restore-complete page when options.snapshotRestore is passed in.

Future state I think this should be in a top level routes definition file similarly to how BPR does it. The top level routes file uses any state necessary to make routing decisions like useAuth in this example:

https://github.com/alan2207/bulletproof-react/blob/master/src/routes/index.tsx

We have similar functionality but we need to isolated it like this.

sgalsaleh (approved):

PR #3418, opened 2022-11-11
add typechecks to SupportBundleRow

rianfowler:
- Adds typechecks to SupportBundleRow

none

<!--
none


web/package.json
```diff
@@ -39,6 +39,7 @@
     "@testing-library/jest-dom": "^5.16.5",
     "@testing-library/react": "^13.3.0",
     "@testing-library/react-hooks": "^8.0.1",
+    "@types/downloadjs": "1.4",
```
rianfowler, line 42: this needs to be upgraded but i'm just adding the types for now.

web/src/components/troubleshoot/SupportBundleRow.tsx
```diff
@@ -1,44 +1,59 @@
...
 
-class SupportBundleRow extends React.Component {
-  state = {
-    downloadingBundle: false,
-    downloadBundleErrMsg: "",
-    errorInsights: [],
-    warningInsights: [],
-    otherInsights: [],
-  };
+let percentage: number;
```
rianfowler, line 19: this was being used as a string and number with math. instead of having the type flip and and forth i opted to make it a number, refactor the places where the type was swapping and use `.toString()` where it's actually displayed.

web/src/components/troubleshoot/SupportBundleRow.tsx
```diff
@@ -1,44 +1,59 @@
...
-    // const sharedIds = bundle.teamShareIds || [];
-    // const isShared = sharedIds.length;
-    // let shareContext;
-
-    // if (notSameTeam) {
-    //   shareContext = <span className="u-marginLeft--normal u-fontSize--normal u-textColor--success">Shared by <span className="u-fontWeight--bold">{bundle.teamName}</span></span>
-    // } else if (isSameTeam && isShared) {
-    //   shareContext = <span className="u-marginLeft--normal u-fontSize--normal u-fontWeight--medium u-textColor--secondary">Shared with Replicated</span>
-    // }
-    // return shareContext;
```
rianfowler, line 41: this is 3 years old so i think we can safely delete it

web/src/components/troubleshoot/SupportBundleRow.tsx
```diff
@@ -1,44 +1,59 @@
...
+
+class SupportBundleRow extends React.Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+
+    this.state = {
+      downloadingBundle: false,
+      sendingBundle: false,
+    };
+  }
```
rianfowler, line 50: added a constructor and removed all of the non boolean default values since they complicate the type definition.

web/src/components/troubleshoot/SupportBundleRow.tsx
```diff
@@ -165,11 +180,12 @@ class SupportBundleRow extends React.Component {
...
-  moveBar(progressData) {
+  moveBar(progressData: SupportBundleProgress) {
     const elem = document.getElementById("supportBundleStatusBar");
-    const calcPercent =
-      (progressData.collectorsCompleted / progressData.collectorCount) * 100;
-    percentage = calcPercent > 98 ? 98 : calcPercent.toFixed();
+    const calcPercent = Math.round(
+      (progressData.collectorsCompleted / progressData.collectorCount) * 100
+    );
+    percentage = calcPercent > 98 ? 98 : calcPercent;
```
rianfowler, line 188: - removed `.toFixed()` which is a number -> string conversion. this allows us to keep `percentage` as a number 
- replaced it with `Math.round` which always returns an integer

web/src/components/troubleshoot/SupportBundleRow.tsx
```diff
@@ -258,7 +274,7 @@ class SupportBundleRow extends React.Component {
                 onClick={() => this.handleBundleClick(bundle)}
               >
                 <div className="flex">
-                  {!this.props.isCustomer && bundle.customer ? (
```
rianfowler, line 261: as far as i can [tell](https://github.com/replicatedhq/kots/blob/939172d803075c2b79ad5d966f573c8e6df44175/pkg/supportbundle/types/types.go#L16-L36) there is no `customer` on the SupportBundle response. This comes straight off of the response and i didn't see anywhere where a customer was added to the response data by the FE code

web/src/components/troubleshoot/TroubleshootContainer.tsx
```diff
@@ -20,7 +20,7 @@ type State = {
   isGeneratingBundle: false;
   generateBundleErrMsg: string;
   loading: boolean;
-  bundleAnalysisProgress: Object;
+  bundleAnalysisProgress?: SupportBundleProgress;
```
rianfowler, line 23: i added this `SupportBundleProgress` type based off of the go types and just threw it on here since i was already looking at the file.

web/src/types/index.ts
```diff
@@ -139,6 +139,39 @@ export type ResourceStates = {
...
+  primary: string;
+  severity: string;
+};
+
+export type SupportBundleProgress = {
+  collectorCount: number;
+  collectorsCompleted: number;
+  message: string;
+};
+
```
rianfowler, line 174: all of this comes from the SupportBundle go types

web/src/components/troubleshoot/SupportBundleRow.tsx
```diff
@@ -375,7 +391,7 @@ class SupportBundleRow extends React.Component {
                     style={{ width: "350px" }}
                   >
                     <span className="u-fontWeight--bold u-fontSize--normal u-textColor--secondary u-marginRight--10">
-                      {percentage + "%"}
+                      {percentage.toString() + "%"}
```
rianfowler, line 394: I think react will handle the type conversion but i wanted to be explicit

web/src/components/troubleshoot/SupportBundleRow.tsx
```diff
@@ -292,7 +308,7 @@ class SupportBundleRow extends React.Component {
                     statusDiv
                   ) : bundle?.analysis?.insights?.length ? (
                     <div className="flex flex1 alignItems--center">
-                      {errorInsights.length > 0 && (
+                      {errorInsights && errorInsights.length > 0 && (
```
rianfowler, line 311: since we're using `undefined` as the default value instead of an empty array, need to check that it actually exists before using the length accessor

web/src/components/troubleshoot/TroubleshootContainer.tsx
```diff
@@ -36,7 +36,6 @@ class TroubleshootContainer extends Component<Props, State> {
       isGeneratingBundle: false,
       generateBundleErrMsg: "",
       loading: false,
-      bundleAnalysisProgress: {},
```
rianfowler, line 39: had to add some optional chaining to support this default: undefined

web/src/components/troubleshoot/SupportBundleRow.tsx
miaawong: sweet.

web/src/components/troubleshoot/SupportBundleRow.tsx
miaawong: we don't need to set default values to non boolean states?

web/src/components/troubleshoot/SupportBundleRow.tsx
miaawong: perfect

web/src/components/troubleshoot/SupportBundleRow.tsx
miaawong: good catch

web/src/components/troubleshoot/TroubleshootContainer.tsx
miaawong: tysm

miaawong (approved):
amazing

PR #3436, opened 2022-11-15
add intermediate step to validate pr tests

cbodonnell:

This PR addresses an issue with branch protection where the `validate-success` step would be skipped if the `validate-*` PR tests did not run, triggering branch protection success.  With these changes, the `validate-success` step will always run and checks that the `validate-pr-tests` step was run successfully and that the `validate-kurl-addon` step was either successful or was skipped (if no changes were made to the addon).

Example of skipped `validate-success` from fork: https://github.com/replicatedhq/kots/actions/runs/3437864499

Example of skipped `validate-success` from branch where tests are passing (but `validate-kurl-addon` skipped): https://github.com/replicatedhq/kots/actions/runs/3473232691

Fixes branch protection in preparation for work on [SC-61272](https://app.shortcut.com/replicated/story/61272/unable-to-run-tests-on-dependabot-prs)

```release-note

```

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -1830,19 +1830,16 @@ jobs:
       - validate-pr-tests
     if: always()
     steps:
-      # https://docs.github.com/en/actions/learn-github-actions/contexts#steps-context
+      # https://docs.github.com/en/actions/learn-github-actions/contexts#jobs-context
       # if the validate-kurl-addon job failed or was cancelled, this job will fail. it's ok if this job was skipped
       - name: fail if validate-kurl-addon job failed
-        if: steps.validate-kurl-addon.outcome == 'failed' || steps.validate-kurl-addon.outcome == 'cancelled'
+        if: jobs.validate-kurl-addon.result == 'failure' || jobs.validate-kurl-addon.result == 'cancelled'
         run: exit 1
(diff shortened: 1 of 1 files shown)


camilamacedo86 (dismissed):

.github/workflows/build-test.yaml
```diff
@@ -1788,8 +1788,7 @@ jobs:
           aws-secret-access-key: '${{ secrets.E2E_SUPPORT_BUNDLE_AWS_SECRET_ACCESS_KEY }}'
 
 
-  # this job will validate that all validate-* jobs succeed and is used for the github branch protection rule
-  validate-success:
+  validate-pr-tests:
```
camilamacedo86, line 1791: We have a bug and I think it is caused by change the name of the test/check.
Could we keep the same name?

cbodonnell: I believe `validate-success` fails because the `validate-kurl-addon` step was skipped and caused `validate-pr-tests` to be skipped as well.  I've pushed another commit to change this so that `validate-kurl-addon` is a checked within `validate-success` since it may or may not run depending on if changes were made to the add-on.

camilamacedo86 (approved):
It makes sense for and shows right. 👍 /lgtm

It seems good improvement and fix. 🥇

sgalsaleh (approved):

PR #3471, opened 2022-12-02
exclude kURL specific collectors from support bundle if not running in kURL

garcialuis:
Fixes errors shown when collecting support bundles in existing clusters. Such as:
```
* cannot collect weave-status: action "get" is not allowed on resource "pods" in the "kube-system" namespace
* cannot collect weave-report: action "get" is not allowed on resource "pods" in the "kube-system" namespace
```

This PR moves collectors that are specific to kURL clusters out of the default support bundle spec so that they are only ran in kURL clusters. 

Fixes # [SC-60507](https://app.shortcut.com/replicated/story/60507/don-t-inject-kurl-specific-collectors-to-support-bundle-spec-if-not-running-in-kurl)

```release-note
Fixes errors when generating support bundle in existing clusters via the CLI. 
```

NONE

### DIFF
--- a/pkg/supportbundle/spec.go
+++ b/pkg/supportbundle/spec.go
@@ -422,22 +422,33 @@ func deduplicatedAnalyzers(supportBundle *troubleshootv1beta2.SupportBundle) *tr
 
 // addDefaultTroubleshoot adds kots.io (github.com/replicatedhq/kots/support-bundle/spec.yaml) spec to the support bundle.
 func addDefaultTroubleshoot(supportBundle *troubleshootv1beta2.SupportBundle, imageName string, pullSecret *troubleshootv1beta2.ImagePullSecrets) *troubleshootv1beta2.SupportBundle {
+	isKurl, err := kurl.IsKurl()
+	if err != nil {
+		logger.Errorf("Failed to check if cluster is kurl: %v", err)
+	}
 	next := supportBundle.DeepCopy()
-	next.Spec.Collectors = append(next.Spec.Collectors, getDefaultCollectors(imageName, pullSecret)...)
-	next.Spec.Analyzers = append(next.Spec.Analyzers, getDefaultAnalyzers()...)
+	next.Spec.Collectors = append(next.Spec.Collectors, getDefaultCollectors(imageName, pullSecret, isKurl)...)
+	next.Spec.Analyzers = append(next.Spec.Analyzers, getDefaultAnalyzers(isKurl)...)
 	return next
 }
 
-func getDefaultCollectors(imageName string, pullSecret *troubleshootv1beta2.ImagePullSecrets) []*troubleshootv1beta2.Collect {
+func getDefaultCollectors(imageName string, pullSecret *troubleshootv1beta2.ImagePullSecrets, isKurl bool) []*troubleshootv1beta2.Collect {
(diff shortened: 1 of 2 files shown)


pkg/supportbundle/defaultspec/spec.yaml
```diff
@@ -203,14 +173,6 @@ spec:
...
-        collectorName: kurl-host-preflights
-        name: kots/kurl/host-preflights
-        hostPath: /var/lib/kurl/host-preflights
-        extractArchive: true
-        image: alpine
-        imagePullPolicy: IfNotPresent
-        timeout: 1m
     - configMap:
         collectorName: kurl-current-config
         name: kurl-current-config # NOTE: this will not live under the kots/ directory like other collectors
```
cbodonnell, line 178: Could we find a way to separate out _all_ of the kurl-specific collectors and analyzers into a separate support bundle spec and load/append those only when needed?  For example, these kurl configmap colelctors.  Same would go for the kurl add-ons (weave, rook, etc).

pkg/supportbundle/defaultspec/spec.yaml
```diff
@@ -6,8 +6,6 @@ spec:
   collectors:
     - clusterInfo: {}
     - clusterResources: {}
-    - ceph: {}
-    - longhorn: {}
```
cbodonnell, line 10: Copying feedback from @divolgin here.  We likely do not want to remove stuff from the default spec because it is used when running:
```
kubectl support-bundle https://kots.io/
```

garcialuis: I added two functions to remove the kURL collectors and analyzers instead, that way the default spec can remain unchanged.

pkg/supportbundle/defaultspec/spec.yaml
garcialuis: I updated the two functions to remove those collectors/analyzers as well.

pkg/supportbundle/spec.go
```diff
@@ -422,22 +422,38 @@ func deduplicatedAnalyzers(supportBundle *troubleshootv1beta2.SupportBundle) *tr
...
 // addDefaultTroubleshoot adds kots.io (github.com/replicatedhq/kots/support-bundle/spec.yaml) spec to the support bundle.
 func addDefaultTroubleshoot(supportBundle *troubleshootv1beta2.SupportBundle, imageName string, pullSecret *troubleshootv1beta2.ImagePullSecrets) *troubleshootv1beta2.SupportBundle {
+	isKurl, err := kurl.IsKurl()
+	if err != nil {
+		logger.Errorf("Failed to check if cluster is kurl: %v", err)
+	}
 	next := supportBundle.DeepCopy()
-	next.Spec.Collectors = append(next.Spec.Collectors, getDefaultCollectors(imageName, pullSecret)...)
+	next.Spec.Collectors = append(next.Spec.Collectors, getDefaultCollectors(imageName, pullSecret, isKurl)...)
 	next.Spec.Analyzers = append(next.Spec.Analyzers, getDefaultAnalyzers()...)
```
cbodonnell, line 431: Could we also pass the `isKurl` result into this function so that we don't have to call it again?

cbodonnell (dismissed):

PR #3472, opened 2022-12-02
Remove head requests that were not necessary

marccampbell:

Remove the HEAD requests on app update checks. It doesn't return anything, and the API has the same results.

This also removed the not implemented helm, git, http upstream types.

```release-note
none
```

none


divolgin (approved):
👍

pkg/replicatedapp/api.go
```diff
@@ -165,29 +165,3 @@ func getApplicationMetadataFromHost(host string, endpoint string, upstream *url.
...
-		return errors.Wrap(err, "failed to execute head request")
-	}
-	defer headResp.Body.Close()
-
-	if headResp.StatusCode == 401 {
-		return errors.New("license was not accepted")
-	}
-
-	if headResp.StatusCode == 403 {
-		return util.ActionableError{Message: "License is expired"}
```
divolgin, line 185: `ActionableError` here is important.  The GET function should have this behavior.

marccampbell: Just confirmed it's still there.

sgalsaleh (approved):

PR #3489, opened 2022-12-09
BREAKING: Update Troubleshoot to 0.51.1

xavpaice:

Updates Troubleshoot.  Note that this update includes a change where IPv4 addresses are no longer redacted by default.  Folks that want this feature will need to adjust their application to include that as a separate redactor.

Fixes # n/a

Note the breaking change to the default redaction of IPv4 addresses in support bundles.

## Steps to reproduce

Create a support bundle.

Run preflights.

```release-note
Note that this update includes a change where IPv4 addresses are no longer redacted by default.  Folks that want this feature will need to adjust their application to include that as a separate redactor.
```

NONE


xavpaice:
FOSSA scan is failing, but the change in Troubleshoot passed it - https://github.com/replicatedhq/troubleshoot/actions/runs/3767118196/jobs/6404338562

xavpaice:
this has sat stale too long, will close and refresh with a new one

PR #3491, opened 2022-12-10
Allow required values in helm charts

divolgin:

This re-adds support for required values in helm charts (the reverted commit) and fixes airgap upload error when app has required config items

Fixes #

```release-note
* Fixes an issue that caused the license upload to fail for applications that include Helm charts with [`required`](https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-required-function) values missing from configuration.
```

NONE

### DIFF
--- a/pkg/pull/pull.go
+++ b/pkg/pull/pull.go
@@ -17,14 +17,16 @@ import (
 	"github.com/replicatedhq/kots/pkg/base"
 	"github.com/replicatedhq/kots/pkg/crypto"
 	"github.com/replicatedhq/kots/pkg/docker/registry"
-	registrytypes "github.com/replicatedhq/kots/pkg/docker/registry/types"
+	dockerregistrytypes "github.com/replicatedhq/kots/pkg/docker/registry/types"
 	"github.com/replicatedhq/kots/pkg/downstream"
 	"github.com/replicatedhq/kots/pkg/k8sdoc"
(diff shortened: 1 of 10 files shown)


pkg/pull/pull.go
```diff
@@ -293,6 +299,27 @@ func Pull(upstreamURI string, pullOptions PullOptions) (string, error) {
...
+		IsReadOnly: pullOptions.RewriteImageOptions.IsReadOnly,
+	}
+
+	needsConfig, err := kotsadmconfig.NeedsConfiguration(pullOptions.AppSlug, pullOptions.AppSequence, pullOptions.AirgapRoot != "", kotsKinds, registrySettings)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to check if version needs configuration")
+	}
+	if needsConfig {
+		return "", ErrConfigNeeded
+	}
```
cbodonnell, line 321: This causes the `Pull(...)` function to return if `needsConfig` and we do not write midstreams (which is expected since we aren't rendering the version).  Unfortunately, this seems to also result in the airgap images not getting pushed since that is part of `writeMidstream(...)`: https://github.com/replicatedhq/kots/blob/main/pkg/pull/pull.go#L579.

divolgin: That makes sense.  Nice catch.

PR #3499, opened 2022-12-14
use decrypt flag

alexanderjophus:

Plugs in the decrypt flag

Fixes #63231

## Steps to reproduce

Create a config with a password, set it to hidden

Currently it's decrypted regardless of if the flag is passed in

```release-note
When using kots get config, now decrypts only if the --decrypt flag has been passed.
```

NONE

### DIFF
--- a/cmd/kots/cli/get-config.go
+++ b/cmd/kots/cli/get-config.go
@@ -68,6 +68,7 @@ func getConfigCmd(cmd *cobra.Command, args []string) error {
 
 	appSlug := v.GetString("appslug")
 	appSequence := v.GetInt64("sequence")
+	decrypt := v.GetBool("decrypt")
 
 	localPort, errChan, err := k8sutil.PortForward(0, 3000, namespace, getPodName, false, stopCh, log)
 	if err != nil {
@@ -129,9 +130,11 @@ func getConfigCmd(cmd *cobra.Command, args []string) error {
 		return errors.Wrap(err, "failed to get config")
 	}
 
-	config.ConfigGroups, err = decryptGroups(clientset, namespace, config.ConfigGroups)
-	if err != nil {
-		return errors.Wrap(err, "failed to decrypt config")
+	if decrypt {
+		config.ConfigGroups, err = decryptGroups(clientset, namespace, config.ConfigGroups)
+		if err != nil {
+			return errors.Wrap(err, "failed to decrypt config")
+		}
 	}
 
 	values := configGroupToValues(config.ConfigGroups)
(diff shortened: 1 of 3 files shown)


cmd/kots/cli/get-config.go
```diff
@@ -209,18 +210,24 @@ func decryptGroups(clientset kubernetes.Interface, namespace string, groups []v1
...
 		for idx, item := range group.Items {
-			if item.Type == "password" {
+			if item.Type == "password" && decrypt {
 				// attempt to decrypt the password's value and default
-				decrypted, err := decryptString(item.Value.String())
-				if err == nil {
+				if item.Value.String() != "" {
+					decrypted, err := decryptString(item.Value.String())
+					if err != nil {
+						continue // don't fail if we can't decrypt
```
alexanderjophus, line 218: Not sure we want continue here. If `value` fails, we still probably want to decode `default`?

cmd/kots/cli/get-config.go
```diff
@@ -234,6 +222,21 @@ func decryptGroups(clientset kubernetes.Interface, namespace string, groups []v1
...
 }
 
+func extractString(item string) multitype.BoolOrString {
+	if item == "" {
+		return multitype.BoolOrString{}
+	}
+	decrypted, err := decryptString(item)
+	if err != nil {
+		fmt.Println(err)
+		return multitype.BoolOrString{} // don't fail if we can't decrypt
```
sgalsaleh, line 232: why not fail?

cbodonnell (approved):

cmd/kots/cli/get-config.go
```diff
@@ -129,7 +130,7 @@ func getConfigCmd(cmd *cobra.Command, args []string) error {
 		return errors.Wrap(err, "failed to get config")
 	}
 
-	config.ConfigGroups, err = decryptGroups(clientset, namespace, config.ConfigGroups)
+	config.ConfigGroups, err = decryptGroups(clientset, namespace, config.ConfigGroups, decrypt)
```
sgalsaleh, line 133: why not just skip this whole function if `decrypt` is false?

cmd/kots/cli/get-config.go
alexanderjophus: Could do, I just wasn't 100% sure what may break, currently we don't do anything with the error

sgalsaleh: Let's fail here please so that customers aren't confused if their passwords were empty

sgalsaleh: Ah I see that we weren't checking the error before, so this is fine

sgalsaleh (approved):

PR #3503, opened 2022-12-15
Expose k8s version info as template functions

sgalsaleh:

Exposes k8s version info as a template functions:

- KubernetesVersion
- KubernetesMajorVersion
- KubernetesMinorVersion

Fixes #

```release-note
Adds [KubernetesVersion](/reference/template-functions-static-context#kubernetesversion), [KubernetesMajorVersion](/reference/template-functions-static-context#kubernetesmajorversion), and [KubernetesMinorVersion](/reference/template-functions-static-context#kubernetesminorversion) template functions
```

https://github.com/replicatedhq/replicated-docs/pull/840

### DIFF
--- a/pkg/template/static_context.go
+++ b/pkg/template/static_context.go
@@ -35,6 +35,7 @@ import (
 	"gopkg.in/yaml.v3"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8sversion "k8s.io/apimachinery/pkg/version"
 	"k8s.io/client-go/kubernetes"
 	certUtil "k8s.io/client-go/util/cert"
 )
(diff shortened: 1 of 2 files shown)


pkg/template/static_context.go
```diff
@@ -584,6 +591,61 @@ func (ctx StaticCtx) yamlEscape(plain string) string {
...
+func (ctx StaticCtx) kubernetesVersion() string {
+	clientset, err := ctx.getClientset()
+	if err != nil {
+		return ""
+	}
+	sv, err := getK8sServerVersion(clientset)
+	if err != nil {
+		return ""
+	}
+	return strings.TrimPrefix(sv.GitVersion, "v")
```
cbodonnell, line 603: Any particular reason for trimming `v`?

cbodonnell (approved):

pkg/template/static_context.go
sgalsaleh: Just to make sure it works with with sprig semver comparison. Apparently, that's what we do for the KOTS version template function. It might work with the "v", but I wanted to keep things consistent.

PR #3504, opened 2022-12-16
add storybook to kots

miaawong:
This PR adds storybook to KOTS to make it easy for developers to build and maintain frontend components. 
This PR creates stories for CodeSnippet and ErrorModal as examples. 

Fixes # https://app.shortcut.com/replicated/story/63465/add-storybook-into-kots

cd into web directory and run `npm run storybook`

```release-note
NONE
```

NONE


web/babel.config.js
```diff
@@ -1,11 +1,11 @@
...
   ],
-  "plugins": [
-    "@babel/plugin-proposal-class-properties",
-    "@babel/plugin-proposal-optional-chaining"
-  ],
-}
\ No newline at end of file
+  plugins: [
+    ["@babel/plugin-proposal-class-properties", { loose: true }],
+    ["@babel/plugin-proposal-optional-chaining", { loose: true }]
```
miaawong, line 9: have to add `loose: true` or it errors out

web/src/stories/Introduction.stories.mdx
```diff
@@ -0,0 +1,219 @@
...
+import Code from "./assets/code-brackets.svg";
+import Colors from "./assets/colors.svg";
+import Comments from "./assets/comments.svg";
+import Direction from "./assets/direction.svg";
+import Flow from "./assets/flow.svg";
+import Plugin from "./assets/plugin.svg";
+import Repo from "./assets/repo.svg";
+import StackAlt from "./assets/stackalt.svg";
+
+<Meta title="Example/Introduction" />
```
miaawong, line 11: this page and the icons can honestly be deleted. lmk if you feel like it's unnecessary

web/src/stories/CodeSnippet.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+import CodeSnippet from "../components/shared/CodeSnippet";
+
+export default {
+  title: "Example/CodeSnippet",
+  component: CodeSnippet,
+} as ComponentMeta<typeof CodeSnippet>;
+
+const Template: ComponentStory<typeof CodeSnippet> = () => (
+  <CodeSnippet
+    language="bash"
```
miaawong, line 13: CodeSnippet and most likely some of the existing components will need rework so that the props will allow storybook to control.

web/babel.config.js
rianfowler: This seems fine if it passes the regression tests. We might split the babel config and use a different one for storybook later if we don't want to use loose mode.

web/src/stories/Introduction.stories.mdx
rianfowler: yeah let's get rid of it.

web/src/stories/CodeSnippet.stories.tsx
rianfowler: Depending on how much it is used we could make a new one with a different interface and swap out the old ones one at a time before deleting the old code. It's extra steps but I think it's worth it if it's really tangled up with other components.

rianfowler (dismissed):

rianfowler (approved):

PR #3509, opened 2022-12-16
Make config nav fixed / scroll independently of config form 

rianfowler:
- pins the config nav so it scrolls independently of of the config 

Fixes # https://app.shortcut.com/replicated/story/59042/pin-the-config-navigation-menu-on-the-config-page

```release-note
- Config Page: updated the config nav so that its position is fixed and doesn't scroll with the config form. 
```

### DIFF
--- a/web/src/features/AppConfig/components/AppConfig.tsx
+++ b/web/src/features/AppConfig/components/AppConfig.tsx
@@ -579,7 +579,7 @@ class AppConfig extends Component<Props, State> {
     }
 
     return (
-      <div className="flex flex-column u-padding--20 alignItems--center">
+      <div className="flex flex-column u-paddingLeft--20 u-paddingBottom--20 u-paddingRight--20 alignItems--center">
         <KotsPageTitle pageName="Config" showAppSlug />
         {fromLicenseFlow && app && (
           <Span size="18" weight="bold" mt="30" ml="38">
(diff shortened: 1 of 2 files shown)


web/src/features/AppConfig/components/AppConfig.tsx
```diff
@@ -579,7 +579,7 @@ class AppConfig extends Component<Props, State> {
     }
 
     return (
-      <div className="flex flex-column u-padding--20 alignItems--center">
+      <div className="flex flex-column u-paddingLeft--20 u-paddingBottom--20 u-paddingRight--20 alignItems--center">
```
rianfowler, line 582: remove the top padding from the inner row containing the nav and the form. This keeps the side-nav from scrolling behind the top nav

web/src/scss/components/watches/WatchConfig.scss
```diff
@@ -4,6 +4,7 @@
   width: 100%;
   max-width: 750px;
   min-width: 500px;
+  padding-top: 20px;
```
rianfowler, line 7: This pushes the config form section down so that it's vertically aligned with the nav when the page is loaded initially.

web/src/scss/components/watches/WatchConfig.scss
```diff
@@ -48,6 +49,9 @@ a:visited {
   width: 250px;
   height: fit-content;
   padding: 10px;
+  top: 60px;
+  position: sticky;
+  max-height: calc(100vh - 190px);
```
rianfowler, line 54: Since this page has a footer and header, we use the view port height minus the height of the other elements on the page to constrain the max height of the side nav's container 


i tested this in firefox and seems to work. never used `calc` before

miaawong (approved):
nice one!

PR #3525, opened 2022-12-21
fix: airgap releases with required config items

cbodonnell:

* This PR reapplies the changes from https://github.com/replicatedhq/kots/pull/3452 along with a change that fixes an airgap upload issue when there are required config values. This is the diff that is relevant to the bug fix: https://github.com/replicatedhq/kots/compare/main...ed18c6b
* Additionally, this PR refactors `WriteMidstream` to a common function in order to remove duplicate code. This commit contains the refactor: https://github.com/replicatedhq/kots/pull/3525/commits/88b5b99c2c557b56e90dc9acabb57b6e41a55317

Fixes [SC-55103](https://app.shortcut.com/replicated/story/55103/cannot-deploy-a-helm-chart-using-kots-install-when-there-are-required-fields)

Used the modified `RegistrySettings` struct as the general type for registry settings instead of having other duplicate structs elsewhere containing the same data.

#### Steps to reproduce

Install an application containing a helm chart with required values, but no defaults.  Add a config option that maps to this value, which is also required and does not have a default value.  The application should install successfully after the user provides the required configuration.

```release-note
Fixes an issue that caused the license upload to fail for applications that include Helm charts with [required](https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-required-function) values missing from configuration
```

NONE


pkg/store/kotsstore/registry_store.go
```diff
@@ -38,18 +38,17 @@ func (s *KOTSStore) GetRegistryDetailsForApp(appID string) (registrytypes.Regist
...
+		Username:   registryUsername.String,
+		Namespace:  registryNamespace.String,
+		IsReadOnly: isReadOnly.Bool,
 	}
 
 	if !registryPasswordEnc.Valid {
 		return registrySettings, nil
 	}
 
-	decodedPassword, err := base64.StdEncoding.DecodeString(registrySettings.PasswordEnc)
```
cbodonnell, line 52: `registrySettings.PasswordEnc` is not used outside of this function, so it does not need to be part of this struct.

pkg/registry/types/types.go
```diff
@@ -1,12 +1,11 @@
...
-	PasswordEnc string
-	Password    string
-	Namespace   string
-	IsReadOnly  bool
+	Hostname   string
+	Username   string
+	Password   string
+	Namespace  string
+	IsReadOnly bool
 }
```
cbodonnell, line 10: See comment: https://github.com/replicatedhq/kots/pull/3525/files#r1054788408

sgalsaleh (approved):

PR #3537, opened 2023-01-04
Miawong/sc 45692/preflight checks pass with warnings breaks out of UI 

miaawong:
#### What this PR does / why we need it: Preflight checks pass with warnings breaks out of the UI. This PR fixes this issue. 
Before: 
image
After fix: 
Screen Shot 2023-01-04 at 12 19 35 PM

Fixes # https://app.shortcut.com/replicated/story/45692/preflight-checks-pass-with-warnings-can-sometimes-break-out-of-the-ui

```release-note
Fixed an issue on the App Version History page where the row layout was broken when displaying preflight check warnings.
```

NONE

### DIFF
--- a/web/src/features/AppVersionHistory/AppVersionHistoryRow.jsx
+++ b/web/src/features/AppVersionHistory/AppVersionHistoryRow.jsx
@@ -207,7 +207,6 @@ class AppVersionHistoryRow extends Component {
     }
 
     const preflightState = this.getPreflightState(version);
-
     let checksStatusText;
     if (preflightState.preflightsFailed) {
       checksStatusText = "Checks failed";
--- a/web/src/features/AppVersionHistory/AppVersionHistoryRow.jsx
+++ b/web/src/features/AppVersionHistory/AppVersionHistoryRow.jsx
@@ -193,6 +193,12 @@ class AppVersionHistoryRow extends Component {
     const editableConfig =
       isCurrentVersion || isLatestVersion || isPendingVersion?.semver;
 
+    const showDeployLogs =
+      isPastVersion ||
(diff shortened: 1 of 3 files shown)


web/src/features/Dashboard/components/DashboardVersionCard.tsx
```diff
@@ -520,13 +521,19 @@ const DashboardVersionCard = (props: Props) => {
...
                       preflightState.preflightsFailed
                         ? "err"
                         : preflightState.preflightState === "warn"
                         ? "warning"
                         : ""
-                    }`}
+                    }
+                     ${
+                       !selectedApp && currentVersion?.status === "deploying"
+                         ? "without-btns"
```
miaawong, line 534: if no selectedApp and currentVersion is deploying, the two buttons will not be rendered. Therefore, the paragraph needs to transform a bit more.

web/src/features/AppVersionHistory/AppVersionHistoryRow.jsx
```diff
@@ -281,7 +287,12 @@ class AppVersionHistoryRow extends Component {
                               : newPreflightResults
                               ? "success"
                               : ""
-                          }`}
+                          } 
+                           ${
+                             !showDeployLogs && !showActions
+                               ? "without-btns"
```
miaawong, line 293: if showDeployLogs and showActions are false, the two buttons will not be rendered. Therefore, the paragraph needs to transform a bit more.

rianfowler (approved):
probably needs a release note like 

```
- Fixed an issue on the App Version History page where the row layout was broken when displaying preflight check warnings 
```

PR #3539, opened 2023-01-10
feat(support-bundle): add troubleshoot label for support-bundle spec secret

DexterYan:
- it adds a label `troubleshoot.io/kind=support-bundle` when  it stores/updates the Troubleshoot support-bundle spec secret

Shortcut: https://app.shortcut.com/replicated/story/65655/troubleshoot-add-label-to-support-bundle-secret

Fixes #

## Steps to reproduce
When you run it under okteto, it will generate a `kotsadm-{app_slug}-supportbundle` secret with following tags
```
  labels:
    kots.io/backup: velero
    kots.io/kotsadm: "true"
    troubleshoot.io/kind: support-bundle
```

```release-note
NONE
```

NONE

### DIFF
--- a/pkg/kotsadm/types/constants.go
+++ b/pkg/kotsadm/types/constants.go
@@ -23,9 +23,7 @@ func GetKotsadmLabels(additionalLabels ...map[string]string) map[string]string {
 	}
 
 	for _, l := range additionalLabels {
-		for k, v := range l {
-			labels[k] = v
-		}
+		labels = MergeLabels(labels, l)
 	}
 
 	return labels
@@ -34,6 +32,10 @@ func GetKotsadmLabels(additionalLabels ...map[string]string) map[string]string {
 func GetTroubleshootLabels(additionalLabels ...map[string]string) map[string]string {
 	labels := map[string]string{
 		TroubleshootKey: TroubleshootValue,
+	}
+
+	for _, l := range additionalLabels {
+		labels = MergeLabels(labels, l)
 	}
 
 	return labels
--- a/pkg/kotsadm/types/constants.go
+++ b/pkg/kotsadm/types/constants.go
@@ -32,8 +32,21 @@ func GetKotsadmLabels(additionalLabels ...map[string]string) map[string]string {
 }
 
 func GetTroubleshootLabels(additionalLabels ...map[string]string) map[string]string {
-	labels := GetKotsadmLabels(additionalLabels...)
-	labels[TroubleshootKey] = TroubleshootValue
+	labels := map[string]string{
+		TroubleshootKey: TroubleshootValue,
+	}
 
 	return labels
+}
+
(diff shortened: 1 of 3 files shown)


pkg/kotsadm/types/constants.go
```diff
@@ -27,3 +30,23 @@ func GetKotsadmLabels(additionalLabels ...map[string]string) map[string]string {
 
 	return labels
 }
+
+func GetTroubleshootLabels(additionalLabels ...map[string]string) map[string]string {
```
cbodonnell, line 34: The `additionalLabels` argument is not used.  Can we remove it?  Or apply the additional labels similar to `GetKotsadmLabels(...)`

pkg/kotsadm/types/constants.go
```diff
@@ -13,6 +13,9 @@ const ExcludeValue = "true"
 const BackupLabel = "kots.io/backup"
 const BackupLabelValue = "velero"
 
+const TroubleshootKey = "troubleshoot.io/kind"
+const TroubleshootValue = "support-bundle"
```
cbodonnell, line 17: It feels like these labels are troubleshoot-specific and could be constants imported from the troubleshoot module. If it ever changes in troubleshoot, we may also have to change it here too so that it gets discovered.

pkg/kotsadm/types/constants.go
DexterYan: Thank you @cbodonnell. Sorry, I missed that part. Pr has been changed to use merge labels function for all additional labels. Tests also included

pkg/kotsadm/types/constants.go
DexterYan: Good idea! I will try to see if I can implement it.

pkg/kotsadm/types/constants_test.go
```diff
@@ -0,0 +1,123 @@
...
+				"kots.io/backup":       "velero",
+				"troubleshoot.io/kind": "support-bundle",
+			},
+		},
+		{
+			name: "pass case with merge troubleshoot and kotadm additional labels",
+			labels: []map[string]string{
+				GetKotsadmLabels(),
+				GetTroubleshootLabels(),
+			},
```
cbodonnell, line 108: This looks the same as the above case, should we add some additional labels as args to `GetKotsadmLabels(...)`?

pkg/kotsadm/types/constants.go
cbodonnell: I think that can be in a separate PR if troubleshoot doesn't export the labels as constants.  It may be something we have to wait for a new release to implement if that is the case.

pkg/kotsadm/types/constants_test.go
DexterYan: yes, we should  add some additional labels for testing merge labels. The test has been changed

cbodonnell (approved):

PR #3543, opened 2023-01-10
Rianfowler 2023 increase slowloading threshold

rianfowler:

support issue: https://github.com/replicated-collab/schrodinger-replicated/issues/64#issuecomment-1377800987

Fixes #

```release-note
- Dashboard: increased threshold for full page loading indicator to display
```


web/src/constants/timers.js
```diff
@@ -0,0 +1 @@
+export const slowLoadingThreshold = 5000;
```
rianfowler, line 1: Customer has a /dashboard that takes over a second to load. Bumping this up to 5 seconds for now. I think long term we'll speed up the endpoint and remove the full page loader for these polling fetches.

miaawong: What would be a way to speed up the endpoint?

miaawong (approved):

web/src/constants/timers.js
rianfowler: Would take some backend work. I'm not sure what `/dashboard` is doing but we'd probably have to profile the tasks it's doing and try to tackle the biggest contributor to response time.

PR #3547, opened 2023-01-12
Miawong/sc 65991/add show all modal to storybook

miaawong:

Fixes # [sc-65991]

```release-note
NONE
```

NONE


web/src/components/snapshots/SnapshotDetails.jsx
```diff
@@ -311,8 +311,8 @@ class SnapshotDetails extends Component {
 
   renderShowAllVolumes = (volumes) => {
     return volumes.map((volume) => {
-      const diffMinutes = dayjs(volume?.finished).diff(
-        dayjs(volume?.started),
+      const diffMinutes = dayjs(volume?.finishedAt).diff(
+        dayjs(volume?.startedAt),
```
miaawong, line 315: started and finished never existed in the volumes object. 
converting more of these jsx to typescript files will help catch more of these

web/src/components/snapshots/SnapshotDetails.jsx
```diff
@@ -980,9 +980,7 @@ class SnapshotDetails extends Component {
           <ShowAllModal
             displayShowAllModal={showAllPostSnapshotScripts}
             toggleShowAllModal={this.toggleShowAllPostScripts}
-            dataToShow={this.renderShowAllPostscripts(
-              this.postSnapshotScripts()
-            )}
+            dataToShow={this.renderShowAllScripts(this.postSnapshotScripts())}
             name="Post-snapshot scripts"
```
miaawong, line 984: renderShowAllPostscripts function never existed in the code base. However, these should render similarly to pre-snapshot scripts. replacing it with the same function here.

web/src/stories/ShowAllModal.stories.tsx
```diff
@@ -0,0 +1,291 @@
...
+
+export default {
+  title: "Example/ShowAllModal",
+  component: ShowAllModal,
+} as ComponentMeta<typeof ShowAllModal>;
+
+export const Volumes = () => {
+  const renderShowAllVolumes = () => {
+    const volumes = [
+      {
```
miaawong, line 15: maybe there's a better place to store fake data?

rianfowler: Looks like this is how BPR does it: https://github.com/alan2207/bulletproof-react/blob/master/src/test/data-generators.ts

something like this could be neat for using with react-query too since it just takes a promise returning function for fetch which doesn't necessarily have to do a network call.

rianfowler (approved):

web/src/stories/ShowAllModal.stories.tsx
miaawong: great find! I'll start doing this for subsequent PRs.

PR #3548, opened 2023-01-13
feat(support-bundle): use labels to search spec secret

DexterYan:
- add function to get support bundle secret spec by troubleshoot labels
- the function is similar to function [LoadFromSecretMatchingLabel](https://github.com/replicatedhq/troubleshoot/blob/c34d80c300425e72f53b00ebfb327b33d04e50c3/pkg/specs/secrets.go#L36) in troubleshoot
- previous dependent pr https://github.com/replicatedhq/kots/pull/3539
[sc-65659](https://app.shortcut.com/replicated/story/65659/troubleshoot-call-troubleshoot-to-search-for-labels-rather-than-provide-secret-uri)

Fixes #

```release-note
Support bundles generated from the admin console now include collectors and analyzers from all discovered support bundle specs in the cluster.
```

NONE


pkg/supportbundle/supportbundle.go
```diff
@@ -403,3 +407,21 @@ func CreateSupportBundleAnalysis(appID string, archivePath string, bundle *types
 
 	return nil
 }
+
+func GetSpecSecretsMatchingLabel(client kubernetes.Interface, labelSelector string, namespace string, key string) (*corev1.Secret, error) {
```
cbodonnell, line 411: Is it possible use the function exported by troubleshoot rather than define it again in kots?

pkg/kotsadm/types/constants_test.go
```diff
@@ -123,3 +123,33 @@ func Test_mergeLabels(t *testing.T) {
...
+			},
+			expect: "foo=foo",
+		},
+		{
+			name: "pass case with troubleshoot labels",
+			labels: map[string]string{
+				"troubleshoot.io/kind": "support-bundle",
+			},
+			expect: "troubleshoot.io/kind=support-bundle",
+		},
```
cbodonnell, line 146: maybe add another case for testing multiple labels render as expected?

pkg/supportbundle/supportbundle.go
DexterYan: In current version of kots, it is using `github.com/replicatedhq/troubleshoot v0.45.0` which only can return the spec string of secret. However, in our kots following steps, it needs the labels of secret.
```
existingSecret.Data[SpecDataKey] = renderedSpec
existingSecret.ObjectMeta.Labels = labels
```
For now, without updating the kots version to the latest, we need add this function. 
I agree with you that we should put this function back to troubleshoot and imported it. I have tested in my local, it is possible and I will create a pr to add this function in latest troubleshoot.

pkg/kotsadm/types/constants_test.go
DexterYan: more tests has been added for testing multiple labels

pkg/supportbundle/spec.go
```diff
@@ -144,7 +144,7 @@ func CreateRenderedSpec(app apptypes.AppType, sequence int64, kotsKinds *kotsuti
 	renderedSpec = b.Bytes()
 
 	secretName := GetSpecSecretName(app.GetSlug())
-	existingSecret, err := clientset.CoreV1().Secrets(util.PodNamespace).Get(context.TODO(), secretName, metav1.GetOptions{})
+	existingSecret, err := GetSpecSecretsMatchingLabel(clientset, kotstypes.LabelsToStrings(kotstypes.GetTroubleshootLabels()), util.PodNamespace, SpecDataKey)
```
cbodonnell, line 147: How might this work in an environment that was running an older version of kots and then upgraded to this version.  The older version may have an existing secret, but may not have the newer troubleshoot label (`troubleshoot.io/kind=support-bundle`).  Would this function return `kuberneteserrors.IsNotFound(err) = true` and kots would try to create a new support bundle secret with the same name (and may fail because it already exists, but with different labels)?

cbodonnell: Could it work if we also use `GetKotsadmLabels()` in addition to the troubleshoot labels?  Or would that select only secrets that have _both_ instead of _either_?

cbodonnell: Here is output from a test of this.  The support bundle failed to run:

Existing secret in cluster:
```
apiVersion: v1
data:
  support-bundle-spec: ...
kind: Secret
metadata:
  creationTimestamp: "2022-07-18T16:02:48Z"
  labels:
    kots.io/backup: velero
    kots.io/kotsadm: "true"
  name: kotsadm-craig-helm-supportbundle
  namespace: cbodonnell
  resourceVersion: "346040016"
  uid: 16f3563f-c931-47e3-850b-95663a5aee0d
type: Opaque
```

Error:
```
{"level":"error","ts":"2023-01-19T15:26:09Z","msg":"could not generate support bundle dependencies: failed to create rendered support bundle spec: failed to create support bundle secret: secrets \"kotsadm-craig-helm-supportbundle\" already exists"}
```

DexterYan: Great catch! Yes, I missed that part. Now I have added AddLabelExistingSpecSecret function to add the troubleshoot labels if the old secret spec exists. It will make sure that migrate all the existing spec secrets will new troubleshoot lablel. Tested in Okteto, looks ok. Could you help to double confirm? Thx!

pkg/kotsadm/types/constants.go
```diff
@@ -52,3 +56,8 @@ func MergeLabels(labels ...map[string]string) map[string]string {
 
 	return mergedLabels
 }
+
+func LabelsToStrings(labelMap map[string]string) string {
```
sgalsaleh, line 60: this functions seems redundant to me. since it really replaces a single line? can't the caller just do `labels.SelectorFromSet(labelMap).String()`?

pkg/supportbundle/spec.go
```diff
@@ -144,7 +144,12 @@ func CreateRenderedSpec(app apptypes.AppType, sequence int64, kotsKinds *kotsuti
 	renderedSpec = b.Bytes()
 
 	secretName := GetSpecSecretName(app.GetSlug())
-	existingSecret, err := clientset.CoreV1().Secrets(util.PodNamespace).Get(context.TODO(), secretName, metav1.GetOptions{})
+	_, err = AddLabelExistingSpecSecret(clientset, util.PodNamespace, secretName, kotstypes.GetTroubleshootLabels())
+	if err != nil {
+		return nil, err
```
sgalsaleh, line 149:
```suggestion
		return nil, errors.Wrap(err, "failed to add troubleshoot labels to existing spec secret")
```

pkg/supportbundle/supportbundle.go
```diff
@@ -403,3 +408,37 @@ func CreateSupportBundleAnalysis(appID string, archivePath string, bundle *types
...
+		if !ok {
+			continue
+		}
+		return &secret, nil
+	}
+
+	return nil, kuberneteserrors.NewNotFound(corev1.Resource("secret"), "support bundle spec")
+}
+
+func AddLabelExistingSpecSecret(client kubernetes.Interface, namespace string, secretName string, labels map[string]string) (*corev1.Secret, error) {
```
sgalsaleh, line 430: this seems like a generic function that would work with any secret. what do you think about moving it to the `k8sutil` package and renaming it to something like `AddLabelsToSecret`?

pkg/supportbundle/spec.go
```diff
@@ -144,7 +144,12 @@ func CreateRenderedSpec(app apptypes.AppType, sequence int64, kotsKinds *kotsuti
 	renderedSpec = b.Bytes()
 
 	secretName := GetSpecSecretName(app.GetSlug())
-	existingSecret, err := clientset.CoreV1().Secrets(util.PodNamespace).Get(context.TODO(), secretName, metav1.GetOptions{})
+	_, err = AddLabelExistingSpecSecret(clientset, util.PodNamespace, secretName, kotstypes.GetTroubleshootLabels())
```
sgalsaleh, line 147: would you mind adding a comment about why this functions is needed? (e.g. to add the troubleshoot labels to existing secrets that were created in older versions of kots)

pkg/kotsadm/types/constants.go
DexterYan: good idea, removed

pkg/supportbundle/spec.go
DexterYan: change has been applied

pkg/supportbundle/supportbundle.go
DexterYan: I have moved that part to k8sutil as AddLabelsToSecret

pkg/supportbundle/spec.go
DexterYan: added a comment

pkg/supportbundle/spec.go
```diff
@@ -144,7 +145,13 @@ func CreateRenderedSpec(app apptypes.AppType, sequence int64, kotsKinds *kotsuti
 	renderedSpec = b.Bytes()
 
 	secretName := GetSpecSecretName(app.GetSlug())
-	existingSecret, err := clientset.CoreV1().Secrets(util.PodNamespace).Get(context.TODO(), secretName, metav1.GetOptions{})
+	// to add the troubleshoot labels to existing support-bundler spec secrets that were created in older versions of kots
```
cbodonnell, line 148: just a small typo
```suggestion
	// to add the troubleshoot labels to existing support-bundle spec secrets that were created in older versions of kots
```

cbodonnell (approved):

pkg/k8sutil/secret.go
```diff
@@ -0,0 +1,28 @@
...
+	corev1 "k8s.io/api/core/v1"
+	kuberneteserrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+func AddLabelsToSecret(client kubernetes.Interface, namespace string, secretName string, labels map[string]string) (*corev1.Secret, error) {
+	existingSecret, err := client.CoreV1().Secrets(namespace).Get(context.TODO(), secretName, metav1.GetOptions{})
+	if err != nil && !kuberneteserrors.IsNotFound(err) {
+		return nil, errors.Wrap(err, "failed to read support bundle secret")
```
sgalsaleh, line 17:
```suggestion
		return nil, errors.Wrap(err, "failed to read secret")
```

pkg/k8sutil/secret.go
```diff
@@ -0,0 +1,28 @@
...
+	} else if kuberneteserrors.IsNotFound(err) {
+		return nil, nil
+	}
+
+	existingSecret.ObjectMeta.Labels = kotstypes.MergeLabels(existingSecret.ObjectMeta.Labels, labels)
+	updatedSecret, err := client.CoreV1().Secrets(namespace).Update(context.TODO(), existingSecret, metav1.UpdateOptions{})
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to update support bundle secret")
```
sgalsaleh, line 25:
```suggestion
		return nil, errors.Wrap(err, "failed to update secret")
```

xavpaice:
Note: https://github.com/replicatedhq/kots/pull/3540 is likely relevant, once merged the ability to search labels is built into Troubleshoot

pkg/supportbundle/spec.go
DexterYan: thank you! Changed

pkg/k8sutil/secret.go
DexterYan: Thank you, I have removed words of support bundle

pkg/k8sutil/secret.go
DexterYan: removed

pkg/supportbundle/spec.go
```diff
@@ -144,7 +145,13 @@ func CreateRenderedSpec(app apptypes.AppType, sequence int64, kotsKinds *kotsuti
...
 
 	secretName := GetSpecSecretName(app.GetSlug())
-	existingSecret, err := clientset.CoreV1().Secrets(util.PodNamespace).Get(context.TODO(), secretName, metav1.GetOptions{})
+	// to add the troubleshoot labels to existing support-bundle spec secrets that were created in older versions of kots
+	_, err = k8sutil.AddLabelsToSecret(clientset, util.PodNamespace, secretName, kotstypes.GetTroubleshootLabels())
+	if err != nil {
+		errors.Wrap(err, "failed to add troubleshoot labels to existing spec secret")
+	}
+
+	existingSecret, err := GetSpecSecretsMatchingLabel(clientset, labels.SelectorFromSet(kotstypes.GetTroubleshootLabels()).String(), util.PodNamespace, SpecDataKey)
```
sgalsaleh, line 154: wouldn't the troubleshoot labels match multiple secrets at some point? KOTS creates a secret for each application version. the `GetSpecSecretsMatchingLabel` might return a different secret than the one that matches `secretName`. if the secret returned is different, then it could use the wrong spec.

chris-sanders: By design there would also be other secrets with labels that we _do_ want collected. kURL and kURL add-ons for example will install support bundles to collect information on those components.

I think if KOTS maintains multiple versions of a spec we should be labeling only the on that's considered 'active' and not 'historic' specs because we do want to collect those specs provided by other projects.

diamonwiggins: > KOTS creates a secret for each application version

I'm curious what do you mean by application version here? Are you saying each separate app install or each sequence of an app release?

sgalsaleh: @diamonwiggins each sequence of an app release. e.g. for each version in the "Version history" page

sgalsaleh: actually, looks like my info is outdated. there will be a secret per app, not app version. still a similar issue though. KOTS should be using specific labels that scope the spec to a single app

pkg/k8sutil/secret.go
```diff
@@ -0,0 +1,28 @@
...
+func AddLabelsToSecret(client kubernetes.Interface, namespace string, secretName string, labels map[string]string) (*corev1.Secret, error) {
+	existingSecret, err := client.CoreV1().Secrets(namespace).Get(context.TODO(), secretName, metav1.GetOptions{})
+	if err != nil && !kuberneteserrors.IsNotFound(err) {
+		return nil, errors.Wrap(err, "failed to read secret")
+	} else if kuberneteserrors.IsNotFound(err) {
+		return nil, nil
+	}
+
+	existingSecret.ObjectMeta.Labels = kotstypes.MergeLabels(existingSecret.ObjectMeta.Labels, labels)
+	updatedSecret, err := client.CoreV1().Secrets(namespace).Update(context.TODO(), existingSecret, metav1.UpdateOptions{})
```
banjoh, line 23: This is probably not a major issue, but I wonder whether always making this update has any implications.

pkg/supportbundle/spec.go
cbodonnell: why are we querying the support bundle spec that kots uses by the new troubleshoot label?  we previously queried by name since it is deterministic.  adding the new label to the secret makes sense so that troubleshoot itself can find it outside of kots, but I'm not sure I'm understanding why we need to query for it that way here.

banjoh: I see what you mean @cbodonnell. So basically this PR should only have the logic of adding the troubleshoot label to the secret.

Gotcha

pkg/k8sutil/secret.go
```diff
@@ -0,0 +1,28 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+func AddLabelsToSecret(client kubernetes.Interface, namespace string, secretName string, labels map[string]string) (*corev1.Secret, error) {
+	existingSecret, err := client.CoreV1().Secrets(namespace).Get(context.TODO(), secretName, metav1.GetOptions{})
+	if err != nil && !kuberneteserrors.IsNotFound(err) {
+		return nil, errors.Wrap(err, "failed to read secret")
+	} else if kuberneteserrors.IsNotFound(err) {
+		return nil, nil
```
cbodonnell, line 19: Returning no secret and no error here causes the caller to panic on `pkg/supportbundle/spec.go` line 180 ([example](https://github.com/replicatedhq/kots/actions/runs/4133866974/jobs/7144456180#step:8:460)).

pkg/supportbundle/spec.go
```diff
@@ -144,7 +144,12 @@ func CreateRenderedSpec(app apptypes.AppType, sequence int64, kotsKinds *kotsuti
 	renderedSpec = b.Bytes()
 
 	secretName := GetSpecSecretName(app.GetSlug())
-	existingSecret, err := clientset.CoreV1().Secrets(util.PodNamespace).Get(context.TODO(), secretName, metav1.GetOptions{})
+	// to add the troubleshoot labels to existing support-bundle spec secrets that were created in older versions of kots
+	existingSecret, err := k8sutil.AddLabelsToSecret(clientset, util.PodNamespace, secretName, kotstypes.GetTroubleshootLabels())
+	if err != nil {
+		errors.Wrap(err, "failed to add troubleshoot labels to existing spec secret")
+	}
+
```
cbodonnell, line 152: I think this was originally added in commit c08ae3d because we were querying for the existing secret by the troubleshoot label only.  We have logic just below this that adds the labels to the secrets (on create or update).  It was added in https://github.com/replicatedhq/kots/pull/3539.

I might be misunderstanding the purpose of this PR, but would this mean than we don't need _any_ of these changes?  Let me know what you think.

banjoh:
To ensure we filter out specs from other apps in a multi-app environment, we need to do the following

1. Query all secrets with `troubleshoot.io/kind=support-bundle` label
2. In code, only merge specs that for the current app **and** any cluster-wide secrets in the returned result

pkg/supportbundle/spec.go
```diff
@@ -251,6 +264,120 @@ func injectDefaults(app apptypes.AppType, b *troubleshootv1beta2.SupportBundle,
...
 }
 
+func addDiscoveredSpecs(
+	supportBundle *troubleshootv1beta2.SupportBundle, app apptypes.AppType,
+) *troubleshootv1beta2.SupportBundle {
+	clientset, err := k8sutil.GetClientset()
+	if err != nil {
+		logger.Errorf("Failed to get kubernetes clientset: %v", err)
+		return supportBundle
+	}
```
cbodonnell, line 274: Can we pass a k8s client to this function instead of instantiating a new one?  The caller already has one, so it should be a small change.  That way we can unit test this function too.

pkg/supportbundle/spec.go
```diff
@@ -251,6 +264,120 @@ func injectDefaults(app apptypes.AppType, b *troubleshootv1beta2.SupportBundle,
...
+	labelSelector := kotstypes.TroubleshootKey + "=" + kotstypes.TroubleshootValue
+	ctx := context.TODO()
+
+	specs := []string{}
+
+	// Get all namespaces
+	nsList, err := client.CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
```
cbodonnell, line 312: Could listing all namespaces fail in a namespace-scoped installation?  Found some existing logic for gathering namespaces here that seems to take that into account: https://github.com/replicatedhq/kots/blob/main/pkg/helm/cache.go#L46-L57

pkg/supportbundle/spec.go
```diff
@@ -251,6 +264,120 @@ func injectDefaults(app apptypes.AppType, b *troubleshootv1beta2.SupportBundle,
...
+
+	// List objects from one namespace at a time so as to isolate errors e.g RBAC
+	// Search secrets
+	for _, ns := range nsList.Items {
+		secrets, err := client.CoreV1().Secrets(ns.Name).List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
+		if err != nil {
+			if kuberneteserrors.IsForbidden(err) || kuberneteserrors.IsUnauthorized(err) {
```
cbodonnell, line 319: Might not need this RBAC check if the above is implemented since kots _should_ be able to list secrets at the cluster level or namespace level (if namespace-scoped). Ref: https://github.com/replicatedhq/kots/blob/main/pkg/kotsadm/objects/kotsadm_objects.go#L22-L65

pkg/supportbundle/spec.go
```diff
@@ -251,6 +264,120 @@ func injectDefaults(app apptypes.AppType, b *troubleshootv1beta2.SupportBundle,
...
+
+			specs = append(specs, string(specData))
+		}
+	}
+
+	// Search config maps
+	for _, ns := range nsList.Items {
+		configmaps, err := client.CoreV1().ConfigMaps(ns.Name).List(context.TODO(), metav1.ListOptions{LabelSelector: labelSelector})
+		if err != nil {
+			if kuberneteserrors.IsForbidden(err) || kuberneteserrors.IsUnauthorized(err) {
```
cbodonnell, line 350: Same as above

pkg/supportbundle/spec_test.go
```diff
@@ -340,3 +348,135 @@ func Test_deduplicatedAnalyzers(t *testing.T) {
...
+			createNamespaces(t, clientset, "kotsadm", "default", "another")
+			got, err := findSupportBundleSpecs(clientset)
+			assert.Equal(t, tt.wantErr, err != nil, "findSupportBundleSecrets() error %v, wantErr %v", err, tt.wantErr)
+			require.NotNil(t, got)
+
+			t.Log(got)
+			specs := []string{}
+			for _, spec := range got {
+				specs = append(specs, spec)
+			}
```
cbodonnell, line 464: This looks like it may have been for debugging.  Is it needed anymore?

cbodonnell (commented):
@banjoh do you know if any of the kURL components install support bundle specs currently?  I'm thinking that we might (or might not) need to update some regression tests if they do since those would be discovered by kots now.

banjoh:
> 

[flannel CNI kurl addon](https://github.com/replicatedhq/kURL/blob/0d341f18f4d5ae6365c031082ea1aa90270c2a7e/addons/flannel/0.21.1/yaml/troubleshoot.yaml) currently is installing a support bundle as a secret. This will be discovered and run once this change is released.

pkg/supportbundle/spec.go
banjoh: Good point. Addressing this now

pkg/supportbundle/spec.go
```diff
@@ -251,6 +264,120 @@ func injectDefaults(app apptypes.AppType, b *troubleshootv1beta2.SupportBundle,
...
+	}
+
+	for _, specData := range specs {
+		sbObject, err := sb.ParseSupportBundleFromDoc([]byte(specData))
+		if err != nil {
+			logger.Errorf("Failed to unmarshal support bundle spec: %v", err)
+			continue
+		}
+
+		// TODO: How about other spec fields e.g URI?
```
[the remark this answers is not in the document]
cbodonnell: Could come in to play with future troubleshoot updates

pkg/supportbundle/spec.go
cbodonnell: I think we should just log an error here in call cases.  It would fail silently if it was a permissions issue without debug logging enabled.  Kots generally should be able to list them in the namespaces it has access to.  If if fails for any reason, we can see the error in the support bundle and understand why.

banjoh: Make sense. Updating that now

cbodonnell (approved):

PR #3549, opened 2023-01-13
read remaining bytes in pkcs12 certificates

alexanderjophus:

Currently we throw away the remaining bytes when reading in certificates, if our certificate is a pkcs12 and the private key is first, we throw away the remaining bytes (which includes the certificate)

Fixes # 61769

## Steps to reproduce
See [SC](https://app.shortcut.com/replicated/story/61769/uploading-pkcs12-certificate-bundle-that-has-a-private-key-in-the-beginning-of-the-file-to-admin-console-fails-silently)

```release-note
Fixes bug where a pkcs12 certification that had the private key listed first would throw an error when uploading it.
```

NONE


pavansokkenagaraj (dismissed):
LGTM! Please get one more review who have actively contributed to `kurl-proxy`

kurl_proxy/cmd/main.go
```diff
@@ -206,10 +206,20 @@ func watchSecret(certs chan cert, name string, secrets corev1.SecretInterface) {
...
+	for {
+		derBlock, certData = pem.Decode(certData)
+		if derBlock == nil {
+			return "", errors.New("no PEM data found in certificate")
+		}
+		if derBlock.Type == "CERTIFICATE" {
+			break
+		}
+		if len(certData) == 0 {
+			return "", errors.New("no PEM data found in certificate")
```
cbodonnell, line 219: nit pick: this wouldn't mean there is no PEM data found, but just none of type `"CERTIFICATE"`? might be useful to change this error message so we know which code path failed. it seems unlikely that it would ever reach this case since we check for a valid cert/key pair before calling `getFingerprint`, but just a thought

cbodonnell (dismissed):

cbodonnell (approved):

PR #3550, opened 2023-01-14
Miawong/sc 57763/add ability to delete support bundles from admin console

miaawong:
#### What this PR does / why we need it: This PR is the frontend implementation that allows users to delete support bundle from admin console. 

https://user-images.githubusercontent.com/28071398/212441098-9471fadd-88a4-4be7-bf03-ef4570af72a8.mov

Fixes # [sc-57763] 

Go to troubleshoot and create a support bundle. 
Delete it

```release-note
Adds the ability to delete support bundles. 
```

None

### DIFF
--- a/web/src/Root.tsx
+++ b/web/src/Root.tsx
@@ -50,7 +50,7 @@ let history = connectHistory(browserHistory);
 const ProtectedRoute = ({
   render,
   location,
-  path
+  path,
 }: {
   render: (props: object) => React.ReactNode;
 } & RouteProps) => {
@@ -74,7 +74,7 @@ const ThemeContext = React.createContext({
     console.log("setThemeState used before being set", themeState);
   },
   getThemeState: (): ThemeState => ({ navbarLogo: null }),
-  clearThemeState: () => {}
+  clearThemeState: () => {},
 });
 
 type AppBranding = {
@@ -107,7 +107,7 @@ const Root = () => {
   const [state, setState] = useReducer(
     (currentState: State, newState: Partial<State>) => ({
       ...currentState,
-      ...newState
+      ...newState,
     }),
     {
       appBranding: null,
@@ -127,8 +127,8 @@ const Root = () => {
       selectedAppName: null,
       snapshotInProgressApps: [],
       themeState: {
-        navbarLogo: null
-      }
+        navbarLogo: null,
+      },
     }
   );
 
@@ -140,7 +140,7 @@ const Root = () => {
   const setThemeState = (newThemeState?: ThemeState) => {
     if (newThemeState) {
       setState({
-        themeState: { ...newThemeState }
+        themeState: { ...newThemeState },
       });
     }
   };
@@ -161,11 +161,11 @@ const Root = () => {
      * Reference object to a blank theme state
      */
     const EMPTY_THEME_STATE = {
-      navbarLogo: null
+      navbarLogo: null,
     };
 
     setState({
-      themeState: { ...EMPTY_THEME_STATE }
+      themeState: { ...EMPTY_THEME_STATE },
     });
   };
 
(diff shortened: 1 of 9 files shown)


web/src/Root.tsx
```diff
@@ -448,208 +449,212 @@ const Root = () => {
...
+                  )}
+                />
+                <Route component={NotFound} />
+              </Switch>
+            </div>
+            <div className="flex-auto Footer-wrapper u-width--full">
+              <Footer appsList={state.appsList} />
+            </div>
+          </Router>
+        </ToastProvider>
```
miaawong, line 657: just added the ToastProvider so the entire app can use it

web/src/components/troubleshoot/SupportBundleList.tsx
```diff
@@ -52,58 +54,42 @@ type State = {
   isGeneratingBundleOpen: boolean;
 };
 
-class SupportBundleList extends React.Component<Props, State> {
-  constructor(props: Props) {
-    super(props);
-    this.state = {
+export const SupportBundleList = (props: Props) => {
```
miaawong, line 57: convert SupportBundleList to functional component so it can use hooks

web/src/components/troubleshoot/SupportBundleRow.tsx
```diff
@@ -39,40 +42,77 @@ type State = {
   sendingBundle: boolean;
   sendingBundleErrMsg?: string;
   warningInsights?: SupportBundleInsight[];
+  timeoutId?: ReturnType<typeof setTimeout>;
 };
 
-class SupportBundleRow extends React.Component<Props, State> {
-  constructor(props: Props) {
-    super(props);
+export const SupportBundleRow = (props: Props) => {
```
miaawong, line 48: convert into functional component

web/src/scss/components/troubleshoot/SupportBundleList.scss
```diff
@@ -86,6 +86,16 @@
     }
   }
 }
+.deleting {
+  animation: blinker 3s linear infinite;
```
miaawong, line 90: adds blinking animation to the bundle that's being deleted

web/src/scss/utilities/base.scss
```diff
@@ -512,3 +512,49 @@ and `background-position: [value]`
   color: $text-color-primary;
   line-height: 1.7;
 }
+
+@mixin keyframes($animation-name) {
```
miaawong, line 516: adds animation to the toast when it pops up and leaves

miaawong (commented):
looks like a lot but it's mostly just convert SupportBundleList and SupportBundleRow into a functional components so we can use hooks.

web/src/components/shared/Toast.tsx
```diff
@@ -0,0 +1,42 @@
...
+        return "tw-bg-[#38cc97]";
+      case "error":
+        return "tw-bg-[#f65c5c]";
+      case "warning":
+        return "tw-bg-[#FFA500]";
+      default:
+        return "tw-bg-[#38cc97]";
+    }
+  };
+
```
rianfowler, line 22:
```suggestion
const toastType = () => {
  switch (type) {
    case "success":
      return "tw-bg-[#38cc97]";
    case "error":
      return "tw-bg-[#f65c5c]";
    case "warning":
      return "tw-bg-[#FFA500]";
    default:
      return "tw-bg-[#38cc97]";
  }
};
const Toast = ({ isToastVisible, type, children }: ToastProps) => {
```

this should define the function when the module is loaded rather than every time this is rendered. only a suggestion since it probably has no real user impact

web/src/scss/components/troubleshoot/SupportBundleList.scss
rianfowler: this is neat

rianfowler (dismissed):

web/src/components/shared/Toast.tsx
miaawong: @rianfowler good point! I'll update it. thanks for pointing it out!

rianfowler (approved):

PR #3551, opened 2023-01-16
use fork of logfmt

alexanderjophus:

Uses a fork of logfmt to allow reading more than 64kb of data in a single log line when parsing a backup

Fixes # 62367

I think there's an underlying issue that we're unable to see as we fail parsing the whole file. 

## Steps to reproduce

Create a backup, ensure that a log line has more than 64kb (or change the code based off this pr to artificially reduce the limit), get the backup logs.

A failure will be logged out, but the request will continue to succeed (note it will not have the full logs, and iirc it won't display any logs)

```release-note
Fixes issue where backup log lines more than 64kb caused the`/api/v1/snapshot/{snapshotName}` route to fail. New limit is 1mb
```

None


cbodonnell:
closed as the upstream PR with the changes was merged, so those will be used: https://github.com/replicatedhq/kots/pull/3603

PR #3572, opened 2023-01-25
Update k3s version in e2e tests to avoid the port forward bug

ousbots:

Update the k3s versions used in the end to end tests and add the `--egress-selector-mode disabled` argument input when the version starts with "v1.22". In the current versions, there is a bug in the k3s control plane when using web sockets connections between the pods (default networking option) that results in the kots cli failing to create a port forward.

Fixes [SC-67356](https://app.shortcut.com/replicated/story/67356/kots-cli-port-forwarding-fails-intermittently)

This does not seem to have affected the v1.21 branch, but I updated it to the latest version anyways.

The bug in the k3s control plane was fixed by [this k3s commit](https://github.com/k3s-io/k3s/pull/6216). I'm pretty sure that's the right commit based on the code, but I didn't bisect and test.

Reasoning for why this is the fix:

The problem seems to have been the k3s control plane incorrectly handling web sockets connections of pods under its control. It took a while to be able to consistently reproduce this issue, but digging through the kots cli and admin code, it was clearly between the api machinery (client side port forward) and the k3s agent (server side tunnel) as the error was created during a TCP dial that normally worked fine.

Next by playing around with the delays, reordering, and dropping packets with tc and watching traffic with tshark, this EOF error could be somewhat rarely triggered and when it happened there were packets sent with bogus sequence number, port re-use, and strangely ordered seemingly unrelated SYN/ACK and LAST_ACK sequences. So from this it seemed likely that the connection management in the control plane was buggy. Testing this by running several loops of kots commands that opened a port forward on a k3s cluster confirmed this by making this issue consistently reproducible with 3-5 kots command loops running in parallel.

The bug fix was back ported to the v1.23, v1.24, and v1.25 branches and this can be confirmed by running the reproduction steps below using these versions. It does not seem to exist in the v1.21 branches and the fix was not back ported to the v1.22 branch, so the action-k3s GitHub action was updated to allow passing command line arguments to k3s and the egress selector is disabled if the version string starts with "v1.22".

This can also be fixed by disabling the web sockets connections between pods and instead using direction connections by passing the k3s server the `--egress-selector-mode disabled` option on startup, which avoids the control plane managing web sockets connections.

## Steps to reproduce
Note: Make sure the gcloud instance name and options are to your preferences in the instructions below (i.e. change repldev-tim to your preferred name, fix the zone and any options in the gcloud commands, etc).

Reproduce:
* Create a VM on gcloud: 
```
gcloud computer instances create repldev-tim --project=replicated-dev-environment --machine-type=n1-standard-v8 --boot-disk-size=200GB --boot-disk-type=pd-ssd --subnet=default --network-tier=PREMIUM --image=ubuntu-2204-jammy-v20230114 --image-project=replicated-dev-environment
```
* SSH to the new VM:
```
ssh $(gcloud compute instances describe repldev-timo --format='get(networkInterfaces[0].accessConfigs[0].natIP)'
```
* Install containerd and tools:
```
sudo mkdir -p /etc/apt/keyrings
sudo curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
sudo curl -fsSLo /etc/apt/keyrings/kubernetes-archive-keyring.gpg https://packages.cloud.google.com/apt/doc/apt-key.gpg
sudo echo \
  "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.gpg] https://download.docker.com/linux/ubuntu \
  $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
sudo apt update
sudo apt install -y docker-ce docker-ce-cli containerd.io docker-compose-plugin kubectl
curl -L https://kots.io/install | bash
``` 
* Start a buggy K3s version: 
```
sudo docker run -d --privileged --name="k3s" -e K3S_KUBECONFIG_OUTPUT="/tmp/output/k3sconfig" -e K3S_KUBECONFIG_MODE=666 -v /tmp/output:/tmp/output -p 6443:6443 -i rancher/k3s:v1.24.4-k3s1 server
```
* Get a test license for a test app (from kots-sentry, your own vandoor deployment, vendor.staging.replicated.com, etc)
* Copy the kube config file:
```
mkdir ~/.kube; cp /tmp/output/k3sconfig ~/.kube/config
```
* Install the test app using the license from the previous step:
```
kubectl kots install -n test --license-file test-license.yaml --skip-preflights test
```
* Open 5+ terminals, ssh into the VM with K3s, and run kubectl commands in a loop that open port forwarding:
```
for i in $(seq 1000); do kubectl kots set config -n test test example_default_value="test1 ${i}"; done
```
* Observe the EOF port forwarding error

Confirm fix:
* Stop and delete the K3s container:
```
sudo docker stop k3s
sudo docker rm k3s
```
* Run the latest version, or one with the fix back ported, of k3s:
```
sudo docker run -d --privileged --name="k3s" -e K3S_KUBECONFIG_OUTPUT="/tmp/output/k3sconfig" -e K3S_KUBECONFIG_MODE=666 -v /tmp/output:/tmp/output -p 6443:6443 -i rancher/k3s:latest server
```
* Or run k3s without the websockets connection management:
```
sudo docker run -d --privileged --name="k3s" -e K3S_KUBECONFIG_OUTPUT="/tmp/output/k3sconfig" -e K3S_KUBECONFIG_MODE=666 -v /tmp/output:/tmp/output -p 6443:6443 -i rancher/k3s:v1.24.4-k3s1 server --egress-selector-mode disabled
```
* Repeat the reproduction steps to run a bunch of kots commands.

```release-note
NONE
```

NONE

### DIFF
--- a/.github/actions/kots-e2e/action.yml
+++ b/.github/actions/kots-e2e/action.yml
@@ -10,9 +10,9 @@ inputs:
   k8s-version:
     description: 'K8s version'
     required: true
-  egress-fix:
-    description: 'Disable k3s egress selector'
-    required: true
+  arguments:
+    description: 'arguments to pass to k8s'
+    required: false
   testim-access-token:
     description: 'Testim access token'
     required: true
@@ -59,7 +59,7 @@ runs:
     - uses: replicatedhq/action-k3s@main
       with:
         version: ${{ inputs.k8s-version }}
-        egress-fix: ${{ inputs.egress-fix }}
+        arguments: ${{ inputs.arguments }}
 
     - name: execute suite "${{ inputs.test-focus }}"
       env:
--- a/.github/actions/kots-e2e/action.yml
+++ b/.github/actions/kots-e2e/action.yml
@@ -10,6 +10,9 @@ inputs:
   k8s-version:
     description: 'K8s version'
     required: true
+  egress-fix:
+    description: 'Disable k3s egress selector'
+    required: true
   testim-access-token:
     description: 'Testim access token'
     required: true
@@ -56,6 +59,7 @@ runs:
     - uses: replicatedhq/action-k3s@main
       with:
         version: ${{ inputs.k8s-version }}
+        egress-fix: ${{ inputs.egress-fix }}
 
     - name: execute suite "${{ inputs.test-focus }}"
       env:
--- a/.github/actions/kots-e2e/action.yml
+++ b/.github/actions/kots-e2e/action.yml
@@ -10,7 +10,7 @@ inputs:
   k8s-version:
     description: 'K8s version'
     required: true
-  arguments:
+  k8s-arguments:
     description: 'arguments to pass to k8s'
     required: false
   testim-access-token:
@@ -59,7 +59,7 @@ runs:
     - uses: replicatedhq/action-k3s@main
       with:
         version: ${{ inputs.k8s-version }}
-        arguments: ${{ inputs.arguments }}
+        arguments: ${{ inputs.k8s-arguments }}
 
     - name: execute suite "${{ inputs.test-focus }}"
       env:
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -431,6 +431,7 @@ jobs:
         with:
           test-focus: 'Regression'
           k8s-version: 'v1.24.9-k3s2'
+          egress-fix: false
           testim-branch: ${{ github.head_ref == 'main' && 'master' || github.head_ref }}
           testim-access-token: '${{ secrets.TESTIM_ACCESS_TOKEN }}'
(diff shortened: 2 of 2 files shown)


cbodonnell (dismissed):

sgalsaleh (approved):

.github/actions/kots-e2e/action.yml
```diff
@@ -10,6 +10,9 @@ inputs:
   k8s-version:
     description: 'K8s version'
     required: true
+  arguments:
+    description: 'arguments to pass to k8s'
+    required: false
```
cbodonnell, line 15: _Really_ minor, but `arguments` is pretty generic since this isn't a k3s action. Could change this to something like `k8s-arguments` for this one? Might make it easier to know what it is when looking at the caller.

cbodonnell (approved):

cbodonnell (approved):

PR #3586, opened 2023-01-30
Miawong/sc 67312/can t generate new support bundle after deleting

miaawong:
#### What this PR does / why we need it: This PR changes the toast message to be 7 seconds. 
Fixes an issue where if a user deletes a support bundle in progress, user can click "generate new support bundle" link. 
Fixes an issue where if a user undo a support bundle in progress, it will continue the progress. 

Fixes # [sc-67312]

```release-note
Fixes an issue where if a user deletes a pending support bundle, user can generate a new support bundle. 
```

NONE

### DIFF
--- a/web/src/components/troubleshoot/SupportBundleList.tsx
+++ b/web/src/components/troubleshoot/SupportBundleList.tsx
@@ -158,19 +158,28 @@ export const SupportBundleList = (props: Props) => {
   }, [props.bundle]);
 
   const prevLoadingBundleId = usePrevious(props.loadingBundleId);
+  const prevDeleteBundleId = usePrevious(deleteBundleId);
 
   useEffect(() => {
+    // if the current bundle to delete is the same as the bundle that is loading
+    // stop the polling
     if (props.loadingBundleId === deleteBundleId) {
       state.pollForBundleAnalysisProgress.stop();
       props.updateState({ loadingBundleId: "", loadingBundle: false });
     }
-    console.log(prevLoadingBundleId, "prev");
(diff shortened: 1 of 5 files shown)


web/src/components/troubleshoot/SupportBundleRow.tsx
```diff
@@ -169,37 +172,48 @@ export const SupportBundleRow = (props: Props) => {
...
-      )
-        .then(async () => {
+      );
+      if (res.ok) {
+        await props.deleteBundleFromList(bundle.id);
+        setIsToastVisible(false);
+        clearInterval(id);
+      } else {
+        console.log(res);
+        setToastMessage("Unable to delete bundle, please try again.");
```
miaawong, line 208: add error toast

rianfowler (approved):

web/src/components/troubleshoot/SupportBundleList.tsx
```diff
@@ -71,12 +64,22 @@ export const SupportBundleList = (props: Props) => {
...
     isToastVisible,
     toastMessage,
-    setIsCancelled,
+    toastType,
+    setIsToastVisible,
+    toastChild,
   } = useContext(ToastContext);
 
+  // rework this so full page refresh is not needed.
+  // const deleteBundleFromList = (deleteId: string) => {
```
miaawong, line 75: Page has to be refetched rather than filtered.
If we filter, it causes a test to fail when installing with CLI. I will work on this later on.

rianfowler (approved):

PR #3593, opened 2023-01-31
improve velero installation experience for existing clusters

sgalsaleh:

- Supports installing Velero without a backup storage location.
- Shows simpler instructions for configuring NFS or host path when Velero is not installed (both in the UI and CLI)

Image 2023-01-27 at 3 15 37 PM

Image 2023-01-31 at 11 44 21 AM

Fixes #

```release-note
TBD
```

NONE


rianfowler (dismissed):
FE looks good.

divolgin (approved):

rianfowler (approved):

pkg/snapshot/store.go
```diff
@@ -158,7 +149,15 @@ func ConfigureStore(ctx context.Context, options ConfigureStoreOptions) (*types.
 	return updatedStore, nil
 }
 
-func updateExistingStore(ctx context.Context, clientset kubernetes.Interface, store *types.Store, options ConfigureStoreOptions) (*types.Store, bool, error) {
+func buildNewStore(ctx context.Context, clientset kubernetes.Interface, options ConfigureStoreOptions) (*types.Store, bool, error) {
+	store, err := GetGlobalStore(ctx, options.KotsadmNamespace, nil)
```
cbodonnell, line 153: `GetGlobalStore` needs a k8s client, so the unit tests may fail unless we make it accept one and pass it through here.  Tests will likely need to be updated too if we go that route.

pkg/k8sutil/util.go
```diff
@@ -0,0 +1,97 @@
...
+			mergedVolumeMounts = append(mergedVolumeMounts, *desiredVolumeMount.DeepCopy())
+		} else if override {
+			// found and should override
+			mergedVolumeMounts[idx] = *desiredVolumeMount.DeepCopy()
+		}
+	}
+	return mergedVolumeMounts
+}
+
+func MergeImagePullSecrets(desired []corev1.LocalObjectReference, existing []corev1.LocalObjectReference, override bool) []corev1.LocalObjectReference {
```
sgalsaleh, line 76: override doesn't do much today since there's only a `Name` field in the `LocalObjectReference`, but it'll become helpful when more fields are added

cbodonnell (approved):

PR #3594, opened 2023-01-31
use currentVersion iconURI to display app icons

pavansokkenagaraj:
Uses app's currentVersion icon URI to display app icons
- when a new version is deployed, deployed version's application.yaml is saved in kotsadm-application-metadata configmap

Fixes #
[SC-57832](https://app.shortcut.com/replicated/story/57832).

```release-note
fix a bug where app icon was shown from latest version instead of current version
```

NONE

### DIFF
--- a/pkg/api/downstream/types/types.go
+++ b/pkg/api/downstream/types/types.go
@@ -54,6 +54,8 @@ type DownstreamVersion struct {
 	NeedsKotsUpgrade           bool                            `json:"needsKotsUpgrade,omitempty"`
 	KOTSKinds                  *kotsutil.KotsKinds             `json:"-"`
 	DownloadStatus             DownloadStatus                  `json:"downloadStatus,omitempty"`
+	Title                      string                          `json:"title,omitempty"`
+	IconURI                    string                          `json:"iconUri"`
 }
 
 type DownloadStatus struct {
--- a/pkg/api/downstream/types/types.go
+++ b/pkg/api/downstream/types/types.go
@@ -54,8 +54,8 @@ type DownstreamVersion struct {
 	NeedsKotsUpgrade           bool                            `json:"needsKotsUpgrade,omitempty"`
 	KOTSKinds                  *kotsutil.KotsKinds             `json:"-"`
 	DownloadStatus             DownloadStatus                  `json:"downloadStatus,omitempty"`
-	Title                      string                          `json:"title,omitempty"`
-	IconURI                    string                          `json:"iconUri"`
+	AppTitle                   string                          `json:"appTitle,omitempty"`
+	AppIconURI                 string                          `json:"appIconUri"`
 }
 
 type DownloadStatus struct {
--- a/pkg/api/downstream/types/types.go
+++ b/pkg/api/downstream/types/types.go
@@ -55,7 +55,7 @@ type DownstreamVersion struct {
 	KOTSKinds                  *kotsutil.KotsKinds             `json:"-"`
 	DownloadStatus             DownloadStatus                  `json:"downloadStatus,omitempty"`
 	AppTitle                   string                          `json:"appTitle,omitempty"`
-	AppIconURI                 string                          `json:"appIconUri"`
+	AppIconURI                 string                          `json:"appIconUri,omitempty"`
 }
 
 type DownloadStatus struct {
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -31,7 +31,9 @@ import (
 	"github.com/replicatedhq/kots/pkg/midstream"
 	"github.com/replicatedhq/kots/pkg/operator/client"
 	operatortypes "github.com/replicatedhq/kots/pkg/operator/types"
+	registrytypes "github.com/replicatedhq/kots/pkg/registry/types"
 	"github.com/replicatedhq/kots/pkg/render"
+	rendertypes "github.com/replicatedhq/kots/pkg/render/types"
 	"github.com/replicatedhq/kots/pkg/reporting"
 	"github.com/replicatedhq/kots/pkg/store"
 	storetypes "github.com/replicatedhq/kots/pkg/store/types"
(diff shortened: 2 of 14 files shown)


garcialuis (commented):
I think the same change needs to be applied to the GitOps configuration page:
https://github.com/replicatedhq/kots/blob/main/web/src/features/Gitops/utils.js#L11
It's still showing the latest version's icon here.

garcialuis (approved):
LGTM

pkg/handlers/metadata.go
```diff
@@ -104,7 +104,20 @@ func GetMetadataHandler(getK8sInfoFn MetadataK8sFn, kotsStore store.Store) http.
 			return
 		}
 		application := obj.(*kotsv1beta1.Application)
-		metadataResponse.IconURI = application.Spec.Icon
+
+		appIconURI, err := kotsStore.GetLatestAppIconURI(appID)
```
sgalsaleh, line 108: if i'm understanding things correctly, this gets the app icon for the currently deployed version, not the latest one. also, i think the better approach to just updated the branding configmap when a version gets deployed, then there's no changes needed to this handler.

pkg/api/downstream/types/types.go
```diff
@@ -54,6 +54,7 @@ type DownstreamVersion struct {
 	NeedsKotsUpgrade           bool                            `json:"needsKotsUpgrade,omitempty"`
 	KOTSKinds                  *kotsutil.KotsKinds             `json:"-"`
 	DownloadStatus             DownloadStatus                  `json:"downloadStatus,omitempty"`
+	IconURI                    string                          `json:"iconUri"`
```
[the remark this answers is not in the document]
pavansokkenagaraj: KOTSKinds *kotsutil.KotsKinds `json:"-"`
We cannot use `KOTSKinds` as it is not exposed to API, the json tag is `"-"`

pkg/handlers/metadata.go
pavansokkenagaraj: >the branding configmap
there is no branding configmap.
we have `kotsadm-application-metadata` CM which holds `application.yaml` which always points to the latest downloaded version CM[which we cannot use]

pkg/api/downstream/types/types.go
```diff
@@ -54,6 +54,8 @@ type DownstreamVersion struct {
 	NeedsKotsUpgrade           bool                            `json:"needsKotsUpgrade,omitempty"`
 	KOTSKinds                  *kotsutil.KotsKinds             `json:"-"`
 	DownloadStatus             DownloadStatus                  `json:"downloadStatus,omitempty"`
+	Title                      string                          `json:"title,omitempty"`
```
sgalsaleh, line 57:
```suggestion
	AppTitle                      string                          `json:"title,omitempty"`
```

pkg/api/downstream/types/types.go
```diff
@@ -54,6 +54,8 @@ type DownstreamVersion struct {
...
+	IconURI                    string                          `json:"iconUri"`
```
sgalsaleh, line 58:
```suggestion
	AppIconURI                    string                          `json:"appIconUri"`
```

pkg/operator/operator.go
```diff
@@ -365,6 +370,10 @@ func (o *Operator) DeployApp(appID string, sequence int64) (deployed bool, deplo
 		return false, errors.Wrap(err, "failed to deploy app")
 	}
 
+	if err := o.ensureKotsadmApplicationMetadataConfigMap(util.PodNamespace, kotsKinds, app.UpstreamURI); err != nil {
```
sgalsaleh, line 373: since we should show branding for the version even if deploying the app fails, how about we move this to the defer statement at the top and execute this regardless of deployment status?

pkg/operator/operator.go
```diff
@@ -191,6 +191,10 @@ func (o *Operator) DeployApp(appID string, sequence int64) (deployed bool, deplo
 		if err != nil {
 			logger.Error(errors.Wrap(err, "failed to update downstream status"))
 		}
+
+		if err := o.ensureKotsadmApplicationMetadataConfigMap(appID, util.PodNamespace); err != nil {
```
sgalsaleh, line 195: this should be at the beginning if the defer, because there are return statements above and it might not get executed

pkg/operator/operator.go
```diff
@@ -770,20 +770,15 @@ func deduplicateSecrets(secretSpecs []string) []string {
...
 }
 
-func (o *Operator) ensureKotsadmApplicationMetadataConfigMap(upstreamURI string, kotsKinds *kotsutil.KotsKinds, namespace string) error {
-	kotsAppSpec, err := kotsKinds.Marshal("kots.io", "v1beta1", "Application")
-	if err != nil {
-		return errors.Wrap(err, "failed to marshal kots app spec")
-	}
-
+func (o *Operator) ensureKotsadmApplicationMetadataConfigMap(app *apptypes.App, sequence int64, namespace string, kotsKinds *kotsutil.KotsKinds, registrySettings registrytypes.RegistrySettings) error {
+	renderedKotsAppSpec, err := o.renderKotsApplicationSpec(app, sequence, namespace, kotsKinds, registrySettings, &render.Renderer{})
```
sgalsaleh, line 774: should check the error here. my preference is to have the render logic happen inside this function since it's not really an operator related function, but i don't have a strong opinion as it is now

pkg/api/downstream/types/types.go
```diff
@@ -54,6 +54,8 @@ type DownstreamVersion struct {
 	NeedsKotsUpgrade           bool                            `json:"needsKotsUpgrade,omitempty"`
 	KOTSKinds                  *kotsutil.KotsKinds             `json:"-"`
 	DownloadStatus             DownloadStatus                  `json:"downloadStatus,omitempty"`
+	AppTitle                   string                          `json:"appTitle,omitempty"`
+	AppIconURI                 string                          `json:"appIconUri"`
```
sgalsaleh, line 58: any reason this is not "omitempty"?

sgalsaleh (dismissed):

sgalsaleh (approved):

PR #3596, opened 2023-02-01
add start of preflights hook

rianfowler:
- adds initial refactored hook for preflights component. This will enable us to migrate the preflights components to functional components so we can use modern react features like hooks. 
- Expect some changes in subsequent PRs as I integrate this into the components. Feedback on the naming conventions etc are appreciated. 

Fixes #

https://app.shortcut.com/replicated/story/68110/create-hook-for-preflight-endpoints

```release-note
none
```

none

### DIFF
--- a/web/src/features/PreflightChecks/api/getPreflightResult.tsx
+++ b/web/src/features/PreflightChecks/api/getPreflightResult.tsx
@@ -12,29 +12,29 @@ async function getPreflightResult({
   slug: string;
   sequence?: number;
 }): Promise<PreflightResponse> {
-  const getUrl = sequence ?
-    `${apiEndpoint}/app/${slug}/sequence/${sequence}/preflight/result` :
-    `${apiEndpoint}/app/${slug}/preflight/result`;
-  const jsonResponse = await fetch(
-    getUrl,
-    {
-      method: "GET",
-      headers: {
-        Authorization: Utilities.getToken(),
-      },
-    }
-  );
+  const getUrl = sequence
+    ? `${apiEndpoint}/app/${slug}/sequence/${sequence}/preflight/result`
+    : `${apiEndpoint}/app/${slug}/preflight/result`;
+  const jsonResponse = await fetch(getUrl, {
+    method: "GET",
+    headers: {
+      Authorization: Utilities.getToken(),
+    },
+  });
 
   if (!jsonResponse.ok) {
-    throw new Error(`Encountered an error while fetching preflight results: Unexpected status code: ${jsonResponse.status}`);
+    throw new Error(
+      `Encountered an error while fetching preflight results: Unexpected status code: ${jsonResponse.status}`
+    );
   }
 
   const response: PreflightResponse = await jsonResponse.json();
 
   // unmarshall these nested JSON strings
   if (typeof response?.preflightResult?.result === "string")
-    response.preflightResult.result =
-      JSON.parse(response.preflightResult.result);
+    response.preflightResult.result = JSON.parse(
+      response.preflightResult.result
+    );
 
   if (typeof response?.preflightProgress === "string")
     response.preflightProgress = JSON.parse(response.preflightProgress);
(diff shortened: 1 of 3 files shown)


web/src/features/PreflightChecks/api/getPreflightResult.tsx
```diff
@@ -0,0 +1,122 @@
...
+        learnMoreUri: responseResult.uri || "",
+        message: responseResult.message || "",
+        title: responseResult.title || "",
+        showCannotFail: responseResult?.strict || false,
+        showFail: responseResult?.isFail || false,
+        showPass: responseResult?.isPass || false,
+        showWarn: responseResult?.isWarn || false,
+      })) || [],
+    showDeploymentBlocked: response?.preflightResult?.result?.results
+      ?.find(result => result?.isFail && result?.strict) ? true : false,
```
rianfowler, line 70: Decided to elevate this so it's easier to handle in the components as a simple boolean instead of forcing the components to search every result on every render. 

https://github.com/replicatedhq/kots/blob/ea5b1e5bed2ac1fdfc6459cb8f6bf604cc8bebc1/web/src/features/PreflightChecks/types.ts#L34-L43

https://github.com/replicatedhq/kots/blob/fbc20947791ce613b30b68b9074292ea0bdcd5aa/web/src/components/PreflightRenderer.jsx#L84-L88

web/src/features/PreflightChecks/api/getPreflightResult.tsx
```diff
@@ -0,0 +1,122 @@
...
+    showPreflightCheckPending: !response?.preflightResult?.result,
+    showPreflightNoChecks: response?.preflightResult?.result
+      ?.results?.length === 0,
+    showPreflightSkipped: response?.preflightResult?.skipped,
+    showRbacError: response?.preflightResult?.result
+      ?.errors?.find(error => error?.isRbac) ? true : false,
```
rianfowler, line 76: Similar thing here with removing the finds from the rendering 

https://github.com/replicatedhq/kots/blob/ea5b1e5bed2ac1fdfc6459cb8f6bf604cc8bebc1/web/src/features/PreflightChecks/types.ts#L29-L32

https://github.com/replicatedhq/kots/blob/fbc20947791ce613b30b68b9074292ea0bdcd5aa/web/src/components/PreflightResultErrors.tsx#L139

web/src/features/PreflightChecks/api/getPreflightResult.tsx
```diff
@@ -0,0 +1,122 @@
...
+  }
+}
+
+function makeRefetchInterval(preflightCheck: PreflightCheck): number | false {
+  if (preflightCheck.showPreflightCheckPending)
```
rianfowler, line 81: As far as I can tell this is all we need to determine if we're polling- whether or not the response has a results array. I will do a lot of testing when i actually hook it into the component. 

https://github.com/replicatedhq/kots/blob/fbc20947791ce613b30b68b9074292ea0bdcd5aa/web/src/components/PreflightResultPage.tsx#L284-L285

https://github.com/replicatedhq/kots/blob/ea5b1e5bed2ac1fdfc6459cb8f6bf604cc8bebc1/web/src/features/PreflightChecks/api/getPreflightResult.tsx#L71

web/src/features/PreflightChecks/types.ts
```diff
@@ -0,0 +1,67 @@
+/* REQUEST TYPES */
```
rianfowler, line 1: dunno what these are yet. Also I copied some of the type definitions you did a while back. I will refactor them out of the main types file when I update the components.

web/src/features/PreflightChecks/api/getPreflightResult.tsx
```diff
@@ -0,0 +1,137 @@
...
+    );
+  }
+
+  const response: PreflightResponse = await jsonResponse.json();
+
+  // unmarshall these nested JSON strings
+  if (typeof response?.preflightResult?.result === "string")
+    response.preflightResult.result = JSON.parse(
+      response.preflightResult.result
+    );
```
miaawong, line 37: wonder if using axios would be helpful here since axios responses are already served as javascript object. You wouldn't need to parse.

miaawong (approved):
amazing 🙌 
flattenPreflightResponse must've been a head banger

web/src/features/PreflightChecks/api/getPreflightResult.tsx
rianfowler: might try that out.

PR #3601, opened 2023-02-03
Add initial install diagrams for style and content suggestions

ousbots:

This PR is for discussing the kots install diagrams.

No issue, this is to hopefully help people understand kots workflows better.

These are incomplete and I'm putting them up here as an example for critique before spending too much more time finishing the diagrams. It took me a while to figure out a good way to display an appropriate level of detail and context without being mostly noise.

There are two charts here. The first is a sequence diagram of the communication between kots, k8s, docker, and kotsadm parts. The second is a higher level flow chart of the kots cli install process.

## Steps to reproduce

To rebuild the images:
```
cd kots/design/diagrams
yarn install
yarn run build
```

```release-note
NONE
```

```
NONE
```


ousbots:
What do you think about the level of detail in these diagrams? Would it be useful to indicate better which function and file each step is in, maybe with boxed and colored backgrounds with this info? How about the presentation of what data kots sends to k8s (and other components) in the sequence diagram, does that make sense? Would more notes describing higher level process and the reason why it's happening help or just be clutter?

ousbots:
Also, these are very incomplete as they're the first iteration of the diagrams that I thought weren't confusing and way too busy

ousbots:
What about the level of Kubernetes details? I wasn't sure if I should explain what each Kubernetes request is doing and why it's used in the context of an install. Is everyone familiar enough with what it means to create things like configmaps, secrets, namespaces, deployments, and services?

design/diagrams/kots_install_sequence.mmd
```diff
@@ -0,0 +1,267 @@
...
+	participant k8s as Cluster k8s API
+	participant kotsadm as Kots Admin API
+	participant kotsdb as Kots Admin Database (rqlite)
+	participant kotsstore as Kots Admin Data Store (minio)
+	participant docker as Docker Registry
+	participant repli as Replicated API
+
+	%% Check that there are sufficient RBAC permissions on the cluster
+	note over kots, k8s: Check that there are sufficient priveledges on the cluster to deploy the app
+	kots -> k8s: Self Subject Access Review of resource attributes [Verb: *, Group: *, Version: *, Resource: *]
```
rianfowler, line 12: `->` doesn't render an arrowhead; `->>` and `-->>` [will](https://mermaid.js.org/syntax/sequenceDiagram.html#messages) though.

rianfowler (approved):
this is pretty good. thanks for putting it together.

PR #3610, opened 2023-02-09
exclude shutdown pods from backup

pavansokkenagaraj:
If pods are in shutdown state and while backing up the instances, hook backup fails as it tries to exec and tries collecting files. This doesn't happen in pod backup since velero backs up only pods in running state

Fixes # [SC-53284: Velero backups will fail if pods in a Shutdown state are present](https://app.shortcut.com/replicated/story/53284)

It is not possible to create a pod with pod.status.phase as `Failed` + pod.status.reason as `Shutdown`
I have added unit tests + was able to validate the code by changing phase to ImagePullErr and check the pods have the velero exclude labels

```release-note
- Fixes a bug where velero snapshot backup failed due to pods in Shutdown state
```

NONE

### DIFF
--- a/pkg/kotsadmsnapshot/backup.go
+++ b/pkg/kotsadmsnapshot/backup.go
@@ -28,7 +28,10 @@ import (
 	veleroclientv1 "github.com/vmware-tanzu/velero/pkg/generated/clientset/versioned/typed/velero/v1"
 	velerolabel "github.com/vmware-tanzu/velero/pkg/label"
 	"go.uber.org/zap"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/kubernetes"
 )
 
@@ -161,7 +164,6 @@ func CreateApplicationBackup(ctx context.Context, a *apptypes.App, isScheduled b
 		return nil, errors.Wrap(err, "failed to create k8s clientset")
 	}
 
-	// excludeShutdownPodsFromBackup will add a hook to the backup to exclude pods that are in the process of shutting down
 	if err := excludeShutdownPodsFromBackup(ctx, clientset, veleroBackup); err != nil {
 		return nil, errors.Wrap(err, "failed to exclude shutdown pods from backup")
 	}
@@ -390,7 +392,6 @@ func CreateInstanceBackup(ctx context.Context, cluster *downstreamtypes.Downstre
 		}
 	}
 
-	// excludeShutdownPodsFromBackup will add a hook to the backup to exclude pods that are in the process of shutting down
 	if err := excludeShutdownPodsFromBackup(ctx, clientset, veleroBackup); err != nil {
 		return nil, errors.Wrap(err, "failed to exclude shutdown pods from backup")
 	}
@@ -963,16 +964,27 @@ func excludeShutdownPodsFromBackup(ctx context.Context, clientset kubernetes.Int
 		}
 	}
 
+	kotsadmLabelSet := labels.Set{
+		kotsadmtypes.KotsadmKey:  kotsadmtypes.KotsadmLabelValue,
+		kotsadmtypes.BackupLabel: kotsadmtypes.BackupLabelValue,
+	}
+
+	selectorMap := map[string]string{
+		"status.phase": string(corev1.PodFailed),
+	}
+
+	failedPodListOptions := metav1.ListOptions{
+		LabelSelector: kotsadmLabelSet.String(),
+		FieldSelector: fields.SelectorFromSet(selectorMap).String(),
+	}
+
 	for _, namespace := range namespaces {
-		pods, err := clientset.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{
-			FieldSelector: "status.phase=Failed",
-		})
+		pods, err := clientset.CoreV1().Pods(namespace).List(ctx, failedPodListOptions)
 		if err != nil {
 			return errors.Wrapf(err, "failed to list pods in namespace %s", namespace)
 		}
-
 		for _, pod := range pods.Items {
-			if pod.Status.Reason == "Shutdown" {
+			if pod.Status.Phase == corev1.PodFailed && pod.Status.Reason == "Shutdown" {
 				logger.Infof("Excluding pod %s in namespace %s from backup", pod.Name, namespace)
 				// add velero.io/exclude-from-backup=true label to pod
 				if pod.Labels == nil {
--- a/pkg/kotsadmsnapshot/backup.go
+++ b/pkg/kotsadmsnapshot/backup.go
@@ -29,6 +29,7 @@ import (
 	velerolabel "github.com/vmware-tanzu/velero/pkg/label"
 	"go.uber.org/zap"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 )
 
 func CreateApplicationBackup(ctx context.Context, a *apptypes.App, isScheduled bool) (*velerov1.Backup, error) {
@@ -153,6 +154,16 @@ func CreateApplicationBackup(ctx context.Context, a *apptypes.App, isScheduled b
 		veleroBackup.Spec.TTL = metav1.Duration{
 			Duration: ttlDuration,
 		}
+	}
+
+	clientset, err := k8sutil.GetClientset()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create k8s clientset")
+	}
+
+	// excludeShutdownPodsFromBackup will add a hook to the backup to exclude pods that are in the process of shutting down
+	if err := excludeShutdownPodsFromBackup(ctx, clientset, veleroBackup); err != nil {
+		return nil, errors.Wrap(err, "failed to exclude shutdown pods from backup")
 	}
 
 	cfg, err := k8sutil.GetClusterConfig()
@@ -377,6 +388,11 @@ func CreateInstanceBackup(ctx context.Context, cluster *downstreamtypes.Downstre
 		veleroBackup.Spec.TTL = metav1.Duration{
 			Duration: ttlDuration,
 		}
+	}
+
+	// excludeShutdownPodsFromBackup will add a hook to the backup to exclude pods that are in the process of shutting down
+	if err := excludeShutdownPodsFromBackup(ctx, clientset, veleroBackup); err != nil {
+		return nil, errors.Wrap(err, "failed to exclude shutdown pods from backup")
 	}
 
 	cfg, err := k8sutil.GetClusterConfig()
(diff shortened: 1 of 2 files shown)


pkg/kotsadmsnapshot/backup.go
```diff
@@ -930,3 +946,59 @@ func prepareIncludedNamespaces(namespaces []string) []string {
...
+// excludeShutdownPodsFromBackup will exclude pods that are in a shutdown state from the backup
+// this is to prevent the hook backup from failing if a pod is in a shutdown state and cannot be backed up
+func excludeShutdownPodsFromBackup(ctx context.Context, clientset kubernetes.Interface, backup *velerov1.Backup) (err error) {
+	namespaces := backup.Spec.IncludedNamespaces
+	if len(namespaces) == 0 {
+		return
+	}
+
+	// if the backup includes all namespaces, we need to get the list of namespaces
+	if namespaces[0] == "*" {
```
sgalsaleh, line 959: `*` is not necessarily the first element in the array. i would suggest looping through the array to detect it.
```suggestion
	if namespaces[0] == "*" {
```

pkg/kotsadmsnapshot/backup.go
```diff
@@ -930,3 +946,59 @@ func prepareIncludedNamespaces(namespaces []string) []string {
...
+
+		for _, pod := range pods.Items {
+			if pod.Status.Reason == "Shutdown" {
+				logger.Infof("Excluding pod %s in namespace %s from backup", pod.Name, namespace)
+				// add velero.io/exclude-from-backup=true label to pod
+				if pod.Labels == nil {
+					pod.Labels = map[string]string{}
+				}
+
+				pod.Labels["velero.io/exclude-from-backup"] = "true"
```
sgalsaleh, line 982: they are constants already defined for the exclude key and value:
```suggestion
				pod.Labels[kotsadmtypes.ExcludeKey] = kotsadmtypes.ExcludeValue
```

pkg/kotsadmsnapshot/backup.go
sgalsaleh: hmm, actually nevermind, i see that the `prepareIncludedNamespaces` already handles that

pkg/kotsadmsnapshot/backup.go
```diff
@@ -930,3 +946,59 @@ func prepareIncludedNamespaces(namespaces []string) []string {
...
+	if namespaces[0] == "*" {
+		namespaces, err = getNamespaces(ctx, clientset)
+		if err != nil {
+			return errors.Wrap(err, "failed to get namespaces")
+		}
+	}
+
+	for _, namespace := range namespaces {
+		pods, err := clientset.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{
+			FieldSelector: "status.phase=Failed",
```
sgalsaleh, line 968: to scope this even further, i would suggest adding a `kots.io/backup: velero` label selector as well (also has constants `kotsadmtypes.BackupLabel` and `kotsadmtypes.BackupLabelValue` defined)

sgalsaleh: in fact, there should be two selections, one with `kots.io/backup: velero` and `kots.io/kotsadm: true`, and the other with `kots.io/backup: velero` and `kots.io/app-slug: <app-slug>` so that only the relevant pods get modified.

pkg/kotsadmsnapshot/backup.go
```diff
@@ -930,3 +946,59 @@ func prepareIncludedNamespaces(namespaces []string) []string {
...
+				}
+			}
+		}
+	}
+
+	return nil
+}
+
+func getNamespaces(ctx context.Context, clientset kubernetes.Interface) (namespaces []string, err error) {
+	ns, err := clientset.CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
```
sgalsaleh, line 995: this won't work with minimal rbac. i would suggest simply returning an empty array if `k8sutil.IsKotsadmClusterScoped` is false.

pkg/kotsadmsnapshot/backup.go
```diff
@@ -379,6 +390,11 @@ func CreateInstanceBackup(ctx context.Context, cluster *downstreamtypes.Downstre
 		}
 	}
 
+	// excludeShutdownPodsFromBackup will add a hook to the backup to exclude pods that are in the process of shutting down
```
sgalsaleh, line 393: this comment isn't necessary since the function definition already has the same comment

pkg/kotsadmsnapshot/backup.go
```diff
@@ -155,6 +156,16 @@ func CreateApplicationBackup(ctx context.Context, a *apptypes.App, isScheduled b
 		}
 	}
 
+	clientset, err := k8sutil.GetClientset()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create k8s clientset")
+	}
+
+	// excludeShutdownPodsFromBackup will add a hook to the backup to exclude pods that are in the process of shutting down
```
sgalsaleh, line 164: ditto

pkg/kotsadmsnapshot/backup.go
```diff
@@ -930,3 +946,59 @@ func prepareIncludedNamespaces(namespaces []string) []string {
 	}
 	return includedNamespaces
 }
+
+// excludeShutdownPodsFromBackup will exclude pods that are in a shutdown state from the backup
+// this is to prevent the hook backup from failing if a pod is in a shutdown state and cannot be backed up
+func excludeShutdownPodsFromBackup(ctx context.Context, clientset kubernetes.Interface, backup *velerov1.Backup) (err error) {
```
sgalsaleh, line 952: i think there's an edge case in the minimal rbac scenario if `*` is part of the included namespaces because the `prepareIncludedNamespaces` would remove all the namespaces and keep `*`, and this whole function would be a no-op. one workaround would be to execute this function before the `prepareIncludedNamespaces` function is called. if you do that, please add a comment.

pkg/kotsadmsnapshot/backup.go
```diff
@@ -930,3 +949,68 @@ func prepareIncludedNamespaces(namespaces []string) []string {
...
+				return errors.Wrapf(err, "failed to update pod %s in namespace %s", pod.Name, pod.Namespace)
+			}
+		}
+	}
+	return nil
+}
+
+// buildShutdownPodListOptions returns a list options object that will match all pods that are in a failed state with shutdown reason
+func buildShutdownPodListOptions() metav1.ListOptions {
+	kotsadmLabelSet := labels.Set{
```
sgalsaleh, line 1003: those labels won't work for application resources. since selectors don't support "OR" operations for different keys, you'll probably need to list pods twice, once with the kotsadm key/value, and another with the app slug key/value.

pavansokkenagaraj: Don't the pods have both the labels?
I thought we were listing only pods with those 2 labels.

sgalsaleh: kots components will have the `kots.io/kotsadm: "true"` label, and application pods will have the `kots.io/app-slug: <app-slug>` label

pavansokkenagaraj: made changes in https://github.com/replicatedhq/kots/pull/3610/commits/4f9e404dc195935e60ac9281f93baa7bef524d53

pkg/kotsadmsnapshot/backup.go
```diff
@@ -930,3 +949,79 @@ func prepareIncludedNamespaces(namespaces []string) []string {
 	}
 	return includedNamespaces
 }
+
+// excludeShutdownPodsFromBackup will exclude pods that are in a shutdown state from the backup
+// this is to prevent the hook backup from failing if a pod is in a shutdown state and cannot be backed up
```
sgalsaleh, line 954: this second statement is not necessary since i believe there are other scenarios that would cause the backup to fail

pkg/kotsadmsnapshot/backup.go
```diff
@@ -930,3 +949,79 @@ func prepareIncludedNamespaces(namespaces []string) []string {
...
+func excludeShutdownPodsFromBackup(ctx context.Context, clientset kubernetes.Interface, backupNamespaces []string, isKotsadmClusterScoped bool) (err error) {
+	failedPodListOptions := buildShutdownPodListOptions()
+
+	for _, namespace := range backupNamespaces {
+		if namespace == "*" {
+			if !isKotsadmClusterScoped {
+				continue
+			} else {
+				// if namespace is *, kubernetes api equivalent is empty string for all namespaces
```
sgalsaleh, line 963:
```suggestion
				// specifying an empty ("") namespace in client-go retrieves resources from all namespaces
```

pkg/kotsadmsnapshot/backup.go
```diff
@@ -312,6 +327,10 @@ func CreateInstanceBackup(ctx context.Context, cluster *downstreamtypes.Downstre
 		includedNamespaces = append(includedNamespaces, kotsadmVeleroBackendStorageLocation.Namespace)
 	}
 
+	// excludeShutdownPodsFromBackup is run before the prepareIncludedNamespaces to ensure that the isKotsadmClusterScoped installs ignore all namespaces
+	if err := excludeShutdownPodsFromBackup(ctx, clientset, includedNamespaces, isKotsadmClusterScoped); err != nil {
+		return nil, errors.Wrap(err, "failed to exclude shutdown pods from backup")
+	}
```
sgalsaleh, line 333: nit pick: let's add a new line here

pkg/kotsadmsnapshot/backup.go
```diff
@@ -312,6 +327,10 @@ func CreateInstanceBackup(ctx context.Context, cluster *downstreamtypes.Downstre
 		includedNamespaces = append(includedNamespaces, kotsadmVeleroBackendStorageLocation.Namespace)
 	}
 
+	// excludeShutdownPodsFromBackup is run before the prepareIncludedNamespaces to ensure that the isKotsadmClusterScoped installs ignore all namespaces
```
sgalsaleh, line 330: this comment won't be needed anymore

pkg/kotsadmsnapshot/backup.go
```diff
@@ -107,6 +111,18 @@ func CreateApplicationBackup(ctx context.Context, a *apptypes.App, isScheduled b
...
 	includedNamespaces = append(includedNamespaces, kotsKinds.KotsApplication.Spec.AdditionalNamespaces...)
 
+	clientset, err := k8sutil.GetClientset()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create k8s clientset")
+	}
+
+	isKotsadmClusterScoped := k8sutil.IsKotsadmClusterScoped(ctx, clientset, kotsadmNamespace)
+
+	// excludeShutdownPodsFromBackup is run before the prepareIncludedNamespaces to ensure that the isKotsadmClusterScoped installs ignore all namespaces
```
sgalsaleh, line 121: this comment won't be needed anymore

pkg/kotsadmsnapshot/backup.go
```diff
@@ -930,3 +948,56 @@ func prepareIncludedNamespaces(namespaces []string) []string {
...
+		return errors.Wrapf(err, "failed to list pods in namespace %s", namespace)
+	}
+
+	if podList == nil {
+		return nil
+	}
+
+	for _, pod := range podList.Items {
+		if pod.Status.Phase == corev1.PodFailed && pod.Status.Reason == "Shutdown" {
+			logger.Infof("Excluding pod %s in namespace %s from backup", pod.Name, namespace)
```
sgalsaleh, line 989:
```suggestion
			logger.Infof("Excluding pod %s in namespace %s from backup because it's in a Shutdown state", pod.Name, namespace)
```

sgalsaleh (approved):

PR #3616, opened 2023-02-10
Add GKE Autopilot tests

ousbots:

Modifies the kots-e2e GitHub action to accept a few new inputs that enable running tests on Google Cloud Kubernetes Engine (GKE) Autopilot clusters. The most important new input is `platform`, which will accept either 'gke' or 'k3s' and run the end to end (e2e) tests on the given platform. There is also another input `gke-service-key` that is used as the GKE authentication token. All other inputs have reasonable default values, but can be used to change the cluster configuration if needed.

The modifications to the e2e action are steps that authenticate to GCloud, create a new autopilot cluster, then generate a kubectl config to the new cluster if the platform input is 'gee'. The test run continues as normal and after the tests are run, even on failure, the cluster is destroyed.

Each job in a workflow will create and destroy the cluster used by the job. While this leads to more clusters being created, it avoids the pitfalls of trying to make sure that when there are multiple jobs run on GKE clusters that there won't be failures synchronizing the creation, use of, and destruction of a cluster that is shared between jobs. The cost impact from this should be minimal.

The platforms are also run in a matrix with the k8s versions, although currently the created cluster uses the latest k8s version on GKE. If there is a test with multiple versions in a matrix with k3s, a single version of gke can be added with the include annotation:
```
test-name:
  strategy:
    matrix:
      k8s_version: [ v1.26.1, v1.25.6 ]
      platform: [ k3s ]
      include:
        - platform: gke
          k8s_version: "v1.26.1"
```

For estimating costs, let's assume that there could be a max of 5 PRs created per day and each PR will run 3 tests (3 clusters created) and each test starts 5 pods each using 1GB of memory and run for 1 hour. Since GKE charges by vCPU and memory usage, this will be 30 vCPUs and 30GB of memory being used for 1 hour per PR and with 5 PRs that's 150 of hourly units to charge each day. Based on the prices found [here](https://cloud.google.com/kubernetes-engine/pricing) that totals to:

75 vCPU hours * $0.0445 / vCPU hour = $3.34
75 GB ram hours * $0.0049225 / ram hour = $0.37
Total daily cost = $3.71
Monthly cost (x20) = $74.20

The real number will be much lower, because this PR only enables a single GKE test and there are probably on average fewer than 5 PRs created per day.

The creation and destruction of a GKE cluster seems to take about 5-10 minutes each and running the test on the cluster adds maybe 1-2 minutes to the test as the GKE pods take longer to spin up than a k3s cluster. So on average this will add around 12-15 minutes to a test, but could take as long as 25 extra minutes. In practice everything usually runs on the faster end of those ranges, but there have been a few times where a GKE command takes a lot longer to run than normal.

The regular validate-smoke-test does not work on autopilot clusters due to a security restriction not allowing hostpaths to be mounted in write mode ([docs](https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-security)), but the validate-air-gap-smoke-test does not need those permissions, so was used instead.

Many of the other testim tests could be modified to either have higher timeouts (deploys seem to take around 2 minutes before being ready) or "wait for" (testim action) the kotsadm web interface to say "Ready" instead of "Unknown". 

Fixes [#SC-62335](https://app.shortcut.com/replicated/story/62335/add-tests-for-kots-install-on-gke-autopilot)

The Testim step timeout for the validate-airgap-smoke-test checking for 'Ready' in the admin page after a deploy was increased to 5 minutes as it typically takes around 2 minutes for a deployment to be ready, but the testim test has a default timeout of 30 seconds (really 28 for some reason).

```release-note
NONE
```

NONE


.github/actions/kots-e2e/action.yml
```diff
@@ -52,11 +55,66 @@ inputs:
...
+        gcloud container clusters create-auto "${{ steps.cluster-name.outputs.cluster_name }}" --project "${{ inputs.gke-project }}" --region "${{ inputs.gke-cluster-region }}" --network "${{ inputs.gke-cluster-network }}"  --subnetwork "${{ inputs.gke-cluster-subnetwork }}"
+
+    - name: Get GKE cluster credentials for kubectl
+      if: inputs.platform == 'gke'
+      uses: google-github-actions/get-gke-credentials@v1
+      with:
+        cluster_name: ${{ steps.cluster-name.outputs.cluster_name }}
+        location: ${{ inputs.gke-cluster-region }}
+        project_id: ${{ inputs.gke-project }}
+
```
cbodonnell, line 113: Could this be its own action and it's called from here?  Can be maintained in this repo or outside, but would simplify the logic in case we add more platforms down the road.

Something like:
```
- name: Create a GKE cluster
  if: inputs.platform == 'gke'
  uses: ./.github/actions/clusters/gke/create
  with:
    ...
- name: Destroy a GKE cluster
  if: ${{ always() && inputs.platform == 'gke' }}
  uses: ./.github/actions/clusters/gke/destroy
  with:
    ...
```

.github/workflows/build-test.yaml
```diff
@@ -431,10 +431,12 @@ jobs:
         with:
           test-focus: 'Regression'
           k8s-version: 'v1.26.1-k3s1'
+          platform: 'k3s'
           testim-branch: ${{ github.head_ref == 'main' && 'master' || github.head_ref }}
           testim-access-token: '${{ secrets.TESTIM_ACCESS_TOKEN }}'
           aws-access-key-id: '${{ secrets.E2E_SUPPORT_BUNDLE_AWS_ACCESS_KEY_ID }}'
           aws-secret-access-key: '${{ secrets.E2E_SUPPORT_BUNDLE_AWS_SECRET_ACCESS_KEY }}'
+          gke-service-key: ${{ secrets.GOOGLE_CLOUD_CREDENTIALS }}
```
cbodonnell, line 439: feels weird that this is passed to every test even if we don't use it for most.  though I can see how it makes it easy to simply toggle or add new platforms.  if you don't pass this, will it just fail since it's not a required input?

.github/workflows/build-test.yaml
```diff
@@ -464,12 +467,14 @@ jobs:
...
           test-focus: 'Smoke Test'
           kots-namespace: 'smoke-test'
+          platform: '${{ matrix.platform }}'
           k8s-version: '${{ matrix.k8s_version }}'
           k8s-arguments: ${{ startsWith(matrix.k8s_version, 'v1.22') && '--egress-selector-mode disabled' || '' }}
           testim-access-token: '${{ secrets.TESTIM_ACCESS_TOKEN }}'
           testim-branch: ${{ github.head_ref == 'main' && 'master' || github.head_ref }}
           aws-access-key-id: '${{ secrets.E2E_SUPPORT_BUNDLE_AWS_ACCESS_KEY_ID }}'
           aws-secret-access-key: '${{ secrets.E2E_SUPPORT_BUNDLE_AWS_SECRET_ACCESS_KEY }}'
+          gke-service-key: ${{ secrets.GOOGLE_CLOUD_CREDENTIALS }}
```
sgalsaleh, line 477: why is this needed with k3s? (same to most other tests that are only run on k3s)

.github/actions/kots-e2e/action.yml
sgalsaleh: +1

.github/workflows/build-test.yaml
ousbots: Yeah, if the credentials aren't passed, the test will fail with an error about being unable to authenticate to the google cloud, which should be obvious to anyone adding gke tests in the future. I've removed all the unneeded `gke-service-key` entries as it does feel weird passing an unused secret

.github/actions/kots-e2e/action.yml
ousbots: 😂 I originally had the gke stuff in it's own action but decided that was making things overly complicated. But I do like the separation, so the gke action file is back

.github/actions/kots-e2e/action.yml
```diff
@@ -109,3 +124,10 @@ runs:
         aws-access-key-id: ${{ inputs.aws-access-key-id }}
         aws-secret-access-key: ${{ inputs.aws-secret-access-key }}
         aws-region: ${{ inputs.aws-region }}
+
+    - name: Delete GKE cluster if created
+      if: ${{ always() && inputs.platform == 'gke' }}
+      uses: ./.github/actions/kots-gke
+      with:
+        service-key: inputs.gke-service-key
+        delete: 'true'
```
sgalsaleh, line 133: how about having two separate scripts, one for creating a gke cluster, and the other for deleting a cluster?

sgalsaleh: e.g. `kots-gke-create` and `kots-gke-delete`

.github/workflows/build-test.yaml
ousbots: Craig pointed this out too, I originally added it so that it wouldn't be needed to add gke tests, but it does feel weird passing around an unused secret, so the excess ones are removed now

.github/actions/kots-e2e/action.yml
ousbots: good idea, they're split now. They both go through the gcloud authentication and kubeconfig generation process even though in the current setup the delete does not need it, since it's in the same workflow as the create, but this will give a little more flexibility for the future and that part only takes seconds

sgalsaleh (approved):

cbodonnell (approved):

e3b0c442 (approved):
My only comment would be, does it make sense to have the `platform` input be a requirement in `kots-e2e` when it seems like gke is an edge case. A sane default might make more sense and reduce the number of workflow jobs that need to be updated.

That's a nitpick detail though, beyond that lgtm

ousbots:
I agree with making the platform not required with a default of 'k3s', that's a good nitpick and implemented now

cbodonnell (approved):

sgalsaleh (approved):

PR #3617, opened 2023-02-10
Miawong/sc 68585/refactor fetch app downstream

miaawong:

Fixes # https://app.shortcut.com/replicated/story/68585/refactor-dashboard-tsx-to-use-custom-hooks

```release-note
NONE
```

NONE

### DIFF
--- a/web/src/features/Dashboard/components/Dashboard.tsx
+++ b/web/src/features/Dashboard/components/Dashboard.tsx
@@ -177,7 +177,6 @@ const Dashboard = (props: Props) => {
     useUpdateDownloadStatus();
   console.log(getUpdateDownloadStatus, "get update dw status");
   const fetchAppDownstream = async () => {
-    console.log("fetch downstream");
     if (!app) {
       return;
     }
@@ -227,7 +226,6 @@ const Dashboard = (props: Props) => {
   };
 
   const startFetchAppDownstreamJob = () => {
-    console.log("start fetch app downstream polling");
     state.fetchAppDownstreamJob.start(fetchAppDownstream, 2000);
   };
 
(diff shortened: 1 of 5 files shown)


web/src/features/Dashboard/components/Dashboard.tsx
```diff
@@ -138,7 +138,6 @@ const Dashboard = (props: Props) => {
       currentVersion: null,
       displayErrorModal: false,
       downstream: null,
-      fetchAppDownstreamJob: new Repeater(),
```
miaawong, line 141: 👋

web/src/features/Dashboard/components/Dashboard.tsx
```diff
@@ -174,58 +173,26 @@ const Dashboard = (props: Props) => {
...
-          ? err.message
-          : "Something went wrong, please try again.";
-      setState({
-        loadingApp: false,
-        gettingAppErrMsg: errorMessage,
-        displayErrorModal: true,
-      });
-    }
+  const onAppDownstreamSuccess = (data: Downstream) => {
+    setState({ downstream: data });
```
miaawong, line 177: in the beginning I wanted to be able to extract downstream and remove the local state. Looks like there is a function that sets downstream too. 
https://github.com/replicatedhq/kots/blob/158eef9895df56f542cacf88cedbe0313e4530d5/web/src/features/Dashboard/components/Dashboard.tsx#L196

will look into it and maybe refactor that later on.

rianfowler (dismissed):

web/src/features/Dashboard/components/Dashboard.tsx
```diff
@@ -174,58 +173,26 @@ const Dashboard = (props: Props) => {
...
+    setTimeout(() => {
+      props.refreshAppData();
+    }, 2000);
```
rianfowler, line 180: minor nit. its best to `clearTimeout` when the component unmounts. I don't think it's a big deal here unless the handlers are going to blow up if they receive data after the component is unmounted https://developer.mozilla.org/en-US/docs/Web/API/clearTimeout

PR #3619, opened 2023-02-13
flatten isHelmManaged hook return value 

rianfowler:

- flatten `isHelmManaged` 

https://app.shortcut.com/replicated/story/68569/refactor-version-history-api-calls-to-use-hooks

```release-note
none
```

none


web/src/components/hooks/useIsHelmManaged.js
```diff
@@ -1,39 +0,0 @@
-import { useQuery } from "react-query";
-import { Utilities } from "../../utilities/utilities";
-
-async function fetchIsHelmManaged({
-  accessToken = Utilities.getToken(),
-  apiEndpoint = process.env.API_ENDPOINT,
-  _fetch = fetch,
```
rianfowler, line 7: not doing this anymore. originally i thought i would unit test this stuff but I think just leaving out any logic is the better path.

web/src/components/hooks/useIsHelmManaged.js
```diff
@@ -1,39 +0,0 @@
...
-    const res = await _fetch(`${apiEndpoint}/is-helm-managed`, {
-      headers: {
-        Authorization: accessToken,
-        "Content-Type": "application/json",
-      },
-      method: "GET",
-    });
-    if (res.ok) {
-      const response = await res.json();
-      return { isHelmManaged: response.isHelmManaged };
```
rianfowler, line 19: decided not to do this kind of transformation in the fetch handler. this is technically a mutation that we should unit test and we can do it in the `select`

web/src/components/hooks/useIsHelmManaged.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+function UseIsHelmManaged({
+  children,
+}: {
+  children: (props: UseQueryResult) => JSX.Element;
+}) {
+  const query = useIsHelmManaged();
+
+  return children(query);
+}
```
rianfowler, line 50: will be glad when i can get rid of this

web/src/features/Dashboard/api/getSelectedAppClusterDashboard.ts
```diff
@@ -47,8 +47,7 @@ export const useSelectedAppClusterDashboard = ({
 }: {
   refetchInterval: number | false;
 }) => {
-  const { data: isHelmManagedResponse } = useIsHelmManaged();
-  const { isHelmManaged } = isHelmManagedResponse || { isHelmManaged: false };
+  const { data: isHelmManaged = false } = useIsHelmManaged();
```
rianfowler, line 50: one thing that kind of bugs me about this is we sometimes have `use-` hooks that you have to pull info off of `{ data: ... }` and some like `useSelectedApp` that just return the data and nothing about the async stuff. I think that inconsistency is bad but not sure how to handle it yet. Maybe we should just pass the `useQuery` interface into `useSelectedApp` but it's composing data from two endpoints.

miaawong (approved):

PR #3621, opened 2023-02-13
Surface HelmChart custom resource decoding and rendering errors better

sgalsaleh:

in some cases, KOTS silently ignores errors in decoding and rendering HelmChart custom resources. this PR improves that by surfacing those errors to the customer for better self-troubleshooting.

Fixes #

```release-note
Fixes an issue where errors related to parsing and rendering HelmChart custom resources are silently ignored.
```

NONE

### DIFF
--- a/pkg/kotsutil/kots.go
+++ b/pkg/kotsutil/kots.go
@@ -33,6 +33,7 @@ import (
 	"github.com/replicatedhq/troubleshoot/pkg/collect"
 	"github.com/replicatedhq/troubleshoot/pkg/docrewrite"
 	velerov1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
+	"gopkg.in/yaml.v2"
 	kuberneteserrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


pkg/kotsutil/kots.go
```diff
@@ -59,6 +60,17 @@ var (
...
+	APIVersion string               `yaml:"apiVersion"`
+	Kind       string               `yaml:"kind"`
+	Metadata   OverlySimpleMetadata `yaml:"metadata"`
+}
+
+type OverlySimpleMetadata struct {
+	Name      string `yaml:"name"`
+	Namespace string `yaml:"namespace"`
+}
+
```
cbodonnell, line 73: Are these types defined in two places now?  If it differs slightly, maybe we can use embedded structs?

cbodonnell (approved):

pkg/kotsutil/kots.go
sgalsaleh: hmm, there is a quirk that's making consolidating the two types a bit annoying, so i'm gonna keep this duplicate for now since it's more of a utility type.

PR #3624, opened 2023-02-13
migrate to minio xl backend

cbodonnell:

This PR migrates from the legacy MinIO FS backend to the new XL backend.  This is necessary because MinIO [dropped support for the legacy backend](https://min.io/docs/minio/linux/operations/install-deploy-manage/migrate-fs-gateway.html).

This PR also adds support for OpenShift and GKE Autopilot to the KOTS CLI `admin-console generate-manifests` command when executed with a Kubernetes cluster context.  This was necessary because cluster context is also necessary to determine whether or not a migration is needed.  The command will still succeed when run without a cluster context, but it will not migrate by default.

Summary of the migration implementation:
* Detect if a migration is necessary when upgrading kots based on the existing minio instance
* If it is needed, add three init containers to the updated minio statefulset:
  * first copies the minio client to a shared emptydir
  * second runs an old minio image and exports data to a shared emptydir
  * third runs a new minio image and imports that data and afterwards swaps the old data directory with the new data directory
* Main container comes up running the new minio image

Fixes [SC-62651](https://app.shortcut.com/replicated/story/62651/unable-to-upgrade-kots-to-use-newer-minio-images)

E2E testing is covered already by the postgres -> rqlite PR tests (these are what initially exposed this issue since it upgrades from v1.57.0)

## Steps to reproduce

```
$ kubectl kots version
Replicated KOTS 1.70.0
$ kubectl kots install $APP_SLUG -n $NAMESPACE
$ make kots all-ttl.sh
$ ./bin/kots admin-console upgrade -n $NAMESPACE --kotsadm-registry=ttl.sh --kotsadm-namespace=$USER --kotsadm-tag=24h
```

```release-note
Upgrades the MinIO image to RELEASE.2023-02-22T18-23-45Z to resolve CVEs
Adds support for OpenShift and GKE Autopilot to the KOTS CLI `admin-console generate-manifests` command when executed with a Kubernetes cluster context
```

Yes, https://github.com/replicatedhq/replicated-docs/pull/956


pkg/kotsadm/objects/scripts/export-minio-data.sh
```diff
@@ -0,0 +1,84 @@
...
+fi
+
+echo "exporting minio iam info"
+$KOTSADM_MINIO_MIGRATION_WORK_DIR/bin/mc admin cluster iam export $KOTSADM_MINIO_LEGACY_ALIAS
+# old instances may throw error: "This 'admin' API is not supported by server in 'mode-server-fs'"
+# ref: https://github.com/minio/mc/issues/4409 and https://github.com/minio/console/issues/863
+if [ $? -ne 0 ]; then
+    echo "exporting minio iam info failed, skipping"
+fi
+set -e
```
cbodonnell, line 70: I've seen these fail with older KOTS versions using old MinIO images.  Even if we don't migrate these, we are still able to create the bucket in the new instance and migrate the data without any lost functionality in KOTS.

pkg/kotsadm/main.go
```diff
@@ -535,6 +558,7 @@ func ensureKotsadm(deployOptions types.DeployOptions, clientset *kubernetes.Clie
 				if err := ensureMinio(deployOptions, clientset); err != nil {
 					return errors.Wrap(err, "failed to ensure minio")
 				}
+				// TODO: This returns before kotsadm-minio is updated
```
cbodonnell, line 561: This appears to return before the minio statefulset gets updated, so it doesn't wait for the initcontainers

cbodonnell: Updated the login in `ensureMinio` to validate the number of initcontainers before returning

pkg/kotsadm/objects/minio_objects.go
```diff
@@ -35,6 +46,33 @@ func MinioStatefulset(deployOptions types.DeployOptions, size resource.Quantity)
 		securityContext = psc
 	}
 
+	cpuRequest, cpuLimit := "50m", "100m"
+	memoryRequest, memoryLimit := "100Mi", "512Mi"
```
cbodonnell, line 50: Would we need to increase the memory limit for the initcontainers since we're running minio and mc in the same container?

pkg/kotsadm/objects/scripts/export-minio-data.sh
cbodonnell: this is now outdated.  we're not migrating config or metadata, only the bucket contents

.github/workflows/build-test.yaml
```diff
@@ -1516,9 +1535,29 @@ jobs:
...
+      - name: download kots binary
+        uses: actions/download-artifact@v2
+        with:
+          name: kots
+          path: bin/
+      - run: chmod +x bin/kots
+      - name: run kots admin-console generate-manifests without k8s context
+        run: |
+          set +e
+          ./bin/kots admin-console generate-manifests --shared-password password
```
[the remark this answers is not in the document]
sgalsaleh: maybe even attempt to deploy them and validate that kotsadm runs?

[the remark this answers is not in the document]
cbodonnell: agreed.  might play around with this test a bit since right now it's cli-only (no k8s).  could probably introduce k8s and just unset `KUBECONFIG` temporarily to test the no-context permutation.

pkg/kotsadm/main.go
```diff
@@ -1011,31 +1032,6 @@ func ReadDeployOptionsFromCluster(namespace string, clientset *kubernetes.Client
 		deployOptions.SharedPassword = sharedPassword
 	}
 
-	if deployOptions.IncludeMinio {
```
sgalsaleh, line 1014: wouldn't removing this cause the `S3AccessKey` and `S3SecretKey` to not be set anymore?

pkg/kotsadm/minio.go
```diff
@@ -97,12 +121,20 @@ func ensureMinioStatefulset(deployOptions types.DeployOptions, clientset *kubern
...
 
 	_, err = clientset.AppsV1().StatefulSets(deployOptions.Namespace).Update(ctx, existingMinio, metav1.UpdateOptions{})
 	if err != nil {
 		return errors.Wrap(err, "failed to update minio statefulset")
 	}
 
+	// ensureMinioStatefulset can return before the init containers are actually added or removed from the pods
+	// if this occurs, the statefulset will show as a false 'Ready' status and kotsadm will start before the migration is complete
+	desiredInitContainers := len(existingMinio.Spec.Template.Spec.InitContainers)
+	if err := waitForMinioInitContainers(ctx, deployOptions.Namespace, clientset, deployOptions.Timeout, desiredInitContainers); err != nil {
```
sgalsaleh, line 134: doesn't `WaitForStatefulSetReady` take care of this?

pkg/kotsadm/objects/scripts/export-minio-data.sh
```diff
@@ -0,0 +1,67 @@
...
+   [ -z $MINIO_ACCESS_KEY ] ||
+   [ -z $MINIO_SECRET_KEY ] ||
+   [ -z $KOTSADM_MINIO_BUCKET_NAME ];
+then
+    echo 's3 migration environment variables not set'
+    exit 1
+fi
+
+# check if the migration has already been completed
+if [ -f /export/.migration ];
```
sgalsaleh, line 28: this check already exists on line 9

pkg/kotsadm/objects/scripts/export-minio-data.sh
```diff
@@ -0,0 +1,67 @@
...
+until curl -s $KOTSADM_MINIO_ENDPOINT/minio/health/ready; do
+    echo "waiting for minio to be ready"
+    sleep 1
+done
+
+# alias the minio instance
+echo "aliasing minio instance"
+until $KOTSADM_MINIO_MIGRATION_WORK_DIR/bin/mc alias set $KOTSADM_MINIO_LEGACY_ALIAS $KOTSADM_MINIO_ENDPOINT $MINIO_ACCESS_KEY $MINIO_SECRET_KEY; do
+    # minio may not actually be ready to accept requests immediately after it is "ready", so this provides a secondary check
+    echo "waiting for minio to be ready"
```
sgalsaleh, line 51:
```suggestion
    echo "attempting to alias minio instance"
```

pkg/kotsadm/objects/scripts/import-minio-data.sh
```diff
@@ -0,0 +1,85 @@
...
+until curl -s $KOTSADM_MINIO_ENDPOINT/minio/health/ready; do
+    echo "waiting for minio to be ready"
+    sleep 1
+done
+
+# alias the minio instance
+echo "aliasing minio instance"
+until $KOTSADM_MINIO_MIGRATION_WORK_DIR/bin/mc alias set $KOTSADM_MINIO_NEW_ALIAS $KOTSADM_MINIO_ENDPOINT $MINIO_ACCESS_KEY $MINIO_SECRET_KEY; do
+    # minio may not actually be ready to accept requests immediately after it is "ready", so this provides a secondary check
+    echo "waiting for minio to be ready"
```
sgalsaleh, line 47:
```suggestion
    echo "attempting to alias minio instance"
```

pkg/kotsadm/objects/scripts/copy-minio-client.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+set -e
+
+# This script copies the minio client binary to a dedicated migration directory
+# This is necessary because the minio client binary is not available in the minio container
+
+# check if the migration has already been completed
+if [ -f /export/.migration ];
```
sgalsaleh, line 9: where does this file get created?

pkg/kotsadm/minio.go
cbodonnell: not always. I observed this function returning before the minio statefulset was updated and ready ready.  it seems that there is a race condition where if the statefulset pods have not begun updating when the check happens, it will see it as ready (which it is, but we need it to actually update before the check happens).  i think it may be because we now update the initcontainers, so k8s might be taking a bit longer to process the change and issue the update

pkg/kotsadm/main.go
cbodonnell: wow nice catch.  i have no idea why this is removed...

pkg/kotsadm/objects/scripts/copy-minio-client.sh
cbodonnell: another good catch.  that got lost somewhere in the back and forth

.github/workflows/build-test.yaml
```diff
@@ -1517,9 +1536,104 @@ jobs:
...
+          set +e
+          if [ ! -d ./admin-console ]; then
+            echo "admin-console directory does not exist"
+            exit -1
+          fi
+          if [ -z "$(ls -A ./admin-console)" ]; then
+            echo "admin-console directory is empty"
+            exit -1
+          fi
+      - name: validate that ./admin-console/minio-statefulset.yaml does not have initContainers since we do not have a k8s context
```
sgalsaleh, line 1573: this is testing that a bug _exists_ which feels weird... how about we just don't? :)

.github/workflows/build-test.yaml
```diff
@@ -1517,9 +1536,104 @@ jobs:
...
+          arguments: ${{ startsWith(matrix.k8s_version, 'v1.22') && '--egress-selector-mode disabled' || '' }}
+      - name: download base kots version
+        run: |
+          curl -LO https://github.com/replicatedhq/kots/releases/download/$BASE_KOTS_VERSION/kots_linux_amd64.tar.gz \
+            && tar zxvf kots_linux_amd64.tar.gz \
+            && mv kots kots-$BASE_KOTS_VERSION
+      - name: install base kots version
+        run: |
+          set +e
+          ./kots-$BASE_KOTS_VERSION install $APP_SLUG/automated -n generate-manifests --shared-password password --port-forward=false
```
sgalsaleh, line 1593: hmm, can we just test the generate-manifests command as a fresh install? would make this test a lot simpler/cleaner.

pkg/kotsadm/main.go
```diff
@@ -466,6 +478,24 @@ func removeKotsadmPostgres(deployOptions types.DeployOptions, clientset *kuberne
 	return nil
 }
 
+func removeMinioXlMigrationScriptsConfigMap(deployOptions types.DeployOptions, clientset *kubernetes.Clientset) error {
```
sgalsaleh, line 481: let's not remove the scripts so that we can use them for troubleshooting purposes if something goes wrong later. it's just a configmap, so it's not a big deal to leave it around.

pkg/kotsadm/types/deployoptions.go
```diff
@@ -34,6 +34,8 @@ type DeployOptions struct {
 	ProgressWriter         io.Writer
 	IncludeMinio           bool
 	IncludeMinioSnapshots  bool
+	MigrateMinioXl         bool
```
sgalsaleh, line 37:
```suggestion
	MigrateToMinioXl         bool
```

pkg/kotsadm/minio.go
```diff
@@ -97,12 +121,20 @@ func ensureMinioStatefulset(deployOptions types.DeployOptions, clientset *kubern
 	existingMinio.Spec.Template.Spec.Volumes = desiredMinio.Spec.Template.Spec.DeepCopy().Volumes
 	existingMinio.Spec.Template.Spec.Containers[0].Image = desiredMinio.Spec.Template.Spec.Containers[0].Image
 	existingMinio.Spec.Template.Spec.Containers[0].VolumeMounts = desiredMinio.Spec.Template.Spec.Containers[0].DeepCopy().VolumeMounts
+	existingMinio.Spec.Template.Spec.InitContainers = desiredMinio.Spec.Template.Spec.DeepCopy().InitContainers
```
sgalsaleh, line 124: what if the kots upgrade command happens to fails after updating the minio statefulset and the migration failed or didn't finish yet, and the user reruns the upgrade command? the init container would be removed because the current minio image in the cluster is the new one, no?

pkg/kotsadm/types/deployoptions.go
```diff
@@ -34,6 +34,8 @@ type DeployOptions struct {
...
+	MigrateMinioXlOldImage string
```
sgalsaleh, line 38:
```suggestion
	CurrentMinioImage string
```

pkg/kotsadm/minio.go
sgalsaleh: the `WaitForStatefulSetReady` checks the `generation` and `observedGeneration` which are immediately updated even if the pods did not roll yet. so, the only race condition would happen if the actual statefulset definition is not updated yet, which seems a bit unlikely.

pkg/kotsadm/objects/images.go
```diff
@@ -37,6 +41,7 @@ func GetAdminConsoleImages(deployOptions types.DeployOptions) map[string]string
 		"kotsadm-migrations": fmt.Sprintf("%s/kotsadm-migrations:%s", kotsadmversion.KotsadmRegistry(deployOptions.RegistryConfig), kotsadmversion.KotsadmTag(deployOptions.RegistryConfig)),
 		"kotsadm":            fmt.Sprintf("%s/kotsadm:%s", kotsadmversion.KotsadmRegistry(deployOptions.RegistryConfig), kotsadmversion.KotsadmTag(deployOptions.RegistryConfig)),
 		"minio":              minioImage,
+		"minio-client":       minioClientImage,
```
sgalsaleh, line 44: this should also be added in the `GetOriginalAdminConsoleImages` function below.

pkg/kotsadm/objects/minio_objects.go
```diff
@@ -217,6 +213,255 @@ func MinioStatefulset(deployOptions types.DeployOptions, size resource.Quantity)
...
+							Key: "secretkey",
+						},
+					},
+				},
+				{
+					Name:  "KOTSADM_MINIO_LEGACY_ALIAS",
+					Value: "legacy",
+				},
+				{
+					Name:  "KOTSADM_MINIO_MIGRATION_WORK_DIR",
```
sgalsaleh, line 281: nit pick: i would've called this `KOTSADM_MINIO_MIGRATION_DIR`

pkg/kotsadm/objects/minio_objects.go
```diff
@@ -217,6 +213,255 @@ func MinioStatefulset(deployOptions types.DeployOptions, size resource.Quantity)
...
+					Value: "/migration-work",
```
sgalsaleh, line 282: nit pick: i would've called this `/migration`

pkg/kotsadm/objects/minio_objects.go
```diff
@@ -217,6 +213,255 @@ func MinioStatefulset(deployOptions types.DeployOptions, size resource.Quantity)
...
+							Key: "secretkey",
+						},
+					},
+				},
+				{
+					Name:  "KOTSADM_MINIO_NEW_ALIAS",
+					Value: "new",
+				},
+				{
+					Name:  "KOTSADM_MINIO_MIGRATION_WORK_DIR",
```
sgalsaleh, line 332: ditto

pkg/kotsadm/objects/scripts/export-minio-data.sh
```diff
@@ -0,0 +1,60 @@
...
+   [ -z $MINIO_ACCESS_KEY ] ||
+   [ -z $MINIO_SECRET_KEY ] ||
+   [ -z $KOTSADM_MINIO_BUCKET_NAME ];
+then
+    echo 's3 migration environment variables not set'
+    exit 1
+fi
+
+# change the working directory to the migration directory
+cd $KOTSADM_MINIO_MIGRATION_WORK_DIR
```
sgalsaleh, line 28: IMO this directory should be cleaned up in case there is left over data from a previously failed migration. Maybe that should happen in the first init container before copying the `mc` binary.

pkg/kotsadm/minio.go
cbodonnell: Hmm, could maybe leverage the task logic to keep track of the status? I'll think about this more

pkg/kotsadm/minio.go
cbodonnell: The check does not seem to be working as expected, so I guess I'll have to dig in more to understand why.

pkg/kotsadm/objects/minio_objects.go
cbodonnell: yup, that's what it was before the additional pvc back-and-forth.  changing it back...

pkg/kotsadm/objects/minio_objects.go
cbodonnell: same as above :)

pkg/kotsadm/minio.go
cbodonnell: Minio StatefulSet is "Ready" and has the updated image, but the pods have not restarted:
Screen Shot 2023-02-21 at 11 32 04 AM

Full loom (this occurs at 10s): https://www.loom.com/share/cb7c89a59a2e45eb8704177943ddfb88

pkg/kotsadm/main.go
cbodonnell: no longer relevant.  was able to fix `WaitForStatefulSetReady` by changing to:
```
if s.Status.ObservedGeneration == s.ObjectMeta.Generation &&
	s.Status.ReadyReplicas == *s.Spec.Replicas &&
	s.Status.UpdateRevision == s.Status.CurrentRevision {
	return nil
}
```

pkg/kotsadm/minio.go
cbodonnell: resolved: https://github.com/replicatedhq/kots/pull/3624#discussion_r1113429594

pkg/kotsadm/minio.go
cbodonnell: couldn't utilize existing store logic since it depends on kotsadm variables like `util.PodNamespace`, so creating/updating tasks from cli-only would require changes.  instead opted for a simpler implementation with a "status" configmap that could also aid in debugging.

.github/workflows/build-test.yaml
```diff
@@ -1517,9 +1536,109 @@ jobs:
...
+          set +e
+          if [ ! -d ./admin-console ]; then
+            echo "admin-console directory does not exist"
+            exit -1
+          fi
+          if [ -z "$(ls -A ./admin-console)" ]; then
+            echo "admin-console directory is empty"
+            exit -1
+          fi
+      - name: install k3s
```
sgalsaleh, line 1573: could you please add new lines before those steps? it's a bit hard to read.

.github/workflows/build-test.yaml
```diff
@@ -1517,9 +1536,109 @@ jobs:
...
+        uses: replicatedhq/action-k3s@main
+        with:
+          version: ${{ matrix.k8s_version }}
+          arguments: ${{ startsWith(matrix.k8s_version, 'v1.22') && '--egress-selector-mode disabled' || '' }}
+      - name: download base kots version
```
sgalsaleh, line 1578: i thought we weren't doing an upgrade scenario?

pkg/kotsadm/objects/scripts/import-minio-data.sh
```diff
@@ -0,0 +1,88 @@
...
+echo "minio stopped"
+
+echo "replacing old data with new data"
+shopt -s dotglob
+mkdir -p $KOTSADM_MINIO_MIGRATION_DIR/old-data
+mv -v /export/* $KOTSADM_MINIO_MIGRATION_DIR/old-data/
+mv -v $KOTSADM_MINIO_MIGRATION_DIR/new-data/* /export/
+
+echo "adding migration complete marker"
+touch /export/.migration
```
sgalsaleh, line 86: how about writing a timestamp to this file?

sgalsaleh (commented):
I'd like to see a test that installs a version of KOTS (e.g. 1.94.2) where MinIO uses the FS backend, and then gets upgraded / migrated using the generated KOTS binary in the PR.

pkg/kotsadm/minio.go
```diff
@@ -327,3 +357,112 @@ func cleanUpMigrationArtifact(clientset kubernetes.Interface, namespace string)
...
+	}
+
+	defer func() {
+		if finalErr == nil {
+			isMinioXlMigrationRunning, err := IsMinioXlMigrationRunning(ctx, clientset, deployOptions.Namespace)
+			if err != nil {
+				logger.Error(errors.Wrap(err, "failed to check if minio xl migration is running"))
+			}
+
+			if isMinioXlMigrationRunning {
```
sgalsaleh, line 417: is this check even needed? can't we just mark as complete regardless of its current status?

pkg/k8sutil/statefulset.go
```diff
@@ -23,7 +23,9 @@ func WaitForStatefulSetReady(ctx context.Context, clientset kubernetes.Interface
 			return nil
 		}
 
-		if s.Status.ObservedGeneration == s.ObjectMeta.Generation && s.Status.ReadyReplicas == *s.Spec.Replicas {
+		if s.Status.ObservedGeneration == s.ObjectMeta.Generation &&
+			s.Status.ReadyReplicas == *s.Spec.Replicas &&
+			s.Status.UpdateRevision == s.Status.CurrentRevision {
```
sgalsaleh, line 28: should we make this change for deployments as well? e.g. the `WaitForDeploymentReady` function.

pkg/kotsadm/objects/scripts/import-minio-data.sh
sgalsaleh:
```suggestion
echo "migration ran at: $(date)" > /export/.migration
```

.github/workflows/build-test.yaml
cbodonnell: `generate-manifests` can perform the minio upgrade if there is a k8s context

pkg/k8sutil/statefulset.go
cbodonnell: looks like it's only for statefulsets: https://github.com/kubernetes/api/blob/v0.26.1/apps/v1/types.go#L477

cbodonnell:
> I'd like to see a test that installs a version of KOTS (e.g. 1.94.2) where MinIO uses the FS backend, and then gets upgraded / migrated using the generated KOTS binary in the PR.

going to rename postgres-to-rqlite test since that does just this

cmd/kots/cli/admin-console-generate-manifests.go
```diff
@@ -77,5 +100,7 @@ func AdminGenerateManifestsCmd() *cobra.Command {
 	cmd.Flags().Bool("minimal-rbac", false, "set to true to use the namespaced role and bindings instead of cluster-level permissions")
 	cmd.Flags().StringSlice("additional-namespaces", []string{}, "Comma separate list to specify additional namespace(s) managed by KOTS outside where it is to be deployed. Ignored without with '--minimal-rbac=true'")
 
+	registryFlags(cmd.Flags())
```
cbodonnell, line 103: added registry flags to `generate-manifests` so that we can test it with ttl images

pkg/kotsadm/objects/minio_objects.go
cbodonnell: likely fine and an OOM situation would be recoverable since it would occur before any data is moved in/out of the PVC, so the memory limit could be patched and the migration would be re-run

cmd/kots/cli/admin-console-generate-manifests.go
```diff
@@ -39,8 +50,20 @@ func AdminGenerateManifestsCmd() *cobra.Command {
...
 				NoProxyEnvValue:      v.GetString("no-proxy"),
 				IncludeMinio:         v.GetBool("with-minio"),
+				MigrateToMinioXl:     migrateToMinioXl,
+				CurrentMinioImage:    currentMinioImage,
 				IsMinimalRBAC:        v.GetBool("minimal-rbac"),
 				AdditionalNamespaces: v.GetStringSlice("additional-namespaces"),
+				IsOpenShift:          isOpenShift,
+				IsGKEAutopilot:       isGKEAutopilot,
+
+				RegistryConfig: kotsadmtypes.RegistryConfig{
```
sgalsaleh, line 60: you need to use the `getRegistryConfig` function which has specific logic for reading the registry flags. (check the other commands for reference)

pkg/kotsadm/minio.go
```diff
@@ -21,10 +23,25 @@ import (
...
 func getMinioYAML(deployOptions types.DeployOptions) (map[string][]byte, error) {
 	docs := map[string][]byte{}
 	s := json.NewYAMLSerializer(json.DefaultMetaFactory, scheme.Scheme, scheme.Scheme)
 
+	if deployOptions.MigrateToMinioXl {
+		var configMap bytes.Buffer
+		if err := s.Encode(kotsadmobjects.MinioXlMigrationScriptsConfigMap(deployOptions.Namespace), &configMap); err != nil {
+			return nil, errors.Wrap(err, "failed to marshal minio migration configmap")
+		}
+		docs["minio-xl-migration-configmap.yaml"] = configMap.Bytes()
```
sgalsaleh, line 42: do you not need the status configmap as well?

cmd/kots/cli/admin-console-generate-manifests.go
cbodonnell: appears that function requires a kubernetes context.  thinking we can set defaults using the flag, and if we get a context, then use that function to update.

pkg/kotsadm/minio.go
cbodonnell: that configmap is created, checked, and updated during `admin-cosole upgrade` only to prevent re-runs of the command from editing an in-progress or failed migration and removing initcontainers.  generated manifests won't have that same cli logic running.  though i suppose the user could re-generate and re-apply manifests during a migration to mess things up?

pkg/kotsadm/objects/scripts/copy-minio-client.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# validate environment variables
+if [ -z $KOTSADM_MINIO_MIGRATION_DIR ]; then
+    echo 'KOTSADM_MINIO_MIGRATION_DIR not set'
+    exit 1
+fi
+
+# clean the migration directory
+shopt -s dotglob
+rm -rfv $KOTSADM_MINIO_MIGRATION_DIR/*
```
sgalsaleh, line 24: hmm, I know i suggested cleaning up this directory earlier, but consider this scenario:

during the migration, the old data is copied out of the pvc successfully, then, copying the new data into the pvc fails. at that point, there is a copy of the old data in the migration emptydir volume, and since kubelet will only restart the container and not the pod, i _believe_ the emptydir will still contain that data after containers restarts, which is valuable IMO and could be utilized to prevent data loss in some cases. what happens if `mc mirror` is called again and the pvc does not contain any data, but there is data in the emptydir volume? does it delete the data from the emptydir volume to match the pvc? or does it only copy missing data?

cbodonnell: i'll have to check and see.  the old minio would be running with a fresh and empty data dir in that case, so ideally it would not remove anything from the destination.  my guess would be it does not be default given there is a `--remove` flag:
```
NAME:                                                                                                            
  mc mirror - synchronize object(s) to a remote site                                                             
                                                                                                                 
USAGE:                                                                                                           
  mc mirror [FLAGS] SOURCE TARGET                                                                                
                                                                                                                 
FLAGS:                                                                                                           
...                             
  --remove                           remove extraneous object(s) on target 
```

.github/workflows/build-test.yaml
```diff
@@ -1532,9 +1551,213 @@ jobs:
...
+      - name: apply the generated manifests
+        run: |
+          set +e
+          kubectl apply -f ./admin-console -n $NAMESPACE
+
+      - name: wait for the kotsadm-minio-0 pod to be created
+        run: |
+          set +e
+          COUNTER=1
+          while [ -z "$(kubectl get pods -n $NAMESPACE | grep kotsadm-minio-0)" ]; do
```
sgalsaleh, line 1628: nit pick, it can be written this way
```suggestion
          while ! kubectl get pods -n $NAMESPACE | grep -q kotsadm-minio-0; do
```

.github/workflows/build-test.yaml
```diff
@@ -1532,9 +1551,213 @@ jobs:
...
+          ./bin/kots admin-console generate-manifests -n $NAMESPACE \
+            --shared-password password \
+            --kotsadm-registry ttl.sh \
+            --kotsadm-namespace automated-${{ github.run_id }} \
+            --kotsadm-tag 24h
+      
+      - name: validate that ./admin-console/minio-statefulset.yaml has initContainers since a migration is needed
+        run: |
+          set +e
+          if [ -z "$(grep -E 'initContainers' ./admin-console/minio-statefulset.yaml)" ]; then
```
sgalsaleh, line 1659: ditto
```suggestion
          if grep -E 'initContainers' ./admin-console/minio-statefulset.yaml; then
```

.github/workflows/build-test.yaml
```diff
@@ -1532,9 +1551,213 @@ jobs:
...
+            echo "admin-console/minio-statefulset.yaml does not have initContainers"
+            exit -1
+          fi
+      
+      - name: apply the generated manifests
+        run: |
+          set +e
+          kubectl apply -f ./admin-console -n $NAMESPACE
+
+      - name: wait for the kotsadm-minio-0 pod to be created
```
sgalsaleh, line 1669: wouldn't the minio pod already exist from the previous installation?

.github/workflows/build-test.yaml
```diff
@@ -1532,9 +1551,213 @@ jobs:
...
+          ./bin/kots admin-console generate-manifests -n $NAMESPACE \
+            --shared-password password \
+            --kotsadm-registry ttl.sh \
+            --kotsadm-namespace automated-${{ github.run_id }} \
+            --kotsadm-tag 24h
+      
+      - name: validate that ./admin-console/minio-statefulset.yaml does not have initContainers since a migration is not needed
+        run: |
+          set +e
+          if [ -n "$(grep -E 'initContainers' ./admin-console/minio-statefulset.yaml)" ]; then
```
sgalsaleh, line 1705:
```suggestion
          if grep -E 'initContainers' ./admin-console/minio-statefulset.yaml; then
```

pkg/kotsadm/objects/scripts/copy-minio-client.sh
sgalsaleh: nice, i hope that's the case

pkg/kotsadm/minio.go
sgalsaleh: ah good point. i don't think we need to worry about that edge case. thanks for the explanation!

.github/workflows/build-test.yaml
```diff
@@ -1532,9 +1551,213 @@ jobs:
...
+      - name: run kots admin-console generate-manifests using new kots version with k8s context
+        run: |
+          set +e
+          ./bin/kots admin-console generate-manifests -n $NAMESPACE \
+            --shared-password password \
+            --kotsadm-registry ttl.sh \
+            --kotsadm-namespace automated-${{ github.run_id }} \
+            --kotsadm-tag 24h
+      
+      - name: validate that ./admin-console/minio-statefulset.yaml does not have initContainers since a migration is not needed
```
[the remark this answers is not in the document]
cbodonnell: yes, if the existing statefulset image is a newer one, the statefulset will be updated to remove those initcontainers.  this update wouldn't happen during `admin-console upgrade` if the status is still `"running"` though

.github/workflows/build-test.yaml
cbodonnell: yup, just need to wait for it to be ready rather than created.  will remove this one and the one below

pkg/kotsadm/configmaps.go
```diff
@@ -174,3 +174,55 @@ func configMapFromFile(deployOptions types.DeployOptions, configMapName string,
...
+	_, err = clientset.CoreV1().ConfigMaps(namespace).Update(context.TODO(), existingConfigMap, metav1.UpdateOptions{})
+	if err != nil {
+		return errors.Wrap(err, "failed to update minio xl migration scripts configmap")
+	}
+
+	return nil
+}
+
+func ensureMinioXlMigrationStatusConfigmap(namespace string, clientset kubernetes.Interface) error {
+	_, err := clientset.CoreV1().ConfigMaps(namespace).Get(context.TODO(), "kotsadm-minio-xl-migration-status", metav1.GetOptions{})
```
cbodonnell, line 206: set the names of these migration configmaps as constants

pkg/kotsadm/objects/scripts/copy-minio-client.sh
cbodonnell: old minio would not have a bucket, so the export couldn't even occur.  this will be a no-op

cmd/kots/cli/admin-console-generate-manifests.go
```diff
@@ -31,6 +34,26 @@ func AdminGenerateManifestsCmd() *cobra.Command {
...
+				Username:          v.GetString("registry-username"),
+				Password:          v.GetString("registry-password"),
+			}
+			if clientset, err := k8sutil.GetClientset(); err == nil {
+				isOpenShift, isGKEAutopilot = k8sutil.IsOpenShift(clientset), k8sutil.IsGKEAutopilot(clientset)
+				migrateToMinioXl, currentMinioImage, _ = kotsadm.IsMinioXlMigrationNeeded(clientset, namespace)
+				if newRegistryConfig, err := getRegistryConfig(v, clientset); err == nil {
+					registryConfig = *newRegistryConfig
+				} else {
+					logger.Error(errors.Wrap(err, "failed to get registry config"))
```
sgalsaleh, line 53: how about moving the `log := logger.NewCLILogger(cmd.OutOrStdout())` below to be above this and just use the `log` variable that's already initialized?

cbodonnell: yup, good call

pkg/kotsadm/objects/scripts/export-minio-data.sh
```diff
@@ -0,0 +1,74 @@
...
+echo "stopping minio"
+kill $MINIO_PID
+
+# wait for minio to exit
+wait $MINIO_PID
+
+echo "minio stopped"
+
+# mark the export as complete
+echo "marking export as complete"
```
sgalsaleh, line 73: this should happen before stopping minio

pkg/kotsadm/objects/scripts/import-minio-data.sh
```diff
@@ -0,0 +1,88 @@
...
+# shutdown minio
+echo "stopping minio"
+kill $MINIO_PID
+
+# wait for minio to exit
+wait $MINIO_PID
+
+echo "minio stopped"
+
+echo "adding migration complete marker"
```
sgalsaleh, line 81: this should happen before stopping minio

pkg/kotsadm/objects/scripts/export-minio-data.sh
```diff
@@ -0,0 +1,74 @@
+#!/bin/bash
+
+set -e
+
+# This script exports bucket content from the old minio instance to the shared migration directory
+
+# check if the migration has already been completed
+if [ -f /export/.migration ];
```
sgalsaleh, line 8: is this check even necessary anymore?

sgalsaleh (approved):

PR #3627, opened 2023-02-14
create hooks needed for renderVersionAction in VersionHistoryPage

rianfowler:
- renames and moves the deploy version call to make it closer to bulletproof react 
- removes dead code related to `isRollback` in the `redeploy` button behavior 
- adds hooks for all of the calls made by `renderVersionAction`
- **None of the new hooks are integrated. We need to add regression testing for this button behavior before updating / refactoring it**

Fixes #
- https://app.shortcut.com/replicated/story/68569/refactor-version-history-api-calls-to-use-hooks

```release-note
none
```

none


web/src/components/apps/AppVersionHistory.tsx
```diff
@@ -796,27 +796,18 @@ class AppVersionHistory extends Component<Props, State> {
     }
   };
 
-  redeployVersion = (version: Version, isRollback = false) => {
```
rianfowler, line 799: this is never called with `true` as a second argument so I'm just getting rid of it. The whole rollback situation is pretty messy right now so we'll probably clean that up once I get the hooks integrated

web/src/features/AppVersionHistory/AppVersionHistoryRow.tsx
```diff
@@ -190,14 +190,20 @@ function AppVersionHistoryRow(props: Props) {
...
     } else if (version.needsKotsUpgrade) {
+      // postUpdateAdminConsole
       actionFn = props.upgradeAdminConsole;
     } else if (version.status === "pending_download") {
+      // postDownloadVersion
       actionFn = props.downloadVersion;
     } else if (version.status === "failed" || version.status === "deployed") {
+      // postRedeployVersion
       actionFn = props.redeployVersion;
     }
```
rianfowler, line 208: going to get rid of all of these props and just hit the hooks that i created. documenting what hooks these will use for future reference.

web/src/features/AppVersionHistory/api/getVersions.tsx
```diff
@@ -5,17 +5,23 @@ import { useParams } from "react-router-dom";
 import { useSelectedApp } from "@features/App";
 import { useMetadata } from "@src/stores";
 import { useIsHelmManaged } from "@src/components/hooks";
+import { App, KotsParams, Metadata, Version } from "@types";
```
rianfowler, line 8: nothing uses this hook yet but I saw some things I wanted to clean up. make it a typed component and got rid using default values to pass in imports.

miaawong (approved):

web/src/components/PreflightResultPage.tsx
```diff
@@ -14,12 +14,13 @@ import PreflightsProgress from "./troubleshoot/PreflightsProgress";
 import Icon from "./Icon";
 
 import {
-  useDeployKotsDownsteam,
   useGetPrelightResults,
   useIgnorePermissionErrors,
   useRerunPreflights,
 } from "@features/PreflightChecks/api";
 
+import { useDeployAppVersion } from "@features/App/api";
```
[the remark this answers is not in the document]
rianfowler: found out we can stop using `version.parentSequence` too and just use `version.sequence`. I added that to the next ticket. I suspect there's other cruft we can get rid of if it doesn't make sense

PR #3633, opened 2023-02-15
fix rerun preflight button logic

rianfowler:
- re-run button is not showing up during failed preflights check in airgapped flow 

Fixes #
https://app.shortcut.com/replicated/story/69593/fix-rerun-button-on-airgapped-preflights-flow

```release-note
- kots/web: fixes and issue where Re-run preflights button would not render during failed preflights in airgapped installation flow 
```

none

### DIFF
--- a/web/src/components/PreflightResultPage.tsx
+++ b/web/src/components/PreflightResultPage.tsx
@@ -219,7 +219,7 @@ function PreflightResultPage(props: Props) {
           )}
         </div>
       )}
-      {!props.fromLicenseFlow && preflightCheck?.showPreflightCheckPending && (
+      {!props.fromLicenseFlow && preflightCheck?.shouldShowRerunPreflight && (
         <div className="flex-auto flex justifyContent--flexEnd u-marginBottom--15">
           <button
             type="button"
--- a/web/src/features/PreflightChecks/api/getPreflightResult.tsx
+++ b/web/src/features/PreflightChecks/api/getPreflightResult.tsx
@@ -136,7 +136,7 @@ function flattenPreflightResponse({
       !response?.preflightResult?.skipped && // not skipped
       (hasFailureOrWarning(response) || hasPreflightErrors(response)), // or it has errors
     shouldShowRerunPreflight:
-      !hasRunningPreflightChecks(response) && // not running
+      !hasRunningPreflightChecks(response) || // not running
       !response?.preflightResult?.skipped, // not skipped
     showDeploymentBlocked:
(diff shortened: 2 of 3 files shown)


web/src/components/PreflightResultPage.tsx
```diff
@@ -219,7 +219,7 @@ function PreflightResultPage(props: Props) {
           )}
         </div>
       )}
-      {!props.fromLicenseFlow && preflightCheck?.showPreflightCheckPending && (
+      {!props.fromLicenseFlow && preflightCheck?.shouldShowRerunPreflight && (
```
rianfowler, line 222: Previously this was a giant ternary. Decided to make an explicit property from the response for this. it's a `shouldShow` because we have to combine the BE state from the response with this `fromLicenseFlow` prop. I think we can get `fromLicenseFlow` from the BE but I'm not sure what endpoint is coming from yet.

so basically the logic was 
`!this.props.fromLicenseFlow && stopPolling` 

- the `stopPolling` part been moved into the selector as `shouldShowRerunPreflight` 
- we're still checking `!this.props.fromLicenseFlow` here 

Screenshot 2023-02-15 at 16 10 00

web/src/features/PreflightChecks/api/getPreflightResult.tsx
```diff
@@ -135,6 +135,9 @@ function flattenPreflightResponse({
     shouldShowConfirmContinueWithFailedPreflights:
       !response?.preflightResult?.skipped && // not skipped
       (hasFailureOrWarning(response) || hasPreflightErrors(response)), // or it has errors
+    shouldShowRerunPreflight:
+      Boolean(response?.preflightResult?.result) || // not running
+      response?.preflightResult?.skipped, // not skipped
```
rianfowler, line 140: so the original logic was 
`!this.props.fromLicenseFlow && stopPolling`. (see screenshot in previous comment - it's a big ternary) 

- `fromLicenseFlow` is still used for now in the component above on 222
- `stopPolling` looked like this: 
https://github.com/replicatedhq/kots/blob/af7db8652d2e24b4022b0eef939c6963fbd7712d/web/src/components/PreflightResultPage.tsx#L377

- `preflightResultData?.result` was basically checking if the the preflight result had any data. we do the same thing here. 

- we're just pulling `skipped` right off the response and ORing it like the old `stopPolling` worked
https://github.com/replicatedhq/kots/blob/af7db8652d2e24b4022b0eef939c6963fbd7712d/web/src/components/PreflightResultPage.tsx#L376

miaawong (approved):

PR #3636, opened 2023-02-16
Add daemonset status informer

ousbots:

Add a status informer for k8s daemonsets so the state of daemonsets is taken into account when displaying the state of an app deployment. Includes a cli test to make sure the daemon set, deployment, and stateful set status informers become ready.

For [SC-60393](https://app.shortcut.com/replicated/story/60393/status-informers-should-support-daemon-sets)

Fixes https://github.com/replicated-collab/h2o-replicated/issues/48

Note that without keeping the previous state of the daemonset, the updating state won't be reported correctly. This is because the observed generation is incremented to match the meta generation before pods finish deploying. So without knowing that the previous state was updating, during an update the observed generation will be less then the meta generation, the status of updating is reported, the observed generation is increased and on the next status informer check, the generations are the same, but the number of ready pods will be less than the desired scheduled and the state will be degraded and not updating. This also happens with the deployment and statefulset status informers.

## Steps to reproduce

Create an app release with the k8s Kind: DaemonSet and in the app manifest make sure there is the following lines:

```
StatusInformers:
  - daemonset/<daemonset name>
```

Where the daemonset name comes from the metadata.name in the k8s daemonset manifest.

This can be done with the example app release by changing the `Kind` in example-deployment.yaml to `DaemonSet` and in kots-app.yaml changing the status informers annotation from `  - deployment/nginx` to `  - daemonset/nginx`.

```release-note
Adds the ability to use status informers for daemonsets
```

https://github.com/replicatedhq/replicated-docs/pull/962

### DIFF
--- a/pkg/appstate/daemonsets.go
+++ b/pkg/appstate/daemonsets.go
@@ -2,7 +2,6 @@ package appstate
 
 import (
 	"context"
-	"fmt"
 	"log"
 	"time"
 
@@ -10,6 +9,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	labels "k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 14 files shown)


pkg/appstate/daemonsets.go
```diff
@@ -0,0 +1,180 @@
...
+// The pods in a daemonset can be identified by the match label set in the daemonset and the
+// "controller-revision-hash" can be used to determine if they are all the in the same daemonset
+// version.
+func (h *daemonSetEventHandler) calculateDaemonSetState(r *appsv1.DaemonSet) types.State {
+	if r == nil {
+		return types.StateUnavailable
+	}
+
+	selector := ""
+	for key, value := range r.Spec.Selector.MatchLabels {
```
sgalsaleh, line 110: you can just do this:
```go
selector  := labels.SelectorFromSet(r.Spec.Selector.MatchLabels).String()
```

pkg/appstate/daemonsets.go
```diff
@@ -0,0 +1,180 @@
...
+		if len(selector) > 0 {
+			selector += ","
+		}
+		selector += fmt.Sprintf("%s=%s", key, value)
+	}
+
+	// Gather all pods by looping until server responds with an empty continue field.
```
sgalsaleh, line 117: we should check that observed generation is equal to generation first

pkg/appstate/daemonsets.go
```diff
@@ -0,0 +1,180 @@
...
+		if len(result.Continue) == 0 {
+			break
+		}
+	}
+
+	// If the pod version labels are not all the same, then the daemonset is updating.
+	currentVersion := ""
+	for _, pod := range pods {
+		version, exists := pod.Labels[DaemonSetPodVersionLabel]
+		if !exists {
```
sgalsaleh, line 136: does this label exist in older kubernetes versions like 1.19?

pkg/appstate/daemonsets.go
```diff
@@ -0,0 +1,180 @@
...
+			log.Printf("failed to find %s label for pod %s", DaemonSetPodVersionLabel, pod.Name)
+			return types.StateUnavailable
+		}
+
+		if len(currentVersion) == 0 {
+			currentVersion = version
+		} else {
+			if version != currentVersion {
```
sgalsaleh, line 144: nit pick: can't this be combined with the `else` above to be `else if version != currentVersion`?

pkg/appstate/daemonsets.go
ousbots: it looks like [this is the commit](https://github.com/kubernetes/kubernetes/commit/25a67fc5cdac9c3b41bd0c7fdfb0a2c24a925894) that introduced the "controller-revision-hash" annotation to daemon sets in 2018 and the [first release](https://github.com/kubernetes/kubernetes/releases/tag/v1.19.0) of k8s v1.19 was in august 2020, so I'm pretty sure it's in all the versions we support.

pkg/appstate/daemonsets.go
```diff
@@ -0,0 +1,168 @@
...
+// "controller-revision-hash" can be used to determine if they are all the in the same daemonset
+// version.
+func (h *daemonSetEventHandler) calculateDaemonSetState(r *appsv1.DaemonSet) types.State {
+	if r == nil {
+		return types.StateUnavailable
+	}
+
+	if r.Status.ObservedGeneration != r.ObjectMeta.Generation {
+		return types.StateUpdating
+	} else {
```
sgalsaleh, line 112: this else is not needed

sgalsaleh (approved):

pkg/appstate/daemonsets.go
```diff
@@ -0,0 +1,182 @@
...
+		validOwner := false
+		for _, owner := range pod.ObjectMeta.OwnerReferences {
+			if owner.Kind == DaemonSetOwnerKind && owner.Name == r.ObjectMeta.Name {
+				validOwner = true
+				break
+			}
+		}
+
+		if !validOwner {
+			log.Printf("skipping pod %s due to invalid owner references for %s", pod.ObjectMeta.Name, r.ObjectMeta.Name)
```
sgalsaleh, line 135:
```suggestion
			log.Printf("skipping pod %s due to invalid owner references for daemonset %s", pod.ObjectMeta.Name, r.ObjectMeta.Name)
```

sgalsaleh (approved):

PR #3667, opened 2023-02-21
Change stateful set status informer updating calculation

ousbots:

Change the status informer for stateful sets to use a comparison between the controller-revision-hash of the pods in the set to determine the updating status. Similar to how the daemon set is done.

Fixes https://github.com/replicated-collab/h2o-replicated/issues/48

Note there are two edge cases with this method:
* When the last pod is restarting, the status changes from updating to degraded and then ready once the last pod is ready.
* If an update is deployed and pods that should be restarted do not go down and new pods do not come up, then this will always report being ready instead of upgrading or degraded.

## Steps to reproduce

Create an app release with the k8s Kind: StatefulSet and in the app manifest make sure there is the following line:

```
StatusInformers:
  - statefulset/<statefulset name>
```

Where the statefulset name comes from the metadata.name in the k8s daemonset manifest.

This can be done with the example app release by changing the `Kind` in example-deployment.yaml to `StatefulSet` and in kots-app.yaml changing the status informers annotation from `  - deployment/nginx` to `  - statefulset/nginx`.

```release-note
NONE
```

### DIFF
--- a/pkg/appstate/statefulsets.go
+++ b/pkg/appstate/statefulsets.go
@@ -116,29 +116,29 @@ func (h *statefulSetEventHandler) calculateStatefulSetState(r *appsv1.StatefulSe
 
 	if r.Status.ObservedGeneration != r.ObjectMeta.Generation {
 		return types.StateUpdating
-	} else {
-		listOptions := metav1.ListOptions{LabelSelector: labels.SelectorFromSet(r.Spec.Selector.MatchLabels).String()}
+	}
 
-		pods, err := h.clientset.CoreV1().Pods(h.targetNamespace).List(context.TODO(), listOptions)
-		if err != nil {
-			log.Printf("failed to get daemonset pod list: %s", err)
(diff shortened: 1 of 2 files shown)


pkg/appstate/statefulsets.go
```diff
@@ -109,21 +110,66 @@ func makeStatefulSetResourceState(r *appsv1.StatefulSet, state types.State) type
...
+		validOwner := false
+		for _, owner := range pod.ObjectMeta.OwnerReferences {
+			if owner.Kind == StatefulSetOwnerKind && owner.Name == r.ObjectMeta.Name {
+				validOwner = true
+				break
+			}
+		}
+
+		if !validOwner {
+			log.Printf("skipping pod %s due to invalid owner reference for %s", pod.ObjectMeta.Name, r.ObjectMeta.Name)
```
sgalsaleh, line 142:
```suggestion
			log.Printf("skipping pod %s due to invalid owner reference for statefulset %s", pod.ObjectMeta.Name, r.ObjectMeta.Name)
```

sgalsaleh (dismissed):

sgalsaleh (approved):

PR #3674, opened 2023-02-22
feat(support-bundle): split support bundle secrets into multiple specs

DexterYan:
- split default `kotsadm-<appslug>-supportbundle` into three seperate specs
- create `kotsadm-<appslug>-supportbundle-cluster-specific` which use addDiscoveredSpecs function to add all existed collectors and analyzers not defined by kots
- create `kotsadm-<appslug>-supportbundle-default` which only contains default collectors and analyzers which will be replaced by newly introduced uri
- create `kotsadm-<appslug>-supportbundle-vendor` which only contains the vendor application collectors and analyzers
- add mergedBundle to return merged specs from multiples specs
- add deduplicatedAfterCollection to remove duplicated AfterCollection
- add --load-cluster-specs instead of secret uri

Shortcut # 
[sc-65663](https://app.shortcut.com/replicated/story/65663/troubleshoot-split-the-support-bundle-secrets-into-multiple-specs)

```release-note
NONE
```

NONE

### DIFF
--- a/pkg/supportbundle/spec.go
+++ b/pkg/supportbundle/spec.go
@@ -114,86 +114,9 @@ func CreateRenderedSpec(app apptypes.AppType, sequence int64, kotsKinds *kotsuti
 	}
 
 	for key, builtBundle := range builtBundles {
-		s := serializer.NewYAMLSerializer(serializer.DefaultMetaFactory, scheme.Scheme, scheme.Scheme)
-		var b bytes.Buffer
-		if err := s.Encode(builtBundle, &b); err != nil {
-			return nil, errors.Wrap(err, "failed to encode support bundle")
-		}
-
-		templatedSpec := b.Bytes()
-
-		renderedSpec, err := helper.RenderAppFile(app, &sequence, templatedSpec, kotsKinds, util.PodNamespace)
+		err := createSupportBundleSpecSecret(app, sequence, kotsKinds, key, builtBundle, opts, clientset)
 		if err != nil {
-			return nil, errors.Wrap(err, "failed render support bundle spec")
-		}
-
-		// unmarshal the spec, look for image replacements in collectors and then remarshal
-		// we do this after template rendering to support templating and then replacement
-		supportBundle, err := kotsutil.LoadSupportBundleFromContents(renderedSpec)
-		if err != nil {
-			return nil, errors.Wrap(err, "failed to unmarshal rendered support bundle spec")
-		}
-
-		var registrySettings registrytypes.RegistrySettings
-		if !util.IsHelmManaged() {
-			s, err := store.GetStore().GetRegistryDetailsForApp(app.GetID())
-			if err != nil {
-				return nil, errors.Wrap(err, "failed to get registry settings for app")
-			}
-			registrySettings = s
-		}
-
-		collectors, err := registry.UpdateCollectorSpecsWithRegistryData(supportBundle.Spec.Collectors, registrySettings, kotsKinds.Installation.Spec.KnownImages, kotsKinds.License, &kotsKinds.KotsApplication)
-		if err != nil {
-			return nil, errors.Wrap(err, "failed to update collectors")
-		}
-		supportBundle.Spec.Collectors = collectors
-		b.Reset()
-		if err := s.Encode(supportBundle, &b); err != nil {
-			return nil, errors.Wrap(err, "failed to encode support bundle")
-		}
-		renderedSpec = b.Bytes()
-		secretName := GetSpecSecretName(app.GetSlug()) + "-" + key
-
-		existingSecret, err := clientset.CoreV1().Secrets(util.PodNamespace).Get(context.TODO(), secretName, metav1.GetOptions{})
-		labels := kotstypes.MergeLabels(kotstypes.GetKotsadmLabels(), kotstypes.GetTroubleshootLabels())
-		if err != nil {
-			if kuberneteserrors.IsNotFound(err) {
-				secret := &corev1.Secret{
-					TypeMeta: metav1.TypeMeta{
-						APIVersion: "v1",
-						Kind:       "Secret",
-					},
-					ObjectMeta: metav1.ObjectMeta{
-						Name:      secretName,
-						Namespace: util.PodNamespace,
-						Labels:    labels,
-					},
-					Data: map[string][]byte{
-						SpecDataKey: renderedSpec,
-					},
-				}
-
-				_, err = clientset.CoreV1().Secrets(util.PodNamespace).Create(context.TODO(), secret, metav1.CreateOptions{})
-				if err != nil {
-					return nil, errors.Wrap(err, "failed to create support bundle secret")
-				}
-
-				logger.Debugf("created %q support bundle spec secret", secretName)
-			} else {
-				return nil, errors.Wrap(err, "failed to read support bundle secret")
-			}
-		} else {
-			if existingSecret.Data == nil {
-				existingSecret.Data = map[string][]byte{}
-			}
-			existingSecret.Data[SpecDataKey] = renderedSpec
-			existingSecret.ObjectMeta.Labels = labels
-
-			_, err = clientset.CoreV1().Secrets(util.PodNamespace).Update(context.TODO(), existingSecret, metav1.UpdateOptions{})
-			if err != nil {
-				return nil, errors.Wrap(err, "failed to update support bundle secret")
-			}
+			return nil, errors.Wrap(err, "failed to create support bundle secret")
 		}
 	}
 
@@ -201,6 +124,92 @@ func CreateRenderedSpec(app apptypes.AppType, sequence int64, kotsKinds *kotsuti
 	// as not to add discovered specs to the default support bundle spec secret.c
 	// use cluster specific support bundle spec to add Spec.AfterCollection as default
 	return addDiscoveredSpecs(builtBundles[kotstypes.ClusterSpecificSupportBundleSpecKey], app, clientset), nil
+}
+
+// createSupportBundleSpecSecret creates a secret containing the support bundle spec
+func createSupportBundleSpecSecret(app apptypes.AppType, sequence int64, kotsKinds *kotsutil.KotsKinds, key string, builtBundle *troubleshootv1beta2.SupportBundle, opts types.TroubleshootOptions, clientset kubernetes.Interface) error {
+	s := serializer.NewYAMLSerializer(serializer.DefaultMetaFactory, scheme.Scheme, scheme.Scheme)
+	var b bytes.Buffer
+	if err := s.Encode(builtBundle, &b); err != nil {
+		return errors.Wrap(err, "failed to encode support bundle")
+	}
+
+	templatedSpec := b.Bytes()
+
+	renderedSpec, err := helper.RenderAppFile(app, &sequence, templatedSpec, kotsKinds, util.PodNamespace)
+	if err != nil {
+		return errors.Wrap(err, "failed render support bundle spec")
+	}
+
+	// unmarshal the spec, look for image replacements in collectors and then remarshal
+	// we do this after template rendering to support templating and then replacement
+	supportBundle, err := kotsutil.LoadSupportBundleFromContents(renderedSpec)
+	if err != nil {
+		return errors.Wrap(err, "failed to unmarshal rendered support bundle spec")
+	}
+
+	var registrySettings registrytypes.RegistrySettings
+	if !util.IsHelmManaged() {
+		s, err := store.GetStore().GetRegistryDetailsForApp(app.GetID())
+		if err != nil {
+			return errors.Wrap(err, "failed to get registry settings for app")
+		}
+		registrySettings = s
+	}
+
+	collectors, err := registry.UpdateCollectorSpecsWithRegistryData(supportBundle.Spec.Collectors, registrySettings, kotsKinds.Installation.Spec.KnownImages, kotsKinds.License, &kotsKinds.KotsApplication)
+	if err != nil {
+		return errors.Wrap(err, "failed to update collectors")
+	}
+	supportBundle.Spec.Collectors = collectors
+	b.Reset()
+	if err := s.Encode(supportBundle, &b); err != nil {
+		return errors.Wrap(err, "failed to encode support bundle")
+	}
+	renderedSpec = b.Bytes()
+	secretName := GetSpecSecretName(app.GetSlug()) + "-" + key
+
+	existingSecret, err := clientset.CoreV1().Secrets(util.PodNamespace).Get(context.TODO(), secretName, metav1.GetOptions{})
+	labels := kotstypes.MergeLabels(kotstypes.GetKotsadmLabels(), kotstypes.GetTroubleshootLabels())
+	if err != nil {
+		if kuberneteserrors.IsNotFound(err) {
+			secret := &corev1.Secret{
+				TypeMeta: metav1.TypeMeta{
+					APIVersion: "v1",
+					Kind:       "Secret",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      secretName,
+					Namespace: util.PodNamespace,
+					Labels:    labels,
+				},
+				Data: map[string][]byte{
+					SpecDataKey: renderedSpec,
+				},
+			}
+
+			_, err = clientset.CoreV1().Secrets(util.PodNamespace).Create(context.TODO(), secret, metav1.CreateOptions{})
+			if err != nil {
+				return errors.Wrap(err, "failed to create support bundle secret")
+			}
+
+			logger.Debugf("created %q support bundle spec secret", secretName)
+		} else {
+			return errors.Wrap(err, "failed to read support bundle secret")
+		}
+	} else {
+		if existingSecret.Data == nil {
+			existingSecret.Data = map[string][]byte{}
+		}
+		existingSecret.Data[SpecDataKey] = renderedSpec
+		existingSecret.ObjectMeta.Labels = labels
+
+		_, err = clientset.CoreV1().Secrets(util.PodNamespace).Update(context.TODO(), existingSecret, metav1.UpdateOptions{})
+		if err != nil {
+			return errors.Wrap(err, "failed to update support bundle secret")
+		}
+	}
+	return nil
 }
 
 // addClusterSpecificSpec adds cluster specific and upload results URI to the support bundle
--- a/pkg/supportbundle/spec.go
+++ b/pkg/supportbundle/spec.go
@@ -10,6 +10,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"reflect"
 	"regexp"
 	"strings"
 	"time"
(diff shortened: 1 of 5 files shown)


pkg/supportbundle/spec.go
```diff
@@ -106,106 +106,200 @@ func CreateRenderedSpec(app apptypes.AppType, sequence int64, kotsKinds *kotsuti
...
-				Data: map[string][]byte{
-					SpecDataKey: renderedSpec,
-				},
+				_, err = clientset.CoreV1().Secrets(util.PodNamespace).Create(context.TODO(), secret, metav1.CreateOptions{})
+				if err != nil {
+					return nil, errors.Wrap(err, "failed to create support bundle secret")
+				}
+
+				logger.Debugf("created %q default support bundle spec secret", app.GetSlug())
+				// return supportBundle, nil
```
cbodonnell, line 186: This looks like it was a bug?  If so, let's just remove it.

pkg/supportbundle/spec.go
```diff
@@ -106,106 +106,200 @@ func CreateRenderedSpec(app apptypes.AppType, sequence int64, kotsKinds *kotsuti
...
+		}
+		supportBundle.Spec.Collectors = collectors
+		b.Reset()
+		if err := s.Encode(supportBundle, &b); err != nil {
+			return nil, errors.Wrap(err, "failed to encode support bundle")
+		}
+		renderedSpec = b.Bytes()
+		secretName := GetSpecSecretName(app.GetSlug())
+		if key != "kotsadm" {
+			secretName = GetSpecSecretName(app.GetSlug() + "-" + key)
```
cbodonnell, line 158: We might want to come up with a different naming scheme since there is a possibility (though likely a very low one) that there could be a naming conflict.  Example, a user has two apps with slugs `my-app` and `my-app-vendor`.

Maybe we can name the new split specs differently to avoid this?
```
kotsadm-<appslug>-supportbundle-cluster-specific
kotsadm-<appslug>-supportbundle-default
kotsadm-<appslug>-supportbundle-vendor
```

pkg/supportbundle/spec.go
```diff
@@ -106,106 +106,200 @@ func CreateRenderedSpec(app apptypes.AppType, sequence int64, kotsKinds *kotsuti
...
-		} else {
-			return nil, errors.Wrap(err, "failed to read support bundle secret")
 		}
 	}
 
-	if existingSecret.Data == nil {
-		existingSecret.Data = map[string][]byte{}
+	// Include discovered support bundle specs and default kotsadm support bundle spec. Perform this action here so
+	// as not to add discovered specs to the default support bundle spec secret.c
+	return addDiscoveredSpecs(builtBundles["kotsadm"], app, clientset), nil
```
cbodonnell, line 206: Can we use a constant for this key? Could do this for the split bundle keys as well, but this one makes the most sense since we use it in a few places.

pkg/supportbundle/spec.go
```diff
@@ -227,41 +322,19 @@ func injectDefaults(app apptypes.AppType, b *troubleshootv1beta2.SupportBundle,
...
-		uploadURL = fmt.Sprintf("%s/api/v1/troubleshoot/%s/%s", opts.Origin, app.GetID(), randomBundleID)
-		redactURL = fmt.Sprintf("%s/api/v1/troubleshoot/supportbundle/%s/redactions", opts.Origin, randomBundleID)
-	} else if opts.InCluster {
-		uploadURL = fmt.Sprintf("%s/api/v1/troubleshoot/%s/%s", fmt.Sprintf("http://kotsadm.%s.svc.cluster.local:3000", util.PodNamespace), app.GetID(), randomBundleID)
-		redactURL = fmt.Sprintf("%s/api/v1/troubleshoot/supportbundle/%s/redactions", fmt.Sprintf("http://kotsadm.%s.svc.cluster.local:3000", util.PodNamespace), randomBundleID)
-	} else {
-		uploadURL = fmt.Sprintf("%s/api/v1/troubleshoot/%s/%s", os.Getenv("API_ADVERTISE_ENDPOINT"), app.GetID(), randomBundleID)
-		redactURL = fmt.Sprintf("%s/api/v1/troubleshoot/supportbundle/%s/redactions", os.Getenv("API_ADVERTISE_ENDPOINT"), randomBundleID)
-	}
+	venderSupportBundle := supportBundle.DeepCopy()
```
cbodonnell, line 325:
```suggestion
	vendorSupportBundle := supportBundle.DeepCopy()
```

pkg/supportbundle/spec.go
```diff
@@ -227,41 +322,19 @@ func injectDefaults(app apptypes.AppType, b *troubleshootv1beta2.SupportBundle,
...
-				RedactURI: redactURL,
-			},
-		},
-	}
+	supportBundle = addClusterSpecificSpec(app, supportBundle, opts, namespacesToCollect, namespacesToAnalyze, imageName, pullSecret)
+	supportBundle = addDefaultSpec(app, supportBundle, opts, namespacesToCollect, namespacesToAnalyze, imageName, pullSecret)
 
-	return supportBundle, nil
+	return map[string]*troubleshootv1beta2.SupportBundle{
+		"vender":           venderSupportBundle,          //vendors' application support-bundle spec
```
cbodonnell, line 333:
```suggestion
		"vendor":           vendorSupportBundle,          //vendors' application support-bundle spec
```

pkg/supportbundle/spec.go
```diff
@@ -106,106 +106,200 @@ func CreateRenderedSpec(app apptypes.AppType, sequence int64, kotsKinds *kotsuti
...
+	supportBundle = deduplicatedCollectors(supportBundle)
+	supportBundle = deduplicatedAnalyzers(supportBundle)
+
+	return supportBundle
 }
 
-// injectDefaults injects the kotsadm default collectors/analyzers in the the support bundle specification.
-func injectDefaults(app apptypes.AppType, b *troubleshootv1beta2.SupportBundle, opts types.TroubleshootOptions, namespacesToCollect []string, namespacesToAnalyze []string) (*troubleshootv1beta2.SupportBundle, error) {
+// createClusterSpecificSupportBundle creates a support bundle spec with only cluster specific collectors, analyzers and upload result URI.
+func createClusterSpecificSupportBundle(app apptypes.AppType, b *troubleshootv1beta2.SupportBundle, opts types.TroubleshootOptions, namespacesToCollect []string, namespacesToAnalyze []string, imageName string, pullSecret *troubleshootv1beta2.ImagePullSecrets) *troubleshootv1beta2.SupportBundle {
```
cbodonnell, line 250: looks like this var isn't used
```suggestion
func createClusterSpecificSupportBundle(app apptypes.AppType, opts types.TroubleshootOptions, namespacesToCollect []string, namespacesToAnalyze []string, imageName string, pullSecret *troubleshootv1beta2.ImagePullSecrets) *troubleshootv1beta2.SupportBundle {
```

pkg/supportbundle/spec.go
```diff
@@ -106,106 +106,200 @@ func CreateRenderedSpec(app apptypes.AppType, sequence int64, kotsKinds *kotsuti
...
+	supportBundle = addDefaultTroubleshoot(supportBundle, imageName, pullSecret)
+	supportBundle = populateNamespaces(supportBundle, namespacesToCollect, namespacesToAnalyze)
+	supportBundle = deduplicatedCollectors(supportBundle)
+	supportBundle = deduplicatedAnalyzers(supportBundle)
+
+	return supportBundle
+}
+
+// createDefaultSupportBundle creates a support bundle spec with only default collectors and analyzers.
+func createDefaultSupportBundle(app apptypes.AppType, b *troubleshootv1beta2.SupportBundle, opts types.TroubleshootOptions, namespacesToCollect []string, namespacesToAnalyze []string, imageName string, pullSecret *troubleshootv1beta2.ImagePullSecrets) *troubleshootv1beta2.SupportBundle {
```
cbodonnell, line 279: same as above
```suggestion
func createDefaultSupportBundle(app apptypes.AppType, opts types.TroubleshootOptions, namespacesToCollect []string, namespacesToAnalyze []string, imageName string, pullSecret *troubleshootv1beta2.ImagePullSecrets) *troubleshootv1beta2.SupportBundle {
```

pkg/supportbundle/spec.go
```diff
@@ -106,106 +106,200 @@ func CreateRenderedSpec(app apptypes.AppType, sequence int64, kotsKinds *kotsuti
...
+				Method:    "PUT",
+				RedactURI: redactURL,
+			},
+		},
 	}
 
-	// Include discovered support bundle specs. Perform this action here so
-	// as not to add discovered specs to the default support bundle spec secret.
-	return addDiscoveredSpecs(supportBundle, app, clientset), nil
+	supportBundle = addDefaultDynamicTroubleshoot(supportBundle, app, imageName, pullSecret)
```
cbodonnell, line 241: This spec seems more like a `"dynamic-default"` rather than a `"cluster-specific"` since it's just the dynamic support bundle spec for this particular instance+app.

cbodonnell (commented):
General comment: Could splitting these out impact the uploaded results?  The `"cluster-specific"` support bundle has `.Spec.AfterCollection`, but the other split bundles do not.  Not sure how that might play out since we don't merge that field here in kots.

pkg/supportbundle/spec.go
DexterYan: Thank you for point it out. It is not a bug in previous code, however, once I switched to use create multiple specs, it will cause early return. I have removed it.

pkg/supportbundle/spec.go
DexterYan: good idea, I have changed the name to be as your suggested, and updated 
```
appNameRE = regexp.MustCompile(`^kotsadm-.*-supportbundle(?:$|.*)`)
```
to filter out all kotsadm objects that have an object name

pkg/supportbundle/spec.go
DexterYan: Agree, I have created those constants of those keys in [pkg/kotsadm/types/constants.go](https://github.com/replicatedhq/kots/pull/3674/files#diff-6e08d80aec35b42dd85376cf4f15fa335bcd4f4a5438e436432d9d0a0b460278)

pkg/supportbundle/spec.go
DexterYan: good catch! changed

pkg/supportbundle/spec.go
DexterYan: fixed

pkg/supportbundle/spec.go
DexterYan: removed

pkg/supportbundle/spec.go
DexterYan: removed

pkg/supportbundle/spec.go
DexterYan: when I was checking the content of dynamic default spec, I saw some collector had cluster specfic information, that's why I moved it into  cluster-specific spec. 
```
collectorName: app-info.json
      data: |-
        {
          "instance_id": "*****",
          "cluster_id": "******",
```

As your pointed out, it is for  kots instance+app, I think it is ok to move it back to default spec, does it sound good to you?

DexterYan:
> 

In current code, the injectDefaults will return builtBundles["kotsadm"] same as the origin one which has combined all those separated specs into one and return to CreateRenderedSpec. It means the we will still have `.Spec.AfterCollection` even we split the specs. However, I would suggest to change the new created specs names as follows:

```
<appslug>-supportbundle-cluster-specific
<appslug>-supportbundle-default
<appslug>-supportbundle-vendor
```
we have addDiscoveredSpecs to combine those specs, and in future, we can retire the origin one, kotsadm-<appslug>-supportbundle.

pkg/supportbundle/spec.go
cbodonnell: i think it's fine to move it back into the default for now

banjoh:
I recommend simplifying this code (in general, not your PR changes) because its hard to make sense of at the moment and non-trivial to maintain. I suggest we embed the split specs like we do the [default spec](https://github.com/replicatedhq/kots/tree/main/pkg/supportbundle/defaultspec). This way, all the code will do is load and store specs in secrets. We also get the opportunity to remove functions such as `removeKurlCollectors` & `removeKurlAnalyzers`.

At the moment, in order for someone to see the resulting spec that KOTS will use, they have to run KOTS to see what gets rendered. The spec is stored in a base64 encoded secret (me injecting a rant. Should we consider a `configmap` or a `stringData` secret?)

banjoh:
> and in future, we can retire the origin one, kotsadm--supportbundle.

To avoid having to come back to address this in future, I wonder whether there is a way to do the following
- split specs
- create/update them
- fetch them again to create the default spec
- add `.Spec.AfterCollection` and return the spec.

I wonder if `addDiscoveredSpecs()` can recreate the default support bundle from the split ones

banjoh:
> create kotsadm-<appslug>-supportbundle-default which only contains default collectors and analyzers which will be replaced by newly introduced uri

The [default spec](https://github.com/replicatedhq/kots/blob/main/pkg/supportbundle/defaultspec/spec.yaml) does not at the moment have a `uri`. The following things would need to be addressed before adding the `uri`
- Better handling of errors when downloading a spec fails. At the moment KOTS will [just crash and never start](https://github.com/replicatedhq/kots/blob/main/pkg/supportbundle/defaultspec/embed.go#L17) if say it was in an airgap environment.
- Only [collectors and analysers are extracted from the spec](https://github.com/replicatedhq/kots/blob/3ebbee86b35da8553c4c75f9a23ffedc31902d1d/pkg/supportbundle/spec.go#L565-L566)

DexterYan:
Thank you @banjoh!  I have created a few sub-tasks to follow those suggestions and will re-write this pr to simplify the code. Also, I agreed with you the current way of debugging support bundle spec secrets needs a lot of steps, I would prefer to use stringData secret. 

The `addDiscoveredSpecs()` can recreate the default support bundle, however, we need to modify the way of searching secrets by not just filtering out the kotsadm secret and also only including the secrets whose names contain the same appSlug.

pkg/supportbundle/spec.go
```diff
@@ -106,52 +106,69 @@ func CreateRenderedSpec(app apptypes.AppType, sequence int64, kotsKinds *kotsuti
...
+		err := createSupportBundleSpecSecret(app, sequence, kotsKinds, key, builtBundle, opts, clientset)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to create support bundle secret")
+		}
+	}
+
+	// Include discovered support bundle specs and multiple kotsadm support bundle specs. Perform this action here so
+	// as not to add discovered specs to the default support bundle spec secret.c
+	// use cluster specific support bundle spec to add Spec.AfterCollection as default
+	return addDiscoveredSpecs(builtBundles[kotstypes.ClusterSpecificSupportBundleSpecKey], app, clientset), nil
```
cbodonnell, line 126: This results in the generated bundle missing all of the collectors/analyzers from the "default" and "vendor" specs since those are filtered out by the `^kotsadm-.*-supportbundle$` regex in `addDiscoveredSpecs(...)`.

We could maybe do something like this (pseudo-code):
```
defaultBundle := builtBundles[kotstypes.DefaultSupportBundleSpecKey]
vendorBundle := builtBundles[kotstypes.VendorSpecificSupportBundleSpecKey]
clusterSpecificBundle := builtBundles[kotstypes.ClusterSpecificSupportBundleSpecKey]

// add discovered collectors and analyzers to cluster-specific
clusterSpecificBundle = addDiscoveredSpecs(clusterSpecificBundle, app, clientset)

// merge all collectors and analyzers (need to make sure AfterCollection is maintained)
mergedBundle := mergeBundles(defaultBundle, vendorBundle, clusterSpecificBundle)

return mergedBundle
```

pkg/supportbundle/spec.go
```diff
@@ -173,39 +190,116 @@ func CreateRenderedSpec(app apptypes.AppType, sequence int64, kotsKinds *kotsuti
...
+			return errors.Wrap(err, "failed to update support bundle secret")
 		}
 	}
+	return nil
+}
 
-	if existingSecret.Data == nil {
-		existingSecret.Data = map[string][]byte{}
+// addClusterSpecificSpec adds cluster specific and upload results URI to the support bundle
+func addClusterSpecificSpec(app apptypes.AppType, b *troubleshootv1beta2.SupportBundle, opts types.TroubleshootOptions) *troubleshootv1beta2.SupportBundle {
```
cbodonnell, line 216: Everything in this function feels like it belongs in the "default" spec.  I think we can leave the cluster-specific spec empty initially and then it will be populated with any discovered specs from the cluster.

pkg/supportbundle/spec.go
cbodonnell: After some discussion, this was not the reason why the bundle was empty.  It was failing to collect with an error:
```
"error collecting support bundle ID from URI: 2mknjabqcz7sx9rorg9poukpg0v: could not bundle from URI: failed to load collector spec: failed to get spec from secret: failed to get secret: secrets \"kotsadm-craig-helm-supportbundle\" not found"
```

DexterYan: Hey @cbodonnell, I have added  mergedBundle to combine three separated specs into one and saved as origin one for backward compatible. Also, clusterSpecificBundle has been changed to use discovery function

pkg/supportbundle/spec.go
DexterYan: AfterCollection has been moved back to default one

DexterYan:
Few changes that has been added that as follows:
1. change appNameRE to `^kotsadm-.*-supportbundle(?:$|.*)` to prevent that cluster-specific-supportbundle has add existing kots spec as duplicated spec after running multiple time
2. change `fmt.Sprintf("kubectl support-bundle %s --redactors=%s\n", GetSpecURI(appSlug), redactors)` to `fmt.Sprintf("kubectl support-bundle --load-cluster-specs --redactors=%s\n", redactors)`

pkg/supportbundle/supportbundle.go
```diff
@@ -119,17 +119,13 @@ func GetSpecSecretName(appSlug string) string {
...
-}
-
 func GetBundleCommand(appSlug string) []string {
 	redactURIs := []string{redact.GetKotsadmRedactSpecURI(), redact.GetAppRedactSpecURI(appSlug)}
 	redactors := strings.Join(redactURIs, ",")
 
 	command := []string{
 		"curl https://krew.sh/support-bundle | bash",
-		fmt.Sprintf("kubectl support-bundle %s --redactors=%s\n", GetSpecURI(appSlug), redactors),
+		fmt.Sprintf("kubectl support-bundle --load-cluster-specs --redactors=%s\n", redactors),
```
banjoh, line 128: While we are here, I suppose we want to change the rest of the commands presented to users. 

[Here are are other references](https://github.com/replicatedhq/kots/search?q=krew.sh%5C%2Fsupport-bundle)

banjoh (commented):
I think we need changes in this https://github.com/replicatedhq/kots/pull/3717 PR here. I suppose once its merged you can rebase with upstream then we can review the PR further.

The comments here still apply though

pkg/supportbundle/supportbundle.go
```diff
@@ -189,11 +185,11 @@ func CreateSupportBundleDependencies(app apptypes.AppType, sequence int64, opts
 
 	supportBundleObj := types.SupportBundle{
 		AppID:      app.GetID(),
-		URI:        GetSpecURI(app.GetSlug()),
 		RedactURIs: redactURIs,
 		Progress: types.SupportBundleProgress{
 			CollectorCount: len(supportBundle.Spec.Collectors),
 		},
+		BundleSpec: supportBundle,
```
cbodonnell, line 192: We should change this back to the `URI` now that we are also preserving the merged `kotsadm-<app-slug>-supportbundle` spec?  It seems to cause the redactions to be omitted because `RedactURIs` are not used when [this code path](https://github.com/replicatedhq/kots/blob/main/pkg/supportbundle/execute.go#L191-L196) is taken.

pkg/supportbundle/supportbundle.go
cbodonnell, line 128: Thinking about this more.  We are already [discovering specs](https://github.com/replicatedhq/kots/pull/3674/files#diff-3681d5e2735b3de144e2433bb7104c9bae6e47ddf4891d7d3017e9b7e2d83e5aR294) and [merging them](https://github.com/replicatedhq/kots/pull/3674/files#diff-3681d5e2735b3de144e2433bb7104c9bae6e47ddf4891d7d3017e9b7e2d83e5aR125) into `kotsadm-<app-slug>-supportbundle`.  Can we leave this as-is so that it would work with users running old versions of the support-bundle binary too?  It seems like `--load-cluster-specs` would not result in any additional collectors/analyzers.  What do you think?

pkg/supportbundle/supportbundle.go
DexterYan: That's a good catch. I have rollback the change to use URI, until we change way of how RedactURIs to use same way like use tags to discover, let's keep it for now

DexterYan:
The #3717 has been rebased on top of this pr, while we are here, I would suggest to review #3717 instead of this pr to make this change perfectly. This pr will be closed. Thank you @cbodonnell and @banjoh!

PR #3681, opened 2023-02-23
add actionlint and clean up deprecations [sc-70302]

e3b0c442:
* Add `actionlint` for static analysis of GitHub Actions workflows
* cleanup linting issues
* update actions to minimum node16 versions
* clean up deprecated usages of certain github commands

Be careful of unintended behavior changes in shell scripts, especially where quotes were added to satisfy shellcheck. I caught one error I had made in the instances where we were populating environment variables from the .image.env, and I'm going to go over it again with a fine-toothed comb, but there may be others.

```release-note
NONE
```

NONE

### DIFF
--- a/.github/workflows/alpha.yaml
+++ b/.github/workflows/alpha.yaml
@@ -13,7 +13,7 @@ jobs:
       tag: ${{ steps.get_tag.outputs.GIT_TAG }}
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Get tags
         id: get_tag
         uses: ./.github/actions/version-tag
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
     - name: Checkout
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
 
     - uses: azure/docker-login@v1
       env:
@@ -36,7 +36,7 @@ jobs:
       env:
         DOCKER_CONFIG: ./.docker
       run: |
-        export $(cat .image.env | sed 's/#.*//g' | xargs) && make -C migrations schema-alpha
+        mapfile -t envs < <(grep -v '#.*' < .image.env) && export "${envs[@]}" && make -C migrations schema-alpha
 
 
   build_web:
@@ -47,21 +47,21 @@ jobs:
       options: --user root
     steps:
     - name: Checkout
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
 
     - name: Setup Node.js environment
-      uses: actions/setup-node@v2
+      uses: actions/setup-node@v3
       with:
         node-version: '17.x'
 
     - name: Build web
       env:
         GIT_TAG: ${{ needs.generate-tag.outputs.tag }}
-      run: export $(cat .image.env | sed 's/#.*//g' | xargs) && make -C web deps build-kotsadm
+      run: mapfile -t envs < <(grep -v '#.*' < .image.env) && export "${envs[@]}" && make -C web deps build-kotsadm
       shell: bash
 
     - name: Upload web artifact
-      uses: actions/upload-artifact@v2
+      uses: actions/upload-artifact@v3
       with:
         name: web
         path: ./web/dist
@@ -72,12 +72,12 @@ jobs:
     needs: [generate-tag]
     steps:
     - name: Setup Go
-      uses: actions/setup-go@v2
+      uses: actions/setup-go@v3
       with:
         go-version: '^1.19.2'
 
     - name: Checkout
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
 
     - name: Cache Go modules
       uses: actions/cache@v3
@@ -93,11 +93,11 @@ jobs:
       env:
         GIT_TAG: ${{ needs.generate-tag.outputs.tag }}
         SCOPE_DSN_PUBLIC: ""
-      run: export $(cat .image.env | sed 's/#.*//g' | xargs) && make -C kurl_proxy test build
+      run: mapfile -t envs < <(grep -v '#.*' < .image.env) && export "${envs[@]}" && make -C kurl_proxy test build
       shell: bash
 
     - name: Upload kurl_proxy artifact
-      uses: actions/upload-artifact@v2
+      uses: actions/upload-artifact@v3
       with:
         name: kurl_proxy
         path: ./kurl_proxy/bin
@@ -107,12 +107,12 @@ jobs:
     runs-on: ubuntu-20.04
     needs: [build_web, build_kurl_proxy, generate-tag]
     steps:
-    - uses: actions/setup-go@v2
+    - uses: actions/setup-go@v3
       with:
         go-version: '^1.19.2'
 
     - name: Checkout
-      uses: actions/checkout@v1 # not @v2 because of: https://github.com/actions/checkout/issues/126
+      uses: actions/checkout@v3
 
     - name: Cache Go modules
       uses: actions/cache@v3
@@ -125,7 +125,7 @@ jobs:
           ${{ runner.os }}-go-kots-
 
     - name: Download web artifact
-      uses: actions/download-artifact@v2
+      uses: actions/download-artifact@v3
       with:
         name: web
         path: ./web/dist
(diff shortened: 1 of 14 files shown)


.github/workflows/alpha.yaml
```diff
@@ -107,12 +107,12 @@ jobs:
...
     needs: [build_web, build_kurl_proxy, generate-tag]
     steps:
-    - uses: actions/setup-go@v2
+    - uses: actions/setup-go@v3
       with:
         go-version: '^1.19.2'
 
     - name: Checkout
-      uses: actions/checkout@v1 # not @v2 because of: https://github.com/actions/checkout/issues/126
+      uses: actions/checkout@v3
```
e3b0c442, line 115: The underlying issue here has been resolved

e3b0c442:
There are still some deprecations, but I mainly wanted to get the ones cleaned up that would trigger actionlint, I'll come back to clean up the remainder in a different PR.

.github/workflows/build-test.yaml
```diff
@@ -448,6 +453,8 @@ jobs:
           chmod +x bin/*
       - uses: ./.github/actions/kots-e2e
         with:
+          # REVIEW: missing required `kots-namespace` parameter. Set to empty string`
+          kots-namespace: ''
```
sgalsaleh, line 457: I believe this should be:
```suggestion
          kots-namespace: 'qakotsregression'
```
Based on https://github.com/replicatedhq/kots/blob/main/e2e/testim/inventory/inventory.go#L28-L30

.github/workflows/build-test.yaml
```diff
@@ -1041,56 +1051,56 @@ jobs:
...
+            ((COUNTER += 1))
             if [ $COUNTER -gt 120 ]; then
               echo "Timed out waiting for app to be installed with correct version label: $APP_VERSION_LABEL"
-              ./bin/kots get apps --namespace $APP_SLUG
+              ./bin/kots get apps --namespace "$APP_SLUG"
               echo "kotsadm logs:"
-              kubectl logs -l app=kotsadm --tail=100 --namespace $APP_SLUG
-              exit -1
+              kubectl logs -l app=kotsadm --tail=100 --namespace "$APP_SLUG"
+              exit 255
```
sgalsaleh, line 1068: I think these can just be `1` instead of `-1`. Pretty sure we put `-1` by mistake.
```suggestion
              exit 1
```

.github/workflows/build-test.yaml
```diff
@@ -1925,78 +1940,77 @@ jobs:
...
-            --namespace $APP_SLUG \
+            --namespace "$APP_SLUG" \
             --kotsadm-registry ttl.sh \
             --kotsadm-namespace automated-${{ github.run_id }} \
             --kotsadm-tag 24h
 
-          # verify that the postgres to rqlite migration was successful
-
-          if ! kubectl logs -l app=kotsadm --namespace $APP_SLUG | grep -q "Migrated from Postgres to rqlite successfully"; then
+          if ! kubectl logs -l app=kotsadm --namespace "$APP_SLUG" | grep -q "Migrated from Postgres to rqlite successfully"; then
```
e3b0c442, line 1980: Need to re-add the comment, I think this was a merge conflict

.github/workflows/build-test.yaml
```diff
@@ -1099,33 +1109,34 @@ jobs:
...
-          ./bin/kots get apps --namespace $APP_SLUG
+          printf "App is installed successfully with the correct version label: %s\n\n" "$LATEST_APP_VERSION_LABEL"
+          ./bin/kots get apps --namespace "$APP_SLUG"
 
       - name: Generate support bundle on failure
         if: failure()
         uses: ./.github/actions/generate-support-bundle
         with:
-          kots-namespace: "$APP_SLUG"
+          # REVIEW: input `kots-namespace` does not exist on `generate-support-bundle` action
```
sgalsaleh, line 1138: It does exist here: https://github.com/replicatedhq/kots/blob/main/.github/actions/generate-support-bundle/action.yml#L22

We should add it to the list of input parameters.

.github/workflows/build-test.yaml
```diff
@@ -1249,33 +1261,34 @@ jobs:
...
           printf "App is installed successfully and is ready\n\n"
-          ./bin/kots get apps --namespace $APP_SLUG
+          ./bin/kots get apps --namespace "$APP_SLUG"
 
       - name: Generate support bundle on failure
         if: failure()
         uses: ./.github/actions/generate-support-bundle
         with:
-          kots-namespace: "$APP_SLUG"
+          # REVIEW: input `kots-namespace` does not exist on `generate-support-bundle` action
```
sgalsaleh, line 1290: Ditto

.github/workflows/build-test.yaml
```diff
@@ -1374,7 +1387,8 @@ jobs:
...
```
sgalsaleh, line 1390: Ditto

.github/workflows/build-test.yaml
```diff
@@ -1442,53 +1457,53 @@ jobs:
...
+          ./bin/kots get apps --namespace "$APP_SLUG"
 
       - name: Generate support bundle on failure
         if: failure()
-        uses: ./.github/actions/kots-e2e
+        uses: ./.github/actions/generate-support-bundle
         with:
-          kots-namespace: "$APP_SLUG"
-          k8s-version: '${{ matrix.k8s_version }}'
+          # REVIEW: input `kots-namespace` does not exist on `generate-support-bundle` action
```
sgalsaleh, line 1505: Ditto

.github/workflows/build-test.yaml
```diff
@@ -1689,42 +1702,43 @@ jobs:
...
           echo "------kotsadm-minio logs"
-          kubectl logs -l app=kotsadm-minio --tail=100 --namespace $NAMESPACE
+          kubectl logs -l app=kotsadm-minio --tail=100 --namespace "$NAMESPACE"
 
       - name: Generate support bundle on failure
         if: failure()
         uses: ./.github/actions/generate-support-bundle
         with:
-          kots-namespace: "$APP_SLUG"
+          # REVIEW: input `kots-namespace` does not exist on `generate-support-bundle` action
```
sgalsaleh, line 1740: Ditto

.github/workflows/build-test.yaml
e3b0c442: Is there a sane default?

.github/workflows/build-test.yaml
```diff
@@ -1766,19 +1780,20 @@ jobs:
           testim-access-token: '${{ secrets.TESTIM_ACCESS_TOKEN }}'
           testim-branch: ${{ github.head_ref == 'main' && 'master' || github.head_ref }}
           kotsadm-image-registry: ttl.sh
-          kotsadm-namespace: automated-${{ github.run_id }}
-          kotsadm-tag: 24h
+          # REVIEW: inputs `kotsadm-namespace` and `kotsadm-tag` do not exist on `kots-e2e` action
```
sgalsaleh, line 1783: Yep, those are not needed for this test. You can remove them.

.github/workflows/kurl-addon-kots-test.yaml
```diff
@@ -41,7 +45,7 @@ jobs:
         uses: ./.github/actions/kurl-addon-kots-generate
         with:
           addon_version: "${{ inputs.addon_version }}"
-          s3_prefix: "${{ github.event.action == 'workflow_dispatch' && 'test/' || inputs.s3_prefix }}"
```
sgalsaleh, line 44: I'm not 100% sure this change is equivalent since the prefix would need to empty if the event action is not a `workflow_dispatch`, but now `s3_prefix` has a default of `test/`. I'll defer to @emosbaugh since I'm not too familiar with what the prefix value should be.

.github/workflows/build-test.yaml
sgalsaleh: I don't think so, let's make it required I guess?

.github/workflows/kurl-addon-kots-test.yaml
e3b0c442: This should be safe. There is no way that expression could have evaluated to anything but an empty string.

sgalsaleh (approved):

PR #3707, opened 2023-03-03
Rianfowler/sc 60124/investigate why eslint is not catching components

rianfowler:
- linting doesn't block issues in js files. migrating some easy js files over to ts 

Fixes # https://app.shortcut.com/replicated/story/60124/investigate-why-eslint-is-not-catching-components-that-are-missing-import-statements

```release-note
none
```

none


web/src/hooks/usePrevious.tsx
```diff
@@ -1,7 +1,7 @@
 import { useEffect, useRef } from "react";
 
-export function usePrevious(value) {
-  const ref = useRef();
+export function usePrevious<T>(value: T): T | null {
+  const ref = useRef<T | null>(null);
```
rianfowler, line 4: logic change here- the default value is set to `null`

web/src/index.tsx
```diff
@@ -4,6 +4,10 @@ import ReplicatedErrorBoundary from "./components/shared/ErrorBoundary";
 import { Root } from "./Root";
 
 const container = document.getElementById("app");
+if (!container) {
+  throw new Error("No container found");
+}
```
rianfowler, line 9: type guard - this should never happen but we're handling it now in case it does

web/yarn.lock
```diff
@@ -4280,6 +4288,15 @@
   resolved "https://registry.npmjs.org/@types/stack-utils/-/stack-utils-2.0.1.tgz"
   integrity sha512-Hl219/BT5fLAaz6NDkSuhzasy49dwQS/DSdu4MdggFB8zcXv7vflBI3xp7FEmkmdDkBUI2bPUNeMttp2knYdxw==
 
+"@types/styled-components@^5.1.26":
```
rianfowler, line 4291: the version of `styled-components` doesn't ship with type definitions so added those

alicenstar (approved):

PR #3708, opened 2023-03-03
fix: crds from subcharts included in helm release secret

cbodonnell:
**TODO:** Add unit tests for CRDs in the parent chart and subcharts

This PR addresses an issue where CRDs from subcharts were being included in the secret that stores the Helm release.  By default, Helm only stores CRDs in the `.chart.files` of the serialized release object for the parent chart and not for subcharts.  KOTS runs the equivalent of a `helm template --include-crds`, which results in the `# Source: ...` for each CRD to reflect the path relative to the subchart, instead of the path relative the parent chart.  In KOTS, this results in all CRDs ending up in the `crds/` directory of the parent chart within the base files.  It's not clear whether Helm intentionally uses a path relative to the subchart for CRDs only, but it results in this KOTS bug because of that.  All other files output from `helm template` have a `# Source: ...` relative to the parent chart.

This PR changes the KOTS logic so that the `--include-crds` option is effectively left out and the CRDs are added to the rendered output separately using the path relative to the parent chart.  This way the CRDs will reside in their respective subchart directories in the KOTS base files and will not be added as additional files to the Helm release secret when installing the chart.  The implementation reflects the Helm [source code](https://github.com/helm/helm/blob/main/pkg/action/action.go#L189-L201), but instead of using `crd.Name`, it uses `crd.Filename`, which is the path relative to the parent chart.

Fixes [SC-70646](https://app.shortcut.com/replicated/story/70646/helm-releases-created-by-kots-are-significantly-larger-than-those-created-when-just-using-helm)

Install an application using native helm that has a subchart with CRDs.  Observe that all CRDs (for the parent and subcharts) and up in the `crds/` directory for the parent chart.  With these changes, CRDs will end up in the `crds/` directory for the chart they belong to.

```release-note
Fixes an issue where CRDs from subcharts were being included in the secret that Helm stores release information in, which could dramatically increase its size
```

NONE

### DIFF
--- a/pkg/base/helm_v3.go
+++ b/pkg/base/helm_v3.go
@@ -34,7 +34,6 @@ func renderHelmV3(chartName string, chartPath string, vals map[string]interface{
 	client.ReleaseName = chartName
 	client.Replace = true
 	client.ClientOnly = true
-	client.IncludeCRDs = true
 
 	client.Namespace = renderOptions.Namespace
 	if client.Namespace == "" {
@@ -70,18 +69,23 @@ func renderHelmV3(chartName string, chartPath string, vals map[string]interface{
 		return nil, nil, errors.Wrap(err, "failed to marshal rendered values")
 	}
 
-	var manifests bytes.Buffer
-	fmt.Fprintln(&manifests, strings.TrimSpace(rel.Manifest))
-	for _, m := range rel.Hooks {
-		fmt.Fprintf(&manifests, "---\n# Source: %s\n%s\n", m.Path, m.Manifest)
-	}
-
 	baseFiles := []BaseFile{}
 	additionalFiles := []BaseFile{
 		{
 			Path:    "values.yaml",
 			Content: valuesContent,
 		},
+	}
+
+	var manifests bytes.Buffer
+	fmt.Fprintln(&manifests, strings.TrimSpace(rel.Manifest))
+	// add hooks
+	for _, m := range rel.Hooks {
+		fmt.Fprintf(&manifests, "---\n# Source: %s\n%s\n", m.Path, m.Manifest)
+	}
+	// add crds
+	for _, crd := range chartRequested.CRDObjects() {
+		fmt.Fprintf(&manifests, "---\n# Source: %s\n%s\n", crd.Filename, string(crd.File.Data[:]))
 	}
 
 	splitManifests := splitManifests(manifests.String())


sgalsaleh (dismissed):

pkg/base/helm_v3.go
```diff
@@ -25,16 +25,15 @@ var (
 
 const NamespaceTemplateConst = "repl{{ Namespace}}"
 
-func renderHelmV3(chartName string, chartPath string, vals map[string]interface{}, renderOptions *RenderOptions) ([]BaseFile, []BaseFile, error) {
+func renderHelmV3(releaseName string, chartPath string, vals map[string]interface{}, renderOptions *RenderOptions) ([]BaseFile, []BaseFile, error) {
```
cbodonnell, line 28: This is only used to set the release name, so renamed it appropriately.

pkg/tests/pull/cases/samechartvariations/wantResults/base/charts/test-chart-release-1/kustomization.yaml
```diff
@@ -1,8 +1,8 @@
...
-- test-chart/templates/test-1.yaml
-- test-chart/templates/test-2.yaml
-- test-chart/templates/test-3.yaml
-- test-chart/templates/test-4.yaml
-- test-chart/templates/test-5.yaml
+- templates/test-1.yaml
+- templates/test-2.yaml
+- templates/test-3.yaml
+- templates/test-4.yaml
+- templates/test-5.yaml
```
cbodonnell, line 8: If the user did not specify a `releaseName`, then the chart templates ended up in `chart-name/templates/...`.  If they did, they would end up in `release-name/chart-name/templates/...` because the chart name prefix was not removed from the base path.

pkg/base/helm_v3.go
```diff
@@ -84,13 +77,24 @@ func renderHelmV3(chartName string, chartPath string, vals map[string]interface{
...
+	}
+
 	splitManifests := splitManifests(manifests.String())
 	manifestName := ""
 	for _, manifest := range splitManifests {
 		submatch := HelmV3ManifestNameRegex.FindStringSubmatch(manifest)
 		if len(submatch) > 0 {
 			// multi-doc manifests will not have the Source comment so use the previous name
-			manifestName = strings.TrimPrefix(submatch[1], fmt.Sprintf("%s/", chartName))
+			manifestName = strings.TrimPrefix(submatch[1], fmt.Sprintf("%s/", chartRequested.Name()))
```
cbodonnell, line 97: Fixes a separate issue where if `spec.chart.releaseName` was specified, the chart name prefix was not removed here since it would search for the release name instead.  This changes it to always search for the chart name.

sgalsaleh (approved):

PR #3717, opened 2023-03-07
feat(support-bundle): simplify implementation handling the support bundle spec

DexterYan:
- split the current support bundle spec into vender spec, default spec and cluster spec, embed them as code under kots/pkg/supportbundle 
- remove injectDefaults to simplify way of generate specs
- seperate default spec into defaultspec and kurlspec, so we can remove removeKurlCollectors & removeKurlAnalyzers
- deduplicatedCollectors and deduplicatedAnalyzers rewritten

Shortcut: #
[sc-70445](https://app.shortcut.com/replicated/story/70445/simplify-implementation-handling-the-support-bundle-spec)

```release-note
NONE
```

NONE

### DIFF
--- a/pkg/supportbundle/staticspecs/embed.go
+++ b/pkg/supportbundle/staticspecs/embed.go
@@ -1,3 +1,7 @@
+// DO NOT CACHE THE PARSED SPEC
+// All getters always parse the bundle from the raw spec (embedded files)
+// to ensure that we fetch the latest specs defined in URI fields (if any)
+
 package staticspecs
 
 import (
@@ -19,40 +23,18 @@ var vendorspec []byte
 //go:embed defaultspec.yaml
 var defaultspec []byte
 
-var spec *troubleshootv1beta2.SupportBundle
-
-func GetVendorSpec() troubleshootv1beta2.SupportBundle {
-	var err error
-	spec, err = supportbundle.ParseSupportBundleFromDoc(vendorspec)
-	if err != nil {
-		panic(err)
-	}
-	return *spec.DeepCopy()
+func GetVendorSpec() (*troubleshootv1beta2.SupportBundle, error) {
+	return supportbundle.ParseSupportBundleFromDoc(vendorspec)
 }
 
-func GetClusterSpecificSpec() troubleshootv1beta2.SupportBundle {
-	var err error
-	spec, err = supportbundle.ParseSupportBundleFromDoc(clusterspec)
-	if err != nil {
-		panic(err)
-	}
-	return *spec.DeepCopy()
+func GetClusterSpecificSpec() (*troubleshootv1beta2.SupportBundle, error) {
+	return supportbundle.ParseSupportBundleFromDoc(clusterspec)
 }
 
-func GetDefaultSpec() troubleshootv1beta2.SupportBundle {
-	var err error
-	spec, err = supportbundle.ParseSupportBundleFromDoc(defaultspec)
-	if err != nil {
-		panic(err)
-	}
-	return *spec.DeepCopy()
+func GetDefaultSpec() (*troubleshootv1beta2.SupportBundle, error) {
+	return supportbundle.ParseSupportBundleFromDoc(defaultspec)
 }
 
-func GetKurlSpec() troubleshootv1beta2.SupportBundle {
-	var err error
-	spec, err = supportbundle.ParseSupportBundleFromDoc(kurlspec)
-	if err != nil {
-		panic(err)
-	}
-	return *spec.DeepCopy()
+func GetKurlSpec() (*troubleshootv1beta2.SupportBundle, error) {
+	return supportbundle.ParseSupportBundleFromDoc(kurlspec)
 }
--- a/pkg/supportbundle/staticspecs/embed.go
+++ b/pkg/supportbundle/staticspecs/embed.go
@@ -1,0 +1,58 @@
+package staticspecs
+
+import (
+	_ "embed"
+
+	troubleshootv1beta2 "github.com/replicatedhq/troubleshoot/pkg/apis/troubleshoot/v1beta2"
+	"github.com/replicatedhq/troubleshoot/pkg/supportbundle"
+)
+
+//go:embed clusterspec.yaml
+var clusterspec []byte
+
+//go:embed kurlspec.yaml
+var kurlspec []byte
+
+//go:embed vendorspec.yaml
+var vendorspec []byte
+
+//go:embed defaultspec.yaml
+var defaultspec []byte
+
+var spec *troubleshootv1beta2.SupportBundle
+
+func GetVendorSpec() troubleshootv1beta2.SupportBundle {
+	var err error
+	spec, err = supportbundle.ParseSupportBundleFromDoc(vendorspec)
+	if err != nil {
+		panic(err)
+	}
+	return *spec.DeepCopy()
+}
+
+func GetClusterSpecificSpec() troubleshootv1beta2.SupportBundle {
+	var err error
+	spec, err = supportbundle.ParseSupportBundleFromDoc(clusterspec)
+	if err != nil {
+		panic(err)
+	}
+	return *spec.DeepCopy()
+}
+
+func GetDefaultSpec() troubleshootv1beta2.SupportBundle {
+	var err error
+	spec, err = supportbundle.ParseSupportBundleFromDoc(defaultspec)
+	if err != nil {
+		panic(err)
+	}
+	return *spec.DeepCopy()
+}
+
+func GetKurlSpec() troubleshootv1beta2.SupportBundle {
+	var err error
+	spec, err = supportbundle.ParseSupportBundleFromDoc(kurlspec)
+	if err != nil {
+		panic(err)
+	}
+	return *spec.DeepCopy()
+}
--- a/pkg/supportbundle/staticspecs/vendorspec.yaml
+++ b/pkg/supportbundle/staticspecs/vendorspec.yaml
@@ -1,0 +1,5 @@
+apiVersion: troubleshoot.sh/v1beta2
+kind: SupportBundle
+metadata:
+  name: cluster-specific-supportbundle
+spec: {}
--- a/pkg/supportbundle/staticspecs/vendorspec.yaml
+++ b/pkg/supportbundle/staticspecs/vendorspec.yaml
@@ -1,5 +1,5 @@
 apiVersion: troubleshoot.sh/v1beta2
 kind: SupportBundle
 metadata:
-  name: cluster-specific-supportbundle
+  name: vendor-supportbundle
 spec: {}
--- a/pkg/supportbundle/supportbundle.go
+++ b/pkg/supportbundle/supportbundle.go
@@ -129,7 +129,7 @@ func GetBundleCommand(appSlug string) []string {
 
 	command := []string{
 		"curl https://krew.sh/support-bundle | bash",
-		fmt.Sprintf("kubectl support-bundle --load-cluster-specs --redactors=%s\n", redactors),
+		fmt.Sprintf("kubectl support-bundle %s --redactors=%s\n", GetSpecURI(appSlug), redactors),
 	}
 
 	return command
--- a/pkg/supportbundle/supportbundle.go
+++ b/pkg/supportbundle/supportbundle.go
@@ -115,12 +115,12 @@ func CreateBundle(requestedID string, appID string, archivePath string) (*types.
 	return store.GetStore().CreateSupportBundle(id, appID, archivePath, marshalledTree)
 }
 
-func GetSpecSecretName(appSlug string) string {
+func GetSpecName(appSlug string) string {
 	return fmt.Sprintf("kotsadm-%s-supportbundle", appSlug)
 }
 
 func GetSpecURI(appSlug string) string {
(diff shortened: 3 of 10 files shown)


pkg/supportbundle/clusterspec/embed.go
```diff
@@ -0,0 +1,39 @@
...
+//go:embed spec.yaml
+var raw []byte
+
+var spec *troubleshootv1beta2.SupportBundle
+
+func init() {
+	var err error
+	spec, err = supportbundle.ParseSupportBundleFromDoc(raw)
+	if err != nil {
+		panic(err)
```
banjoh, line 19: I think we need a better way to handle these errors. KOTS should not crash if there is a problem loading specs. I suggest we have the specs lazily loaded in the `Getters` and return an error if there is a failure. This is how I would do this

```go
var (
	defaultSpec *troubleshootv1beta2.SupportBundle
)

func GetDefault() (*troubleshootv1beta2.SupportBundle, error) {
	// Some worth improvements
	// * Thread safety
	// * sync.Pool for spec objects. Maybe? Perhaps making object copies is intentional.
	if (spec == nil) {
		var err error
		defaultSpec, err = supportbundle.ParseSupportBundleFromDoc(raw)
		if err != nil {
			return nil, err
		}
	}
	return defaultSpec.DeepCopy(), nil
}
```

banjoh: This change would require additional application error handling like showing some information in Admin console's troubleshoot page. There is an [error message response object](https://github.com/replicatedhq/kots/blob/main/web/src/components/troubleshoot/GenerateSupportBundleModal.tsx#L210) used by the web UI to show that there was a problem. Perhaps this will just work but its worth testing

pkg/supportbundle/staticspecs/embed.go
```diff
@@ -0,0 +1,58 @@
...
+//go:embed kurlspec.yaml
+var kurlspec []byte
+
+//go:embed vendorspec.yaml
+var vendorspec []byte
+
+//go:embed defaultspec.yaml
+var defaultspec []byte
+
+var spec *troubleshootv1beta2.SupportBundle
```
banjoh, line 22: This variable is unnecessary

Once removed, your getters code will look so

```go
func GetVendorSpec() troubleshootv1beta2.SupportBundle {
	spec, err := supportbundle.ParseSupportBundleFromDoc(vendorspec)
	if err != nil {
		panic(err)
	}
	return *spec
}
```

pkg/supportbundle/staticspecs/embed.go
```diff
@@ -0,0 +1,58 @@
+package staticspecs
+
+import (
+	_ "embed"
+
+	troubleshootv1beta2 "github.com/replicatedhq/troubleshoot/pkg/apis/troubleshoot/v1beta2"
+	"github.com/replicatedhq/troubleshoot/pkg/supportbundle"
+)
+
```
banjoh, line 9: Please add the following comment here. I reckon someone (like me) would one day try to optimise this code by caching the parsed spec.

```go
// DO NOT CACHE THE PARSED SPEC
// All getters always parse the bundle from the raw spec (embedded files)
// to ensure that we fetch the latest specs defined in URI fields (if any)
```

pkg/supportbundle/staticspecs/vendorspec.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: troubleshoot.sh/v1beta2
+kind: SupportBundle
+metadata:
+  name: cluster-specific-supportbundle
```
cbodonnell, line 4: Rename to `vendor-supportbundle`

DexterYan: Thank you @cbodonnell, good catch! Changed

pkg/supportbundle/staticspecs/embed.go
DexterYan: Thank you @banjoh, I have remove unnecessary code and return error back to spec.go to handle.

pkg/supportbundle/staticspecs/embed.go
DexterYan: Add

pkg/supportbundle/supportbundle.go
```diff
@@ -129,7 +129,7 @@ func GetBundleCommand(appSlug string) []string {
 
 	command := []string{
 		"curl https://krew.sh/support-bundle | bash",
-		fmt.Sprintf("kubectl support-bundle %s --redactors=%s\n", GetSpecURI(appSlug), redactors),
+		fmt.Sprintf("kubectl support-bundle --load-cluster-specs --redactors=%s\n", redactors),
```
cbodonnell, line 132: I ran this command in Okteto, which is a namespace-scoped KOTS instance, and it returned the following:
```
$ kubectl support-bundle --load-cluster-specs --redactors=configmap/cbodonnell/kotsadm-redact-spec/redact-spec,configmap/cbodonnell/kotsadm-craig-helm-redact-spec/redact-spec
Error: no specs found in cluster
```

Running with `-n cbodonnell` allowed it to run.  Does this mean that the support-bundle command is namespace scoped?

cbodonnell: It also appears that some of the analyzers were not de-duplicated when running that way?
```
Check PASS
Title: Required Kubernetes Version
Message: Your cluster meets the recommended and required versions of Kubernetes

------------
Check PASS
Title: kotsadm Status
Message: At least 1 replica of the Admin Console API is running and ready

------------
Check PASS
Title: Required Kubernetes Version
Message: Your cluster meets the recommended and required versions of Kubernetes

------------
Check PASS
Title: kotsadm Status
Message: At least 1 replica of the Admin Console API is running and ready

------------
```

DexterYan: According the code https://github.com/replicatedhq/troubleshoot/blob/546ffde14b5c7bcc883ceead3303a1f6e7dbffc2/cmd/troubleshoot/cli/run.go#L117. And a error,
I think it is using namespace "" if you don't input any namespace, but according to https://troubleshoot.sh/docs/support-bundle/discover-cluster-specs/#usage, it should search everything in the cluster. Which made me also quite confusing
```
2023/03/16 17:26:30 failed to load redactor specs from config maps: failed to search for configmaps in the cluster: configmaps is forbidden: User "system:serviceaccount:okteto:0cc534d8-64a0-4abb-b6dd-f45e055ac018" cannot list resource "configmaps" in API group "" at the cluster scope
```

DexterYan: for de-duplicated, I think our deduplicatedAnalyzers https://github.com/replicatedhq/kots/blob/29c817c403807f6dd6b43ef305abdb925dd019b4/pkg/supportbundle/spec.go#L514 is not actually deduplicated. 

The right one should be like https://github.com/replicatedhq/kots/blob/1df1fb704daa0b55c809e864fe4bf432ef68b9ac/pkg/supportbundle/spec.go#L743. I will rewrite deduplicatedAnalyzers and deduplicatedCollectors according that one, if you can also confirm that our origin way is not right?

DexterYan: I have tested in my ubuntu cluster with `curl https://kurl.sh/latest | sudo bash`,

It will search all the namespace.
```
 kubectl support-bundle --load-cluster-specs -n kube-flannel
 kubectl support-bundle --load-cluster-specs
```
has same result

DexterYan: Error: no specs found in cluster, I think it is complain about the redactors spec cannot find

cbodonnell: > I will rewrite deduplicatedAnalyzers and deduplicatedCollectors according that one, if you can also confirm that our origin way is not right?

I agree, that looks like it's only checking for a handful of specific collectors/analyzers in those functions.  However, I only saw duplicated analyzers when running from the CLI with the `kubectl support-bundle` and not when using KOTS to generate the bundle.  Do you see the same behavior?

> Error: no specs found in cluster, I think it is complain about the redactors spec cannot find

Both of those redactor configmaps do exist in the namespace.  I tried adding some logging to [this section](https://github.com/replicatedhq/troubleshoot/blob/main/cmd/troubleshoot/cli/run.go#L134-L145) in troubleshoot and I see that it's not finding the support bundle specs in secrets/configmaps when I don't pass a namespace.

```
$ ./bin/support-bundle --load-cluster-specs --redactors=configmap/cbodonnell/kotsadm-redact-spec/redact-spec,configmap/cbodonnell/kotsadm-craig-helm-redact-spec/redact-spec
*** found 0 support bundle specs in secrets
*** found 0 support bundle specs in configmaps
*** found 0 support bundle specs in cluster
Error: no specs found in cluster
$ ./bin/support-bundle -n cbodonnell --load-cluster-specs --redactors=configmap/cbodonnell/kotsadm-redact-spec/redact-spec,configmap/cbodonnell/kotsadm-craig-helm-redact-spec/redact-spec
*** found 1 support bundle specs in secrets
*** found 3 support bundle specs in configmaps
*** found 4 support bundle specs in cluster
```

I don't necessarily want to hold up the review for this PR because of this.  We could change the commands back to the original and I think it would still work since we keep the old `kotsadm-<app-slug>-supportbundle` secret around for backwards compatibility.  I'll leave that up to you.

cbodonnell: The [helm-managed test is failing](https://github.com/replicatedhq/kots/actions/runs/4412456556/jobs/7747529870#step:8:418) is also because it's expecting the old command, so this test will need to be updated.  I can do this once we're all set here or we can pair on it if you want to learn how to do it.

DexterYan: Aha! I tested ` ./bin/support-bundle --load-cluster-specs` against okteto cluster and got same error `Error: no specs found in cluster`. Then when I open the debug mode `./bin/support-bundle --load-cluster-specs --debug`, I found the issue is actually that okteto forbidden user to do a search across all the namespace, so it is not our script issue. Could you try it again with debug mode?


```
E0317 16:11:10.525947    3226 run.go:137] failed to load support bundle spec from secrets: failed to search for secrets in the cluster: secrets is forbidden: User "system:serviceaccount:okteto:0cc534d8-64a0-4abb-b6dd-f45e055ac018" cannot list resource "secrets" in API group "" at the cluster scope
E0317 16:11:10.693767    3226 run.go:143] failed to load support bundle spec from secrets: failed to search for configmaps in the cluster: configmaps is forbidden: User "system:serviceaccount:okteto:0cc534d8-64a0-4abb-b6dd-f45e055ac018" cannot list resource "configmaps" in API group "" at the cluster scope
E0317 16:11:10.862615    3226 run.go:175] failed to load redactor specs from config maps: failed to search for secrets in the cluster: secrets is forbidden: User "system:serviceaccount:okteto:0cc534d8-64a0-4abb-b6dd-f45e055ac018" cannot list resource "secrets" in API group "" at the cluster scope
E0317 16:11:11.036565    3226 run.go:181] failed to load redactor specs from config maps: failed to search for configmaps in the cluster: configmaps is forbidden: User "system:serviceaccount:okteto:0cc534d8-64a0-4abb-b6dd-f45e055ac018" cannot list resource "configmaps" in API group "" at the cluster scope

============ Collectors summary =============
Suceeded (S), eXcluded (X), Failed (F)
=============================================
No collectors executed

============ Redactors summary =============
No redactors executed

============= Analyzers summary =============
Suceeded (S), eXcluded (X), Failed (F)
=============================================
No analyzers executed

Duration: 0ms
Error: no specs found in cluster
```

DexterYan: for the duplicate analyzer, I tried `kubectl support-bundle --load-cluster-specs cmd` in a purl kurl cluster. Seems it is fine. I uploaded analysis.json for your reference. OK, I will rewrite those deduplicate function
[analysis.json.txt](https://github.com/replicatedhq/kots/files/10997678/analysis.json.txt)

DexterYan: For deduplicatedCollectors and deduplicatedAnalyzers, I have rewriten them with more test cases. Could you also test it again?

cbodonnell: Thanks @DexterYan, I'll take a look at the updated deduplicate functions.

> I found the issue is actually that okteto forbidden user to do a search across all the namespace

This is the issue that we will want to resolve.  KOTS instances can use cluster-scoped RBAC (like kURL) or they can use namespace-scoped RBAC (only applies to existing clusters - some docs [here](https://docs.replicated.com/vendor/packaging-rbac#min-rbac) and [here](https://docs.replicated.com/enterprise/installing-general-requirements#namespace-scoped)).  If an instance is namespace-scoped, we will want to provide a `support-bundle` command that will run successfully for those users. Maybe in those cases we need to add the `-n <namespace>` flag?  There is an example of how to check if KOTS is cluster or namespace-scoped [here](https://github.com/replicatedhq/kots/blob/main/pkg/supportbundle/spec.go#L308-L320).

DexterYan: --load-cluster-specs has been rollbacked to use secret uri due to current issue existed in our `kubectl support-bundle --load-cluster-specs`. PRs have been created in troubleshoot to fix them. We would prefer to approach current change in kots without waiting for support-bundle changes

PR #3719, opened 2023-03-07
remove accent color

rianfowler:
- accent color is the same as primary color. rather than giving the vendors another override, we're just getting rid of accent color

Screenshot 2023-03-07 at 11 30 18
Screenshot 2023-03-07 at 11 30 57

Fixes # https://app.shortcut.com/replicated/story/70719/cleanup-various-parts-of-custom-branding

```release-note
none
```

none

### DIFF
--- a/web/src/scss/utilities/forms.scss
+++ b/web/src/scss/utilities/forms.scss
@@ -34,10 +34,10 @@
   }
 
   &.is-active {
-    border-color: var(--accent-color);
-    background-color: lighten($accent-color, 43.5%);
+    border-color: var(--primary-color);
+    background-color: $primary-light-color;
     &:hover {
-      border-color: var(--accent-color);
+      border-color: var(--primary-color);
     }
   }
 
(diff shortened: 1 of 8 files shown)


web/src/scss/utilities/forms.scss
```diff
@@ -34,10 +34,10 @@
   }
 
   &.is-active {
-    border-color: var(--accent-color);
-    background-color: lighten($accent-color, 43.5%);
```
rianfowler, line 38: since the value of `$accent-color` is the same as `$primary-color`, we can just use 

`$primary-light-color: lighten($primary-color, 43.5%);`

erquhart: Might have an issue if the color is already light, but that shouldn't block this PR.

erquhart (approved):

web/src/scss/utilities/forms.scss
rianfowler: Yeah I think we're going to go through and remove most of the colors from kots- there too many variations to make theming easy for the vendor. 

What I'd really like to do is jus migrate to a design system and delegate theming to that. doesn't seem like kots was built with theming in mind

PR #3732, opened 2023-03-10
add type checking to Toggle and make it a functional component

rianfowler:
- migrating component to TS 

Fixes # https://app.shortcut.com/replicated/story/60124/investigate-why-eslint-is-not-catching-components-that-are-missing-import-statements

```release-note
none
```

none


web/src/components/shared/Toggle.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+interface Props {
+  items: {
+    onClick: () => void;
+    isActive: boolean;
+    title: string;
+  }[];
+}
+
+export default function Tooltip(props: Props) {
```
rianfowler, line 12: decided to just make this an FP since it's so small. not really a fan of default exports but just leaving this as is for now

rianfowler: also dunno why this the file is Toggle and the component is called Tooltip

PR #3742, opened 2023-03-15
KOTS CLI: when app port-forward port is not available, failure message are printed once and retry in silent

pavansokkenagaraj:
when app port-forward port is not available, failure message prints repeatedly

Fixes # https://app.shortcut.com/replicated/story/65158/kots-cli-when-app-port-forward-port-is-not-available-failure-message-prints-repeatedly

image

## Steps to reproduce

1. In replicated application spec, set service/ports which don't exist/defined
image
2. deploy application
3. run `kubectl kots admin-console` to expose an admin-console on a port
4. and can see the repeatable errors
5. image

```release-note
When enabling access to admin-console using command `kubectl kots admin-console`, if application port-forward port is not available, failure message are printed once and retried in silent
```

### DIFF
--- a/pkg/k8sutil/portforward.go
+++ b/pkg/k8sutil/portforward.go
@@ -330,19 +330,20 @@ func PortForward(localPort int, remotePort int, namespace string, getPodName fun
 						prevServiceForwardErr := prevServiceForwardErrMap[desiredAdditionalPort.ServiceName]
 						if prevServiceForwardErr == nil || prevServiceForwardErr.Error() != err.Error() {
 							log.Error(errors.Wrap(err, fmt.Sprintf("failed to execute kubectl port-forward -n %s svc/%s %d:%d", namespace, desiredAdditionalPort.ServiceName, desiredAdditionalPort.LocalPort, desiredAdditionalPort.ServicePort)))
+							prevServiceForwardErrMap[desiredAdditionalPort.ServiceName] = err
 						}
-						prevServiceForwardErrMap[desiredAdditionalPort.ServiceName] = err
 						continue // try again
 					}
 					if serviceStopCh == nil {
 						// we didn't do the port forwarding, probably because the pod isn't ready.
 						// try again next loop
 						// The API doesn't return ports that aren't ready, so this is possibly rbac?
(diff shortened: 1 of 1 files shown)


pkg/k8sutil/portforward.go
```diff
@@ -326,14 +327,24 @@ func PortForward(localPort int, remotePort int, namespace string, getPodName fun
...
 					}
+					
+					prevServiceForwardErrMap[desiredAdditionalPort.ServiceName] = nil
 					if serviceStopCh == nil {
 						// we didn't do the port forwarding, probably because the pod isn't ready.
 						// try again next loop
 						// The API doesn't return ports that aren't ready, so this is possibly rbac?
-						log.Error(errors.Errorf("failed to forward port. check that you have permission to run kubectl port-forward -n %s svc/%s %d:%d", namespace, desiredAdditionalPort.ServiceName, desiredAdditionalPort.LocalPort, desiredAdditionalPort.ServicePort))
+						prevServiceForwardErr := prevServiceForwardErrMap[desiredAdditionalPort.ServiceName]
+						if prevServiceForwardErr == nil || prevServiceForwardErr.Error() != err.Error() {
```
cbodonnell, line 344: won't `prevServiceForwardErr` always be `nil` here since we [set it to `nil`](https://github.com/replicatedhq/kots/pull/3742/files#diff-0504f350f64aa3345543945b4fd6c1ffca0a16283faf1fd21e3c4ea92590ae6fR338) before this block?

pkg/k8sutil/portforward.go
```diff
@@ -326,14 +327,22 @@ func PortForward(localPort int, remotePort int, namespace string, getPodName fun
...
+							log.Error(errors.Errorf("failed to forward port. check that you have permission to run kubectl port-forward -n %s svc/%s %d:%d", namespace, desiredAdditionalPort.ServiceName, desiredAdditionalPort.LocalPort, desiredAdditionalPort.ServicePort))
+						}
+						prevServiceForwardErrMap[desiredAdditionalPort.ServiceName] = err
```
cbodonnell, line 345: will this always be set to `nil` since `err` must be `nil` to get here?  if so, then will the above log will always be printed?

pavansokkenagaraj: oh, this would have been a run time err when err is nil. Fixed in the next commit.
https://github.com/replicatedhq/kots/pull/3742/commits/4c0fb3119c78a75ae037a532b90d201ac6b781ad

cbodonnell (approved):

PR #3749, opened 2023-03-17
change gke test to not use testim

ousbots:

Change the GKE test to not use Testim, freeing up testim resources and running just a little bit faster. Apologies for not doing this when originally creating the original GKE test, but at the time I didn't understand the testing systems well and thought the e2e tests through testim was necessary.

Note that this test uses the "helm-install-order" test app because bigger ones trigger auto-scaling and potential cluster repair making the test take longer.

```release-note
NONE
```

NONE


cbodonnell (approved):
looks good to me

PR #3760, opened 2023-03-21
migrate helm release version secrets to helm namespace

pavansokkenagaraj:
Helm upgrade fails for charts installed with KOTS versions earlier than 1.95.0

Helm Release Version Secrets were created in kots installed namespace earlier to kots v1.95.0
and in kots v1.95.0 after secrets started to respect kots HelmChart namespace, 
helm upgrade expects these secrets to be their respective namespace

Fixes # https://app.shortcut.com/replicated/story/72389/helm-upgrade-fails-for-charts-installed-with-kots-versions-earlier-than-1-95-0

```release-note
fix a bug where helm release secret was created in kotsadm namespace instead helm release namespace
```

NONE

### DIFF
--- a/pkg/helm/update_release.go
+++ b/pkg/helm/update_release.go
@@ -16,7 +16,8 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
-func MigrateExistingHelmReleaseSecrets(clientset kubernetes.Interface, releaseName string, releaseNS string) error {
+// MigrateExistingHelmReleaseSecrets will move all helm release secrets from the kotsadm namespace to the release namespace
+func MigrateExistingHelmReleaseSecrets(clientset kubernetes.Interface, releaseName string, releaseNamespace string, kotsadmNamespace string) error {
 	selectorLabels := map[string]string{
 		"owner": "helm",
 		"name":  releaseName,
@@ -26,7 +27,7 @@ func MigrateExistingHelmReleaseSecrets(clientset kubernetes.Interface, releaseNa
 	}
 
 	// list all helm releases secrets for given release name
-	secretList, err := clientset.CoreV1().Secrets("").List(context.TODO(), listOpts)
+	secretList, err := clientset.CoreV1().Secrets(kotsadmNamespace).List(context.TODO(), listOpts)
 	if err != nil {
 		return errors.Wrapf(err, "failed to list release secrets for %s", releaseName)
 	}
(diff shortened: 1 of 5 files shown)


cbodonnell (dismissed):

pkg/helm/update_release.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	// set release namespace to the new namespace
+	release.Namespace = releaseNamespace
+	releaseStr, err := encodeRelease(release)
+	if err != nil {
+		return errors.Wrapf(err, "failed to encode release")
+	}
+
+	newReleaseSecret := corev1.Secret{
```
sgalsaleh, line 75: Hmm, any reason to not just use the existing secret? Who knows what Helm will add to it later. That also seems to be how this article did it https://blog.random.io/moving-helm-release-to-another-kubernetes-namespace/

pkg/helm/update_release.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+		return errors.Wrapf(err, "failed to list release secrets for %s", releaseName)
+	}
+
+	if len(secretList.Items) == 0 {
+		return nil
+	}
+
+	for _, secret := range secretList.Items {
+		if secret.Namespace != releaseNamespace {
```
sgalsaleh, line 51: this if statement is not needed, no? if you really want to have the function to be a no-op if the two namespaces match, i'd suggest you check for this at the beginning of the function and just return?

pkg/operator/client/deploy.go
```diff
@@ -522,11 +524,27 @@ func (c *Client) installWithHelm(helmDir string, targetNamespace string, kotsCha
 		installDir := filepath.Join(chartsDir, dir.Name)
 		args := []string{"upgrade", "-i", dir.ReleaseName, installDir, "--timeout", "3600s"}
 
+		upgradeNamespace := "default"
```
sgalsaleh, line 527: too many variables that refer to a namespace which is pretty confusing. my recommendation is:

- pass the `targetNamespace` to the `getSortedCharts` function and if the `dir.Namespace` is empty, set it to the `targetNamespace` (if it's not `.`.
- change the if statement below to just be `if dir.Namespace != util.AppNamespace()`. e.g.: let's not define 2 more variables that refer to a namespace

sgalsaleh: from what i can see and know, it should never be `.` when you get to this the `installWithHelm` function, so you don't need to check for that

PR #3763, opened 2023-03-24
Replace troubleshoot.io with troubleshoot.sh

xavpaice:
Relates to: https://github.com/replicatedhq/troubleshoot/issues/1070

This changes references to `troubleshoot.io` in Support Bundle types to `troubleshoot.sh`.  The reason for this is that the Troubleshoot project does not own troubleshoot.io.

Fixes [Troubleshoot #1070](https://github.com/replicatedhq/troubleshoot/issues/1070)

Note that this changes the label stored on secrets that contain support bundles.  This should be reviewed alongside https://github.com/replicatedhq/troubleshoot/pull/1075

```release-note
Support-bundle specs stored in-cluster prior to this change were stored with a label `"troubleshoot.io/kind=support-bundle"`.  The label was used as a default setting for the Troubleshoot command to search for support bundles.  As of release 0.60.0, Troubleshoot searches for `"troubleshoot.sh/kind=support-bundle"`.  The label has been updated in this change to match the new search value.
```

NONE

### DIFF
--- a/pkg/kotsadm/types/constants.go
+++ b/pkg/kotsadm/types/constants.go
@@ -13,7 +13,7 @@ const ExcludeValue = "true"
 const BackupLabel = "kots.io/backup"
 const BackupLabelValue = "velero"
 
-const TroubleshootKey = "troubleshoot.io/kind"
+const TroubleshootKey = "troubleshoot.sh/kind"
 const TroubleshootValue = "support-bundle"
 
 const DefaultSupportBundleSpecKey = "default"
--- a/pkg/kotsadm/types/constants_test.go
+++ b/pkg/kotsadm/types/constants_test.go
@@ -49,7 +49,7 @@ func Test_getTroubleshootLabels(t *testing.T) {
 				},
 			},
 			expectLabels: map[string]string{
-				"troubleshoot.io/kind": "support-bundle",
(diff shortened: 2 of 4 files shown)


pkg/kotsadm/types/constants.go
```diff
@@ -13,7 +13,7 @@ const ExcludeValue = "true"
 const BackupLabel = "kots.io/backup"
 const BackupLabelValue = "velero"
 
-const TroubleshootKey = "troubleshoot.io/kind"
+const TroubleshootKey = "troubleshoot.sh/kind"
```
banjoh, line 16: As per https://github.com/replicatedhq/troubleshoot/pull/1075#issuecomment-1482656575, we need to maintain both labels and instead do the following

* When users upgrade to a newer KOTS version, update secrets and configmaps that have `"troubleshoot.io/kind"` label to `"troubleshoot.sh/kind"` label.
* Maintain searching using both labels for sometime. Use cases such as searching for labels from objects restored from a backup need to be supported.

banjoh (changes requested):

xavpaice:
KOTS appears to be using `kots.io/kind: supportbundle` as it's label rather than the default, which is fine.  That makes this change no longer required.

PR #3767, opened 2023-03-25
write KOTS Kind manifests to the kotsKinds app archive directory

ousbots:

This creates the `kotsKinds` directory in the application archive which contains the KOTS Kind manifests that are rendered during the pull and rewrite kotsadm workflows.

The KOTS manifests, except the Installation manifest, are rendered in the `renderKotsKinds` function that mostly mimics the existing rendering behavior. The Installation manifest is created in the

This also makes all kotsv1beta1.HelmChart fields public and adds the struct to the KotsKind Marshaler for serialization.

The `LoadKotsKindsFromPath` function can now be used on the `kotsKinds` folder to load the KotsKinds struct fully rendered without further processing needed.

Fixes https://app.shortcut.com/replicated/story/63863/failed-license-upload-should-not-create-multiple-customer-instances-for-reporting

The `MappedChartValue` struct requires a custom JSON marshaler to build the value and also marshal the value without being wrapped in a struct. This provides two bonuses:
 * The written HelmChart Kots Kind is much more readable and in the same form as the upstream versions
 * Testing is easier as there doesn't need to be a custom equality for cases where one `MappedChartValue` has a built value and the other doesn't.

For example the differences in the values field output is:
With the custom marshaler:
```
values:
  field1: testvalue1
  field2: testvalue2
```
Without the custom marshaler:
```
values:
  value: 
    valueType: children
    children:
      - field1: 
          value: testvalue1
          valueType: string
      - field2:
          value: testvalue2
          valueType: string
```

## Steps to reproduce

* Create an app with some fields in the manifest templated and a config.
* Deploy the app
* Download the app archive `kubectl kots download ...`
* Verify that there is no 'kotsKinds' folder

* Deploy the app using this branch of kots
* Download the app archive `kubectl kots download ...`
* Verify that all the upstream KOTS Kind manifests created have been fully rendered and are in the 'kotsKinds' folder.

```release-note
The KOTS app archive now includes the folder 'kotsKinds' with all rendered upstream app manifests.
```

I think so, theres probably a doc about the kots file structure. I still need to create that PR.


pkg/base/replicated.go
```diff
@@ -152,7 +157,100 @@ func renderReplicated(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (
...
+		}
+
+		renderedKotsKinds.HelmCharts = append(renderedKotsKinds.HelmCharts, &kotsutil.KotsKindsFile{Path: kotsHelmChartPaths[i], Content: chartBytes})
+	}
+
+	return &commonBase, helmBases, renderedKotsKinds, nil
+}
+
+func renderKotsKinds(upstreamFiles []upstreamtypes.UpstreamFile, renderedConfig *kotsv1beta1.Config, renderOptions *RenderOptions, builder *template.Builder) (*kotsutil.KotsKindsBundle, error) {
+	kotsKinds := kotsutil.KotsKindsBundle{}
```
sgalsaleh, line 173: can't this just be a `map[string][]byte`? where the key is the file path and the value is the rendered content?

pkg/base/replicated.go
```diff
@@ -152,7 +157,100 @@ func renderReplicated(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (
...
+			// Use the rendered Config instead of the upstream.
+			configBytes, err := yaml.Marshal(renderedConfig)
+			if err != nil {
+				return nil, errors.Wrap(err, "failed to marshal rendered config")
+			}
+			kotsKinds.Config = &kotsutil.KotsKindsFile{Path: upstreamFile.Path, Content: configBytes}
+			continue
+
+		case "kots.io/v1beta1, Kind=ConfigValues":
+			// Skip the ConfigValues because the Config section was already rendered.
```
sgalsaleh, line 201: ConfigValues is different than Config. we should include the config values as well if we want to rely on loading kots kinds from this new directory later.

pkg/base/replicated.go
```diff
@@ -152,7 +157,100 @@ func renderReplicated(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (
...
+			continue
+
+		case "kots.io/v1beta1, Kind=HelmChart":
+			// Skip HelmCharts because they get rendered later.
```
sgalsaleh, line 205: can't we move rendering the HelmChart custom resources here and have them be used later instead of rendering and adding them later?

pkg/base/replicated.go
```diff
@@ -228,6 +326,10 @@ func renderReplicatedHelmChart(kotsHelmChart *kotsv1beta1.HelmChart, upstreamFil
 		return nil, errors.Wrap(err, "failed to render local values for chart")
 	}
 
+	// Update the kots helmchart to only contain the rendered values.
+	kotsHelmChart.Spec.Values = mergedValues
```
sgalsaleh, line 330: why is this needed? doesn't this change existing behavior?

pkg/base/write.go
```diff
@@ -48,6 +48,43 @@ func (b *Base) WriteBase(options WriteOptions) error {
 	return nil
 }
 
+func (b *Base) WriteKotsKinds(options WriteOptions) error {
```
sgalsaleh, line 51: this should not be a method of `Base`.

pkg/base/write.go
```diff
@@ -48,6 +48,43 @@ func (b *Base) WriteBase(options WriteOptions) error {
...
+	renderDir := filepath.Join(options.BaseDir, b.Path)
```
sgalsaleh, line 52: this shouldn't exist in the base directory. should be a separate top level `kotsKinds` directory.

pkg/kotsutil/kots.go
```diff
@@ -564,6 +570,380 @@ func LoadKotsKindsFromPath(fromDir string) (*KotsKinds, error) {
 	return &kotsKinds, nil
 }
 
+// KotsKindsBundle is used for gathering rendered KOTS Kinds manifests for writing.
+type KotsKindsBundle struct {
```
sgalsaleh, line 574: as i said earlier, this can be replaced with `map[string][]byte`

pkg/kotsutil/kots.go
```diff
@@ -564,6 +570,380 @@ func LoadKotsKindsFromPath(fromDir string) (*KotsKinds, error) {
...
+	}
+
+	if b.LintConfig != nil {
+		files = append(files, b.LintConfig)
+	}
+
+	return files
+}
+
+func (b *KotsKindsBundle) KotsKinds() (*KotsKinds, error) {
```
sgalsaleh, line 683: this can be replaced with a modified version of the `LoadKotsKindsFromPath` function

pkg/upload/archive.go
```diff
@@ -39,6 +39,11 @@ func createUploadableArchive(rootPath string) (string, error) {
 		paths = append(paths, skippedFilesPath)
 	}
 
+	kotsKindsPath := path.Join(rootPath, "kotsKinds")
```
sgalsaleh, line 42: let's move this before `skippedFiles`

pkg/store/kotsstore/version_store.go
```diff
@@ -226,6 +226,11 @@ func (s *KOTSStore) CreateAppVersionArchive(appID string, sequence int64, archiv
 		paths = append(paths, skippedFilesPath)
 	}
 
+	kotsKindsPath := filepath.Join(archivePath, "kotsKinds")
```
sgalsaleh, line 229: let's move this before `skippedFiles`

pkg/base/write.go
ousbots: I meant to clean that up before submitting the final version of this PR, sorry!

pkg/base/write.go
ousbots: I was abusing the base struct to organize and write the kots kinds data, this isn't applicable anymore, but the Basedir was 'tmpdir/kotskinds'

pkg/base/replicated.go
ousbots: yeah, and it's easier to code that way too, I was just trying to copy the style of the other code that managed file contents 🤦

pkg/kotsutil/kots.go
ousbots: 👍 I broke that functionality out into a KotsKind method

pkg/base/replicated.go
ousbots: nevermind, I forgot about the comment to move the helm rendering into the `renderKotsKinds` function, this won't apply anymore after I finish that

pkg/base/replicated.go
```diff
@@ -152,7 +157,100 @@ func renderReplicated(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (
...
+	for _, upstreamFile := range upstreamFiles {
+		document := &Document{}
+		if err := yaml.Unmarshal(upstreamFile.Content, document); err != nil {
+			continue
+		}
+
+		decode := scheme.Codecs.UniversalDeserializer().Decode
+		_, gvk, err := decode(upstreamFile.Content, nil, nil)
+		if err != nil {
+			if document.APIVersion == "kots.io/v1beta1" {
```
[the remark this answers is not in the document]
ousbots: I added the troubleshoot and velero ones, but didn't see any others, am I missing any still?

pkg/base/replicated.go
ousbots: nevermind, since it's user supplied config values, I see the value of keeping that data umerged with the vendor supplied config

kotskinds/Makefile
```diff
@@ -6,7 +6,7 @@ generate: controller-gen client-gen
 		object:headerFile=./hack/boilerplate.go.txt \
 		paths=./apis/...
 	$(CONTROLLER_GEN) \
-		crd \
+		crd:allowDangerousTypes=true \
```
sgalsaleh, line 9: why is this needed?

pkg/tests/base/cases/configcontext/kotsKinds/license.yaml
ousbots, line 1: Is this something that shouldn't be included in a git repository (the signed license field)? The test can skip license files

pkg/tests/base/cases/configlicensesideeffect/kotsKinds/license.yaml
```diff
@@ -0,0 +1,47 @@
...
...eHVSWEYxT0VkeWF5dFFVRGQyUkdSd2JFUjNjWFpQV2t4RmRYazNkamhuUm01U09WUlVSV3ByTlVvNWRuWlVTR2RtU25VemVubEVPR2xLWTBSRE5YcHFPVnh1YjFGSlJFRlJRVUpjYmkwdExTMHRSVTVFSUZCVlFreEpReUJMUlZrdExTMHRMVnh1SWl3aWEyVjVVMmxuYm1GMGRYSmxJam9pWlhsS2VtRlhaSFZaV0ZJeFkyMVZhVTlwU2pCUldIQjJXVE5LVms1NmFGaFNSMlJzVVRKb2NtTklXa1ZVVlRsRldqQktXVTFGUmtaVFJFNUZVMGhLYkUxclRUTkxNSEJFVkROR2VGTnROVVJVVlRWVlltMDFiVnBGUm5sWldIQjZaRVJqTVZaSGFFeFBXRUpVVWtacmRrd3diek5aTUZaSlVteFdWRXd5T1VoV1JXeHNWa1ZPTUZSSE1WWlJNR04zVkd4R2JGa3pTblJUUm1zMFZVWk9hMVpWU2pCVU1WbDNZbXQwY0ZSclZuQmpia0poVFZjNWFtSldiSEZaYTNob1UyeHNWV0pGUmtWWGJVWnZWakZLVUZkcWJGSmhXRVp1V2xkb1EyRnVRak5TUjNNd1lWWkpOVTVXVmxkV1ZUVnlUMGhLYjFsVlRYbGhiVGcwVjBkYWVGbHFWbFppYlhoeFpFWkZkMDU1Y3pCaFZsSkpWRVpPTm1WRk1IcGxWWFJ2VFVaR1ZtRXdWVFJSVnpsSFVsaEtVRTFZUmxCU01WcFJVMVJDTmxsV2FIcFdWWEJ0WTBSU2JFMVVRazlPVjNSU1ZucFdUMU5XWTNaU1ZYUkZVMGhzYlU5VmJGaGtNMUl3WTFWc1lXTlhSakJTYTA1RVlVWmtjbUo2VmtSU00wSllUREkxUmsxWVl6SmxWM1JKVlZoQk1sVXhTbEppU0Zwd1VrVXdNRlpFVWt0VU1rWnNVVmQwYzFSV1VrMVVWV055V1RCYVRHSXpaRTlUVm05NVlraE9SR1JzVG5aUmFrWmFaVmRPVGxOVlNteGFiRXB1Wld0U2RVMHhSVGxRVTBselNXMWtjMkl5U21oaVJYUnNaVlZzYTBscWIybFpiVkpzV2xSVk1rNVVXWGRaTWxwcFRrUk9hazlYU1hsUFIwcHRUMVJvYkZsWFRtaGFiVVV5VGtSWmFXWlJQVDBpZlE9PSJ9
```
ousbots, line 47: Same with this license, is this a secret that shouldn't be in the repository?

kotskinds/apis/kots/v1beta1/helmchart_types.go
```diff
@@ -26,7 +26,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// +kubebuilder:validation:Type=""
```
sgalsaleh, line 29: i think we need to keep this. vendors use this custom resource to define certain values to build airgap images and now they would be required to use this custom struct instead of it being any yaml key:value interface.

kotskinds/schemas/helmchart-kots-v1beta1.json
```diff
@@ -22,7 +22,37 @@
       "properties": {
         "builder": {
           "type": "object",
-          "additionalProperties": {}
+          "additionalProperties": {
```
sgalsaleh, line 25: proves my point above ^

pkg/base/replicated.go
```diff
@@ -152,7 +157,149 @@ func renderReplicated(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (
...
+			if err != nil {
+				return nil, nil, errors.Wrap(err, "failed to marshal rendered config")
+			}
+			fileContentsMap[upstreamFile.Path] = configBytes
+
+		case "kots.io/v1beta1, Kind=ConfigValues":
+			fileContentsMap["configvalues.yaml"] = upstreamFile.Content
+			kotsKinds.ConfigValues = decoded.(*kotsv1beta1.ConfigValues)
+
+		case "kots.io/v1beta1, Kind=HelmChart":
```
sgalsaleh, line 196: we shouldn't do any processing for this custom resource. it should just be the same as the others. the goal is to have visibility into the rendered version of the custom resource, and this does more than that. what if we want to check if `optionalValues` rendered correctly?

pkg/base/replicated.go
```diff
@@ -152,7 +157,149 @@ func renderReplicated(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (
...
+			}
+
+			if kotsKinds.HelmCharts == nil {
+				kotsKinds.HelmCharts = &kotsv1beta1.HelmChartList{}
+			}
+
+			kotsKinds.HelmCharts.Items = append(kotsKinds.HelmCharts.Items, decodedChart.(*kotsv1beta1.HelmChart))
+			fileContentsMap[upstreamFile.Path] = []byte(renderedChart)
+
+		default:
```
sgalsaleh, line 242: would this basically trigger for every single file in the upstream? that seems like wasted performance to render and this discard.

pkg/base/replicated.go
```diff
@@ -152,7 +157,149 @@ func renderReplicated(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (
...
+	kotsKinds := kotsutil.EmptyKotsKinds()
+
+	for _, upstreamFile := range upstreamFiles {
+		document := &Document{}
+		if err := yaml.Unmarshal(upstreamFile.Content, document); err != nil {
+			continue
+		}
+
+		decode := scheme.Codecs.UniversalDeserializer().Decode
+		decoded, gvk, err := decode(upstreamFile.Content, nil, nil)
```
sgalsaleh, line 174: instead of decoding, i would suggest utilizing [this function](https://github.com/replicatedhq/kots/blob/237fa185de3a4dde7fca15c4f0b6635b99a1bbda/pkg/base/base.go#L279) and `continue` if it's not a kots kind.

pkg/base/replicated.go
```diff
@@ -152,7 +157,149 @@ func renderReplicated(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (
...
+		if err != nil {
+			if document.APIVersion == "kots.io/v1beta1" || document.APIVersion == "troubleshoot.sh/v1beta2" || document.APIVersion == "velero.io/v1" {
+				return nil, nil, errors.Wrapf(err, "failed to decode kots yaml %s", upstreamFile.Path)
+			}
+			continue
+		}
+
+		switch gvk.String() {
```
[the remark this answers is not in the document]
sgalsaleh: the config is a special case, so you do need to check that and not render it (like you currently do)

pkg/kotsutil/kots.go
```diff
@@ -564,6 +585,53 @@ func LoadKotsKindsFromPath(fromDir string) (*KotsKinds, error) {
 	return &kotsKinds, nil
 }
 
+func KotsKindsFromMap(kotsKindsMap map[string][]byte) (*KotsKinds, error) {
```
sgalsaleh, line 588: is this only used for testing? i'd suggest writing the files to a directory and calling the `LoadKotsKindsFromPath` so that it uses the same workflow other callers do.

pkg/base/replicated.go
ousbots: yeah, fixed now. originally this upstreamFiles input was filtered to be only kots manifests and I forgot to move the filter into this function during one of the rewrites 🤦

pkg/base/replicated.go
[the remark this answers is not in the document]
ousbots: I removed the rendering of the `kotsutil.KotsKinds` struct from this `renderKotsKinds` function since it wasn't being used. It can be generated with the `kotsutil.KotsKindsFromMap` function if needed after this function call and avoids writing and parsing extra files

pkg/kotsutil/kots.go
```diff
@@ -396,6 +405,74 @@ func (o KotsKinds) Marshal(g string, v string, k string) (string, error) {
 	return "", errors.Errorf("unknown gvk %s/%s, Kind=%s", g, v, k)
 }
 
+func (k *KotsKinds) addKotsKind(content []byte) error {
```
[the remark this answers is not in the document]
ousbots: This function is used by the `kotsutil.KotsKindsFromMap` function which is useful for testing and could also be used to convert the output of `base.renderKotsKinds` for later use whenever consolidating all the KotsKinds loading happens

pkg/base/replicated.go
sgalsaleh: hmm are they not merged in with the vendor's defaults at this point?

kotskinds/schemas/helmchart-kots-v1beta1.json
sgalsaleh: this file needs to be regenerated

pkg/base/replicated.go
```diff
@@ -152,7 +156,54 @@ func renderReplicated(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (
...
+
+func renderKotsKinds(upstreamFiles []upstreamtypes.UpstreamFile, renderedConfig *kotsv1beta1.Config, renderOptions *RenderOptions, builder *template.Builder) (map[string][]byte, error) {
+	fileContentsMap := make(map[string][]byte)
+
+	for _, upstreamFile := range upstreamFiles {
+		gvk := OverlySimpleGVK{}
+		if err := yaml.Unmarshal(upstreamFile.Content, &gvk); err != nil {
+			continue
+		}
+		gvkString := gvk.APIVersion + "," + gvk.Kind
```
sgalsaleh, line 170: the switch statement below includes a space, but this doesn't seem to. i would also suggest using `fmt`:
```suggestion
		gvkString := fmt.Sprintf("%s, %s", gvk.APIVersion, gvk.Kind)
```

pkg/base/replicated.go
```diff
@@ -202,28 +253,7 @@ func renderReplicatedHelmChart(kotsHelmChart *kotsv1beta1.HelmChart, upstreamFil
 	}
 	helmUpstream.Name = kotsHelmChart.GetDirName()
 
-	mergedValues := kotsHelmChart.Spec.Values
```
sgalsaleh, line 205: we should put this back since it's not done anymore when the kots kinds are rendered, right?

pkg/kotsutil/kots.go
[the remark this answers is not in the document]
sgalsaleh: i don't like the fact that we are reiterating all the kots kinds again. this will have to be maintained/updated whenever we add a new kind. how about iterating over the map, writing the files to a tmp directory, and calling the `LoadKotsKindsFromPath` function instead?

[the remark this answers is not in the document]
sgalsaleh: oh nevermind, i just noticed you moved things here instead. i'll resolve.

pkg/kotsutil/kots.go
```diff
@@ -564,6 +585,53 @@ func LoadKotsKindsFromPath(fromDir string) (*KotsKinds, error) {
...
+		}
+	}
+
+	return &kotsKinds, nil
+}
+
+func WriteKotsKinds(kotsKinds map[string][]byte, rootDir string, overwrite bool) error {
+	_, err := os.Stat(rootDir)
+	if err == nil {
+		if overwrite {
```
sgalsaleh, line 603: this should always be true. there shouldn't be an option to _not_ overwrite. always cleanup/remove the existing directory.

pkg/kotsutil/kots.go
```diff
@@ -564,6 +585,53 @@ func LoadKotsKindsFromPath(fromDir string) (*KotsKinds, error) {
...
+			if err := os.RemoveAll(rootDir); err != nil {
+				return errors.Wrap(err, "failed to remove previous content in base")
+			}
+		} else {
+			return fmt.Errorf("directory %s already exists", rootDir)
+		}
+	}
+
+	if os.IsNotExist(err) {
```
sgalsaleh, line 612: this check isn't needed with my above suggestion

pkg/kotsutil/kots.go
```diff
@@ -396,6 +405,74 @@ func (o KotsKinds) Marshal(g string, v string, k string) (string, error) {
...
+			k.License = decoded.(*kotsv1beta1.License)
+		case "kots.io/v1beta1, Kind=Identity":
+			k.Identity = decoded.(*kotsv1beta1.Identity)
+		case "kots.io/v1beta1, Kind=IdentityConfig":
+			k.IdentityConfig = decoded.(*kotsv1beta1.IdentityConfig)
+		case "kots.io/v1beta1, Kind=Installation":
+			k.Installation = *decoded.(*kotsv1beta1.Installation)
+		case "kots.io/v1beta1, Kind=HelmChart":
+			if k.HelmCharts == nil {
+				k.HelmCharts = &kotsv1beta1.HelmChartList{}
```
sgalsaleh, line 447: why not just use the same type defined in the KotsKinds struct:
```suggestion
				k.HelmCharts = []*kotsv1beta1.HelmChart{}
```
that way you don't have to worry about using `.Items`

sgalsaleh: you actually don't even have to check if it's nil. the previous code should just work, no?
```
kotsKinds.HelmCharts = append(kotsKinds.HelmCharts, decoded.(*kotsv1beta1.HelmChart))
```

pkg/base/render.go
```diff
@@ -27,16 +27,16 @@ type RenderOptions struct {
...
-func RenderUpstream(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (base *Base, helmBases []Base, err error) {
+func RenderUpstream(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (base *Base, helmBases []Base, kotsKinds map[string][]byte, err error) {
 	if u.Type == "helm" {
 		base, err = RenderHelm(u, renderOptions)
 		return
 	}
 
 	if u.Type == "replicated" {
-		base, helmBases, err = renderReplicated(u, renderOptions)
+		base, helmBases, kotsKinds, err = renderReplicated(u, renderOptions)
```
sgalsaleh, line 37:
```suggestion
		base, helmBases, renderedKotsKinds, err = renderReplicated(u, renderOptions)
```

pkg/base/render.go
```diff
@@ -27,16 +27,16 @@ type RenderOptions struct {
 
 // RenderUpstream is responsible for any conversions or transpilation steps are required
 // to take an upstream and make it a valid kubernetes base
-func RenderUpstream(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (base *Base, helmBases []Base, err error) {
+func RenderUpstream(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (base *Base, helmBases []Base, kotsKinds map[string][]byte, err error) {
```
sgalsaleh, line 30:
```suggestion
func RenderUpstream(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (base *Base, helmBases []Base, renderedKotsKinds map[string][]byte, err error) {
```

kotskinds/Makefile
ousbots: removed now, but was added when the MappedChartValue struct's fields were exported (which was a bad idea)

pkg/tests/base/cases/configcontext/kotsKinds/license.yaml
ousbots: nevermind, there are already a few licenses in the test files

kotskinds/schemas/helmchart-kots-v1beta1.json
ousbots: I'm not sure why this is still here, I've regenerated these files and the only place these additional properties appears is in the `MappedChartValue` struct and the `// +kubebuilder:validation:Type=""` notation doesn't seem to affect the generation.

pkg/kotsutil/kots.go
ousbots: Regretfully I think there's no other way to add generated k8s apimachinery types and methods to the `[]*kotsv1beta1.HelmChart{}` type.

kotskinds/apis/kots/v1beta1/helmchart_types.go
```diff
@@ -40,6 +40,15 @@ type MappedChartValue struct {
...
 
+func (m MappedChartValue) MarshalJSON() ([]byte, error) {
+	val, err := m.getBuiltValue()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get built value")
+	}
+
+	return json.Marshal(val)
+}
+
```
ousbots, line 51: this is the only change to the apis, so I'm not sure why the `additionalProperties` on the helmchart schema are being generated.

pkg/base/replicated.go
```diff
@@ -152,7 +157,62 @@ func renderReplicated(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (
...
+
+func renderKotsKinds(upstreamFiles []upstreamtypes.UpstreamFile, renderedConfig *kotsv1beta1.Config, renderOptions *RenderOptions, builder *template.Builder) (map[string][]byte, error) {
+	fileContentsMap := make(map[string][]byte)
+
+	for _, upstreamFile := range upstreamFiles {
+		gvk := OverlySimpleGVK{}
+		if err := yaml.Unmarshal(upstreamFile.Content, &gvk); err != nil {
+			continue
+		}
+		gvkString := fmt.Sprintf("%s,Kind=%s", gvk.APIVersion, gvk.Kind)
```
sgalsaleh, line 171: nit pick: can we just follow the `gvk.String()` pattern here and have a space like so:
```suggestion
		gvkString := fmt.Sprintf("%s, Kind=%s", gvk.APIVersion, gvk.Kind)
```

pkg/base/replicated.go
```diff
@@ -152,7 +157,62 @@ func renderReplicated(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (
...
-	return &commonBase, helmBases, nil
+	return &commonBase, helmBases, renderedKotsKinds, nil
+}
+
+func renderKotsKinds(upstreamFiles []upstreamtypes.UpstreamFile, renderedConfig *kotsv1beta1.Config, renderOptions *RenderOptions, builder *template.Builder) (map[string][]byte, error) {
+	fileContentsMap := make(map[string][]byte)
+
+	for _, upstreamFile := range upstreamFiles {
+		gvk := OverlySimpleGVK{}
+		if err := yaml.Unmarshal(upstreamFile.Content, &gvk); err != nil {
```
sgalsaleh, line 168: what if the file is a multi-yaml document that contains multiple kots kinds? when loading kots kinds, we use `util.ConvertToSingleDocs` to convert to single docs.

pkg/base/replicated.go
```diff
@@ -206,24 +266,27 @@ func renderReplicatedHelmChart(kotsHelmChart *kotsv1beta1.HelmChart, upstreamFil
...
 		} else {
-			for k, v := range optionalValues.Values {
+			for k, v := range optionalValue.Values {
 				mergedValues[k] = v
 			}
 		}
 	}
 
-	helmValues, err := kotsHelmChart.Spec.GetHelmValues(mergedValues)
+	helmValues, err := kotsHelmChart.Spec.GetHelmValues(kotsHelmChart.Spec.Values)
```
sgalsaleh, line 289: is this right? shouldn't this be:
```suggestion
	helmValues, err := kotsHelmChart.Spec.GetHelmValues(mergedValues)
```

sgalsaleh: i'm worried if all tests pass with the current change

pkg/pull/pull.go
```diff
@@ -277,7 +277,8 @@ func Pull(upstreamURI string, pullOptions PullOptions) (string, error) {
 		IsGKEAutopilot:      k8sutil.IsGKEAutopilot(clientset),
 		IncludeMinio:        pullOptions.IncludeMinio,
 	}
-	if err := upstream.WriteUpstream(u, writeUpstreamOptions); err != nil {
+	installationManifest, err := upstream.WriteUpstream(u, writeUpstreamOptions)
```
sgalsaleh, line 280: the `Installation` manifest gets updated later in the function when processing and writing the midstream. the `WriteUpstream` shouldn't return a manifest here, we should just run `kotsKinds.Marshal(....)` for the installation crd at the end before writing the kots kinds.

pkg/base/replicated.go
ousbots: I only added expected kotskinds to the existing tests and didn't modify any of the cases. It looks like there aren't any tests that use HelmChart.Spec.OptionalValues, `rg OptionalValue pkg/tests` returns nothing.

pkg/base/replicated.go
ousbots: good call, changed it to use the utility doc splitting function and loop over the results

pkg/pull/pull.go
```diff
@@ -277,7 +277,8 @@ func Pull(upstreamURI string, pullOptions PullOptions) (string, error) {
 		IsGKEAutopilot:      k8sutil.IsGKEAutopilot(clientset),
 		IncludeMinio:        pullOptions.IncludeMinio,
 	}
-	if err := upstream.WriteUpstream(u, writeUpstreamOptions); err != nil {
+	err = upstream.WriteUpstream(u, writeUpstreamOptions)
```
sgalsaleh, line 280: why is this change necessary? felt cleaner when the `err` was assigned inline

kotskinds/apis/kots/v1beta1/installation_types.go
```diff
@@ -61,6 +61,23 @@ type Installation struct {
 	Status InstallationStatus `json:"status,omitempty"`
 }
 
+// Merge known images by overwriting existing image entries with newer ones.
+func (i *Installation) MergeKnownImages(newImages []InstallationImage) {
```
sgalsaleh, line 65: instead of this function, wouldn't it be simpler to add a function [here](https://github.com/replicatedhq/kots/blob/754fcd7d5c759713a1a2ada99920eb4ba657f193/pkg/upstream/installation.go) that reads the installation file before writing kots kinds?

pkg/pull/pull.go
ousbots: my bad, I wasn't looking at the overall diff, just making progressive changes where I modified this line, then later modified it back to "original", but accidentally with a different style

pkg/base/replicated.go
```diff
@@ -152,7 +157,64 @@ func renderReplicated(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (
...
+
+			switch gvkString {
+			case "kots.io/v1beta1, Kind=Config":
+				// Use the rendered Config instead of the upstream.
+				kotsKinds := kotsutil.KotsKinds{Config: renderedConfig}
+				configBytes, err := kotsKinds.Marshal("kots.io", "v1beta1", "Config")
+				if err != nil {
+					return nil, errors.Wrap(err, "failed to marshal rendered config")
+				}
+				fileContentsMap[upstreamFile.Path] = []byte(configBytes)
```
sgalsaleh, line 186: since `configvalues.yaml` is hardcoded, i don't see why this can't just be:
```suggestion
				fileContentsMap["config.yaml"] = []byte(configBytes)
```

sgalsaleh: actually. never mind.

pkg/base/replicated_test.go
```diff
@@ -588,14 +592,84 @@ metadata:
...
+        secrets:
+          secretName-1: MTIz
+          secretName-2: MTIz
+          secretName-3: MTIz
+    name: secrets
+    title: Secrets
+status: {}
+`,
+				),
+				"userdata/configvalues.yaml": []byte(`apiVersion: kots.io/v1beta1
```
sgalsaleh, line 646: i thought that the config values are written at the top level directory as `configvalues.yaml` https://github.com/replicatedhq/kots/pull/3767/files#diff-b8cf2cbeb02e811d9f2e22ca6e1a73949feb7d771328121b1d1b4404007631d6R190

why are they under `userdata/` here?

pkg/base/replicated_test.go
```diff
@@ -937,6 +1046,51 @@ spec:
...
+    chartVersion: 10.13.8
+  helmVersion: v3
+  useHelmInstall: true
+  weight: 42
+  exclude: ""
+  values:
+    postgresqlPassword: abc123
+`,
+				),
+				"userdata/config.yaml": []byte(`apiVersion: kots.io/v1beta1
```
sgalsaleh, line 1083: ditto

pkg/base/replicated.go
```diff
@@ -152,7 +157,64 @@ func renderReplicated(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (
...
 
-	return &commonBase, helmBases, nil
+	return &commonBase, helmBases, renderedKotsKinds, nil
+}
+
+func renderKotsKinds(upstreamFiles []upstreamtypes.UpstreamFile, renderedConfig *kotsv1beta1.Config, renderOptions *RenderOptions, builder *template.Builder) (map[string][]byte, error) {
+	fileContentsMap := make(map[string][]byte)
+
+	for _, upstreamFile := range upstreamFiles {
+		for _, doc := range util.ConvertToSingleDocs(upstreamFile.Content) {
```
sgalsaleh, line 167: i wonder if Go would evaluate `ConvertToSingleDocs` on every iteration

pkg/pull/pull.go
```diff
@@ -277,7 +277,7 @@ func Pull(upstreamURI string, pullOptions PullOptions) (string, error) {
 		IsGKEAutopilot:      k8sutil.IsGKEAutopilot(clientset),
 		IncludeMinio:        pullOptions.IncludeMinio,
 	}
-	if err := upstream.WriteUpstream(u, writeUpstreamOptions); err != nil {
+	if err = upstream.WriteUpstream(u, writeUpstreamOptions); err != nil {
```
sgalsaleh, line 280:
```suggestion
	if err := upstream.WriteUpstream(u, writeUpstreamOptions); err != nil {
```

pkg/rewrite/rewrite.go
```diff
@@ -320,6 +321,21 @@ func Rewrite(rewriteOptions RewriteOptions) error {
 		return errors.Wrap(err, "failed to write rendered")
 	}
 
+	installationManifest, err := kotsutil.LoadInstallationFromPath(filepath.Join(u.GetOverlaysDir(writeUpstreamOptions), "userdata", "installation.yaml"))
```
sgalsaleh, line 324: ditto

pkg/pull/pull.go
```diff
@@ -557,6 +557,22 @@ func Pull(upstreamURI string, pullOptions PullOptions) (string, error) {
 		}
 	}
 
+	installationManifest, err := kotsutil.LoadInstallationFromPath(filepath.Join(u.GetUpstreamDir(writeUpstreamOptions), "userdata", "installation.yaml"))
```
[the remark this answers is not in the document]
ousbots: uhhh ... I think this was just a thoughtless change, a function to load the installation manifest existed, so I just used it and adjusted the output without thinking what that really meant 🤦

pkg/base/replicated.go
```diff
@@ -152,7 +157,64 @@ func renderReplicated(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (
...
+				kotsKinds := kotsutil.KotsKinds{Config: renderedConfig}
+				configBytes, err := kotsKinds.Marshal("kots.io", "v1beta1", "Config")
+				if err != nil {
+					return nil, errors.Wrap(err, "failed to marshal rendered config")
+				}
+				fileContentsMap[upstreamFile.Path] = []byte(configBytes)
+
+			case "kots.io/v1beta1, Kind=ConfigValues":
+				// ConfigValues do not need rendering since they should already be valid values.
+				fileContentsMap["configvalues.yaml"] = upstreamFile.Content
```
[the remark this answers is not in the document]
ousbots: I think this was an attempt to avoid the "userdata" subfolder, but I don't really remember and you mentioned a much better way of generating unique names

pkg/base/replicated_test.go
ousbots: The tests actually convert the raw data to a KotsKind struct for comparison to avoid issues like different yaml list ordering, so these filenames are ignored, but I updated them for consistency

pkg/kotsutil/kots.go
```diff
@@ -272,6 +273,14 @@ func (o KotsKinds) Marshal(g string, v string, k string) (string, error) {
...
 				}
 				return string(b.Bytes()), nil
+			case "HelmChart":
+				if o.HelmCharts == nil {
+					return "", nil
+				}
+				var b bytes.Buffer
+				if err := s.Encode(o.HelmCharts, &b); err != nil {
+					return "", errors.Wrap(err, "failed to encode helmchart")
+				}
```
cbodonnell, line 283: Do we need a `return string(b.Bytes()), nil` here?

cbodonnell: I also called this case `"HelmChartList"` since that's the type that is being marshalled.

pkg/base/replicated.go
cbodonnell: it shouldn't as it's evaluated before iteration begins:
```
for _, doc := range util.ConvertToSingleDocs(upstreamFile.Content)
```
https://go.dev/play/p/ySZvs5lt_4k

pkg/kotsutil/kots.go
ousbots: definitely, great catch

pkg/base/replicated.go
sgalsaleh: yeah, looks like this improvement was added in Go 1.17

pkg/kotsutil/kots.go
```diff
@@ -396,6 +406,90 @@ func (o KotsKinds) Marshal(g string, v string, k string) (string, error) {
 	return "", errors.Errorf("unknown gvk %s/%s, Kind=%s", g, v, k)
 }
 
+func (k *KotsKinds) addKotsKind(content []byte) error {
```
sgalsaleh, line 409: plural would be more accurate
```suggestion
func (k *KotsKinds) addKotsKinds(content []byte) error {
```

pkg/pull/pull.go
```diff
@@ -557,6 +557,21 @@ func Pull(upstreamURI string, pullOptions PullOptions) (string, error) {
 		}
 	}
 
+	installationBytes, err := ioutil.ReadFile(filepath.Join(u.GetUpstreamDir(writeUpstreamOptions), "userdata", "installation.yaml"))
+	if err != nil {
+		return "", errors.Wrap(err, "failed to read installation file")
+	}
+
+	// Ensure the installation manifest filename is unique.
```
sgalsaleh, line 565: no need. vendors can't/won't create installation manifests. let's just use `installation.yaml` like we do in the `upstream` directory.

pkg/rewrite/rewrite.go
```diff
@@ -101,7 +101,8 @@ func Rewrite(rewriteOptions RewriteOptions) error {
 		IsOpenShift:          k8sutil.IsOpenShift(clientset),
 		IsGKEAutopilot:       k8sutil.IsGKEAutopilot(clientset),
 	}
-	if err := upstream.WriteUpstream(u, writeUpstreamOptions); err != nil {
+	err = upstream.WriteUpstream(u, writeUpstreamOptions)
```
sgalsaleh, line 104: nit pick: i'd the previous format:
```suggestion
	if err := upstream.WriteUpstream(u, writeUpstreamOptions); err != nil {
```

pkg/rewrite/rewrite.go
```diff
@@ -320,6 +321,20 @@ func Rewrite(rewriteOptions RewriteOptions) error {
 		return errors.Wrap(err, "failed to write rendered")
 	}
 
+	installationBytes, err := ioutil.ReadFile(filepath.Join(u.GetUpstreamDir(writeUpstreamOptions), "userdata", "installation.yaml"))
+	if err != nil {
+		return errors.Wrap(err, "failed to read installation file")
+	}
+
+	// Ensure the installation manifest filename is unique.
```
sgalsaleh, line 329: ditto

pkg/rewrite/rewrite.go
```diff
@@ -320,6 +320,18 @@ func Rewrite(rewriteOptions RewriteOptions) error {
...
 
+	installationBytes, err := ioutil.ReadFile(filepath.Join(u.GetUpstreamDir(writeUpstreamOptions), "userdata", "installation.yaml"))
+	if err != nil {
+		return errors.Wrap(err, "failed to read installation file")
+	}
+
+	installationFilename := kotsutil.GenUniqueKotsKindFilename(renderedKotsKinds, "installation")
+	renderedKotsKinds[installationFilename] = []byte(installationBytes)
+
+	if err := kotsutil.WriteKotsKinds(renderedKotsKinds, u.GetKotsKindsDir(writeUpstreamOptions)); err != nil {
```
sgalsaleh, line 331: let's be consistent and move this to be done before `WriteRenderedApp` since that's how it's done in `pull`?

PR #3771, opened 2023-03-27
report when gitops is enabled/disabled

pavansokkenagaraj:
report when gitops is enabled/disabled

Fixes # https://app.shortcut.com/replicated/story/55917/report-when-gitops-is-enabled

```release-note
report when gitops is enabled/disabled
```

NONE

### DIFF
--- a/pkg/api/reporting/types/types.go
+++ b/pkg/api/reporting/types/types.go
@@ -11,6 +11,12 @@ type ReportingInfo struct {
 	K8sVersion         string         `json:"k8s_version"`
 	KOTSInstallID      string         `json:"kots_install_id"`
 	KURLInstallID      string         `json:"kurl_install_id"`
+	GitOpsReport       *GitOpsReport  `json:"gitops_report"`
+}
+
+type GitOpsReport struct {
+	IsEnabled bool   `json:"is_enabled"`
+	Provider  string `json:"provider"`
 }
 
 type DownstreamInfo struct {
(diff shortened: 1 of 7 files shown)


pkg/reporting/app.go
```diff
@@ -327,3 +330,19 @@ func getDownstreamInfo(appID string) (*types.DownstreamInfo, error) {
...
+	gitOpsConfig, err := gitops.GetDownstreamGitOpsConfig(clientset, appID, clusterID)
+	if err != nil {
+		logger.Debugf("failed to get gitops config: %v", err.Error())
+		return nil
+	}
+
+	if gitOpsConfig != nil {
+		return &types.GitOpsReport{
+			Provider:  gitOpsConfig.Provider,
+			IsEnabled: gitOpsConfig.IsConnected,
```
sgalsaleh, line 344: do we care if it's connected or not? isn't the criteria here to report if gitops is configured? if so, we might need to change this.

pkg/api/reporting/types/types.go
```diff
@@ -11,6 +11,12 @@ type ReportingInfo struct {
 	K8sVersion         string         `json:"k8s_version"`
 	KOTSInstallID      string         `json:"kots_install_id"`
 	KURLInstallID      string         `json:"kurl_install_id"`
+	GitOpsReport       *GitOpsReport  `json:"gitops_report"`
+}
+
+type GitOpsReport struct {
```
sgalsaleh, line 17: can we not create a new type for this and just have it be part of the original `ReportingInfo` type?

PR #3793, opened 2023-04-10
upgrade to kustomize 5

cbodonnell:

Updates Kustomize from v4.5.7 to v5.0.1 which resolves CVE-2022-27664, CVE-2022-41723, CVE-2022-41723, and CVE-2022-28948 with high severity and CVE-2022-41717 with medium severity

Fixes [SC-69470](https://app.shortcut.com/replicated/story/69470/cve-2022-28948-aka-upgrade-to-kustomize-5-x)

Added a pod to the `pkg/tests/pull/cases/subchart-crds` test case chart at `templates/my-chart-pod.yaml` with the following contents:
```
apiVersion: v1
kind: Pod
metadata:
  name: my-chart-pod
spec:
  containers:
  - name: my-chart-container
    image: nginx
    ports:
    - containerPort: 80
  initContainers: null
```

#### Steps to reproduce

Run the updated unit test with and without the code changes.  Without, it will fail with the following:
```
Error: updating name reference in 'spec/initContainers/env/valueFrom/configMapKeyRef/name' field of 'Pod.v1.[noGrp]/my-chart-pod.[noNs]': considering field 'spec/initContainers/env/valueFrom/configMapKeyRef/name' of object Pod.v1.[noGrp]/my-chart-pod.[noNs]: expected sequence or mapping node\n
```

```release-note
Updates Kustomize from v4.5.7 to v5.0.1 which resolves CVE-2022-27664, CVE-2022-41723, CVE-2022-41723, and CVE-2022-28948 with high severity and CVE-2022-41717 with medium severity.
```

NONE


pkg/base/helm.go
```diff
@@ -434,116 +435,38 @@ func getAllBasePaths(prefix string, base Base) []string {
...
-			updatedBaseFiles = append(updatedBaseFiles, baseFile)
-			continue // ignore invalid resources
-		}
-
-		// ignore crds
-		if manifest.Kind == "CustomResourceDefinition" {
-			updatedBaseFiles = append(updatedBaseFiles, baseFile)
+	decoder := serializer.NewCodecFactory(scheme.Scheme).UniversalDeserializer()
+	for i := range baseFiles {
+		obj, gvk, err := decoder.Decode(baseFiles[i].Content, nil, nil)
```
sgalsaleh, line 445: what if we don't have the right schema loaded to decode the manifests. were there any issues using the `GetGVKWithNameAndNs` function?

sgalsaleh: Ah, I see you're using this below to inject the namespace. But still, the issue remains.

cbodonnell: good point.  moved away from decoding objects to simply deleting `nil` map entries.  these were what was causing `null` to get populated in it's place (which I believe is still valid yaml, but Kustomize does not like it)

cbodonnell:
Copying PR feedback: We need to run this logic for non-helm charts when _creating_ the rendered directory: https://github.com/replicatedhq/kots/blob/6e358766c5c424a757ff1361bb87191438c95ad8/pkg/kustomize/app.go#L34-L54

pavansokkenagaraj (approved):
LGTM.
Is there a way to run and detect nil fields  in a single pass before calling `kotsutil.RemoveEmptyMappingFields` ?

cbodonnell:
> Is there a way to run and detect nil fields in a single pass before calling `kotsutil.RemoveEmptyMappingFields` ?

We can maybe try to do a text search first, but that also could fail to detect if spacing/indentations are different.  I cannot think about a great way to handle it without unmarshalling the yaml.

sgalsaleh (approved):

PR #3800, opened 2023-04-12
Validation of Config Options 

pavansokkenagaraj:
add validation for config options

Fixes # https://app.shortcut.com/replicated/story/28650/kotsadm-validation-of-config-options-entered-by-user

1. add a new validation struct ConfigOptions CRD
```yaml
---
apiVersion: kots.io/v1beta1
kind: Config
metadata:
  name: config-sample
spec:
  groups:
  - name: example_settings
    title: My Example Configs
    items:
    - name: password
      title: password
      type: password
      validation:
        regex:
          pattern: "p([a-z]+)ch"
          message: "type peach and be happy"
```
2. Add regex validation for the item 
3. Item validation is done only for types `text`, `textArea`, `password`, `file`
4. Validate item only if item is not hidden, when=false and not repeatable item
5. Validation error messages have the below format:
```json
{
	"validationErrors": [
		{
			"name": "example_settings",
			"title": "My Example Configs",
			"item_errors": [
				{
					"name": "email_text",
					"type": "text",
					"value": "",
					"validation_errors": [
						{
							"validation_error_message": "Value does not match regex",
							"regex": {
								"pattern": "^(?:(?:(?:(?:[a-zA-Z]|\\d|[!#\\$%\u0026'\\*\\+\\-\\/=\\?\\^_`{\\|}~]|[\\x{00A0}-\\x{D7FF}\\x{F900}-\\x{FDCF}\\x{FDF0}-\\x{FFEF}])+(?:\\.([a-zA-Z]|\\d|[!#\\$%\u0026'\\*\\+\\-\\/=\\?\\^_`{\\|}~]|[\\x{00A0}-\\x{D7FF}\\x{F900}-\\x{FDCF}\\x{FDF0}-\\x{FFEF}])+)*)|(?:(?:\\x22)(?:(?:(?:(?:\\x20|\\x09)*(?:\\x0d\\x0a))?(?:\\x20|\\x09)+)?(?:(?:[\\x01-\\x08\\x0b\\x0c\\x0e-\\x1f\\x7f]|\\x21|[\\x23-\\x5b]|[\\x5d-\\x7e]|[\\x{00A0}-\\x{D7FF}\\x{F900}-\\x{FDCF}\\x{FDF0}-\\x{FFEF}])|(?:(?:[\\x01-\\x09\\x0b\\x0c\\x0d-\\x7f]|[\\x{00A0}-\\x{D7FF}\\x{F900}-\\x{FDCF}\\x{FDF0}-\\x{FFEF}]))))*(?:(?:(?:\\x20|\\x09)*(?:\\x0d\\x0a))?(\\x20|\\x09)+)?(?:\\x22))))@(?:(?:(?:[a-zA-Z]|\\d|[\\x{00A0}-\\x{D7FF}\\x{F900}-\\x{FDCF}\\x{FDF0}-\\x{FFEF}])|(?:(?:[a-zA-Z]|\\d|[\\x{00A0}-\\x{D7FF}\\x{F900}-\\x{FDCF}\\x{FDF0}-\\x{FFEF}])(?:[a-zA-Z]|\\d|-|\\.|~|[\\x{00A0}-\\x{D7FF}\\x{F900}-\\x{FDCF}\\x{FDF0}-\\x{FFEF}])*(?:[a-zA-Z]|\\d|[\\x{00A0}-\\x{D7FF}\\x{F900}-\\x{FDCF}\\x{FDF0}-\\x{FFEF}])))\\.)+(?:(?:[a-zA-Z]|[\\x{00A0}-\\x{D7FF}\\x{F900}-\\x{FDCF}\\x{FDF0}-\\x{FFEF}])|(?:(?:[a-zA-Z]|[\\x{00A0}-\\x{D7FF}\\x{F900}-\\x{FDCF}\\x{FDF0}-\\x{FFEF}])(?:[a-zA-Z]|\\d|-|\\.|~|[\\x{00A0}-\\x{D7FF}\\x{F900}-\\x{FDCF}\\x{FDF0}-\\x{FFEF}])*(?:[a-zA-Z]|[\\x{00A0}-\\x{D7FF}\\x{F900}-\\x{FDCF}\\x{FDF0}-\\x{FFEF}])))\\.?$",
								"message": "Email is invalid"
							}
						}
					]
				}
			]
		}
	]
}
```
6. `kuebctl kots set config` will return the validation error in the below format
image

```release-note
- Adds support for validating the value of a config item. An item of type `text`, `textarea`, `password`, and `file` can now be matched against a regex pattern. For more information, please refer to the [Config Validation](/reference/custom-resource-config#validation) page.
```

YES 
https://github.com/replicatedhq/replicated-docs/pull/1049
https://github.com/replicatedhq/replicated-docs/pull/1071


cmd/kots/cli/set-config.go
```diff
@@ -132,13 +134,26 @@ func SetConfigCmd() *cobra.Command {
...
-				Error string `json:"error"`
+				Error            string                                        `json:"error"`
+				ValidationErrors []kotsadmvalidator.ConfigGroupValidationError `json:"validationErrors,omitempty"`
 			}{}
 			_ = json.Unmarshal(respBody, &response)
 
 			if resp.StatusCode != http.StatusOK {
 				if resp.StatusCode == http.StatusNotFound {
 					return errors.Errorf("app with slug %s not found", appSlug)
+				} else if resp.StatusCode == http.StatusBadRequest {
```
sgalsaleh, line 145: this check is not necessary. can just check for the existence of validation errors.

cmd/kots/cli/set-config.go
```diff
@@ -132,13 +134,26 @@ func SetConfigCmd() *cobra.Command {
...
+					if len(response.ValidationErrors) > 0 {
+						validationErrors, err := yaml.Marshal(response.ValidationErrors)
+						if err != nil {
+							return errors.Wrap(err, "failed to marshal config")
+						}
+						fmt.Println(response.Error)
+						fmt.Print(string(validationErrors))
```
sgalsaleh, line 152: let's use our logger that's already initialized with `log` instead of `fmt`

pkg/handlers/config.go
```diff
@@ -635,11 +648,11 @@ func (h *Handler) CurrentAppConfig(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if renderedConfig != nil {
-		configGroups = renderedConfig.Spec.Groups
+		currentAppConfigResponse.ConfigGroups = renderedConfig.Spec.Groups
+		currentAppConfigResponse.ValidationErrors = kotsadmvalidator.ValidateConfigSpec(renderedConfig.Spec)
 	}
 
 	currentAppConfigResponse.Success = true
-	currentAppConfigResponse.ConfigGroups = configGroups
```
sgalsaleh, line 642: this means that the config groups will be null instead of an empty array when returned to the UI which might break

pkg/kotsadmconfig/validation/config.go
```diff
@@ -0,0 +1,115 @@
...
+			Value:            item.Value,
+			ValidationErrors: validationErrors,
+		}
+	}
+
+	return nil
+}
+
+func getItemValue(value multitype.BoolOrString, itemType string) (string, error) {
+	switch itemType {
```
sgalsaleh, line 94: what about type `bool` and type `select_one`?

reference: https://docs.replicated.com/reference/custom-resource-config#type

sgalsaleh: there is also `select_many`, `select`, and an empty `""` type which defaults to `text`

pkg/handlers/config.go
```diff
@@ -190,6 +195,15 @@ func (h *Handler) UpdateAppConfig(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	validationErrors := kotsadmvalidator.ValidateConfigSpec(kotsv1beta1.ConfigSpec{Groups: updateAppConfigRequest.ConfigGroups})
+	if len(validationErrors) > 0 {
+		updateAppConfigResponse.Error = "failed to validate config values"
+		updateAppConfigResponse.ValidationErrors = validationErrors
+		logger.Error(errors.Wrap(err, updateAppConfigResponse.Error))
```
cbodonnell, line 202: Do we also want to consider logging the validation errors so they end up in support bundles?  Might not be needed since we are going to surface client-side, but could be interesting to see.

pkg/kotsadmconfig/validation/validator.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	return validationErrs
+}
+
+func validateRegex(value string, regexValidator *kotsv1beta1.RegexValidator) *ValidationError {
+	regex, err := regexp.Compile(regexValidator.Pattern)
+	if err != nil {
+		return &ValidationError{
+			ValidationErrorMessage: fmt.Sprintf("Invalid regex: %s", err.Error()),
```
sgalsaleh, line 55: Why isn't this:
```suggestion
			ValidationErrorMessage: regexValidator.Message,
```

pkg/handlers/config.go
cbodonnell: Do we also want to consider logging the validation errors so they end up in support bundles?  Might not be needed since we are going to surface client-side, but could be interesting to see.

pkg/kotsadmconfig/validation/types.go
```diff
@@ -0,0 +1,24 @@
...
+type ConfigItemValidationError struct {
+	Name             string                 `json:"name"`
+	Type             string                 `json:"type"`
+	Value            multitype.BoolOrString `json:"value"`
+	ValidationErrors []ValidationError      `json:"validation_errors"`
+}
+
+type ValidationError struct {
+	ValidationErrorMessage string                      `json:"validation_error_message"`
+	RegexValidator         *kotsv1beta1.RegexValidator `json:"regex,omitempty"`
```
sgalsaleh, line 23: This field isn't needed IMO

pkg/kotsadmconfig/validation/types.go
```diff
@@ -0,0 +1,24 @@
...
+}
+
+type ConfigItemValidationError struct {
+	Name             string                 `json:"name"`
+	Type             string                 `json:"type"`
+	Value            multitype.BoolOrString `json:"value"`
+	ValidationErrors []ValidationError      `json:"validation_errors"`
+}
+
+type ValidationError struct {
```
sgalsaleh, line 21: This whole type is not needed IMO. Just return a new error with the message defined by the vendor.

kotskinds/apis/kots/v1beta1/config_types.go
```diff
@@ -56,12 +56,22 @@ type ConfigItem struct {
...
 	Templates     []RepeatTemplate       `json:"templates,omitempty"`
 	ValuesByGroup ValuesByGroup          `json:"valuesByGroup,omitempty"`
+	Validation    *ConfigItemValidator   `json:"validation,omitempty"`
 	// Props       map[string]interface{} `json:"props,omitempty"`
 	// DefaultCmd  *ConfigItemCmd         `json:"default_cmd,omitempty"`
 	// ValueCmd    *ConfigItemCmd         `json:"value_cmd,omitempty"`
 	// DataCmd     *ConfigItemCmd         `json:"data_cmd,omitempty"`
 }
 
+type ConfigItemValidator struct {
```
sgalsaleh, line 66: for naming consistency:
```suggestion
type ConfigItemValidation struct {
```

pkg/handlers/config.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/replicatedhq/kots/pkg/crypto"
 	"github.com/replicatedhq/kots/pkg/helm"
 	kotsadmconfig "github.com/replicatedhq/kots/pkg/kotsadmconfig"
+	kotsadmvalidator "github.com/replicatedhq/kots/pkg/kotsadmconfig/validation"
```
sgalsaleh, line 25: nit pick:
```suggestion
	configvalidation "github.com/replicatedhq/kots/pkg/kotsadmconfig/validation"
```

pkg/handlers/config.go
```diff
@@ -1147,6 +1148,15 @@ func (h *Handler) SetAppConfigValues(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	validationErrors := kotsadmvalidator.ValidateConfigSpec(renderedConfig.Spec)
+	if len(validationErrors) > 0 {
+		setAppConfigValuesResponse.Error = "failed to validate config values"
+		setAppConfigValuesResponse.ValidationErrors = validationErrors
+		logger.Error(errors.Wrap(err, setAppConfigValuesResponse.Error))
```
cbodonnell, line 1155: Same as my other comment.  The `renderedConfig` above this creates a new error to log instead of wrapping.

pkg/kotsadmconfig/validation/types.go
```diff
@@ -0,0 +1,24 @@
+package validation
+
+import (
+	kotsv1beta1 "github.com/replicatedhq/kots/kotskinds/apis/kots/v1beta1"
+	"github.com/replicatedhq/kots/kotskinds/multitype"
+)
+
+type ConfigGroupValidationError struct {
```
sgalsaleh, line 8: let's move these to a separate `types` package inside the `validation` package.

pkg/kotsadmconfig/validation/config.go
```diff
@@ -0,0 +1,115 @@
...
+	case TextItemType, TextAreaItemType:
+		return value.StrVal, nil
+	case PasswordItemType:
+		// if decrypting succeeds, use the decrypted value
+		if updatedValue, err := util.DecryptConfigValue(value.String()); err == nil {
+			return updatedValue, nil
+		} else {
+			return value.String(), nil
```
cbodonnell, line 102: are there cases where the password will not be encrypted?

pkg/kotsadmconfig/validation/config.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+	case FileItemType:
+		decodedBytes, err := util.Base64DecodeInterface(value.StrVal)
+		if err != nil {
+			return "", errors.Wrapf(err, "failed to base64 decode file item value")
+		}
+		return string(decodedBytes), err
+	case HeadingItemType, LabelItemType:
+		return "", nil
```
cbodonnell, line 111: Do we need to have this in here, or can these just fall through to the `default` and error out with an `invalid item type for validation X` message? We already filter these out as part of the `isValidatableConfigItem`, so getting here would be a bug?

kotskinds/apis/kots/v1beta1/config_types.go
```diff
@@ -56,12 +56,22 @@ type ConfigItem struct {
 	CountByGroup  map[string]int         `json:"countByGroup,omitempty"`
 	Templates     []RepeatTemplate       `json:"templates,omitempty"`
 	ValuesByGroup ValuesByGroup          `json:"valuesByGroup,omitempty"`
+	Validation    *ConfigItemValidator   `json:"validation,omitempty"`
```
cbodonnell, line 59: Do we want to consider allowing multiple validators per config item?  What if we add more types in the future and a vendor wants to use more that one?  Maybe it can all be accomplished with a sufficiently complex regex, but just a thought...

pkg/kotsadmconfig/validation/validator.go
pavansokkenagaraj: vendor's message will be very ambiguous.
adding another message field controlled by us would be more accurate.
for regex, there can be 2 errors
- when value doesn't match regex: `value does not match regex` --> easier to know the string was not matched
- and, when regex itself is invalid: Invalid regex: <<error from the regex lib>>
eg: when Pattern: "["  err would be:
`"Invalid regex: error parsing regexp: missing closing ]: `[`"` --> makes easier for debugging the validation err

pkg/kotsadmconfig/validation/types.go
pavansokkenagaraj: like mentioned here, https://github.com/replicatedhq/kots/pull/3800#discussion_r1164687355
+ adding the details would help for user to understand when using cli `kubectl kots set config`

pkg/handlers/config.go
pavansokkenagaraj: yeah, I should have logged the `setAppConfigValuesResponse.Error`

pkg/kotsadmconfig/validation/config.go
pavansokkenagaraj: Config spec doesn't contain `valuePlainText`. So,  we cannot know before hand if the value is not encrypted.
for eg: on on first visit of config page, liveConfig will have encrypted values. but when a password value is update, it will be in decrypted format.
In set config cli cmd, this is managed as the values sent are ConfigValues and they are merged into Config spec

so, I had to do this wonkyish stuff to accommodate all the scenarios

kotskinds/apis/kots/v1beta1/config_types.go
pavansokkenagaraj: Yes

pkg/kotsadmconfig/validation/types.go
pavansokkenagaraj: This wouldn't be needed as the `validator` is the package from where all the dependent functions will be referred from and the structs are more generic. 
The chances of cyclic dependency is low as `validator` package only refers to generic  `kotsv1beta1` `util` packages

pkg/handlers/config.go
pavansokkenagaraj: setting `currentAppConfigResponse.ConfigGroups = []kotsv1beta1.ConfigGroup{}` before so it works as it is

pkg/handlers/config.go
pavansokkenagaraj: We could log according the customer response

pkg/kotsadmconfig/validation/config.go
pavansokkenagaraj: changing this func to `getValidatableItemValue()` so only validator supported type's value are returned
+, we check if item type is one of [text, textarea, password, file] before calling `getValidatableItemValue`

pkg/kotsadmconfig/validation/config.go
pavansokkenagaraj: thanks, Moved to defaults, makes it more cleaner

pkg/handlers/config.go
pavansokkenagaraj:
```
{"level":"error","ts":"2023-04-12T23:11:43Z","msg":"failed to validate config values, validation errors: [{Name:example_settings Title:My Example Configs ItemErrors:[{Name:password Type:password Value:{Type:0 BoolVal:false StrVal:} ValidationErrors:[{ValidationErrorMessage:Value does not match regex RegexValidator:0xc00046ede0}]} {Name:email_text Type:text Value:{Type:0 BoolVal:false StrVal:} ValidationErrors:[{ValidationErrorMessage:Value does not match regex RegexValidator:0xc00046ee00}]} {Name:cve_text_area Type:textarea Value:{Type:0 BoolVal:false StrVal:} ValidationErrors:[{ValidationErrorMessage:Value does not match regex RegexValidator:0xc00046ee20}]} {Name:jwt_file Type:file Value:{Type:0 BoolVal:false StrVal:} ValidationErrors:[{ValidationErrorMessage:Value does not match regex RegexValidator:0xc00046ee40}]}]}]"}
```

logging with a standard formatter `%+v` should be good for debugging

pkg/kotsadmconfig/validation/types.go
pavansokkenagaraj: Finally decided to move to a separate package under `/pkg/kotskinds/types`
So all the ItemType Const and other config related structs can re-use the pacakge

pkg/kotsadmconfig/validation/config.go
sgalsaleh: What about the `""` type? That defaults to `text`?

pkg/kotsadmconfig/validation/regex.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+	return nil, nil
+}
+
+func buildRegexValidationError(validatorMessage string, errorMsg string) *configtypes.ValidationError {
+	if validatorMessage == "" {
+		validatorMessage = regexMatchError
+	}
+	return &configtypes.ValidationError{
+		Reason:  errorMsg,
```
sgalsaleh, line 37: again, why is this `Reason` field needed? We know that the only reason a validation error would be returned is if the Regex didn't match.

pkg/kotsadmconfig/validation/regex.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	matched := regex.MatchString(input)
+	if !matched {
+		return buildRegexValidationError(v.Message, regexMatchError), nil
+	}
+	return nil, nil
+}
+
+func buildRegexValidationError(validatorMessage string, errorMsg string) *configtypes.ValidationError {
```
sgalsaleh, line 32: this function isn't needed, just return the message defined by the vendor

pkg/kotsadmconfig/validation/regex.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+type regexValidator struct {
+	*kotsv1beta1.RegexValidator
+}
+
+func (v *regexValidator) Validate(input string) (*configtypes.ValidationError, error) {
+	regex, err := regexp.Compile(v.Pattern)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Invalid regex pattern. failed to compile regex")
```
sgalsaleh, line 22:
```suggestion
		return nil, errors.Wrapf(err, "failed to compile regex")
```

pkg/kotsadmconfig/validation/validator.go
```diff
@@ -0,0 +1,44 @@
...
+	validatableItemTypesMap = map[string]bool{
+		configtypes.TextItemType:     true,
+		configtypes.PasswordItemType: true,
+		configtypes.TextAreaItemType: true,
+		configtypes.FileItemType:     true,
+	}
+)
+
+func isValidatableConfigItem(item kotsv1beta1.ConfigItem) bool {
+	return item.Validation != nil && !item.Hidden && item.When != "false" && !item.Repeatable && validatableItemTypesMap[item.Type]
```
sgalsaleh, line 18: can we break these into separate if statements to follow the conventions of Go?

pkg/kotsadmconfig/validation/validator.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func validate(value string, itemValidator kotsv1beta1.ConfigItemValidation) ([]configtypes.ValidationError, error) {
```
sgalsaleh, line 21:
```suggestion
func validate(value string, itemValidation kotsv1beta1.ConfigItemValidation) ([]configtypes.ValidationError, error) {
```

pkg/kotsadmconfig/validation/regex.go
pavansokkenagaraj: If the vendor's message is empty, we need a default response

pkg/kotsadmconfig/validation/config.go
pavansokkenagaraj: Thanks, added the changes + tests

pkg/kotsadmconfig/validation/regex.go
pavansokkenagaraj: https://github.com/replicatedhq/kots/blob/281e12f35ca7b706afd9b25aa067840ef7f874dc/pkg/kotsadmconfig/validation/regex.go#L26-L33

added inline code to use default message

pkg/kotsadmconfig/validation/validator.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+func isValidatableConfigItem(item kotsv1beta1.ConfigItem) bool {
+	if item.Validation == nil ||
+		item.Hidden ||
+		item.When == "false" ||
+		item.Repeatable ||
+		!validatableItemTypesMap[item.Type] {
+		return false
+	}
```
cbodonnell, line 25: I'm not sure if this is what was intended by @sgalsaleh's comment, but this would produce the desired behavior

cmd/kots/cli/set-config.go
```diff
@@ -228,3 +234,19 @@ func getConfigValuesFromArgs(v *viper.Viper, args []string) ([]byte, error) {
...
+		for _, itemValidationError := range groupValidationError.ItemErrors {
+			log.Errorf("    Name: %s", itemValidationError.Name)
+			log.Errorf("    Type: %s", itemValidationError.Type)
+			log.Errorf("    Errors:")
+			for _, validationError := range itemValidationError.ValidationErrors {
+				log.Errorf("      - %s", validationError.Message)
+			}
+		}
+	}
+}
```
cbodonnell, line 252: Can you update the PR description to show an example of this?  Would help to review the UX

pavansokkenagaraj: added a screenshot
image

cmd/kots/cli/set-config.go
```diff
@@ -228,3 +234,19 @@ func getConfigValuesFromArgs(v *viper.Viper, args []string) ([]byte, error) {
...
 	return b, nil
 }
+
+func logConfigValidationErrors(groupValidationErrors []configtypes.ConfigGroupValidationError, log *logger.CLILogger) {
+	log.Errorf("Following config items have validation errors:")
+	for _, groupValidationError := range groupValidationErrors {
+		log.Errorf("Group - %s(%s)", groupValidationError.Title, groupValidationError.Name)
+		log.Errorf("  Items:")
+		for _, itemValidationError := range groupValidationError.ItemErrors {
+			log.Errorf("    Name: %s", itemValidationError.Name)
```
sgalsaleh, line 244: Let's use `title` instead of `name`.

cmd/kots/cli/set-config.go
```diff
@@ -228,3 +234,19 @@ func getConfigValuesFromArgs(v *viper.Viper, args []string) ([]byte, error) {
...
+			log.Errorf("    Type: %s", itemValidationError.Type)
```
sgalsaleh, line 245: Let's remove this.

pkg/kotsadmconfig/validation/validator.go
pavansokkenagaraj: for the func isValidatableConfigItem
earlier I had this 
func isValidatableConfigItem(item kotsv1beta1.ConfigItem) bool {
	return item.Validation != nil && !item.Hidden && item.When != "false" && !item.Repeatable && validatableItemTypesMap[item.Type]
}
So, I switched it to if statement.
multiple conditional statements on if are allowed.
Switching to `or` instead of `and` is more readable, I think

sgalsaleh: I meant separate if conditions. Separate if conditions is the Go convention and is much easier to read than this.

cmd/kots/cli/set-config.go
pavansokkenagaraj: `kot set config`/`ConfigValues` use `name` to update values.
Isn't using name be more appropriate?

pkg/kotsadmconfig/validation/validator.go
pavansokkenagaraj: ok

kotskinds/apis/kots/v1beta1/config_types.go
```diff
@@ -56,12 +56,26 @@ type ConfigItem struct {
...
 	// DefaultCmd  *ConfigItemCmd         `json:"default_cmd,omitempty"`
 	// ValueCmd    *ConfigItemCmd         `json:"value_cmd,omitempty"`
 	// DataCmd     *ConfigItemCmd         `json:"data_cmd,omitempty"`
 }
 
+type ConfigItemValidation struct {
+	Regex *RegexValidator `json:"regex,omitempty"`
+}
+
+type BaseValidator struct {
```
sgalsaleh, line 70: Why is this base validator needed? It's not helping to implement an interface and it's not adding much value here. I'd suggest just moving the `Message` field to the `RegexValidator`

cmd/kots/cli/set-config.go
```diff
@@ -132,14 +133,19 @@ func SetConfigCmd() *cobra.Command {
...
 			}{}
 			_ = json.Unmarshal(respBody, &response)
 
 			if resp.StatusCode != http.StatusOK {
 				if resp.StatusCode == http.StatusNotFound {
 					return errors.Errorf("app with slug %s not found", appSlug)
 				} else {
+					if len(response.ValidationErrors) > 0 {
+						logConfigValidationErrors(response.ValidationErrors, log)
+						return (errors.New(response.Error))
```
sgalsaleh, line 147: Is it always the case that there will be a response error when there are validation errors?
```suggestion
						return errors.New(response.Error)
```

cmd/kots/cli/set-config.go
sgalsaleh: hmm, yeah maybe. ok

cmd/kots/cli/set-config.go
sgalsaleh: Would it be possible to not have all the dots at the beginning of each line? The dot indicates a new action and I feel like this entire error just belong to a single section.

kotskinds/apis/kots/v1beta1/config_types.go
pavansokkenagaraj: When there are more validators, and if there are common fields, they can be added here.
Maybe this is to early to optimize. I will remove it.

cmd/kots/cli/set-config.go
pavansokkenagaraj: all the CLILogger functions start with a `dot`
Should I add a new logger function without `dot` or use `fmt.Print` to print them and wouldn;t be in red

pavansokkenagaraj: updated validation message for cli:
image

pkg/kotsadmconfig/validation/config.go
```diff
@@ -0,0 +1,109 @@
...
+	configtypes "github.com/replicatedhq/kots/pkg/kotsadmconfig/types"
+	"github.com/replicatedhq/kots/pkg/util"
+)
+
+func ValidateConfigSpec(configSpec kotsv1beta1.ConfigSpec) ([]configtypes.ConfigGroupValidationError, error) {
+	var configGroupErrors []configtypes.ConfigGroupValidationError
+	for _, configGroup := range configSpec.Groups {
+		configGroupError, err := validateConfigGroup(configGroup)
+		if err != nil {
+			return nil, err
```
sgalsaleh, line 16: let's wrap the error and tie it to the group name/title for easier troubleshooting/debugging

pkg/kotsadmconfig/validation/validator.go
```diff
@@ -0,0 +1,65 @@
...
+	return true
+}
+
+func validate(value string, itemValidation kotsv1beta1.ConfigItemValidation) ([]configtypes.ValidationError, error) {
+	var validationErrs []configtypes.ValidationError
+	validators := buildValidators(itemValidation)
+	for _, v := range validators {
+		validationErr, err := v.Validate(value)
+		if err != nil {
+			return nil, err
```
sgalsaleh, line 48: error wrap

pkg/handlers/config.go
```diff
@@ -190,6 +196,22 @@ func (h *Handler) UpdateAppConfig(w http.ResponseWriter, r *http.Request) {
...
+	validationErrors, err := configvalidation.ValidateConfigSpec(kotsv1beta1.ConfigSpec{Groups: updateAppConfigRequest.ConfigGroups})
+	if err != nil {
+		updateAppConfigResponse.Error = "failed to validate config spec."
+		logger.Error(errors.Wrap(err, updateAppConfigResponse.Error))
+		JSON(w, http.StatusInternalServerError, updateAppConfigResponse)
+		return
+	}
+
+	if len(validationErrors) > 0 {
+		updateAppConfigResponse.Error = "failed to validate config values"
```
sgalsaleh, line 208:
```suggestion
		updateAppConfigResponse.Error = "invalid config values"
```

pkg/print/config.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"github.com/fatih/color"
+	configtypes "github.com/replicatedhq/kots/pkg/kotsadmconfig/types"
+	"github.com/replicatedhq/kots/pkg/logger"
+)
+
+func ConfigValidationErrors(log *logger.CLILogger, groupValidationErrors []configtypes.ConfigGroupValidationError) {
+	errPrint := color.New(color.FgHiRed)
+	errPrint.Println("Following config items have validation errors:")
```
sgalsaleh, line 11: why not use the `log` parameter that's passed? you can build the string and the print it once using the logger. also, i'd suggest adding a new line at the beginning and end for better readability.

sgalsaleh (approved):

pkg/kotsadmconfig/validation/validator.go
cbodonnell: un-resolving this conversation since it seems like this wasn't addressed.  let me know if there was a particular reason to leave it as-is

cbodonnell (approved):

cbodonnell (approved):

PR #3802, opened 2023-04-13
template namespace and helmUpgradeFlags

cbodonnell:

This PR adds the ability to use template functions in the HelmChart resource `spec.namespace` and `spec.helmUpgradeFlags` fields so that those fields can be set dynamically for the deployment.

This is more of a temporary solution as we work to support template functions in all fields for all kots kinds.

Fixes [SC-71700](https://app.shortcut.com/replicated/story/71700/namespace-and-helmupgradeflags-field-in-helmchart-custom-resource-do-not-support-templating)

Added a new unit test for the operator that will validate that helm chart namespace and upgrade flags are templated before deploying.

Run operator tests:
```
go test -tags='testing netgo containers_image_ostree_stub exclude_graphdriver_devicemapper exclude_graphdriver_btrfs containers_image_openpgp' -installsuffix netgo ./pkg/operator
```

Without the fix:
```
• [FAILED] [0.853 seconds]
Operator DeployApp() when there is a helm chart with template functions [It] installs the helm chart using the templated namespace and upgrade flags
...
  [FAILED] Expected
      <string>: repl{{ ConfigOption \"deploy_namespace\" }}
  to equal
      <string>: my-namespace
```

With the fix:
```
ok      github.com/replicatedhq/kots/pkg/operator       5.769s
```

#### Steps to reproduce

Use template functions in these fields with and without this change.

```release-note
Adds support for template functions to the `namespace` and `helmUpgradeFlags` fields of the [HelmChart](/reference/custom-resource-helmchart) custom resource.
```

https://github.com/replicatedhq/replicated-docs/pull/1075

### DIFF
--- a/pkg/kotsutil/kots.go
+++ b/pkg/kotsutil/kots.go
@@ -273,7 +273,7 @@ func (o KotsKinds) Marshal(g string, v string, k string) (string, error) {
 					return "", errors.Wrap(err, "failed to encode identityconfig")
 				}
 				return string(b.Bytes()), nil
-			case "HelmChart":
+			case "HelmChartList":
 				if o.HelmCharts == nil {
 					return "", nil
 				}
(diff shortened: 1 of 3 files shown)


pkg/kotsutil/kots.go
```diff
@@ -273,7 +273,7 @@ func (o KotsKinds) Marshal(g string, v string, k string) (string, error) {
 					return "", errors.Wrap(err, "failed to encode identityconfig")
 				}
 				return string(b.Bytes()), nil
-			case "HelmChart":
+			case "HelmChartList":
```
cbodonnell, line 276: Changed this since we're actually marshalling a `HelmChartList` when this is called.

pkg/kotsutil/kots.go
```diff
@@ -445,7 +445,12 @@ func (k *KotsKinds) addKotsKinds(content []byte) error {
 			k.Installation = *decoded.(*kotsv1beta1.Installation)
 		case "kots.io/v1beta1, Kind=HelmChart":
 			if k.HelmCharts == nil {
-				k.HelmCharts = &kotsv1beta1.HelmChartList{}
+				k.HelmCharts = &kotsv1beta1.HelmChartList{
+					TypeMeta: metav1.TypeMeta{
+						APIVersion: "kots.io/v1beta1",
+						Kind:       "HelmChartList",
+					},
+				}
```
cbodonnell, line 453: Instantiate with `apiVersion` and `kind` so that `LoadHelmChartsFromContents(...)` below will work

pkg/operator/operator_test.go
cbodonnell, line 1: Refactored this test file so that archive files can be specified for individual tests

pavansokkenagaraj (approved):
lgtm!

PR #3804, opened 2023-04-14
Apply strict security contexts in OpenShift when specified

sgalsaleh:

Applies strict security contexts in OpenShift when specified as they were ignored before.

Fixes [SC-74591](https://app.shortcut.com/replicated/story/74591/strict-pod-security-context-configurations-aren-t-applied-in-openshift-environments-when-passing-the-strict-security-context)

```release-note
Fixes an issue where strict security context configurations were not applied in OpenShift environments when the `--strict-security-context` flag is passed to the [kots install](https://docs.replicated.com/reference/kots-cli-install) or [kots admin-console upgrade](https://docs.replicated.com/reference/kots-cli-admin-console-upgrade) commands.
```

NONE


pkg/k8sutil/pod.go
```diff
@@ -112,3 +113,44 @@ func PodsHaveTheSameOwner(pods []corev1.Pod) bool {
 
 	return true
 }
+
+func SecurePodContext(user int64, group int64, isStrict bool) *corev1.PodSecurityContext {
```
sgalsaleh, line 117: This function was moved from another file with a minor modification to accept a group as a parameter.

pkg/k8sutil/pod.go
```diff
@@ -112,3 +113,44 @@ func PodsHaveTheSameOwner(pods []corev1.Pod) bool {
...
+	} else {
+		context = corev1.PodSecurityContext{
+			RunAsUser: pointer.Int64(user),
+			FSGroup:   pointer.Int64(group),
+		}
+	}
+
+	return &context
+}
+func SecureContainerContext(isStrict bool) *corev1.SecurityContext {
```
sgalsaleh, line 140: ditto

cbodonnell (approved):

PR #3810, opened 2023-04-17
Miawong/sc 73295/insecure storage of jwt

miaawong:

Fixes # https://app.shortcut.com/replicated/story/73295/insecure-storage-of-jwt

## Steps to reproduce

1. Login to admin console 
2. check application tab of dev tool, go to localstorage
3. verify that JWT is not present 

```release-note
Stores JWT inside of an httpOnly cookie to prevent XSS attacks. 
```

### DIFF
--- a/pkg/handlers/cors.go
+++ b/pkg/handlers/cors.go
@@ -5,8 +5,10 @@ import (
 )
 
 func CORS(w http.ResponseWriter, r *http.Request) {
+	origin := r.Header.Get("Origin")
 	w.Header().Set("Access-Control-Allow-Methods", "GET, POST, OPTIONS, DELETE, PUT")
-	w.Header().Set("Access-Control-Allow-Origin", "*")
+	w.Header().Set("Access-Control-Allow-Origin", origin)
+	w.Header().Set("Access-Control-Allow-Credentials", "true")
 	w.Header().Set("Access-Control-Allow-Headers", "content-type, origin, accept, authorization")
 	w.Header().Set("Access-Control-Expose-Headers", "content-disposition")
 }
--- a/pkg/handlers/login.go
+++ b/pkg/handlers/login.go
@@ -112,15 +112,27 @@ func (h *Handler) Login(w http.ResponseWriter, r *http.Request) {
 
 	responseToken := fmt.Sprintf("Bearer %s", signedJWT)
 
-	expire := time.Now().Add(30 * time.Minute)
-	cookie := &http.Cookie{
+	expire := time.Now().Add(24 * time.Hour)
+	tokenCookie := http.Cookie{
 		Name:     "signed-token",
 		Value:    responseToken,
 		Expires:  expire,
 		Path:     "/",
 		HttpOnly: true,
 	}
-	http.SetCookie(w, cookie)
+
+	originURL, err := url.Parse(r.Header.Get("Origin"))
+	if err != nil {
+		logger.Error(errors.Wrap(err, "failed to parse origin url"))
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	if originURL.Scheme == "https" {
+		tokenCookie.Secure = true
+	}
+
+	http.SetCookie(w, &tokenCookie)
 
 	JSON(w, http.StatusOK, loginResponse)
 }
--- a/pkg/handlers/login.go
+++ b/pkg/handlers/login.go
@@ -112,27 +112,16 @@ func (h *Handler) Login(w http.ResponseWriter, r *http.Request) {
 
 	responseToken := fmt.Sprintf("Bearer %s", signedJWT)
 
-	expire := time.Now().Add(SessionTimeout)
-	tokenCookie := http.Cookie{
-		Name:     "signed-token",
-		Value:    responseToken,
-		Expires:  expire,
-		Path:     "/",
-		HttpOnly: true,
-	}
-
-	originURL, err := url.Parse(r.Header.Get("Origin"))
+	expiration := time.Now().Add(SessionTimeout)
+	origin := r.Header.Get("Origin")
+	tokenCookie, err := session.GetSessionCookie(responseToken, expiration, origin)
 	if err != nil {
-		logger.Error(errors.Wrap(err, "failed to parse origin url"))
+		logger.Error(errors.Wrap(err, "failed to get session cookie"))
 		w.WriteHeader(http.StatusInternalServerError)
 		return
 	}
 
-	if originURL.Scheme == "https" {
-		tokenCookie.Secure = true
-	}
-
-	http.SetCookie(w, &tokenCookie)
+	http.SetCookie(w, tokenCookie)
 
 	JSON(w, http.StatusOK, loginResponse)
 }
--- a/pkg/handlers/login.go
+++ b/pkg/handlers/login.go
@@ -110,20 +110,17 @@ func (h *Handler) Login(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	loginResponse.Token = fmt.Sprintf("Bearer %s", signedJWT)
+	responseToken := fmt.Sprintf("Bearer %s", signedJWT)
 
 	expire := time.Now().Add(30 * time.Minute)
-	fmt.Println("LG: Within Login(), setting cookie.")
 	cookie := &http.Cookie{
 		Name:     "signed-token",
-		Value:    signedJWT,
+		Value:    responseToken,
 		Expires:  expire,
 		Path:     "/",
 		HttpOnly: true,
 	}
-
 	http.SetCookie(w, cookie)
-	fmt.Println("LG: Cookie: ", cookie)
 
 	JSON(w, http.StatusOK, loginResponse)
 }
--- a/pkg/handlers/login.go
+++ b/pkg/handlers/login.go
@@ -112,6 +112,19 @@ func (h *Handler) Login(w http.ResponseWriter, r *http.Request) {
 
 	loginResponse.Token = fmt.Sprintf("Bearer %s", signedJWT)
 
+	expire := time.Now().Add(30 * time.Minute)
+	fmt.Println("LG: Within Login(), setting cookie.")
+	cookie := &http.Cookie{
+		Name:     "signed-token",
+		Value:    signedJWT,
+		Expires:  expire,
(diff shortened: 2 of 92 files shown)


pkg/handlers/cors.go
```diff
@@ -5,8 +5,10 @@ import (
 )
 
 func CORS(w http.ResponseWriter, r *http.Request) {
+	origin := r.Header.Get("Origin")
```
St0rmz1, line 8: If I'm reading this correctly, we are setting origin based on the HTTP GET. This means we are trusting the client is sending an accurate header. Its easy to manipulate the headers in a HTTP request. AKA, we are trusting what is essentially user defined input. Are the headers getting validated in some other area?

garcialuis: Hi @St0rmz1 The origin header will be present on any cross-origin requests, not just GET requests. The [origin header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin) is set by the browser before sending the request to the server, it is a [forbidden header](https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name) which means that the client has no way of setting or modifying the value.

web/src/utilities/utilities.js
```diff
@@ -583,8 +576,11 @@ export const Utilities = {
   },
 
   isLoggedIn() {
-    const hasToken = this.getToken();
-    return !!hasToken;
+    if (localStorage.getItem("isLoggedIn") === "true") {
```
miaawong, line 579: No longer store JWT in localStorage. We set isLoggedIn in localStorage when user successfully logs in. https://github.com/replicatedhq/kots/blob/28fef60f1e0f0589b5925f0caf945f06cecd4da4/web/src/features/Auth/components/SecureAdminConsole.tsx#L55

rianfowler (dismissed):
nice work

pkg/handlers/cors.go
cbodonnell: @garcialuis @St0rmz1 , to better understand this myself.  It seems like we previously had `Access-Control-Allow-Origin: *` since we cannot know the value of the origin header because KOTS may be running behind any arbitrary enterprise domain?  As I see it, this essentially produces the same behavior as the wildcard, but without the wildcard since the browser would not allow sending the session cookie along with `Access-Control-Allow-Origin: *`.

pkg/handlers/login.go
```diff
@@ -110,7 +110,29 @@ func (h *Handler) Login(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	loginResponse.Token = fmt.Sprintf("Bearer %s", signedJWT)
+	responseToken := fmt.Sprintf("Bearer %s", signedJWT)
+
+	expire := time.Now().Add(24 * time.Hour)
```
cbodonnell, line 115: Looks like the token [expires after 12 hours](https://github.com/replicatedhq/kots/blob/main/pkg/handlers/login.go#L44), so the cookie expiration should probably align with that unless there is a reason not to.

pkg/handlers/logout.go
```diff
@@ -9,12 +9,20 @@ import (
 )
 
 func (h *Handler) Logout(w http.ResponseWriter, r *http.Request) {
-	if r.Header.Get("Authorization") == "" {
-		w.WriteHeader(http.StatusNoContent)
-		return
+
+	auth := r.Header.Get("authorization")
```
cbodonnell, line 13: do we need to check the `Authorization` header if we're sending the token in a cookie?

pkg/handlers/session.go
```diff
@@ -47,21 +47,26 @@ func parseClusterAuthorization(authHeader string) (authorization, error) {
 	}, nil
 }
 
-func requireValidSession(kotsStore store.Store, w http.ResponseWriter, r *http.Request) (*sessiontypes.Session, error) {
+func requireValidSession(kotsStore store.Store, w http.ResponseWriter, r *http.Request) (sess *sessiontypes.Session, err error) {
 	if r.Method == "OPTIONS" {
 		return nil, nil
 	}
 
 	auth := r.Header.Get("authorization")
```
cbodonnell, line 55: same question regarding the `Authorization` header

cbodonnell, line 55: same question regarding the `Authorization` header

pkg/handlers/login.go
```diff
@@ -110,7 +110,29 @@ func (h *Handler) Login(w http.ResponseWriter, r *http.Request) {
...
+	tokenCookie := http.Cookie{
+		Name:     "signed-token",
+		Value:    responseToken,
+		Expires:  expire,
+		Path:     "/",
+		HttpOnly: true,
+	}
+
+	originURL, err := url.Parse(r.Header.Get("Origin"))
```
cbodonnell, line 124: this seems like a similar scenario as [above](https://github.com/replicatedhq/kots/pull/3810/files#r1169465032).  we need to trust the `Origin` header to know if they are using https since it can vary from one enterprise instance to another?

pkg/handlers/login.go
```diff
@@ -110,7 +110,29 @@ func (h *Handler) Login(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	loginResponse.Token = fmt.Sprintf("Bearer %s", signedJWT)
+	responseToken := fmt.Sprintf("Bearer %s", signedJWT)
```
cbodonnell, line 113: `Bearer` might not be necessary here since we're just storing it in a cookie, but I see that it does provide value in the `Parse(...)` function where we determine if it's a session token or KOTS cli token.  Probably not worth changing, but just a random thought...

pkg/handlers/cors.go
garcialuis: Yes that's correct, this achieves the same behavior. Without this the browser would not allow the request to go through since it is not allowed to use the `Access-Control-Allow-Credentials: true` header while `Access-Control-Allow-Origin is *`

pkg/handlers/login.go
garcialuis: I updated the cookie to use the same `SessionTimeout` as the token.

pkg/handlers/logout.go
garcialuis: Since this is only used by the UI it is safe to remove, I went ahead and removed it now.

pkg/handlers/session.go
garcialuis: We need to keep the check for the authorization header in this case because this function is also used to validate sessions for requests done via the CLI. Those will still use a `Kots` token set in the authorization header.

pkg/handlers/login.go
garcialuis: Yes, since the domain can change from customer to customer we have to use the value from the `origin` header to determine if the cookie needs to be set to `secure`.

pkg/handlers/login.go
garcialuis: Right, this way the parse function does not need to change.

pkg/handlers/logout.go
```diff
@@ -2,19 +2,33 @@ package handlers
...
+	if err == http.ErrNoCookie {
 		w.WriteHeader(http.StatusNoContent)
 		return
 	}
+	auth := signedTokenCookie.Value
+	// delete cookie by setting expiration to past
+	expiration := time.Now().Add(-1 * time.Hour)
+	tokenCookie, err := session.GetSessionCookie(auth, expiration, r.Header.Get("Origin"))
+	if err != nil {
+		logger.Error(err)
```
cbodonnell, line 25: we should probably wrap this error like we do in the login method

cbodonnell (dismissed):

cbodonnell (approved):

pkg/handlers/login.go
```diff
@@ -110,7 +110,18 @@ func (h *Handler) Login(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	loginResponse.Token = fmt.Sprintf("Bearer %s", signedJWT)
```
sgalsaleh, line 113: Let's remove the `Token` field from the login response type?

pkg/handlers/logout.go
```diff
@@ -2,19 +2,34 @@ package handlers
...
 	"github.com/replicatedhq/kots/pkg/session"
 	"github.com/replicatedhq/kots/pkg/store"
 )
 
 func (h *Handler) Logout(w http.ResponseWriter, r *http.Request) {
-	if r.Header.Get("Authorization") == "" {
+
+	signedTokenCookie, err := r.Cookie("signed-token")
+
+	if err == http.ErrNoCookie {
```
sgalsaleh, line 17: what if there is an error but it isn't `http.ErrNoCookie`. should we log it and mark it as an internal server error?

pkg/handlers/session.go
```diff
@@ -47,21 +47,27 @@ func parseClusterAuthorization(authHeader string) (authorization, error) {
 	}, nil
 }
 
-func requireValidSession(kotsStore store.Store, w http.ResponseWriter, r *http.Request) (*sessiontypes.Session, error) {
+func requireValidSession(kotsStore store.Store, w http.ResponseWriter, r *http.Request) (sess *sessiontypes.Session, err error) {
```
sgalsaleh, line 50: i don't see why naming the response parameters here is necessary

pkg/handlers/login.go
garcialuis: Addressed in PR https://github.com/replicatedhq/kots/pull/3846

pkg/handlers/session.go
garcialuis: Updated in https://github.com/replicatedhq/kots/pull/3846

pkg/handlers/logout.go
garcialuis: As of now the [request](https://go.dev/src/net/http/request.go) Cookie function only returns `nil` or `http.ErrNoCookie` as errors:
```
// Cookie returns the named cookie provided in the request or
// ErrNoCookie if not found.
// If multiple cookies match the given name, only one cookie will
// be returned.
func (r *Request) Cookie(name string) (*Cookie, error) {
	if name == "" {
		return nil, ErrNoCookie
	}
	for _, c := range readCookies(r.Header, name) {
		return c, nil
	}
	return nil, ErrNoCookie
}
```
Would adding a check for other types of errors be to future-proof in case there are others returned in the future?

PR #3818, opened 2023-04-19
add check for when property at group and item level when determining required error state 

rianfowler:
- fixes issue with required fields blocking when they're hidden 

 https://app.shortcut.com/replicated/story/28650/kotsadm-validation-of-config-options-entered-by-user

```release-note
none
```

none

### DIFF
--- a/web/src/features/AppConfig/components/AppConfig.tsx
+++ b/web/src/features/AppConfig/components/AppConfig.tsx
@@ -495,9 +495,10 @@ class AppConfig extends Component<Props, State> {
     const hasRequiredValidationError =
       showErrorsForEmptyValues &&
       Boolean(
-        groups?.find((group) =>
-          Boolean(
+        groups?.find(
+          (group) =>
             group?.when !== "false" &&
+            Boolean(
               group?.items?.find(
                 (item) =>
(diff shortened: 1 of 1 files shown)


web/src/features/AppConfig/components/AppConfig.tsx
```diff
@@ -495,12 +495,22 @@ class AppConfig extends Component<Props, State> {
     const hasRequiredValidationError =
       showErrorsForEmptyValues &&
       Boolean(
-        groups.find((group) =>
-          Boolean(group?.items.find((item) => item.required && !item.value))
+        groups?.find(
+          (group) =>
+            group?.when !== "false" &&
```
rianfowler, line 500: we have to check when at the group and item level. and hidden at just the item level per https://docs.replicated.com/reference/custom-resource-config#when

pavansokkenagaraj:
@rianfowler Please remove ```Fixes # https://app.shortcut.com/replicated/story/72034/restore-hangs-when-using-rabbitmq-clusters``` from the PR description.
as the fix for that issue is different and still in WIP

rianfowler:
[sc-28650]

PR #3825, opened 2023-04-21
add warning if push-images airgap file doesn't exist

ousbots:
Add a warning to `kubectl kots push-images <airgap bundle>` if the airgap bundle specified is both not a valid file and not a valid URL. Without the warning the command appears to have successfully pushed the images.

Fixes #

## Steps to reproduce

`kubectl kots admin-console push-images www.google.com ttl.sh` runs successfully with no warning that 'www.google.com' is not a valid airgap bundle and no airgap images were pushed.

```release-note
NONE
```

NONE


garcialuis (approved):

garcialuis:
LGTM, thanks @ousbots!

PR #3826, opened 2023-04-21
Fix: multiple instances of the same resource get deduplicated even if they have a different namespace

sgalsaleh:

Fixes an issue where multiple copies of the same Kubernetes resource (e.g. same `kind` and `name`) get deduplicated even if they have a different namespace, which results in only one of them getting deployed to the cluster.

Fixes #

```release-note
Fixes an issue where multiple copies of the same Kubernetes resource (e.g. same `kind` and `name`) get deduplicated even if they have a different namespace, which results in only one of them getting deployed to the cluster.
```

NONE


pkg/util/splitter.go
```diff
@@ -0,0 +1,65 @@
+package util
```
sgalsaleh, line 1: this file is a modified version of https://github.com/marccampbell/yaml-toolbox/blob/master/pkg/splitter/splitter.go

cbodonnell (approved):

PR #3828, opened 2023-04-24
make default selected node type primary. force to secondary if no ha

rianfowler:
- sets the default selected note type to primary 
- sets the node type to secondary if HA is unavailable 

Fixes # https://app.shortcut.com/replicated/story/48139/make-join-script-for-primaries-default-in-kots-add-node-ui#activity-75269

## HA cluster

https://www.loom.com/share/899553949a5940a0a2bc045cc1762d26

## Non HA cluster 

https://www.loom.com/share/e8fe70a77d3e406984b79274aa6a4f30

```release-note
Embedded cluster management page now defaults primary node installation instructions when adding a new node to a high availability cluster.
```

none


web/src/components/apps/ClusterNodes.jsx
```diff
@@ -64,6 +64,8 @@ export class ClusterNodes extends Component {
       const response = await res.json();
       this.setState({
         kurl: response,
+        // if cluster doesn't support ha, then primary will be disabled. Force into secondary
+        selectedNodeType: !response.ha ? "secondary" : this.state.selectedNodeType,
```
rianfowler, line 68: `selectedNodeType: primary` is not supported if `kurl.ha = false` so force the selection to secondary when this is the case. Otherwise leave the value as is. 

https://github.com/replicatedhq/kots/blob/9a6623cbc35da2cae4c52b72f5a4a2a0359576a4/web/src/components/apps/ClusterNodes.jsx#L326-L362

web/src/components/apps/ClusterNodes.jsx
```diff
@@ -64,6 +64,10 @@ export class ClusterNodes extends Component {
       const response = await res.json();
       this.setState({
         kurl: response,
+        // if cluster doesn't support ha, then primary will be disabled. Force into secondary
+        selectedNodeType: !response.ha
+          ? "secondary"
+          : this.state.selectedNodeType,
```
rianfowler, line 70: `selectedNodeType: primary` is not supported if `kurl.ha = false` so force the selection to secondary when this is the case. Otherwise leave the value as is. 

https://github.com/replicatedhq/kots/blob/9a6623cbc35da2cae4c52b72f5a4a2a0359576a4/web/src/components/apps/ClusterNodes.jsx#L326-L362

miaawong (approved):

PR #3830, opened 2023-04-24
change missing airgap bundle/URL in 'push-images' to be a failure

ousbots:
Return an error instead of a warning and continuing when the airgap bundle/URL is missing or invalid

Fixes #

```release-note

```

### DIFF
--- a/cmd/kots/cli/admin-console-push-images.go
+++ b/cmd/kots/cli/admin-console-push-images.go
@@ -104,7 +104,7 @@ func AdminPushImagesCmd() *cobra.Command {
 				}
 			} else if os.IsNotExist(err) {
 				if _, err := url.ParseRequestURI(imageSource); err != nil {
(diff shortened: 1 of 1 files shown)


cmd/kots/cli/admin-console-push-images.go
```diff
@@ -104,7 +104,7 @@ func AdminPushImagesCmd() *cobra.Command {
 				}
 			} else if os.IsNotExist(err) {
 				if _, err := url.ParseRequestURI(imageSource); err != nil {
-					log.Errorf("the airgap bundle %s does not exist or is not a valid URL, only public KOTS images will be pushed", imageSource)
+					return errors.Wrapf(err, "the airgap bundle %s does not exist", imageSource)
```
sgalsaleh, line 107:
```suggestion
					// Don't print URI parsing errors as this format is only used internally by KOTS
					return errors.New("File %q does not exist", imageSource)
```

sgalsaleh (approved):

PR #3832, opened 2023-04-25
fix: template helm chart namespace an upgrade flags

cbodonnell:

This PR fixes an issue where quotes are stripped from HelmChart resources prior to deployment, which can lead to unexpected behavior and failed deployments.  This is because the entire resource was being marshalled, rendered, and then unmarshalled.

This PR changes this so that only the desired fields are rendered (`namespace` and `helmUpgradeFlags`) to avoid unexpected results from marshalling and unmarshalling.

Fixes [SC-75352](https://app.shortcut.com/replicated/story/75352/rendering-of-helmchart-resource-template-functions-removes-quotes-causing-unexpected-behavior)

## Steps to reproduce

Include a HelmChart with a spec.optionalValues that contains a when condition evaluated by a template function. The surrounding quotes will be stripped and the resource will fail to be parsed because it is the wrong type (bool instead of string). This will cause the deployment to fail.

```
apiVersion: kots.io/v1beta1
kind: HelmChart
...
  optionalValues:
    - when: "repl{{ HasLocalRegistry }}"
```

```release-note
Fixes an issue where quotes were being stripped from fields in HelmChart resources, which could lead to unexpected behavior and failed deployments.
```

NONE


pkg/operator/operator_test.go
```diff
@@ -487,6 +487,12 @@ spec:
   helmUpgradeFlags:
     - --set
     - extraValue=repl{{ ConfigOption "deploy_extra_value" }}
+  optionalValues:
+  - when: "repl{{ HasLocalRegistry }}"
+    recursiveMerge: true
+    values:
+      global:
+        registry: ''
```
cbodonnell, line 495: Without these changes this test will fail with the following error since `optionalValues` would also be rendered and the result would not be a string since quotes are stripped.
```
failed to load helm charts: failed to decode kots app data of length 472: json: cannot unmarshal bool into Go struct field OptionalValue.items.spec.optionalValues.when of type string
```

pkg/operator/operator.go
```diff
@@ -256,19 +256,19 @@ func (o *Operator) DeployApp(appID string, sequence int64) (deployed bool, deplo
...
+			if err != nil {
+				return false, errors.Wrapf(err, "failed to render namespace %s for chart %s", helmChart.Spec.Namespace, helmChart.GetReleaseName())
+			}
+			kotsKinds.HelmCharts.Items[i].Spec.Namespace = renderedNamespace
+			for j, upgradeFlag := range helmChart.Spec.HelmUpgradeFlags {
+				renderedUpgradeFlag, err := builder.String(upgradeFlag)
+				if err != nil {
+					return false, errors.Wrapf(err, "failed to render upgrade flag %s for chart %s", upgradeFlag, helmChart.GetReleaseName())
+				}
+				kotsKinds.HelmCharts.Items[i].Spec.HelmUpgradeFlags[j] = renderedUpgradeFlag
```
sgalsaleh, line 270: maybe `continue` if the rendered value is empty?

cbodonnell: seems reasonable.  we don't do any validating that `helmUpgradeFlags` contains empty values today, but doing this would ensure that.

sgalsaleh: ah you're right. i think it's fine the way you have it then.

pkg/operator/operator.go
```diff
@@ -256,19 +256,19 @@ func (o *Operator) DeployApp(appID string, sequence int64) (deployed bool, deplo
...
-		kotsKinds.HelmCharts, err = kotsutil.LoadHelmChartsFromContents([]byte(renderedHelmCharts))
-		if err != nil {
-			return false, errors.Wrap(err, "failed to load helm charts")
+		for i, helmChart := range kotsKinds.HelmCharts.Items {
+			renderedNamespace, err := builder.String(helmChart.Spec.Namespace)
+			if err != nil {
+				return false, errors.Wrapf(err, "failed to render namespace %s for chart %s", helmChart.Spec.Namespace, helmChart.GetReleaseName())
+			}
+			kotsKinds.HelmCharts.Items[i].Spec.Namespace = renderedNamespace
+			for j, upgradeFlag := range helmChart.Spec.HelmUpgradeFlags {
```
sgalsaleh, line 265: nit pick: add a new line before for better readability.

sgalsaleh (approved):

PR #3836, opened 2023-04-25
Create sub-command 'admin-console copy-public-images'

ousbots:

Create the subcommand `copy-public-images` of `admin-console` to copy the public admin console images to a private registry.

Previously the `kots admin-console push-images` command invoked this functionality when passed the argap bundle of "docker.io". Instead of hiding the command behind a secret invocation of another command, this makes copying the public admin console images an explicit command. This also fixes the bug where the `kots admin-console push-images` command would push the public images and exit successfully if the provided airgap bundle didn't exist.

Fixes https://app.shortcut.com/replicated/story/72920/kots-admin-console-push-images-does-not-surface-any-errors-if-airgap-bundle-provided-is-missing

Check out this slack thread for more context on this decision: https://replicated.slack.com/archives/C02TQTV9G2V/p1682356184970139

## Steps to reproduce

To reproduce the missing error when the provided airgap bundle doesn't exist, try the following command:

```
kubectl kots admin-console push-images google.com ttl.sh
```

Before this change, this would result in the public admin console images being pushed to ttl.sh and the command would exit successfully. Now, this will return an error about the file missing.

```release-note
The command `kots admin-console push-images` will now return an error if the provided airgap bundle file is missing.
```

NONE

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -2612,8 +2612,6 @@ jobs:
         run: |
           set +e
 
-          touch not-an-app.airgap
-
           ./bin/kots admin-console push-images \
             not-an-app.airgap localhost:1234/not-a-namespace \
             --registry-username not-a-username \
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -2612,8 +2612,8 @@ jobs:
         run: |
           set +e
 
-          ./bin/kots admin-console push-images \
-            not-an-app.airgap localhost:1234/not-a-namespace \
+          ./bin/kots admin-console copy-public-images \
+            localhost:1234/not-a-namespace \
             --registry-username not-a-username \
             --registry-password not-a-password > output.txt 2>&1
 
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -2612,8 +2612,10 @@ jobs:
         run: |
           set +e
 
-          ./bin/kots admin-console copy-public-images \
-            localhost:1234/not-a-namespace \
+          touch not-an-app.airgap
+
+          ./bin/kots admin-console push-images \
+            not-an-app.airgap localhost:1234/not-a-namespace \
             --registry-username not-a-username \
             --registry-password not-a-password > output.txt 2>&1
 
--- a/cmd/kots/cli/admin-console-copy-public-images.go
+++ b/cmd/kots/cli/admin-console-copy-public-images.go
@@ -1,19 +1,12 @@
 package cli
 
 import (
-	"fmt"
 	"os"
 
 	"github.com/pkg/errors"
-	"github.com/replicatedhq/kots/pkg/docker/registry"
-	dockerregistry "github.com/replicatedhq/kots/pkg/docker/registry"
-	registrytypes "github.com/replicatedhq/kots/pkg/docker/registry/types"
 	"github.com/replicatedhq/kots/pkg/kotsadm"
-	kotsadmtypes "github.com/replicatedhq/kots/pkg/kotsadm/types"
-	"github.com/replicatedhq/kots/pkg/logger"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
-	kuberneteserrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
 func AdminCopyPublicImagesCmd() *cobra.Command {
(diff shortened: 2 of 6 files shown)


pkg/kotsadm/copy_images.go
```diff
@@ -20,7 +20,7 @@ import (
 )
 
 // Copies Admin Console images from public registry to private registry
-func CopyImages(sourceEndpoint string, options types.PushImagesOptions, kotsNamespace string) error {
+func CopyImages(options types.PushImagesOptions, kotsNamespace string) error {
```
ousbots, line 23: This was only ever called with a `sourceEndpoint` of "docker.io" and even if it was different than "docker.io", it wasn't used as the image source, so I removed it.

cmd/kots/cli/admin-console-push-images.go
```diff
@@ -23,86 +23,29 @@ import (
 func AdminPushImagesCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:           "push-images [airgap filename] [registry host]",
-		Short:         "Push admin console images",
-		Long:          "Push admin console images from airgap bundle to a private registry",
+		Short:         "Push app images",
+		Long:          "Push app images from airgap bundle to a private registry",
```
sgalsaleh, line 27:
```suggestion
		Long:          "Push images from airgap bundle to a private registry",
```

cmd/kots/cli/admin-console-push-images.go
```diff
@@ -23,86 +23,29 @@ import (
 func AdminPushImagesCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:           "push-images [airgap filename] [registry host]",
-		Short:         "Push admin console images",
-		Long:          "Push admin console images from airgap bundle to a private registry",
+		Short:         "Push app images",
```
sgalsaleh, line 26:
```suggestion
		Short:         "Push images from airgap bundle to a private registry",
```

.github/workflows/build-test.yaml
```diff
@@ -2612,8 +2612,8 @@ jobs:
         run: |
           set +e
 
-          ./bin/kots admin-console push-images \
```
sgalsaleh, line 2615: let's not update this test. this is not the functionality that should be using the new `copy-public-images` command

cmd/kots/cli/admin-console-copy-public-images.go
```diff
@@ -0,0 +1,60 @@
...
+
+			return nil
+		},
+	}
+
+	cmd.Flags().String("registry-username", "", "user name to use to authenticate with the registry")
+	cmd.Flags().String("registry-password", "", "password to use to authenticate with the registry")
+	cmd.Flags().Bool("skip-registry-check", false, "skip the connectivity test and validation of the provided registry information")
+
+	cmd.Flags().String("source-registry", "docker.io", "source registry for public images")
```
sgalsaleh, line 54: where is this used?

cmd/kots/cli/admin-console-push-images.go
```diff
@@ -23,86 +23,29 @@ import (
...
-			}
-
 			if _, err := os.Stat(imageSource); err == nil {
-				err := kotsadm.PushImages(imageSource, options)
+				options, err := genAndCheckPushOptions(endpoint, cmd)
 				if err != nil {
-					return errors.Wrap(err, "failed to push images")
+					return err
 				}
-			} else if os.IsNotExist(err) {
```
sgalsaleh, line 104: Isn't the whole point to show an error if the file does not exist?

ousbots: this was originally to redirect to copying the public images, now if the stat in the guard above returns any error the command fails

cmd/kots/cli/admin-console-copy-public-images.go
ousbots: argh, I'm trying to get better at removing these code bread crumbs 🤦. Before removing the `sourceEndpoint` parameter from `CopyImages`, I was using this to default it to "docker.io", but that's not needed now

.github/workflows/build-test.yaml
ousbots: I didn't actually look at what the test was for, just assumed it was trying to push the public images since the provided airgap file didn't exist. Instead it's trying to go far enough that the registry check happens, I think I've fixed it, but we'll see when it runs

.github/workflows/build-test.yaml
```diff
@@ -2612,6 +2612,8 @@ jobs:
         run: |
           set +e
 
+          touch not-an-app.airgap
```
sgalsaleh, line 2615: remove this? not sure why it was added. maybe for testing?

cmd/kots/cli/admin-console-push-images.go
```diff
@@ -124,6 +67,66 @@ func AdminPushImagesCmd() *cobra.Command {
 	return cmd
 }
 
+func genAndCheckPushOptions(endpoint string, cmd *cobra.Command) (*kotsadmtypes.PushImagesOptions, error) {
+	log := logger.NewCLILogger(cmd.OutOrStdout())
```
sgalsaleh, line 71: let's pass this as a parameter instead of initializing it here.

cmd/kots/cli/admin-console-push-images.go
```diff
@@ -124,6 +67,66 @@ func AdminPushImagesCmd() *cobra.Command {
...
+	v := viper.GetViper()
```
sgalsaleh, line 72: ditto

.github/workflows/build-test.yaml
ousbots: this test will fail with an error about "not-an-app.airgap" missing, but the intention of the test is for that command to get far enough that the registry check fails, which is my understanding of the log grep right below it. I'm trying to provide the command with just enough to get that far

cmd/kots/cli/admin-console-push-images.go
```diff
@@ -124,6 +67,66 @@ func AdminPushImagesCmd() *cobra.Command {
...
+
+	hostname, err := getHostnameFromEndpoint(endpoint)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed get hostname from endpoint")
+	}
+
+	username := v.GetString("registry-username")
+	password := v.GetString("registry-password")
+	if username == "" && password == "" {
+		namespace, err := getNamespaceOrDefault(v.GetString("namespace"))
```
sgalsaleh, line 82: ditto

sgalsaleh (approved):

PR #3843, opened 2023-04-26
Fix snapshot restore cancellation when there are multiple apps installed 

rianfowler:
- the app passed into the restore component was hardcoded to the first index. the partial snapshots flow was updated to pass the currently selected app slug into the component and then that slug is used to perform operations on the selected app. 

Fixes # https://app.shortcut.com/replicated/story/75454/with-multi-app-admin-console-cancelling-second-app-s-snapshot-restore-makes-the-api-call-to-first-app-s-app-slug-snapshot

- Please review for sufficient validation. 
- When the bug manifests, it would kick back to the wrong app and not cancel the restore
- This validation ensures that the app returns to the page for the app that has its restore attempted and cancelled
- The snapshots page will not render for an app that has an in progress snapshot. It will show the in progress snapshot page. So we should be able to validate that the snapshot was correctly cancelled simply by verifying that the correct snapshots page renders after cancellation 

https://www.loom.com/share/684c078f08df4de0a7db2a021b6eb3bd

New test completed successfully here: 

https://app.testim.io/#/project/wpYAooUimFDgQxY73r17/branch/rianfowler%2Fsc-75454%2Fwith-multi-app-admin-console-cancelling-second/test/Hadhz75RDWClUxQ7?result-id=t8YNr3MvtnDcY3pe

```release-note
Fixes an issue with multiple app installations where cancelling a restore of an application snapshot (partial snapshot) would sometimes not work. 
```

none


web/src/Root.tsx
```diff
@@ -608,7 +608,7 @@ const Root = () => {
                   )}
                 />
                 <ProtectedRoute
-                  path={["/snapshots/:tab?"]}
+                  path={["/snapshots/:tab?/:slug?"]}
```
rianfowler, line 611: this slug is already part of the url but we don't pass it into react router as a variable. this slug is needed to know what the selected app is in the snapshots page.

snapshots are a root page so there's no selected app in the way you have one on the dashboard page. it's selected from a drop down.

miaawong (approved):
nice

.github/workflows/build-test.yaml
```diff
@@ -708,7 +708,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        k8s_version: [ v1.23.15-k3s1,v1.24.9-k3s2,v1.25.6-k3s1,v1.26.1-k3s1 ]
+        k8s_version: [ v1.26.1-k3s1 ] # there's no need to run this test on multiple k3s versions
```
cbodonnell, line 711: is this an intentional change? makes sense to only run the new test below on one version, but this one might be good to run on multiple supported versions

rianfowler:
```suggestion
        k8s_version: [ v1.23.15-k3s1,v1.24.9-k3s2,v1.25.6-k3s1,v1.26.1-k3s1 ]
```

PR #3851, opened 2023-05-01
Improvements to application resource management

sgalsaleh:

- Introduces a known deterministic order (inspired by Helm) for applying and deleting resources.
- Uses the [weight](https://docs.replicated.com/reference/custom-resource-helmchart#weight) field from the HelmChart custom resource to determine the order in which to uninstall charts that have `useHelmInstall: true`. Charts are uninstalled by weight in descending order, with higher weights uninstalled first.
- Refactors how applications are deployed/un-deployed.
- Application Helm charts are now uninstalled first, then other native Kubernetes manifests.
- Runs `helm uninstall` for Helm charts using the native installation workflow when un-deploying an application.

Fixes #

```release-note
- Introduces a deterministic order for applying and deleting Kubernetes manifests based on the resource kind.
- Uses the [weight](https://docs.replicated.com/reference/custom-resource-helmchart#weight) field from the HelmChart custom resource to determine the order in which to uninstall charts that have `useHelmInstall: true`. Charts are uninstalled by weight in descending order, with higher weights uninstalled first.
- Application Helm charts are now uninstalled first, then other Kubernetes manifests.
- Fixes an issue where snapshot restores hanged if RabbitMQ cluster custom resources are used.
- Fixes an issue where Helm releases were not uninstalled when un-deploying an application using the the [kots remove](/reference/kots-cli-remove) command and passing the `--undeploy` flag.
- Fixes an issue where Helm charts that were deployed to a different namespace than were KOTS is running, and were installed using the native Helm installation workflow, were not uninstalled when they were removed from subsequent application releases.
```

https://github.com/replicatedhq/replicated-docs/pull/1117


pavansokkenagaraj (dismissed):

pkg/operator/client/delete.go
```diff
@@ -0,0 +1,392 @@
...
+		err := clientset.CoreV1().PersistentVolumeClaims(c.TargetNamespace).Delete(context.TODO(), pvc, opts)
+		if err != nil && !kuberneteserrors.IsNotFound(err) {
+			return errors.Wrapf(err, "failed to delete pvc %s", pvc)
+		}
+	}
+
+	return nil
+}
+
+func getLabelSelector(appLabelSelector *metav1.LabelSelector) string {
```
cbodonnell, line 378: i think this function might already exist in `"k8s.io/apimachinery/pkg/labels"` as `SelectorFromSet(ls labels.Set) labels.Selector`?  we use that in a bunch of places

cbodonnell (dismissed):

cbodonnell (approved):

PR #3852, opened 2023-05-01
show last updated in details modal

miaawong:
#### What this PR does / why we need it: This PR adds 'last updated' time in the resource status modal to indicate when the dated was last automatically fetched.
Screenshot 2023-05-01 at 3 07 21 PM

Fixes # https://app.shortcut.com/replicated/story/63472/add-fetch-refetch-disclosures-to-details-modal

```release-note
Display 'last updated' time in the resource status modal to indicate when the data was last automatically fetched.
```

NONE

### DIFF
--- a/web/src/features/Dashboard/components/Dashboard.tsx
+++ b/web/src/features/Dashboard/components/Dashboard.tsx
@@ -102,6 +102,8 @@ type State = {
   viewAirgapUpdateError: boolean;
   viewAirgapUploadError: boolean;
   slowLoader: boolean;
+  lastUpdated: number;
+  lastUpdatedDate: Date;
 };
 
 const Dashboard = (props: Props) => {
@@ -144,6 +146,8 @@ const Dashboard = (props: Props) => {
       viewAirgapUpdateError: false,
       viewAirgapUploadError: false,
       slowLoader: false,
+      lastUpdated: 0,
+      lastUpdatedDate: new Date(),
     }
   );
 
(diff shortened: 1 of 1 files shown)


web/src/features/Dashboard/components/Dashboard.tsx
```diff
@@ -513,10 +517,25 @@ const Dashboard = (props: Props) => {
...
           metrics: selectedAppClusterDashboardResponse.metrics,
         },
+        lastUpdated: 0,
+        lastUpdatedDate: new Date(),
       });
     }
   }, [selectedAppClusterDashboardResponse]);
 
+  useEffect(() => {
+    const interval = setInterval(() => {
```
miaawong, line 527: this interval checks every second for the time that has passed since appStatus is updated.

web/src/features/Dashboard/components/Dashboard.tsx
```diff
@@ -513,10 +517,25 @@ const Dashboard = (props: Props) => {
             selectedAppClusterDashboardResponse.prometheusAddress,
           metrics: selectedAppClusterDashboardResponse.metrics,
         },
+        lastUpdated: 0,
```
miaawong, line 520: resets everything once there's new data.

miaawong:
I'm not too confident about this implementation. Feedback or other alternatives are welcomed.

web/src/features/Dashboard/components/Dashboard.tsx
rianfowler: this is probably fine. 

- another way to do it would be to use `setState` in an interval that just updates the current time. (make sure to set your date with a function like [this](https://eight-bites.blog/en/2021/05/setinterval-setstate/) to avoid stale state: 

```
import React, { useEffect, useState } from 'react'

export const Sample = () => {
  let timer
  const [count, setCount] = useState(0)
  
  const updateCount = () => {
    timer = !timer && setInterval(() => {
      console.log('ticking')
+     setCount(prevCount => prevCount + 1) // new
    }, 1000)
  }
  
  useEffect(() => {
    updateCount()
    
    return () => clearInterval(timer)
  }, [])
  
  
  return <div>count is: {count}</div>
}
```

- the fetch handler could set the fetch time and the current time. 

- then you could have a `useEffect` that's watching both and updates the display string when they change. that way all of the logic is in taht effect and the jsx is just rendering the string.

rianfowler: ah nevermind. that doesn't make sense. i think the link has an example of how you could do it with one use effect. either way this is fine.

rianfowler (approved):

PR #3856, opened 2023-05-02
update CLI support bundle command

miaawong:
#### What this PR does / why we need it: This PR updates all CLI support bundle commands in admin console. 

Screenshot 2023-05-24 at 9 40 22 AM

Fixes # https://app.shortcut.com/replicated/story/69757/show-customer-improved-cli-support-bundle-command-for-collecting-support-bundles

```release-note
Update improved CLI support-bundle commands
```

### DIFF
--- a/integration/replicated/tests/kitchen-sink/expected/upstream/manifests/support-bundle.yaml
+++ b/integration/replicated/tests/kitchen-sink/expected/upstream/manifests/support-bundle.yaml
@@ -19,7 +19,7 @@ spec:
         collectorName: ping-google
         image: flungo/netutils
         command: ["ping"]
-        args: ["www.google.com"]
+        args: [["-w", "1", "www.google.com"]
         timeout: 5s
     - http:
         collectorName: echo-ip
--- a/integration/replicated/tests/kitchen-sink/expected/upstream/manifests/support-bundle.yaml
+++ b/integration/replicated/tests/kitchen-sink/expected/upstream/manifests/support-bundle.yaml
@@ -19,7 +19,7 @@ spec:
         collectorName: ping-google
         image: flungo/netutils
         command: ["ping"]
-        args: [["-w", "1", "www.google.com"]
+        args: ["-w", "1", "www.google.com"]
         timeout: 5s
     - http:
         collectorName: echo-ip
--- a/integration/replicated/tests/kitchen-sink/expected/upstream/manifests/support-bundle.yaml
+++ b/integration/replicated/tests/kitchen-sink/expected/upstream/manifests/support-bundle.yaml
@@ -19,7 +19,7 @@ spec:
         collectorName: ping-google
         image: flungo/netutils
         command: ["ping"]
-        args: ["-w", "1", "www.google.com"]
+        args: ["www.google.com"]
         timeout: 5s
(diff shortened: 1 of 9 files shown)


web/src/components/UploadAirgapBundle.jsx
```diff
@@ -313,37 +314,9 @@ class UploadAirgapBundle extends React.Component {
     }, 1000);
   };
 
-  getSupportBundleCommand = async (slug) => {
-    const res = await fetch(
-      `${process.env.API_ENDPOINT}/troubleshoot/app/${slug}/supportbundlecommand`,
```
rianfowler, line 318: Should we preserve this call and have this copy updated in the troubleshoot app?

rianfowler (dismissed):
added one comment

web/src/components/UploadAirgapBundle.jsx
miaawong: @rianfowler That might be a good idea. Who would reach out to?

rianfowler: I would just tag the requester and jason on the ticket. something like "we can hardcode this on the FE but it looks like it's currently getting it from a troubleshoot api."

banjoh: I think this makes sense since the response will have the correct troubleshoot command.

banjoh (dismissed):

web/src/components/UploadAirgapBundle.jsx
```diff
@@ -27,7 +27,10 @@ class UploadAirgapBundle extends React.Component {
     fileUploading: false,
     registryDetails: {},
     preparingOnlineInstall: false,
-    supportBundleCommand: "kubectl support-bundle --load-cluster-specs",
+    supportBundleCommand: [
```
banjoh, line 30: Nit pick: Can this be DRY'd by using a constant to store this commands array?

integration/replicated/tests/kitchen-sink/expected/upstream/manifests/support-bundle.yaml
```diff
@@ -19,7 +19,7 @@ spec:
         collectorName: ping-google
         image: flungo/netutils
         command: ["ping"]
-        args: ["www.google.com"]
+        args: [["-w", "1", "www.google.com"]
```
banjoh, line 22: This does't look right

pkg/supportbundle/supportbundle.go
```diff
@@ -129,7 +129,7 @@ func GetBundleCommand(appSlug string) []string {
 
 	command := []string{
 		"curl https://krew.sh/support-bundle | bash",
-		fmt.Sprintf("kubectl support-bundle %s --redactors=%s\n", GetSpecURI(appSlug), redactors),
+		fmt.Sprintf("kubectl support-bundle --load-cluster-specs --redactors=%s\n", redactors),
 	}
```
banjoh, line 133: Redactors are also discovered but the support-bundle spec, but I'm not sure whether KOTS labels them at the moment. I think we can address this in a separate GH issue.

banjoh (approved):

PR #3858, opened 2023-05-02
feat: native helm without kustomize

cbodonnell:

Fixes #

```release-note

```


pkg/diff/diff.go
```diff
@@ -1,11 +1,12 @@
-package kustomize
+package diff
```
cbodonnell, line 1: Moved diffing logic to a new package mainly to avoid circular dependency issues

pkg/helmdeploy/write.go
```diff
@@ -0,0 +1,212 @@
...
+	for _, downstream := range opts.Downstreams {
+		for _, helmChart := range opts.KotsKinds.V1Beta2HelmCharts.Items {
+			renderedPath, err := renderHelmChart(opts, downstream, &helmChart)
+			if err != nil {
+				return errors.Wrap(err, "failed to render helm chart")
+			}
+
+			if err := processImages(opts, renderedPath); err != nil {
+				return errors.Wrap(err, "failed to process images")
+			}
```
cbodonnell, line 122: This is the first place where we have the rendered chart manifests, so we push images here.  Is there a better place to do this, or could this cause any problems?  This would normally be done as part of the midstream flow, but that doesn't exist in the new helm deploy.

pkg/operator/client/client.go
```diff
@@ -247,88 +250,147 @@ func (c *Client) deployManifests(deployArgs operatortypes.DeployAppArgs) (*deplo
...
+	v1Beta1ChartsDir := filepath.Join(curV1Beta1HelmDir, "charts")
+	v1Beta2ChartsDir := filepath.Join(curV1Beta2HelmDir, "helm")
+
+	if len(deployArgs.V1Beta1ChartsArchive) > 0 {
+		v1Beta1HelmCharts := []v1beta1.HelmChart{}
+		if deployArgs.KotsKinds != nil && deployArgs.KotsKinds.V1Beta1HelmCharts != nil {
+			v1Beta1HelmCharts = deployArgs.KotsKinds.V1Beta1HelmCharts.Items
+		}
+
+		// TODO: can we get rid of the OrderedDir type and just use kots.io HelmCharts?
```
cbodonnell, line 334: Left this note here since since the `OrderedDir` type mimics the `v1beta1.HelmChart` type, but one is used to install and the other is used to uninstall.

pkg/operator/client/types/types.go
```diff
@@ -0,0 +1,48 @@
...
+// OrderedDir represents a kots.io/v1beta1 HelmChart deployment
+type OrderedDir struct {
+	Name         string
+	Weight       int64
+	ChartName    string
+	ChartVersion string
+	ReleaseName  string
+	Namespace    string
+	UpgradeFlags []string
+}
```
cbodonnell, line 13: Left this note here since since the `OrderedDir` type mimics the `v1beta1.HelmChart` type, but one is used to install and the other is used to uninstall.

pkg/operator/client/types/types.go
```diff
@@ -0,0 +1,48 @@
+package types
+
+// TODO: can this type be removed? all fields exist in HelmChart
```
cbodonnell, line 3: Left this note here since since the `OrderedDir` type mimics the `v1beta1.HelmChart` type, but one is used to install and the other is used to uninstall.

pkg/base/replicated.go
```diff
@@ -392,19 +403,29 @@ func findAllKotsHelmCharts(upstreamFiles []upstreamtypes.UpstreamFile, builder t
...
+			}
+			kotsV1Beta2HelmCharts = append(kotsV1Beta2HelmCharts, *helmChart)
 		}
-
-		kotsHelmCharts = append(kotsHelmCharts, helmChart)
 	}
 
-	return kotsHelmCharts, nil
+	// filter out any charts where a cooresponding v1beta2 HelmChart exists
+	filteredCharts := kotsutil.FilterV1Beta1ChartsWithV1Beta2Charts(kotsV1Beta1HelmCharts, kotsV1Beta2HelmCharts)
```
sgalsaleh, line 423: why add v1beta2 helm charts then filter them? how about just not adding them in the first place?

cbodonnell: we're filtering the v1beta1 resources if a matching v1beta2 resource exists for the same helm release.  we collect all v1beta2 resources in addition to v1beta1 so that we can make this determination.  should this logic be moved?

pkg/operator/client/deploy.go
```diff
@@ -216,38 +218,55 @@ func (c *Client) ensureResourcesPresent(deployArgs operatortypes.DeployAppArgs)
...
-func (c *Client) installWithHelm(helmDir string, kotsCharts []v1beta1.HelmChart) (*commandResult, error) {
-	version := "3"
-	chartsDir := filepath.Join(helmDir, "charts")
-
-	orderedDirs, err := getSortedCharts(chartsDir, kotsCharts, c.TargetNamespace, false)
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to get sorted helm charts")
-	}
+func (c *Client) installWithHelm(v1Beta1ChartsDir, v1beta2ChartsDir string, kotsCharts []kotsutil.HelmChartInterface) (*commandResult, error) {
+	// sort the charts by weight and then by name
```
sgalsaleh, line 222: why not just use the `getSortedCharts` function?

pkg/operator/client/deploy.go
```diff
@@ -216,38 +218,55 @@ func (c *Client) ensureResourcesPresent(deployArgs operatortypes.DeployAppArgs)
...
+		} else if kotsChart.GetAPIVersion() == "kots.io/v1beta2" {
+			installDir := filepath.Join(v1beta2ChartsDir, kotsChart.GetDirName())
+			chartPath := filepath.Join(installDir, fmt.Sprintf("%s-%s.tgz", kotsChart.GetChartName(), kotsChart.GetChartVersion()))
+			valuesPath := filepath.Join(installDir, "values.yaml")
+			args = []string{"upgrade", "-i", kotsChart.GetReleaseName(), chartPath, "-f", valuesPath, "--timeout", "3600s"}
+		} else {
+			return nil, errors.Errorf("unknown api version %s", kotsChart.GetAPIVersion())
+		}
+
+		chartNamespace := kotsChart.GetNamespace()
```
sgalsaleh, line 248: if the `getSortedCharts` is used, then this should just be `dir.Namespace`

pkg/operator/client/deploy.go
```diff
@@ -216,38 +218,55 @@ func (c *Client) ensureResourcesPresent(deployArgs operatortypes.DeployAppArgs)
...
+	for _, kotsChart := range kotsCharts {
+		var args []string
+		if kotsChart.GetAPIVersion() == "kots.io/v1beta1" {
+			installDir := filepath.Join(v1Beta1ChartsDir, kotsChart.GetDirName())
+			args = []string{"upgrade", "-i", kotsChart.GetReleaseName(), installDir, "--timeout", "3600s"}
+		} else if kotsChart.GetAPIVersion() == "kots.io/v1beta2" {
+			installDir := filepath.Join(v1beta2ChartsDir, kotsChart.GetDirName())
+			chartPath := filepath.Join(installDir, fmt.Sprintf("%s-%s.tgz", kotsChart.GetChartName(), kotsChart.GetChartVersion()))
+			valuesPath := filepath.Join(installDir, "values.yaml")
+			args = []string{"upgrade", "-i", kotsChart.GetReleaseName(), chartPath, "-f", valuesPath, "--timeout", "3600s"}
```
sgalsaleh, line 243: how about separating out the shared args? e.g.: move `"upgrade", "-i", kotsChart.GetReleaseName()` to the array initialization, and the `"--timeout", "3600s"` to afterwards

pkg/operator/client/deploy.go
```diff
@@ -365,19 +384,27 @@ func findChartNameAndVersion(chartDir string) (string, string, error) {
...
-func (c *Client) uninstallWithHelm(helmDir string, kotsCharts []v1beta1.HelmChart) error {
+func (c *Client) uninstallWithHelm(kotsCharts []kotsutil.HelmChartInterface) error {
 	version := "3"
-	chartsDir := filepath.Join(helmDir, "charts")
 
-	orderedDirs, err := getSortedCharts(chartsDir, kotsCharts, c.TargetNamespace, true)
-	if err != nil {
-		return errors.Wrap(err, "failed to get sorted helm charts")
-	}
+	// uninstall in reverse order
```
sgalsaleh, line 390: ditto, why not just use `getSortedCharts`?

pkg/operator/client/deploy.go
```diff
@@ -365,19 +384,27 @@ func findChartNameAndVersion(chartDir string) (string, string, error) {
...
+	})
+
+	for _, chart := range kotsCharts {
+		args := []string{"uninstall", chart.GetReleaseName()}
 
-	for _, dir := range orderedDirs {
-		args := []string{"uninstall", dir.ReleaseName}
-		if dir.Namespace != "" {
-			args = append(args, "-n", dir.Namespace)
+		chartNamespace := chart.GetNamespace()
```
sgalsaleh, line 401: ditto, would be `dir.Namespace` if `getSortedCharts` is used

pkg/operator/client/deploy.go
```diff
@@ -390,80 +417,63 @@ func (c *Client) uninstallWithHelm(helmDir string, kotsCharts []v1beta1.HelmChar
...
 
 	return nil
 }
 
-func getRemovedCharts(prevDir string, curDir string, previousKotsCharts []v1beta1.HelmChart, curKotsCharts []v1beta1.HelmChart) ([]v1beta1.HelmChart, error) {
-	if prevDir == "" {
-		return []v1beta1.HelmChart{}, nil
-	}
+// getRemovedCharts returns a list of helm release names that were removed in the current version
+func getRemovedCharts(prevV1Beta1Dir, curV1Beta1Dir string, previousV1Beta1KotsCharts, currentV1Beta1KotsCharts []kotsutil.HelmChartInterface, prevV1Beta2Dir, curV1Beta2Dir string, previousV1Beta2KotsCharts, currentV1Beta2KotsCharts []kotsutil.HelmChartInterface) ([]kotsutil.HelmChartInterface, error) {
```
sgalsaleh, line 428: let's move all these parameters to a new struct. something like `GetRemovedChartsOptions`

pkg/operator/client/deploy.go
```diff
@@ -481,6 +491,60 @@ func getRemovedCharts(prevDir string, curDir string, previousKotsCharts []v1beta
...
+			if kotsChart.GetChartVersion() == chartVersion && kotsChart.GetChartName() == chartName && kotsChart.GetDirName() == dirName {
+				matching = append(matching, kotsChart)
+			}
+		}
+	}
+
+	return matching, nil
+}
+
+func findMatchingV1Beta2Charts(chartsDir string, kotsCharts []kotsutil.HelmChartInterface) ([]kotsutil.HelmChartInterface, error) {
```
sgalsaleh, line 524: why does this need to be a different function? why not just rename and use the `findMatchingV1Beta1Charts` function?

pkg/operator/client/types/types.go
sgalsaleh: in the most recent changes introduced in 1.98.2, OrderedDir is also used for uninstall. so this can probably be removed

pkg/operator/operator.go
```diff
@@ -755,7 +770,7 @@ func (o *Operator) UndeployApp(a *apptypes.App, d *downstreamtypes.Downstream, i
 		KustomizeVersion:     kotsKinds.KotsApplication.Spec.KustomizeVersion,
 		AdditionalNamespaces: kotsKinds.KotsApplication.Spec.AdditionalNamespaces,
 		Manifests:            base64EncodedManifests,
-		Charts:               chartArchive,
+		V1Beta1ChartsArchive: chartArchive,
```
sgalsaleh, line 773: if `getSortedCharts` is used, then you'd need to pass the v1beta2 archive as well.

pkg/pull/pull.go
```diff
@@ -332,13 +333,13 @@ func Pull(upstreamURI string, pullOptions PullOptions) (string, error) {
 		renderDir = filepath.Join(pullOptions.RootDir, u.Name)
 	}
 
-	newHelmCharts, err := kotsutil.LoadHelmChartsFromPath(renderDir)
```
sgalsaleh, line 335: shouldn't we enforce this for v1beta2 as well? e.g.: can't migration from old replicated Helm to v1beta1 _or_ v1beta2 native Helm

pkg/helmdeploy/write.go
```diff
@@ -0,0 +1,212 @@
...
+		return nil
+	}
+
+	var dockerHubRegistryCreds registry.Credentials
+	dockerhubSecret, _ := registry.GetDockerHubPullSecret(opts.Clientset, util.PodNamespace, opts.ProcessImageOptions.Namespace, opts.ProcessImageOptions.AppSlug)
+	if dockerhubSecret != nil {
+		dockerHubRegistryCreds, _ = registry.GetCredentialsForRegistryFromConfigJSON(dockerhubSecret.Data[".dockerconfigjson"], registry.DockerHubRegistryName)
+	}
+
+	_, err := midstream.RewriteBaseImages(opts.ProcessImageOptions, renderedPath, opts.KotsKinds, opts.KotsKinds.License, dockerHubRegistryCreds, opts.Log)
```
sgalsaleh, line 206: I don't love the name of this function... I wonder if we can refactor some of that code at some point, since neither `midstream` nor `RewriteBaseImages` make sense in this context.

pkg/helmdeploy/write.go
```diff
@@ -0,0 +1,212 @@
+package helmdeploy
```
sgalsaleh, line 1: How about we move the `helmdeploy`, `kustomize`, `diff` packages into a single package and name it `app-archive`?

pkg/handlers/rendered_contents.go
```diff
@@ -94,9 +95,16 @@ func (h *Handler) GetAppRenderedContents(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
-	_, chartsFilesMap, err := kustomize.GetRenderedChartsArchive(archivePath, d.Name, kotsKinds.GetKustomizeBinaryPath())
+	_, v1Beta1ChartsFilesMap, err := kustomize.GetRenderedChartsArchive(archivePath, d.Name, kotsKinds.GetKustomizeBinaryPath())
```
sgalsaleh, line 98: The name `GetRenderedChartsArchive` doesn't distinguish between v1beta1 and v1beta2, how about renaming it to:
```suggestion
	_, v1Beta1ChartsFilesMap, err := kustomize.GetRenderedV1Beta1ChartsArchive(archivePath, d.Name, kotsKinds.GetKustomizeBinaryPath())
```

pkg/operator/client/types/types.go
cbodonnell: refactored getSortedCharts to handle v1beta1 and v1beta2 charts

pkg/helmdeploy/write.go
```diff
@@ -0,0 +1,212 @@
...
+
+	err = os.WriteFile(filepath.Join(renderedPath, "all.yaml"), manifests.Bytes(), 0644)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to write all.yaml")
+	}
+
+	return renderedPath, nil
+}
+
+// processImages will pull any public images from online and copy them to the configured private registry
```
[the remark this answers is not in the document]
cbodonnell: are private images pulled via proxy in these scenarios?

pkg/operator/client/deploy.go
```diff
@@ -365,19 +426,59 @@ func findChartNameAndVersion(chartDir string) (string, string, error) {
...
-	version := "3"
-	chartsDir := filepath.Join(helmDir, "charts")
+func findChartTgz(dir string) (string, error) {
+	files, err := ioutil.ReadDir(dir)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to read dir")
+	}
+
+	for _, file := range files {
+		if filepath.Ext(file.Name()) == ".tgz" {
```
sgalsaleh, line 436: can it be `.tar.gz`?

pkg/diff/helm.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/pkg/errors"
+	"github.com/replicatedhq/kots/pkg/util"
+)
+
+// GetRenderedV1Beta2FileMap returns a map of the rendered v1beta2 charts to be deployed
+func GetRenderedV1Beta2FileMap(deployedVersionArchive, downstream string) (map[string][]byte, error) {
```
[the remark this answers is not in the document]
cbodonnell: agree.  going to see about reorganizing them per your [comment below](https://github.com/replicatedhq/kots/pull/3858#discussion_r1189148732)

pkg/operator/client/deploy.go
cbodonnell: looks like helm will support any extension so long as the data is a tarball: https://github.com/helm/helm/blob/main/pkg/chart/loader/archive.go

in this particular case, I think we are safe because we are the ones that [save the upstream chart archive as `.tgz`](https://github.com/replicatedhq/kots/blob/cbo/sc-74872/implement-the-native-helm-without-kustomize/pkg/helmdeploy/write.go#L42).  we can make this more generic if you want and try to read the file as a tarball regardless of extension?

pkg/operator/client/deploy.go
cbodonnell: good idea :)

pkg/operator/client/client.go
```diff
@@ -247,88 +247,148 @@ func (c *Client) deployManifests(deployArgs operatortypes.DeployAppArgs) (*deplo
...
+			v1Beta1ChartsDir = filepath.Join(prevV1Beta1HelmDir, "charts")
+		}
+		v1Beta2ChartsDir := ""
+		if prevV1Beta2HelmDir != "" {
+			v1Beta2ChartsDir = filepath.Join(prevV1Beta2HelmDir, "helm")
 		}
-		defer os.RemoveAll(tmpDir)
 
-		err = ioutil.WriteFile(path.Join(tmpDir, "archive.tar.gz"), deployArgs.Charts, 0644)
+		err := c.uninstallWithHelm(v1Beta1ChartsDir, v1Beta2ChartsDir, kotsCharts)
```
sgalsaleh, line 329: hmmm, shouldn't this be
```suggestion
		err := c.uninstallWithHelm(v1Beta1ChartsDir, v1Beta2ChartsDir, removedCharts)
```

i'm also looking at the current code, and there might be a bug where if a single chart is removed, all charts are uninstalled and re-installed. we should fix that if that's the case.

cbodonnell: per some internal discussions, we are going to modify the `getSortedCharts` method to use the kots HelmChart resources as the source for the deployments/undeployments rather than the presence of a directory

pkg/operator/client/deploy.go
cbodonnell: nice catch.  looks like this was from before the interface was created.  removing...

cbodonnell: it's different since v1beta1 charts are in a directory, but the v1beta2 chart is just the tgz archive plus a user-provided values.yaml file

cbodonnell: changed it to be a common function that will act based on the `apiVersion`

pkg/pull/pull.go
cbodonnell: yeah I'll add that in for v1beta2 charts as well

pkg/helmdeploy/write.go
cbodonnell: seems we might have to do this refactor to get all of those other functions into a common `apparchive` pkg.  I'll have a go at this...

pkg/operator/client/deploy.go
cbodonnell: > okay, as long as we have tests that cover this :)

what's the case you're concerned about?  i'll see about adding tests

pkg/diff/helm.go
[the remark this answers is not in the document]
cbodonnell: did the refactor

pkg/diff/diff.go
cbodonnell: this comment is now outdated

pkg/operator/client/deploy.go
cbodonnell: thinking of changing this to be more generic since helm does not search based on extension, but whether the file can be read as a tarball.  we should probably do the same

.gitignore
```diff
@@ -28,7 +28,7 @@ supportbundle.tar.gz
 # for the embed experiment
 kubernetes/
 kots.db
-archives/
```
cbodonnell, line 31: changed this since it was actually filtering out files in `pkg/archives/` too

pkg/operator/client/deploy.go
cbodonnell: updated in https://github.com/replicatedhq/kots/pull/3858/commits/9afb7206f6c87dd0e7b1961eb2e93c26f612d1b4

pkg/apparchive/helm-v1beta2.go
```diff
@@ -140,20 +165,40 @@ type HelmWriteOptions struct {
...
 	for _, downstream := range opts.Downstreams {
 		for _, helmChart := range opts.KotsKinds.V1Beta2HelmCharts.Items {
-			renderedPath, err := renderHelmChart(opts, downstream, &helmChart)
+			// template the chart with the values to the rendered dir for the downstream
+			renderedPath := path.Join(opts.RenderedDir, downstream, "helm", helmChart.GetDirName())
+			if err := templateHelmChartWithValuesToDir(opts.HelmDir, &helmChart, "values.yaml", renderedPath, opts.Log.Debug); err != nil {
+				return errors.Wrap(err, "failed to template helm chart for rendered dir")
+			}
+
+			if !opts.ProcessImageOptions.RewriteImages || opts.ProcessImageOptions.AirgapRoot != "" {
```
cbodonnell, line 182: I added this back since we only want to process images for online.  Is that correct?

pkg/apparchive/helm-v1beta2.go
```diff
@@ -0,0 +1,251 @@
...
+		valuesPath := path.Join(chartDir, "values.yaml")
+		if err := ioutil.WriteFile(valuesPath, []byte(renderedValuesContent), 0644); err != nil {
+			return errors.Wrap(err, "failed to write values file")
+		}
+	}
+
+	return nil
+}
+
+type HelmWriteOptions struct {
```
sgalsaleh, line 158:
```suggestion
type HelmV1Beta2WriteOptions struct {
```

pkg/apparchive/helm-v1beta2.go
sgalsaleh: right, but `|| opts.ProcessImageOptions.AirgapRoot != ""` isn't needed here because you can't have an airgap install/bundle without an on-prem registry (e.g. `RewriteImages` will always be `true` in airgap)

pkg/apparchive/helm-v1beta2.go
```diff
@@ -0,0 +1,288 @@
...
+		if err != nil {
+			return errors.Wrap(err, "failed to marshal builder values")
+		}
+
+		renderedBuilderValuesContent, err := builder.RenderTemplate(fmt.Sprintf("%s-builder-values", helmChart.GetDirName()), string(builderValuesContent))
+		if err != nil {
+			return errors.Wrap(err, "failed to render builder values")
+		}
+
+		builderValuesPath := path.Join(chartDir, "builder-values.yaml")
```
sgalsaleh, line 149: instead of storing it in the `helm` directory where it's not really relevant for the user to see, can we move rendering the builder values to be in the same place where processing the images happen?

pkg/apparchive/helm-v1beta2.go
sgalsaleh: oh, you're right. my bad.

pkg/archives/archives.go
```diff
@@ -127,3 +127,26 @@ func ExtractTGZArchiveFromReader(tgzReader io.Reader, destDir string) error {
...
+		return false
+	}
+	defer gzipReader.Close()
+
+	tarReader := tar.NewReader(gzipReader)
+
+	// try to read the first file header from the tar archive
+	_, err = tarReader.Next()
+	if err != nil {
+		if err == io.EOF {
```
sgalsaleh, line 145: why is this check necessary if it's returning the same thing? :)

pkg/operator/client/deploy.go
```diff
@@ -481,6 +584,52 @@ func getRemovedCharts(prevDir string, curDir string, previousKotsCharts []v1beta
...
+			chartDir := filepath.Join(chartsDir, dirName)
+
+			var chartName, chartVersion string
+			if kotsChart.GetAPIVersion() == "kots.io/v1beta1" {
+				chartName, chartVersion, err = findChartNameAndVersion(chartDir)
+				if err != nil {
+					return nil, errors.Wrapf(err, "failed to find chart name and version in %s", chartDir)
+				}
+			} else if kotsChart.GetAPIVersion() == "kots.io/v1beta2" {
+				archivePath, err := findChartTgz(chartDir)
```
sgalsaleh, line 611: can you add a comment here that v1beta2 charts are tgz? maybe add a comment to the above block for v1beta1 that they're not tgz and already extracted?

pkg/pull/pull.go
```diff
@@ -351,14 +358,24 @@ func Pull(upstreamURI string, pullOptions PullOptions) (string, error) {
...
 				if prevChart.Spec.Chart.Name != newChart.Spec.Chart.Name {
 					continue
 				}
 				if prevChart.Spec.UseHelmInstall != newChart.Spec.UseHelmInstall {
 					return "", errors.Errorf("deployment method for chart %s has changed", newChart.Spec.Chart.Name)
 				}
 			}
+
+			for _, newChart := range v1Beta2HelmCharts {
+				if prevChart.Spec.Chart.Name != newChart.Spec.Chart.Name {
```
sgalsaleh, line 371: should this be release name instead of chart name?

cbodonnell: I would agree.  The v1beta1 check uses chart name, but that could be a bug.  I can change both.

pkg/apparchive/helm-v1beta2.go
cbodonnell: yes. as discussed, going to move the image processing to the midstream flow.  this is when the helm archive will be written to the `helm` directory and then it will be rendered with builder values to process images

pkg/helmdeploy/write.go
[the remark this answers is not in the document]
cbodonnell: yes

pkg/apparchive/helm-v1beta2.go
```diff
@@ -0,0 +1,299 @@
...
+
+	err = os.WriteFile(filepath.Join(outputDir, "all.yaml"), manifests.Bytes(), 0644)
+	if err != nil {
+		return errors.Wrap(err, "failed to write all.yaml")
+	}
+
+	return nil
+}
+
+func processOnlineImages(opts WriteV1Beta2HelmChartsOptions, helmChart *kotsv1beta2.HelmChart, chartDir string) error {
```
sgalsaleh, line 250: since this is a shared package, let's name functions specifically. how about:
```suggestion
func processV1Beta2HelmChartImages(opts WriteV1Beta2HelmChartsOptions, helmChart *kotsv1beta2.HelmChart, chartDir string) error {
```

sgalsaleh (dismissed):

sgalsaleh (approved):

PR #3861, opened 2023-05-03
set httpOnly auth cookie when logging in through OIDC

garcialuis:
Sets the signedToken as an httpOnly cookie when the user logs in via OIDC. 
The cookie is used to authenticate the user's session.

Fixes #

```release-note
NONE
```

NONE

### DIFF
--- a/pkg/handlers/login.go
+++ b/pkg/handlers/login.go
@@ -373,17 +373,17 @@ func (h *Handler) OIDCLoginCallback(w http.ResponseWriter, r *http.Request) {
 
 	expire := time.Now().Add(30 * time.Minute)
 
-	// token cookie
-	tokenCookie := http.Cookie{
+	// cookie for identity service
+	identityServiceCookie := http.Cookie{
 		Name:    "identity-service-login",
 		Value:   "true",
 		Expires: expire,
 		Path:    "/",
 	}
 	if u.Scheme == "https" {
-		tokenCookie.Secure = true
+		identityServiceCookie.Secure = true
 	}
-	http.SetCookie(w, &tokenCookie)
+	http.SetCookie(w, &identityServiceCookie)
 
 	signedTokenCookie, err := session.GetSessionCookie(responseToken, expire, redirectURL)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/handlers/login.go
```diff
@@ -385,6 +385,17 @@ func (h *Handler) OIDCLoginCallback(w http.ResponseWriter, r *http.Request) {
...
+	signedTokenCookie, err := session.GetSessionCookie(responseToken, expire, "")
+	if err != nil {
+		logger.Error(errors.Wrap(err, "failed to get session cookie"))
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	if u.Scheme == "https" {
+		signedTokenCookie.Secure = true
+	}
+	http.SetCookie(w, signedTokenCookie)
```
cbodonnell, line 397: what's the reason for passing an empty string to the function? also, doesn't the function handle whether or not `Secure` is set for the cookie?

cbodonnell: could maybe instead pass `u.String()` and let the function handle it?

garcialuis: The third parameter for that function is the origin url and that's how it determines if it's secure or not. I left it empty and check the `u.Scheme` instead so that it was consistent with the others in this function but I can pass in the `redirectURL` and it would be the same result too.

cbodonnell (approved):

cbodonnell (approved):

pkg/handlers/login.go
```diff
@@ -385,6 +385,14 @@ func (h *Handler) OIDCLoginCallback(w http.ResponseWriter, r *http.Request) {
 	}
 	http.SetCookie(w, &tokenCookie)
 
+	signedTokenCookie, err := session.GetSessionCookie(responseToken, expire, redirectURL)
```
sgalsaleh, line 388: this doesn't look right. the identity service token cookie is already set above, and that's the one that should be used.

web/src/features/Auth/components/SecureAdminConsole.tsx
```diff
@@ -251,17 +250,15 @@ class SecureAdminConsole extends React.Component<Props, State> {
   async componentDidMount() {
     window.addEventListener("keydown", this.submitForm);
 
-    const token = Utilities.getCookie("token");
+    const token = Utilities.getCookie("identity-service-login");
```
sgalsaleh, line 253:
```suggestion
    const isIdentityServiceLogin = Utilities.getCookie("identity-service-login");
```

cbodonnell (approved):

sgalsaleh (approved):

PR #3865, opened 2023-05-03
remove containerName: rqlite from sb spec

pavansokkenagaraj:
- use `containerName: rqlite` in all the rqlite statefulset spec
- remove `containerName: rqlite` from sb spec
reason: if the container name for rqlite is changed, the collector will fail.
relying on troubleshoot to select the default container name for collection is better.
https://troubleshoot.sh/docs/collect/exec/#containername-optional
```
The name of the container in which to execute the command. if not specified, the first container in the pod will be used.
```

Fixes # https://app.shortcut.com/replicated/story/76050/rqlite-db-dump-is-no-longer-included-in-support-bundle

```release-note
Fixes a bug where the rqlite collector was unable to collect a data dump if the name of the rqlite container was changed.
```

NONE

### DIFF
--- a/pkg/supportbundle/defaultspec/spec.yaml
+++ b/pkg/supportbundle/defaultspec/spec.yaml
@@ -23,7 +23,6 @@ spec:
     - exec:
         collectorName: kotsadm-rqlite-db
         name: kots/admin_console
-        containerName: rqlite
         selector:
           - app=kotsadm-rqlite
         command:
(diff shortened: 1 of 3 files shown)


pkg/supportbundle/defaultspec/spec.yaml
```diff
@@ -23,7 +23,6 @@ spec:
     - exec:
         collectorName: kotsadm-rqlite-db
         name: kots/admin_console
-        containerName: rqlite
```
garcialuis, line 26: The rqlite pod will always just have one container, right? That's why it's safe to assume that `rqlite` will be the container in which the command will be ran?

garcialuis (approved):

PR #3874, opened 2023-05-10
update sourcemap type to reduce bundle size

miaawong:
#### What this PR does / why we need it: Source maps were previously separated from main bundle, however, the option we selected to generate the source map was the slowest. This PR reduces the bundle size and load time. 
https://webpack.js.org/configuration/devtool/
Firefox
Screenshot 2023-05-09 at 8 00 40 PM
Chrome
Screenshot 2023-05-09 at 7 59 12 PM

Fixes # https://app.shortcut.com/replicated/story/75720/remove-source-maps-from-main-bundle-in-dev-build-of-admin-console

## Steps to reproduce

1. Run admin console and inspect the network tab for main.[number].js 

```release-note
NONE
```

NONE


rianfowler (approved):
this is great- it's like half the size!

rianfowler (approved):
nope it's more like less than 10% the size

PR #3888, opened 2023-05-17
fix: exclude and when for v1beta2 charts

cbodonnell:

This PR fixes an issue where v1beta2 HelmChart `spec.optionalValues[i].when` field did not have template functions rendered before it was parsed as a boolean.  This could lead to errors like the following:
```
{"level":"error","ts":"2023-05-17T19:50:25Z","msg":"rewrite directory: failed to write helm v1beta2 charts: failed to parse when conditional on optional values: strconv.ParseBool: parsing \"repl{{ HasLocalRegistry }}\": invalid syntax"}
```

This PR changes the logic so that the `when` field is rendered prior to parsing.

This PR also fixes an issue where the `exclude` field was not taken into account for v1beta2 charts.

Fixes #

v1beta1 HelmChart resources are templated before this merging of values occurs as part of [this function](https://github.com/replicatedhq/kots/blob/main/pkg/base/replicated.go#L401-L404), which is why these errors don't occur in that flow.  In the new helm install flow, we only need to template the `when` condition to avoid a failure to parse.  We'll then render the entire `values.yaml` file prior to saving in the `helm` directory, so that all template functions are rendered prior to the `helm upgrade`.

```release-note
NONE
```

NONE

### DIFF
--- a/pkg/apparchive/helm-v1beta2.go
+++ b/pkg/apparchive/helm-v1beta2.go
@@ -79,6 +79,18 @@ func WriteV1Beta2HelmCharts(opts WriteV1Beta2HelmChartsOptions) error {
 
 	for _, v1Beta2Chart := range opts.KotsKinds.V1Beta2HelmCharts.Items {
 		helmChart := v1Beta2Chart
+
+		if !helmChart.Spec.Exclude.IsEmpty() {
+			exclude, err := helmChart.Spec.Exclude.Boolean()
+			if err != nil {
+				return errors.Wrap(err, "failed to parse exclude boolean")
+			}
+
+			if exclude {
+				continue
+			}
+		}
+
 		chartDir := path.Join(opts.HelmDir, helmChart.GetDirName())
 		if err := os.MkdirAll(chartDir, 0744); err != nil {
 			return errors.Wrap(err, "failed to create chart dir")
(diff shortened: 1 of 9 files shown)


pkg/apparchive/helm-v1beta2.go
cbodonnell, line 1: v1beta1 HelmChart resources are templated before this merging of values occurs.  Here, we only need to template the `when` condition to avoid a failure to parse.  We then render the entire values.yaml file prior to saving in the `helm` directory

pkg/pull/pull.go
```diff
@@ -562,12 +562,18 @@ func Pull(upstreamURI string, pullOptions PullOptions) (string, error) {
 		return "", errors.Wrapf(err, "failed to remove unused helm midstreams")
 	}
 
+	renderedKotsKinds, err := kotsutil.KotsKindsFromMap(renderedKotsKindsMap)
```
cbodonnell, line 565: This function was previously only used for tests, but does allow us to leverage the already rendered kots kinds here.

pkg/rewrite/rewrite.go
```diff
@@ -320,6 +320,11 @@ func Rewrite(rewriteOptions RewriteOptions) error {
 		return errors.Wrap(err, "failed to write downstreams")
 	}
 
+	kotsKinds, err := kotsutil.LoadKotsKindsFromPath(rewriteOptions.UpstreamPath)
```
cbodonnell, line 323: this was just moved down since we wouldn't need it above anymore

sgalsaleh (approved):

PR #3896, opened 2023-05-18
don't uninstall and reinstall helm charts when the version changes

sgalsaleh:

Fixes an issue where Helm charts deployed using the native Helm installation method were uninstalled then reinstalled when the chart version changes / is updated.

Fixes #

```release-note
Fixes an issue where Helm charts deployed using the native Helm installation method were uninstalled then reinstalled when the chart version changes / is updated.
```

NONE

### DIFF
--- a/pkg/operator/client/deploy_test.go
+++ b/pkg/operator/client/deploy_test.go
@@ -2,11 +2,14 @@ package client
 
 import (
 	"encoding/base64"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"reflect"
 	"testing"
 
+	"github.com/mholt/archiver/v3"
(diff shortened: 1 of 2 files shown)


pkg/operator/client/deploy_test.go
```diff
@@ -1033,3 +1038,1545 @@ version: ver2
...
+							Name:         "chart1",
+							ChartVersion: "0.0.1",
+						},
+						ReleaseName: "chart1-release",
+						Namespace:   "ns1",
+					},
+				},
+			},
+		},
+		// ---- MIX OF v1beta1 and v1beta2 ---- //
```
cbodonnell, line 2285: What do you think about adding a test case where a chart was previously v1beta1 and is now v1beta2 and we validate that it is not removed?

cbodonnell (approved):

PR #3903, opened 2023-05-23
feat: deployment phases and wait-for annotations

cbodonnell:

This PR adds support for new annotations that can be included on native Kubernetes resources deployed by KOTS:
* `kots.io/creation-phase` and `kots.io/deletion-phase` - allow for resources to be grouped into phases when being created and deleted. KOTS will start with the lowest phase number and continue until all phases are complete.  The default phase is `"0"` if one is not specified.  Positive and negative values are allowed in this field.
* `kots.io/wait-for-ready` - makes KOTS wait for a deployed resource to be "ready" before continuing.  This leverages existing status informer logic to determine the meaning of "ready" for a given resources type.  If we don't have a readiness check for a given resource, KOTS will wait until the resource is queryable from the Kubernetes api server.
* `kots.io/wait-for-properties` - makes KOTS wait for one or more properties to match a desired value before continuing.

The goal of this feature is to give users the ability to orchestrate the deployment of native Kubernetes resources.

**Example usage:**

Deployment
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: my-deployment
  annotations:
    kots.io/creation-phase: "2"
    kots.io/wait-for-ready: "true"
```

Custom Resource Definition
```
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: myresources.example.com
  annotations:
    kots.io/creation-phase: "-3"
    kots.io/deletion-phase: "-2"
    kots.io/wait-for-ready: "true"
...
```

Custom Resource
```
kind: MyResource
metadata:
  name: my-resource
  annotations:
    kots.io/creation-phase: "-2"
    kots.io/deletion-phase: "-3"
    kots.io/wait-for-properties: '.status.tasks.extract=true,.status.tasks.transform=true,.status.tasks.load=true'
...
```

Fixes [SC-77370](https://app.shortcut.com/replicated/story/77370/implement-kots-wait-for-ready-feature-for-native-kubernetes-manifests)

```release-note
Adds support for `kots.io/creation-phase` and `kots.io/deletion-phase` annotations for controlling the order in which native Kubernetes resources are created and deleted, respectively.
Adds support for a `kots.io/wait-for-ready` annotation, which will cause the app manager to wait for a native Kubernetes resource to be ready before continuing with the deployment.
Adds support for a `kots.io/wait-for-properties` annotation, which will cause the app manager to wait for one or more properties to match a desired value before continuing with the deployment.
```

Yes, docs PR to come...

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -2948,6 +2948,152 @@ jobs:
           aws-secret-access-key: '${{ secrets.E2E_SUPPORT_BUNDLE_AWS_SECRET_ACCESS_KEY }}'
 
 
+  validate-deployment-orchestration:
+    runs-on: ubuntu-20.04
+    needs: [ enable-tests, can-run-ci, build-push-kotsadm-image, build-kurl-proxy, build-migrations, push-minio, push-mc, push-rqlite ]
+    strategy:
+      fail-fast: false
+      matrix:
+        k8s_version: [ v1.27.1-k3s1 ]
+    env:
+      APP_SLUG: deployment-orchestration
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - uses: azure/setup-helm@v3
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - uses: replicatedhq/action-k3s@main
+        with:
+          version: ${{ matrix.k8s_version }}
+          arguments: ${{ startsWith(matrix.k8s_version, 'v1.22') && '--egress-selector-mode disabled' || '' }}
+
+      - name: download kots binary
+        uses: actions/download-artifact@v3
+        with:
+          name: kots
+          path: bin/
+
+      - run: chmod +x bin/kots
+
+      - name: run the test
+        run: |
+          set +e
+          echo ${{ secrets.DEPLOYMENT_ORCHESTRATION_LICENSE }} | base64 -d > license.yaml
+          ./bin/kots \
+            install "$APP_SLUG/automated" \
+            --license-file license.yaml \
+            --no-port-forward \
+            --namespace "$APP_SLUG" \
+            --shared-password password \
+            --kotsadm-registry ttl.sh \
+            --kotsadm-namespace automated-${{ github.run_id }} \
+            --kotsadm-tag 24h
+
+          EXIT_CODE=$?
+          if [ $EXIT_CODE -ne 0 ]; then
+            echo "------pods:"
+            kubectl -n "$APP_SLUG" get pods
+            echo "------kotsadm logs"
+            kubectl logs -l app=kotsadm --tail=100 --namespace "$APP_SLUG"
+            exit $EXIT_CODE
+          fi
+
+          function wait_for_log {
+            local counter=1
+            local timeout=30
+            local log_pattern="$1"
+
+            while ! kubectl logs deploy/kotsadm -c kotsadm -n "$APP_SLUG" | grep -i "$log_pattern"; do
+              ((counter += 1))
+              if [ $counter -gt $timeout ]; then
+                echo "Timed out waiting for log $log_pattern"
+                exit 1
+              fi
+              sleep 1
+            done
+          }
+
+          wait_for_log "applying phase -9999"
+
+          wait_for_log "applying phase -3"
+          wait_for_log "waiting for resource apiextensions.k8s.io/v1/CustomResourceDefinition/myresources.example.com in namespace $APP_SLUG to be ready"
+
+          wait_for_log "applying phase -2"
+
+          # validate that phase -1 has not deployed yet since we're waiting on the CR status fields
+          if kubectl logs deploy/kotsadm -c kotsadm -n "$APP_SLUG" | grep -i "applying phase -1"; then
+            printf "phase -1 was deployed before phase -2 completed"
+            kubectl logs deploy/kotsadm -c kotsadm -n "$APP_SLUG"
+            exit 1
+          fi
+
+          wait_for_log "waiting for resource example.com/v1beta1/MyResource/my-resource in namespace $APP_SLUG to have property .status.tasks.extract=true"
+
+          # set .status.tasks.extract=true
+          kubectl patch myresources.example.com my-resource -n "$APP_SLUG" -p '{"status": {"tasks": {"extract": true}}}' --type=merge
+
+          wait_for_log "waiting for resource example.com/v1beta1/MyResource/my-resource in namespace $APP_SLUG to have property .status.tasks.transform=true"
+
+          # set .status.tasks.transform=true
+          kubectl patch myresources.example.com my-resource -n "$APP_SLUG" -p '{"status": {"tasks": {"transform": true}}}' --type=merge
+
+          wait_for_log "waiting for resource example.com/v1beta1/MyResource/my-resource in namespace $APP_SLUG to have property .status.tasks.load=true"
+
+          # set .status.tasks.load=true
+          kubectl patch myresources.example.com my-resource -n "$APP_SLUG" -p '{"status": {"tasks": {"load": true}}}' --type=merge
+
+          # now validate that the remaining phases are deployed (-1, 0, and 1)
+          wait_for_log "applying phase -1"
+          wait_for_log "waiting for resource apps/v1/Deployment/nginx-1 in namespace $APP_SLUG to be ready"
+
+          wait_for_log "applying phase 0"
+          wait_for_log "waiting for resource apps/v1/Deployment/nginx-2 in namespace $APP_SLUG to be ready"
+
+          wait_for_log "applying phase 1"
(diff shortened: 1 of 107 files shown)


cbodonnell:
Image pull secrets would be part of the default deployment phase.  Do we need to make them part of phase `-9999` to ensure they are deployed first, similar what we do when adding them as helm hooks?

pkg/operator/types/types.go
```diff
@@ -103,6 +128,61 @@ func (r Resource) GetNamespace() string {
...
+		if !ok {
+			return nil
+		}
+
+		waitForProperties := []WaitForProperty{}
+		for _, property := range strings.Split(annotationValue, ",") {
+			parts := strings.SplitN(property, "=", 2)
+			if len(parts) != 2 {
+				continue
+			}
```
cbodonnell, line 175: Should we log here so we know if something was skipped because it was formatted incorrectly?

pkg/appstate/wait.go
```diff
@@ -0,0 +1,228 @@
...
+
+		if !kuberneteserrors.IsNotFound(err) {
+			if matches, err := resourcePropertyMatchesValue(r, path, desiredValue); err != nil {
+				return errors.Wrap(err, "failed to check if resource property matches value")
+			} else if matches {
+				return nil
+			}
+		}
+
+		time.Sleep(WaitForResourceInterval)
```
miguelvr, line 202: is there a timeout? can it be configured?

cbodonnell: no timeout as it stands now, but is something we have been discussing

pkg/appstate/wait.go
```diff
@@ -0,0 +1,228 @@
...
+	buf := new(bytes.Buffer)
+	err := parser.Execute(buf, r.Object)
+	if err != nil {
+		// don't return an error here since the field may not exist yet
+		logger.Warnf("failed to execute jsonpath: %v", err)
+	}
+
+	if buf.String() == desiredValue {
+		return true, nil
+	}
```
miguelvr, line 225: just equality might be limiting. I can see other use cases like `>= 0` 

have you considered using [CEL](https://github.com/google/cel-spec)? It's being adopted by the k8s project for native resources, so it should be a safe bet. It would also make sense in other places throughout the codebase, such as analyzers (for preflight checks)

.github/workflows/build-test.yaml
```diff
@@ -2948,6 +2948,152 @@ jobs:
...
+                echo "Timed out waiting for log $log_pattern"
+                exit 1
+              fi
+              sleep 1
+            done
+          }
+
+          wait_for_log "applying phase -9999"
+
+          wait_for_log "applying phase -3"
```
sgalsaleh, line 3021: i'm being picky, and i don't have a solution off the top of my head, but if the API applies -3 _then_ -9999, this test wouldn't catch that, because both logs would be present when checking for the -9999 log. same for https://github.com/replicatedhq/kots/pull/3903/files#diff-bd4e577d8cce18da4cecc900fca127f9a0d57d706c9cdb3507b778c1d574a54fR3075-R3079

pkg/operator/client/resource.go
```diff
@@ -111,60 +114,109 @@ func decodeManifests(manifests []string) types.Resources {
...
 		}
+		phases = append(phases, phase)
 	}
 
-	for _, kind := range creationOrder {
-		sortedResources = append(sortedResources, resourcesByKind[kind]...)
+	return phases
+}
+
+func getSortedPhases(resourcesByPhase map[string]types.Resources) []string {
```
sgalsaleh, line 160: nit picking, how about:
```golang
phases := []int64{}
for phase := range resourcesByPhase {
  parsed, _ := strconv.ParseInt(phase, 10, 64)
  phases = append(phases, parsed)
}
sort.Slice(sortedPhases, func(i, j int) bool {
  return sortedPhases[i] < sortedPhases[j]
})
```
i feel that is easier to read and has a better performance as a plus.

pkg/appstate/wait.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+func WaitForDaemonSetToBeReady(clientset kubernetes.Interface, namespace, name string) error {
+	for {
+		r, err := clientset.AppsV1().DaemonSets(namespace).Get(context.TODO(), name, metav1.GetOptions{})
+		if err != nil && !kuberneteserrors.IsNotFound(err) {
+			return errors.Wrap(err, "failed to get existing daemonset")
+		}
+
+		if !kuberneteserrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
cbodonnell: i'd agree since we already are checking `!kuberneteserrors.IsNotFound(err)` above.  will make this change

.github/workflows/build-test.yaml
sgalsaleh: now that i finished reviewing the PR, I see that you already added a unit test for validating the ordering of phases, so it's not necessary to cover it here.

pkg/operator/client/deploy.go
```diff
@@ -144,75 +192,56 @@ func (c *Client) ensureResourcesPresent(deployArgs operatortypes.DeployAppArgs)
...
-			logger.Infof("applied resource %s/%s/%s/%s in namespace %s", group, version, kind, name, namespace)
-		} else {
-			logger.Info("applied unidentified resource.")
+			if resource.ShouldWaitForProperties() {
+				for _, prop := range resource.GetWaitForProperties() {
+					logger.Infof("waiting for resource %s/%s/%s/%s in namespace %s to have property %s=%s", group, version, kind, name, namespace, prop.Path, prop.Value)
+					err := appstate.WaitForProperty(namespace, name, resource.GVK, prop.Path, prop.Value)
+					if err != nil {
+						logger.Infof("failed to wait for resource %s/%s/%s/%s in namespace %s to have property %s=%s", group, version, kind, name, namespace, prop.Path, prop.Value)
+						return nil, errors.Wrapf(err, "failed to wait for resource to have property %s=%s", prop.Path, prop.Value)
```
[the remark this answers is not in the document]
cbodonnell: if the function returns an error, it looks like we just [set the apply result to that error](https://github.com/replicatedhq/kots/blob/main/pkg/operator/client/client.go#L149).

ex:
Screen Shot 2023-05-25 at 3 06 48 PM

pkg/operator/client/resource.go
cbodonnell: keeping this as-is so as not to change the return type

pkg/operator/types/types.go
```diff
@@ -103,6 +128,62 @@ func (r Resource) GetNamespace() string {
...
+}
+
+func (r Resource) ShouldWaitForProperties() bool {
+	if r.Unstructured != nil {
+		annotations := r.Unstructured.GetAnnotations()
+		if annotations == nil {
+			return false
+		}
+		_, ok := annotations[WaitForPropertiesAnnotation]
+		return ok
```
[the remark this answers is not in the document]
cbodonnell: make sense.  this would return true in the case of an empty annotation value.  it would then try to parse out the key-value pairs, would find none, and would not enter the loop to wait for them.  probably a small performance gain if we return false for an empty label, so i'll make this edit

pkg/appstate/wait.go
cbodonnell: thanks for sharing this.  likely something we'll look at for a future iteration on this feature

pkg/appstate/wait.go
cbodonnell: we're sticking with no timeout for now as we don't currently have any concept of timeouts during the deployment of native k8s manifests, so we will likely want to look at it more holistically outside of just this feature.  does sound like something that would not be difficult to implement here, so I could see that happening if there is enough demand for it

pkg/operator/client/deploy.go
```diff
@@ -144,75 +192,56 @@ func (c *Client) ensureResourcesPresent(deployArgs operatortypes.DeployAppArgs)
...
-
-			deployRes.applyResult.hasErr = true
-			return &deployRes, nil
-		}
+			if resource.ShouldWaitForReady() {
+				logger.Infof("waiting for resource %s/%s/%s/%s in namespace %s to be ready", group, version, kind, name, namespace)
+				err := appstate.WaitForResourceToBeReady(namespace, name, resource.GVK)
+				if err != nil {
+					logger.Infof("failed to wait for resource %s/%s/%s/%s in namespace %s to be ready", group, version, kind, name, namespace)
+					return nil, errors.Wrap(err, "failed to wait for resource to be ready")
```
sgalsaleh, line 229: why not use the same error message from the log above?

pkg/operator/client/deploy.go
sgalsaleh, line 240: ditto

pkg/operator/client/deploy.go
cbodonnell: makes sense since it will show in the UI for debugging

sgalsaleh (approved):

PR #3905, opened 2023-05-24
bug: setting https proxy to the right value

ricardomaraschini:

In order to handle two distinct proxies that may exist in the system separately, it is necessary to treat them individually. The presence of a second environment variable, `PROXY_HTTPS_ADDRESS`, indicates the configured https proxy.

This PR ensures the appropriate utilization of the correct env variable for the https proxy.

Depends on https://github.com/replicatedhq/kURL/pull/4538.

Fixes # [sc-70304]

```release-note
Kots starts to support different proxies for HTTP and HTTPS access.
```

### DIFF
--- a/deploy/kurl/kotsadm/template/base/deployment/tmpl-kotsadm-proxy.yaml
+++ b/deploy/kurl/kotsadm/template/base/deployment/tmpl-kotsadm-proxy.yaml
@@ -11,6 +11,6 @@ spec:
         - name: HTTP_PROXY
           value: "${PROXY_ADDRESS}"
         - name: HTTPS_PROXY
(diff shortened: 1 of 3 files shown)


camilamacedo86 (approved):
It shows OK for me 👍

emosbaugh (changes requested):
in order to make this backwards compatible could you set PROXY_HTTPS_ADDRESS = PROXY_HTTP_ADDRESS when unset?

ricardomaraschini:
> in order to make this backwards compatible could you set PROXY_HTTPS_ADDRESS = PROXY_HTTP_ADDRESS when unset?

That makes total sense. Done.

emosbaugh (approved):

PR #3922, opened 2023-06-02
fix creation of config values for multi-doc releases

garcialuis:
Fixes two issues for releases that combined their manifests into a multi-doc yaml file:
- KOTS was failing to create config values.
- Config spec section of the multi-doc yaml was missing from the rendered kotsKinds file.

This was the error logged by kotsadm:
```
  • Pulling upstream
  • failed to create config values: failed to decode values: Object 'Kind' is missing in ''
  • Creating base
  • Creating midstreams
  • Creating downstream "this-cluster"
```
As a result the `upstream/userdata/config.yaml` was missing when viewing or downloading the files.

Under the `kotsKinds` directory the config spec would be missing from the multi-doc release file:
```
---
apiVersion: troubleshoot.sh/v1beta2
kind: SupportBundle
metadata:
  name: support-bundle
spec:
...
---
        <-- config spec should be here
---
apiVersion: kots.io/v1beta1
kind: LintConfig
metadata:
  name: default-lint-config
spec:
...
```

Fixes # [SC-77605](https://app.shortcut.com/replicated/story/77605/combining-all-kots-objects-into-multi-doc-yaml-fails)

The `findTemplateContextDataInRelease` function was only being used to retrieve the config, so it was renamed to `findConfigInRelease` and removed the code that was no longer being used.

Both `findConfigInRelease` and `getKotsKinds ` functions were not splitting the manifests to single doc file, a call to `util.ConvertToSingleDocs` was added to both.

```release-note
Fixes an issue where the ConfigValues were not being saved for multi-doc yaml releases.
Fixes an issue where the Config spec would be missing from the rendered multi-doc release yaml within the kotsKinds folder.
```

NONE

### DIFF
--- a/pkg/base/replicated.go
+++ b/pkg/base/replicated.go
@@ -55,7 +55,7 @@ func renderReplicated(u *upstreamtypes.Upstream, renderOptions *RenderOptions) (
 		return nil, nil, nil, errors.Wrap(err, "failed to create new config context template builder")
 	}
 
-	kotsKinds, err := getKotsKinds(u, renderOptions.Log)
+	kotsKinds, err := getKotsKinds(u)
 	if err != nil {
 		return nil, nil, nil, errors.Wrap(err, "failed to find config file")
 	}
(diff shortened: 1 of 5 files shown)


pavansokkenagaraj (approved):
LGTM!

sgalsaleh (approved):

PR #3950, opened 2023-06-28
build kots with apko+melange in presubmit

imjasonh:

This PR adds a presubmit GitHub Actions workflow to build and publish a kots image from the PR's source, including relevant other packages like `helm`, `kustomize`, and `kubectl`. For now, only the latest `kubectl` is included in the image, but we should be able to add older supported `kubectl`s relatively soon.

To do this, the PR adds an `apko.yaml` and `melange.yaml` to be used with (you guessed it) [apko](https://github.com/chainguard-dev/apko) and [melange](https://github.com/chainguard-dev/melange) to build kots into an APK and to install that APK into an image.

None that I'm aware of.

None

## Steps to reproduce

See apko_melange_build.md

None at this time, though this could be extended to produce user-visible images by this process either on pushes to `main` or at release-time.

```release-note
NONE
```

Yes, added in apko_melange_build.md

cc @denhamparry

### DIFF
--- a/apko_melange_build.md
+++ b/apko_melange_build.md
@@ -1,0 +1,50 @@
+# Building KOTS with apko + melange
+
+## What?
+
+- [`melange`](https://github.com/chainguard-dev/melange) is a tool for reproducibly building APK packages from source
+- [`apko`](https://github.com/chainguard-dev/apko) is a tool for reproducibly building container images from APK packages
+
+## Why?
+
+Building with `apko` and `melange` produces smaller, more reproducible images, which can be easier to operate and easier to keep free of vulnerabilities.
+
+## How?
+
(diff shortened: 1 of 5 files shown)


apko_melange_build.md
```diff
@@ -0,0 +1,50 @@
+# Building KOTS with apko + melange
+
```
denhamparry, line 2: Please add a section to say that this is currently used for testing the kots image and is not to be considered production ready. This ensures that someone doesn't attempt to use the image built in the pipeline within their production environments.

denhamparry (commented):
Comment back from our call today, all looks great but we need to confirm that the image being built shouldn't be used in production instances 🙌

PR #3959, opened 2023-07-05
build kots with apko+melange in presubmit

cbodonnell:

This PR adds a presubmit GitHub Actions workflow to build and publish a kots image from the PR's source, including relevant other packages like `helm`, `kustomize`, and `kubectl`. For now, only the latest `kubectl` is included in the image, but we should be able to add older supported `kubectl`s relatively soon.

To do this, the PR adds an `apko.yaml` and `melange.yaml` to be used with (you guessed it) [apko](https://github.com/chainguard-dev/apko) and [melange](https://github.com/chainguard-dev/melange) to build kots into an APK and to install that APK into an image.

None that I'm aware of.

None

## Steps to reproduce

See apko_melange_build.md

None at this time, though this could be extended to produce user-visible images by this process either on pushes to `main` or at release-time.

```release-note
NONE
```

Yes, added in apko_melange_build.md

cc @denhamparry

### DIFF
--- a/.github/workflows/presubmit-image.yaml
+++ b/.github/workflows/presubmit-image.yaml
@@ -1,0 +1,24 @@
+on:
+  pull_request:
+
+jobs:
+  presubmit-image:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version: '1.20.x'
+      - uses: chainguard-dev/actions/melange-build@main
+        with:
+          config: melange.yaml
+          archs: x86_64
+          sign-with-temporary-key: true
+      - uses: chainguard-images/actions/apko-publish@main
+        with:
+          config: apko.yaml
+          archs: x86_64
+          tag: ttl.sh/kots
+          vcs-url: true
--- a/melange.yaml
+++ b/melange.yaml
@@ -1,0 +1,72 @@
+package:
+  name: kots-head
+  version: 0.0.1
+  epoch: 0
+  description: Kubernetes Off-The-Shelf (KOTS) Software
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  contents:
+    repositories:
+      - https://packages.wolfi.dev/os
+    keyring:
+      - https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
+    packages:
+      - ca-certificates-bundle
+      - busybox
+      - git
+      - go
+      - nodejs
+      - yarn
+
+pipeline:
+  - runs: |
+      set -x
+      export DESTDIR="${{targets.destdir}}"
+      mkdir -p "${DESTDIR}"
+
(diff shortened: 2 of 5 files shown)


cbodonnell:
@denhamparry @imjasonh 

The new workflow itself looks good.  We need to integrate this into our existing `build-test` workflow and ensure that the built image is used by the tests.  It's possible it will require a couple of small tweaks, but we should be able to take this one from here.  Thanks for putting this together!

imjasonh:
> It's possible it will require a couple of small tweaks, but we should be able to take this one from here.

Glad to hear it! Don't hesitate to reach out if there's any trouble. 🚀

melange.yaml
```diff
@@ -0,0 +1,72 @@
...
+      export VERSION=${{package.version}}
+      export GIT_TAG=${{package.version}}
+
+      # Set environment variables from repository
+      source .image.env
+
+      KOTS_KUSTOMIZE_BIN_DIR=/usr/local/bin
+
+      # TODO: fix pact build error on arm https://github.com/pact-foundation/pact-js-core/issues/264
+      export PACT_SKIP_BINARY_INSTALL=true
```
e3b0c442, line 53: This should no longer be necessary as we should be on the latest version of pact-js which bundles the multi-arch Rust core.

I'll submit a PR for kotsadm today to address that.

.github/workflows/presubmit-image.yaml
```diff
@@ -0,0 +1,24 @@
...
+      - uses: chainguard-dev/actions/melange-build@main
+        with:
+          config: melange.yaml
+          archs: x86_64
+          sign-with-temporary-key: true
+      - uses: chainguard-images/actions/apko-publish@main
+        with:
+          config: apko.yaml
+          archs: x86_64
+          tag: ttl.sh/kots
```
cbodonnell, line 23: It would probably make integration easier if we match the chainguard image name to our current image name `kotsadm`

cbodonnell: Additionally, we will usually push the image with a namespace to avoid conflicts on ttl.sh. For example `ttl.sh/<some-identifier/kotsadm:24h`

.github/workflows/presubmit-image.yaml
```diff
@@ -0,0 +1,24 @@
...
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version: '1.20.x'
+      - uses: chainguard-dev/actions/melange-build@main
+        with:
+          config: melange.yaml
+          archs: x86_64
+          sign-with-temporary-key: true
+      - uses: chainguard-images/actions/apko-publish@main
```
cbodonnell, line 19: I see that this action pushes the image with a digest instead of a tag. Our tests currently leverages a `--kotsadm-tag` flag to override the image used during the install. For example: https://github.com/replicatedhq/kots/blob/v1.100.3/.github/workflows/build-test.yaml#L988-L990

PR #3960, opened 2023-07-06
remove validate-helm-managed test

cbodonnell:
#### What this PR does / why we need it:

This PR removes the `validate-helm-managed` test from the PR workflow.  This was a beta feature leveraging the admin console helm chart that we no longer support.

#### Which issue(s) this PR fixes:

Fixes https://github.com/replicatedhq/kots/actions/runs/5468214876/jobs/9956531472#step:8:208

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #3961, opened 2023-07-06
setup-kustomize in ci-test-kots

cbodonnell:

This PR fixes an issue where unit tests relying on kustomize as a runtime dependency were failing.  It seems that the GH runners no longer shipped with it be default, so this PR adds a setup step to install the appropriate binary.

The PR also includes the changes from https://github.com/replicatedhq/kots/pull/3960 since both are necessary to get all tests passing for the `validate-success` requirement.

Fixes https://github.com/replicatedhq/kots/actions/runs/5468790382/jobs/9972117503#step:4:129

```release-note
NONE
```


e3b0c442 (dismissed):
Bonus points for checksum validation

garcialuis (approved):

PR #3964, opened 2023-07-06
Read Preflights from Helm Charts

diamonwiggins:

Adds support to load troubleshoot specs from the rendered app archive generated by KOTS, so that running Preflights included in a Helm Chart is supported. By default, preflights included in any helm charts in the release take precedence over the KOTS Kinds Preflights. 

Fixes #

```release-note
KOTS now uses preflights first from any Helm chart archives in the app, and if it cannot find any, it will default to the prior behavior of searching for a `kind: Preflight` file in the root of the release
```


pkg/preflight/preflight.go
```diff
@@ -262,20 +285,35 @@ func GetPreflightCommand(appSlug string) []string {
...
+
+	builtPreflight := &troubleshootv1beta2.Preflight{}
+	if tsKinds.PreflightsV1Beta2 != nil {
+		for _, v := range tsKinds.PreflightsV1Beta2 {
+			builtPreflight = troubleshootpreflight.ConcatPreflightSpec(builtPreflight, &v)
+		}
+	} else {
+		builtPreflight = kotsKinds.Preflight.DeepCopy()
+		if builtPreflight == nil {
+			builtPreflight = &troubleshootv1beta2.Preflight{
```
sgalsaleh, line 302: initializing with the default spec should probably be moved outside of this if/else block.

pkg/kotsutil/troubleshoot.go
```diff
@@ -0,0 +1,47 @@
...
+
+			return nil
+		})
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to walk rendered dir")
+	}
+
+	ctx := context.Background()
+	tsKinds, err := troubleshootloader.LoadSpecs(ctx, troubleshootloader.LoadOptions{
+		RawSpec: strings.Join(renderedFiles, "---"),
```
sgalsaleh, line 41: should this be like this instead?
```suggestion
		RawSpec: strings.Join(renderedFiles, "\n---\n"),
```

pkg/preflight/preflight.go
```diff
@@ -276,6 +299,20 @@ func CreateRenderedSpec(app *apptypes.App, sequence int64, origin string, inClus
...
 	}
 
+	tsKinds, err := kotsutil.LoadTSKindsFromPath(filepath.Join(archiveDir, "rendered"))
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("failed to load troubleshoot kinds from path: %s", filepath.Join(archiveDir, "rendered")))
+	}
+
+	if tsKinds.PreflightsV1Beta2 != nil {
+		for _, v := range tsKinds.PreflightsV1Beta2 {
+			builtPreflight = troubleshootpreflight.ConcatPreflightSpec(builtPreflight, &v)
```
sgalsaleh, line 309: doesn't this mean that collectors and analyzers from both preflight specs (the CRD and the secret) will be concatenated and executed? That doesn't seem to be the behavior for running preflights in the UI (the code [here](https://github.com/replicatedhq/kots/pull/3964/files#diff-e3a33423651aa73db11fec570357c6b99345bd78263668309b315244691765c4R85))

diamonwiggins: ah, yeah. good catch. I think I had moved that out of the if/else to address one of your earlier comments, but I'll need to rearrange this again.

pkg/kotsutil/troubleshoot.go
```diff
@@ -0,0 +1,47 @@
...
+
+			return nil
+		})
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to walk rendered dir")
+	}
+
+	ctx := context.Background()
+	tsKinds, err := troubleshootloader.LoadSpecs(ctx, troubleshootloader.LoadOptions{
+		RawSpec: strings.Join(renderedFiles, "\n---\n"),
```
banjoh, line 41: Nit pick: You could use `RawSpecs` which accepts a list of strings

pkg/preflight/preflight.go
```diff
@@ -262,8 +285,10 @@ func GetPreflightCommand(appSlug string) []string {
 	return comamnd
 }
 
-func CreateRenderedSpec(app *apptypes.App, sequence int64, origin string, inCluster bool, kotsKinds *kotsutil.KotsKinds) error {
-	builtPreflight := kotsKinds.Preflight.DeepCopy()
+func CreateRenderedSpec(app *apptypes.App, sequence int64, origin string, inCluster bool, kotsKinds *kotsutil.KotsKinds, archiveDir string) error {
+	var builtPreflight *troubleshootv1beta2.Preflight
+
+	builtPreflight = kotsKinds.Preflight.DeepCopy()
```
sgalsaleh, line 291: This is how I'm envisioning it should be:
```golang
	tsKinds, err := kotsutil.LoadTSKindsFromPath(filepath.Join(archiveDir, "rendered"))
	if err != nil {
		return errors.Wrap(err, fmt.Sprintf("failed to load troubleshoot kinds from path: %s", filepath.Join(archiveDir, "rendered")))
	}

	if tsKinds.PreflightsV1Beta2 != nil {
		for _, v := range tsKinds.PreflightsV1Beta2 {
			builtPreflight = troubleshootpreflight.ConcatPreflightSpec(builtPreflight, &v)
		}
	} else {
		builtPreflight = kotsKinds.Preflight.DeepCopy()
	}

	if builtPreflight == nil {
		builtPreflight = &troubleshootv1beta2.Preflight{
			TypeMeta: v1.TypeMeta{
				Kind:       "Preflight",
				APIVersion: "troubleshoot.sh/v1beta2",
			},
			ObjectMeta: v1.ObjectMeta{
				Name: "default-preflight",
			},
		}
	}
```

sgalsaleh (approved):

PR #3971, opened 2023-07-13
run tests on c11y matrix

cbodonnell:

This PR changes the on-PR tests to leverage the compatibility matrix instead of our custom `action-k3s` GH action.

Mostly uses `k3s`, with `kind` where necessary to cover k8s 1.27.  Picked three of our tests use EKS 1.27: `smoke-test`, `minimal-rbac`, and `backup-and-restore`.

https://app.shortcut.com/replicated/story/82931/make-kots-ci-use-compatibility-matrix

```release-note
NONE
```

NONE

### DIFF
--- a/.github/actions/kots-e2e/action.yml
+++ b/.github/actions/kots-e2e/action.yml
@@ -153,7 +153,7 @@ runs:
     - name: Remove Cluster
       id: remove-cluster
       uses: replicatedhq/replicated-actions/remove-cluster@v1
-      if: ${{ inputs.k8s-cluster-skip-teardown != 'true' && outputs.cluster-id }}
+      if: ${{ inputs.k8s-cluster-skip-teardown != 'true' && steps.create-cluster.outputs.cluster-id != '' }}
       continue-on-error: true
       with:
         api-token: ${{ inputs.replicated-api-token }}
--- a/.github/actions/kots-e2e/action.yml
+++ b/.github/actions/kots-e2e/action.yml
@@ -90,7 +90,7 @@ runs:
 
     - name: Create a c11y matrix cluster 
       id: create-cluster 
-      uses: replicatedhq/replicated-actions/create-cluster@v0
+      uses: replicatedhq/replicated-actions/create-cluster@v1
       if: ${{ inputs.k8s-distribution == 'k3s' || inputs.k8s-distribution == 'kind' || inputs.k8s-distribution == 'eks' }}
       with: 
         api-token: ${{ inputs.replicated-api-token }}
@@ -152,7 +152,7 @@ runs:
 
     - name: Remove Cluster
       id: remove-cluster
-      uses: replicatedhq/replicated-actions/remove-cluster@v0
+      uses: replicatedhq/replicated-actions/remove-cluster@v1
       if: ${{ inputs.k8s-cluster-skip-teardown != 'true' && (inputs.k8s-distribution == 'k3s' || inputs.k8s-distribution == 'kind' || inputs.k8s-distribution == 'eks') }}
       continue-on-error: true
       with:
--- a/.github/actions/kots-e2e/action.yml
+++ b/.github/actions/kots-e2e/action.yml
@@ -7,15 +7,21 @@ inputs:
   kots-namespace:
     description: 'KOTS namespace'
     required: true
-  k8s-version:
-    description: 'K8s version'
-    required: true
-  platform:
-    description: 'Platform to run tests on (k3s, gke)'
+  k8s-distribution:
+    description: 'Kubernetes distribution'
     default: 'k3s'
     required: false
-  k8s-arguments:
-    description: 'arguments to pass to k8s'
+  k8s-version:
+    description: 'Kubernetes version'
+    default: 'v1.26'
+    required: false
+  k8s-cluster-ttl:
+    description: 'TTL for c11y matrix clusters'
+    default: '2h'
+    required: false
+  k8s-cluster-skip-teardown:
+    description: 'Skip cluster teardown'
+    default: 'false'
     required: false
   testim-access-token:
     description: 'Testim access token'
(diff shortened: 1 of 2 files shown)


.github/workflows/build-test.yaml
```diff
@@ -2014,13 +2220,15 @@ jobs:
...
-          k8s-arguments: ${{ startsWith(matrix.k8s_version, 'v1.22') && '--egress-selector-mode disabled' || '' }}
+          k8s-distribution: ${{ matrix.cluster.distribution }}
+          k8s-version: ${{ matrix.cluster.version }}
           testim-access-token: '${{ secrets.TESTIM_ACCESS_TOKEN }}'
           testim-branch: ${{ github.head_ref == 'main' && 'master' || github.head_ref }}
           kotsadm-image-registry: ttl.sh
           aws-access-key-id: '${{ secrets.E2E_SUPPORT_BUNDLE_AWS_ACCESS_KEY_ID }}'
           aws-secret-access-key: '${{ secrets.E2E_SUPPORT_BUNDLE_AWS_SECRET_ACCESS_KEY }}'
+          replicated-api-token: '${{ secrets.C11Y_MATRIX_TOKEN }}'
+          k8s-cluster-skip-teardown: true
```
cbodonnell, line 2231: This one and the "target kots version" test was a little tricky.  We run the test using the `e2e` action, but also need the cluster to be around for the CLI step that comes after.  May need to tweak the `e2e` action to export the cluster ID so we can delete it?

.github/workflows/build-test.yaml
```diff
@@ -2066,12 +2276,15 @@ jobs:
...
-          k8s-version: '${{ matrix.k8s_version }}'
-          k8s-arguments: ${{ startsWith(matrix.k8s_version, 'v1.22') && '--egress-selector-mode disabled' || '' }}
+          k8s-distribution: ${{ matrix.cluster.distribution }}
+          k8s-version: ${{ matrix.cluster.version }}
           testim-access-token: '${{ secrets.TESTIM_ACCESS_TOKEN }}'
           testim-branch: ${{ github.head_ref == 'main' && 'master' || github.head_ref }}
           aws-access-key-id: '${{ secrets.E2E_SUPPORT_BUNDLE_AWS_ACCESS_KEY_ID }}'
           aws-secret-access-key: '${{ secrets.E2E_SUPPORT_BUNDLE_AWS_SECRET_ACCESS_KEY }}'
+          replicated-api-token: '${{ secrets.C11Y_MATRIX_TOKEN }}'
+          k8s-cluster-skip-teardown: true
```
cbodonnell, line 2286: Same as above

.github/workflows/build-test.yaml
```diff
@@ -2116,12 +2331,15 @@ jobs:
...
```
cbodonnell, line 2341: Same as above

.github/actions/kots-e2e/action.yml
```diff
@@ -7,15 +7,21 @@ inputs:
...
+    description: 'Kubernetes version'
+    default: 'v1.26'
+    required: false
+  k8s-cluster-ttl:
+    description: 'TTL for c11y matrix clusters'
+    default: '2h'
+    required: false
+  k8s-cluster-skip-teardown:
+    description: 'Skip cluster teardown'
+    default: 'false'
```
cbodonnell, line 24: Add an output of this action for the c11y `cluster-id`

.github/actions/kots-e2e/action.yml
```diff
@@ -58,23 +64,42 @@ inputs:
...
+
+    - name: Create a c11y matrix cluster 
+      id: create-cluster 
+      uses: replicatedhq/replicated-actions/create-cluster@v1
+      if: ${{ inputs.k8s-distribution == 'k3s' || inputs.k8s-distribution == 'kind' || inputs.k8s-distribution == 'eks' }}
+      with: 
+        api-token: ${{ inputs.replicated-api-token }}
+        kubernetes-distribution: ${{ inputs.k8s-distribution }}
+        kubernetes-version: ${{ inputs.k8s-version }}
+        cluster-name: ${{ github.ref_name }}-${{ inputs.k8s-distribution }}-${{ inputs.k8s-version }}
```
sgalsaleh, line 99: how about using the run id instead of ref_name? ref_name is the branch name IIRC, and it would be great to differentiate between runs.

.github/actions/kots-e2e/action.yml
```diff
@@ -125,6 +150,15 @@ runs:
         aws-secret-access-key: ${{ inputs.aws-secret-access-key }}
         aws-region: ${{ inputs.aws-region }}
 
+    - name: Remove Cluster
+      id: remove-cluster
+      uses: replicatedhq/replicated-actions/remove-cluster@v1
+      if: ${{ inputs.k8s-cluster-skip-teardown != 'true' && (inputs.k8s-distribution == 'k3s' || inputs.k8s-distribution == 'kind' || inputs.k8s-distribution == 'eks') }}
```
sgalsaleh, line 156: how about just checking if that outputted `cluster-id` is not empty from the `create-cluster` step instead of checking for individual distributions?

.github/workflows/build-test.yaml
```diff
@@ -918,18 +952,30 @@ jobs:
...
+      - name: Create Cluster 
+        id: create-cluster 
+        uses: replicatedhq/replicated-actions/create-cluster@v1 
         with:
-          version: ${{ matrix.k8s_version }}
-          arguments: ${{ startsWith(matrix.k8s_version, 'v1.22') && '--egress-selector-mode disabled' || '' }}
+          api-token: ${{ secrets.C11Y_MATRIX_TOKEN }}
+          kubernetes-distribution: ${{ matrix.cluster.distribution }}
+          kubernetes-version: ${{ matrix.cluster.version }}
+          cluster-name: ${{ github.ref_name }}-${{ matrix.cluster.distribution }}-${{ matrix.cluster.version }}
```
sgalsaleh, line 975: ditto on the cluster name

.github/workflows/build-test.yaml
```diff
@@ -2035,14 +2244,25 @@ jobs:
...
           fi
 
+      - name: Remove Cluster
+        id: remove-cluster
+        uses: replicatedhq/replicated-actions/remove-cluster@v1
+        continue-on-error: true
+        with:
+          api-token: ${{ secrets.C11Y_MATRIX_TOKEN }}
+          cluster-id: ${{ steps.kots-e2e.outputs.cluster-id }}
+
```
sgalsaleh, line 2254: why the extra new lines here? :)

.github/workflows/build-test.yaml
```diff
@@ -2063,15 +2283,19 @@ jobs:
...
-          k8s-version: '${{ matrix.k8s_version }}'
-          k8s-arguments: ${{ startsWith(matrix.k8s_version, 'v1.22') && '--egress-selector-mode disabled' || '' }}
+          k8s-distribution: ${{ matrix.cluster.distribution }}
+          k8s-version: ${{ matrix.cluster.version }}
           testim-access-token: '${{ secrets.TESTIM_ACCESS_TOKEN }}'
           testim-branch: ${{ github.head_ref == 'main' && 'master' || github.head_ref }}
           aws-access-key-id: '${{ secrets.E2E_SUPPORT_BUNDLE_AWS_ACCESS_KEY_ID }}'
           aws-secret-access-key: '${{ secrets.E2E_SUPPORT_BUNDLE_AWS_SECRET_ACCESS_KEY }}'
+          replicated-api-token: '${{ secrets.C11Y_MATRIX_TOKEN }}'
+          k8s-cluster-skip-teardown: true
```
sgalsaleh, line 2297: why?

.github/actions/kots-e2e/action.yml
cbodonnell: nice idea - that output was just added today!

.github/workflows/build-test.yaml
cbodonnell: the min/target/range kots version tests all have an additional cli step after that do some validation, so we need to keep the cluster around for that and then remove it using the `cluster-id` output.  we could re-work these tests so that this is all wrapped up in the testim suite, but that seemed out of scope

sgalsaleh (approved):

PR #3996, opened 2023-08-04
upgrade to kustomize 5.1.1

cbodonnell:

Upgrades Kustomize to 5.1.1.  As of 5.1.x, kustomize will error if a kustomization file is "empty".  This PR simply adds an annotation to the base kustomization file's metadata so that it is not considered "empty" by [this check](https://github.com/kubernetes-sigs/kustomize/blob/api/v0.14.0/api/types/kustomization.go#L304-L316).  This does not appear to change any behavior of the actual transformations since those use the `commonAnnotations` field.

Fixes [SC-84429](https://app.shortcut.com/replicated/story/84429/upgrade-kustomize-to-v5-1-1)

Explored an alternative solution where we don't even write base kustomization files that are empty, but this felt more error prone since we write these files in a few different places and would have to check each beforehand.  Additionally, we would have to make sure that any midstream and downstream kustomizations don't get written if a base was not written and also make sure that we don't try to run `kustomize build` when there may not be any files.

```release-note
NONE
```

NONE

### DIFF
--- a/pkg/pull/pull.go
+++ b/pkg/pull/pull.go
@@ -32,7 +32,6 @@ import (
 	kotsv1beta1 "github.com/replicatedhq/kotskinds/apis/kots/v1beta1"
 	k8sjson "k8s.io/apimachinery/pkg/runtime/serializer/json"
 	"k8s.io/client-go/kubernetes/scheme"
-	kustomizetypes "sigs.k8s.io/kustomize/api/types"
 )
 
 type PullOptions struct {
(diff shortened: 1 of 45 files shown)


pkg/pull/pull.go
```diff
@@ -721,27 +720,6 @@ func writeDownstreams(options PullOptions, overlaysDir string, m *midstream.Mids
 	return nil
 }
 
-func writeCombinedDownstreamBase(downstreamName string, bases []string, renderDir string) error {
```
cbodonnell, line 724: This and the one below it were not used

sgalsaleh (approved):

PR #4006, opened 2023-08-17
replace windows line endings before splitting docs

cbodonnell:

This PR fixes an issue where KOTS was not handling CRLF line endings (`\r\n`) when splitting multi-doc yaml files, which could result in missing kustomize patches.  This PR modifies the logic of the `ConvertToSingleDocs` utility function so that windows line endings are replaced with unix line endings before splitting documents.  It also renames the function and creates an equivalent utility function for splitting strings.

This PR also consolidates on the utility functions so that all cases of multi-doc splitting use these functions to support CRLF line endings across the board.

Fixes [SC-78355](https://app.shortcut.com/replicated/story/78355/kots-does-not-handle-crlf-line-terminators-when-splitting-multidoc-yaml-which-can-result-in-missing-kustomize-patches)

Based on an implementation of the same thing in kustomize: https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/kio/byteio_reader.go#L213-L215

```release-note
Fixes an issue where the admin console was not correctly processing multi-doc yaml files containing windows line endings.
```

NONE


pkg/util/util.go
```diff
@@ -128,8 +129,10 @@ func CompareStringArrays(arr1, arr2 []string) bool {
 	return true
 }
 
-func ConvertToSingleDocs(doc []byte) [][]byte {
+func YAMLBytesToSingleDocs(doc []byte) [][]byte {
 	singleDocs := [][]byte{}
+	// replace all windows line endings with unix line endings
+	doc = bytes.ReplaceAll(doc, []byte("\r\n"), []byte("\n"))
```
cbodonnell, line 135: this is the fix

pkg/util/util.go
```diff
@@ -140,6 +143,20 @@ func ConvertToSingleDocs(doc []byte) [][]byte {
 	return singleDocs
 }
 
+func YAMLStringToSingleDocs(doc string) []string {
```
sgalsaleh, line 146: can't we just convert the input to bytes and use the above function instead of having to duplicate this entire function?

sgalsaleh (approved):

pkg/util/util.go
cbodonnell: made some minor changes to calling functions to make this happen in https://github.com/replicatedhq/kots/pull/4006/commits/054efb746ec00582dbc882c836dbd0bde8905bee

sgalsaleh (approved):

sgalsaleh (approved):

PR #4011, opened 2023-08-21
reserve capacity for gke autopilot test

cbodonnell:

The `validate-gke-install` test has become increasingly flaky and fails more often than not.  This happens because the `kotsadm` pod is preempted by GKE when the test application pods are deployed and the autopilot cluster has not yet scaled up.  This PR changes the test script to deploy a handful of lower-priority pods that will cause the cluster to begin scaling up before KOTS is deployed.  These pods are also assigned a lower priority class, so they will be the target of preemption, should it occur.  This follows the pattern outlined in GKE documentation: https://cloud.google.com/kubernetes-engine/docs/how-to/capacity-provisioning

More context: https://replicated.slack.com/archives/C02TQTV9G2V/p1692312543883379

Fixes #

```release-note
NONE
```

NONE

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -1130,12 +1130,13 @@ jobs:
           kind: Job
           metadata:
             name: capacity-res-job
-            labels:
-              app: capacity-res-job
           spec:
             parallelism: 4
             backoffLimit: 0
             template:
+              metadata:
+                labels:
+                  app: capacity-res-job
               spec:
                 priorityClassName: low-priority
                 terminationGracePeriodSeconds: 0
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -1107,50 +1107,8 @@ jobs:
       - name: reserve capacity
         run: |
           # https://cloud.google.com/kubernetes-engine/docs/how-to/capacity-provisioning
-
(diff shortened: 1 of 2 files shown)


.github/workflows/build-test.yaml
```diff
@@ -1104,6 +1104,55 @@ jobs:
 
       - run: chmod +x bin/kots
 
+      - name: reserve capacity
+        run: |
+          # https://cloud.google.com/kubernetes-engine/docs/how-to/capacity-provisioning
+
+          echo 'apiVersion: scheduling.k8s.io/v1
```
sgalsaleh, line 1111: can we move this into a yaml file instead?

.github/workflows/build-test.yaml
```diff
@@ -1104,6 +1104,55 @@ jobs:
...
+          apiVersion: scheduling.k8s.io/v1
+          kind: PriorityClass
+          metadata:
+            name: default-priority
+          value: 0
+          preemptionPolicy: PreemptLowerPriority
+          globalDefault: true
+          description: "The global default priority."' | kubectl apply -f -
+
+          echo 'apiVersion: batch/v1
```
sgalsaleh, line 1129: ditto

sgalsaleh (approved):

PR #4021, opened 2023-08-29
remove collectd from default specs

danj-replicated:
Removes collectd from the default specs as per: https://app.shortcut.com/replicated/story/80665
This should speed up default bundle collection by removing a collector that has a 5 minute timeout.

```release-note
users who rely on collectd from the default kots troubleshoot spec may need to enable it manually
```

none


banjoh (approved):

CpuID (approved):
tests pass now, failures were unrelated to the changes here

PR #4026, opened 2023-09-05
smoke test latest cmx versions

cbodonnell:

This PR adds the remaining CMX distributions to our on-PR testing and also adds logic so that we will automatically inherit new CMX distros and versions that are added.

One larger change was to move all of the "compatibility tests" over to a separate re-usable workflow.  This will make it easier for us to create two separate pipelines for stable vs alpha CMX distros.  For example, we are excluding AKS for the moment because it is still alpha, but this could allow us to run those tests on AKS in a non-blocking way.

Fixes #

```release-note
NONE
```

NONE


.github/actions/cmx-versions/index.js
```diff
@@ -0,0 +1,81 @@
...
+        if (distroName === 'helmvm' || distroName === 'kurl') {
+            // excluding the embedded distributions
+            return;
+        }
+
+        if (distroName === 'aks') {
+            // excluding aks for now
+            // TODO: add aks back in when we support some sort of filtering
+            return;
+        }
```
cbodonnell, line 52: AKS has started to become unstable and we also have limited capacity.  Thinking we should instead run one of the other one-off tests on AKS for now. Thoughts?

.github/workflows/compatibility-tests.yaml
```diff
@@ -0,0 +1,802 @@
+name: compatibility-tests
```
cbodonnell, line 1: Moved all of the "compatibility tests" over to this re-usable workflow so that we can potentially have a dual pipeline that runs Stable vs Alpha distros (AKS for example). This way we can test in a non-blocking way by making the alpha run of this workflow a non-required job.

e2e/minio/install.go
```diff
@@ -77,7 +77,7 @@ func (m *Minio) Install(helmCLI *helm.CLI, kubeconfig string) {
 		fmt.Sprintf("--set=buckets[0].name=%s,buckets[0].policy=none,buckets[0].purge=false", m.GetBucket()),
 		m.options.ReleaseName,
 		"minio/minio",
-		"--version=v4.0.3",
+		"--version=v5.0.13",
```
cbodonnell, line 80: Old version doesn't install on the latest openshift

e2e/velero/cli.go
```diff
@@ -65,3 +69,16 @@ aws_secret_access_key=%s
 		accessKey, secretKey,
 	)), 0644)
 }
+
+func patchNodeAgentDaemonset(kubeconfig string) (*gexec.Session, error) {
```
cbodonnell, line 73: this comes from velero docs on openshift recommendations https://velero.io/docs/v1.9/restic/

.github/workflows/compatibility-tests.yaml
```diff
@@ -0,0 +1,802 @@
...
+        with:
+          name: kots
+          path: bin/
+
+      - run: chmod +x bin/kots
+
+      - name: create namespace and dockerhub secret
+        run: |
+          kubectl create ns "$APP_NAME"
+          kubectl create secret docker-registry kotsadm-dockerhub --docker-server index.docker.io --docker-username "${{ secrets.E2E_DOCKERHUB_USERNAME }}" --docker-password "${{ secrets.E2E_DOCKERHUB_PASSWORD }}" --namespace "$APP_NAME"
```
cbodonnell, line 400: Opted to create the dockerhub secret this way since KOTS needs to be deployed first before `kots docker ensure-secret` can run successfully.  Since we do many headless installs in our CLI tests, images would be checked before the secret can be created in those cases.

.github/actions/cmx-versions/index.js
```diff
@@ -0,0 +1,81 @@
...
+
+        if (response.status === 200) {
+            const payload = await response.json();
+            clusterVersions = payload['cluster-versions'];
+        } else {
+            throw new Error(`Request failed with status code ${response.status}`);
+        }
+    } catch (error) {
+        console.error(`Error: ${error.message}`);
+        core.setFailed(error.message);
```
sgalsaleh, line 30: is a `return` needed after this?

e2e/velero/cli.go
sgalsaleh: wouldn't this affect _all_ tests though? is there a way to only enable this for openshift?

.github/workflows/compatibility-tests.yaml
sgalsaleh: wouldn't it make more sense to have the CMX action mark a version as `alpha` or `stable` and then we could keep all these tests in `build-test.yaml` and then decide if a test should fail or not based on that?

.github/actions/cmx-versions/index.js
cbodonnell: good call. it might be ok, but won't hurt

e2e/velero/cli.go
cbodonnell: sounds good. can add some additional logic to e2e enable for OpenShift only

.github/workflows/compatibility-tests.yaml
cbodonnell: We could request that change, but IMO _we_ as the user of CMX need to be able to control what we accept to some level based on our definition of stable

cbodonnell: make sense - any ideas how we could make tests conditionally pass or fail if we went this route? i see `continue-on-error`, but that seems like it would still mark the job as failed?

.github/workflows/build-test.yaml
```diff
@@ -1153,98 +1189,14 @@ jobs:
           cluster-id: ${{ steps.create-cluster.outputs.cluster-id }}
 
 
-  validate-gke-install:
```
cbodonnell, line 1156: removed this as we're testing on CMX GKE - any concerns?

.github/workflows/compatibility-tests.yaml
cbodonnell: looking at the docs here, it seems it will allow the workflow to pass: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error

.github/workflows/build-test.yaml
sgalsaleh: nope

.github/actions/cmx-versions/index.js
```diff
@@ -0,0 +1,84 @@
...
+            } else {
+                const currentVersion = latestMinorVersions[distroName][majorMinor];
+                if (semverGt(parsed, semverCoerce(currentVersion))) {
+                    latestMinorVersions[distroName][majorMinor] = version;
+                }
+            }
+        });
+
+        Object.keys(latestMinorVersions[distroName]).forEach((minorVersion) => {
+            let stage = 'stable';
```
sgalsaleh, line 68: ideally, this should come from `https://api.replicated.com/vendor/v3/cluster/versions`

sgalsaleh (dismissed):

.github/actions/cmx-versions/index.js
cbodonnell: that would be nice. maybe we can put that in as a feature request? i think there's maybe some value to the client determining their own definition of "stable", but agree that for this use-case it would make it easier.  also trying to think ahead for if we make this a re-usable action.  users should probably be able to pass an `include` or `exclude` to determine the versions they want or don't want, respectively.  a conversation for later though...

sgalsaleh (approved):

PR #4027, opened 2023-09-05
Use empty dir for tmp dir in embedded clusters as well

sgalsaleh:

Updates the kotsadm deployment/statefulset in kurl to use an EmptyDir volume for the `tmp` directory. This is consistent with the behavior on existing cluster installs. The main reason this is done is to avoid permissions issues on the host that might be caused by the end customer's environment.

Fixes #

```release-note
- Fixes an issue where uploading the application air gap bundle could fail due to a permissions issue when creating files under the `/tmp` directory inside the `kotsadm` pod/container. This is only applicable to embedded cluster installs.
```

NONE


adamancini (approved):
lgtm

divolgin (approved):

cbodonnell (approved):

PR #4036, opened 2023-09-13
fix: race condition when updating registry settings

cbodonnell:
(cherry picked from commit 004d72d53117f79c5796a7fdb3f252a925906b0f)

This PR fixes a race condition where the backend would return a response from the PUT `/api/v1/app/{appSlug}/registry` handler before the image re-writing task started.  This could result in a scenario where the UI polls the `/api/v1/app/{appSlug}/imagerewritestatus` endpoint before there is a task status, so the backend will return a response indicating that image re-writing is complete even though it has not even begun.  This causes the UI to not display the spinner and status messages, as expected.

Example of this in our tests: https://app.testim.io/#/project/wpYAooUimFDgQxY73r17/branch/ricardomaraschini%2Fsc-86964%2Fkots-test-workflow-for-helmvm/test/aPhWnwQiqtBxx2sn/step/wbZJmRizYrvNWuVc/viewer/screenshots?result-id=XCLtQZlJnzWmYF5k&path=w5blTCkqUInoZuUc%3AwbZJmRizYrvNWuVc

See testim link above...

Changed the Registry Settings step in `validate-smoke-test` to mark an error if the loading indicator does not show to add coverage for the UI ([link](https://app.testim.io/#/project/wpYAooUimFDgQxY73r17/branch/cbo%2Ffix-registry-settings-race-condition/test/14c1fW5zraPrMSf/step/gGSEp6fqGG60lrBU?result-id=fvzXZmvvwQ4HfoeN&path=5mb6Ivw1Us27tpN1%3AgGSEp6fqGG60lrBU)).

## Steps to reproduce

It's a race condition, so only happens occasionally.  To reproduce, change the registry settings and save.  The UI will not always show the loading indicator and status messages.

```release-note
Fixes an issue where updating the registry settings would not always display the loading indicator and status messages in the UI
```

NONE

### DIFF
--- a/.github/actions/kots-gke-create/reserve-capacity.yaml
+++ b/.github/actions/kots-gke-create/reserve-capacity.yaml
@@ -22,7 +22,7 @@ metadata:
   name: capacity-res-job
 spec:
   parallelism: 4
-  backoffLimit: 0
+  backoffLimit: 6
   template:
     metadata:
       labels:
--- a/pkg/handlers/registry.go
+++ b/pkg/handlers/registry.go
@@ -210,21 +210,31 @@ func (h *Handler) UpdateAppRegistry(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	skipImagePush := updateAppRegistryRequest.IsReadOnly
+	if foundApp.IsAirgap {
+		// TODO: pushing images not yet supported in airgapped instances.
+		skipImagePush = true
+	}
+
+	latestSequence, err := store.GetStore().GetLatestAppSequence(foundApp.ID, true)
+	if err != nil {
+		logger.Error(errors.Wrapf(err, "failed to get latest app sequence for app %s", foundApp.Slug))
+		updateAppRegistryResponse.Error = err.Error()
+		JSON(w, http.StatusInternalServerError, updateAppRegistryResponse)
+		return
+	}
+
+	// set task status before starting the goroutine so that the UI can show the status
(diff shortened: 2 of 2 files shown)


pkg/handlers/registry.go
```diff
@@ -210,21 +210,31 @@ func (h *Handler) UpdateAppRegistry(w http.ResponseWriter, r *http.Request) {
...
+		skipImagePush = true
+	}
+
+	latestSequence, err := store.GetStore().GetLatestAppSequence(foundApp.ID, true)
+	if err != nil {
+		logger.Error(errors.Wrapf(err, "failed to get latest app sequence for app %s", foundApp.Slug))
+		updateAppRegistryResponse.Error = err.Error()
+		JSON(w, http.StatusInternalServerError, updateAppRegistryResponse)
+		return
+	}
```
cbodonnell, line 225: Moved this bit out of the goroutine since we don't want to set an initial task status of `"running"` if the query fails.

.github/actions/kots-gke-create/reserve-capacity.yaml
```diff
@@ -22,7 +22,7 @@ metadata:
   name: capacity-res-job
 spec:
   parallelism: 4
-  backoffLimit: 0
+  backoffLimit: 6
```
cbodonnell, line 25: Unrelated to the issue, but should fix flakiness we've been seeing in this test as it will allow this job, which reserves GKE autopilot capacity, to be retried.

divolgin (approved):

PR #4040, opened 2023-09-14
Remove authorization header from custom metrics API

divolgin:

Custom metrics API no longer requires license ID in the authorization header. This also means that custom metrics can no longer be reported in multi-app installs.  The main reason for this change for consistency with Replicated SDK, where license ID is not necessarily known.

Fixes #

```release-note
- Custom metrics API no longer requires authorization header.
```

https://github.com/replicatedhq/replicated-docs/pull/1441


pkg/handlers/handlers.go
```diff
@@ -339,14 +339,9 @@ func RegisterUnauthenticatedRoutes(handler *Handler, kotsStore store.Store, debu
 	loggingRouter.Path("/api/v1/troubleshoot/supportbundle/{bundleId}/redactions").Methods("PUT").HandlerFunc(handler.SetSupportBundleRedactions)
 	loggingRouter.Path("/api/v1/preflight/app/{appSlug}/sequence/{sequence}").Methods("POST").HandlerFunc(handler.PostPreflightStatus)
 
-	// This the handler for license API and should be called by the application only.
+	// This these handlers for should be called by the application only.
```
cbodonnell, line 342: nit pick on the comment
```suggestion
	// These handlers should be called by the application only.
```

cbodonnell (dismissed):

cbodonnell (approved):

PR #4046, opened 2023-09-19
Log install preflight errors/warnings in both the CLI and kotsadm pod

banjoh:

When deployments fail due to preflights having warnings/errors, users do not get to see which of the preflights failed. This change logs all the executed preflights whenever they prevent a deployment. We print all preflights so as to also have users see other preflights that got run so they can perform additional review of their preflights.

**Screenshot of the CLI logs after preflights failing/warning**
image

Fixes #

```release-note
Log preflight checks to the CLI and kotsadm logs whenever there are checks that fail leading to a failed deployment
```


cmd/kots/cli/install.go
```diff
@@ -47,6 +46,10 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+var client = &http.Client{
```
banjoh, line 49: Previously, there was not timeout. Just being defensive here.

pkg/preflight/execute.go
```diff
@@ -121,7 +126,7 @@ func execute(appID string, sequence int64, preflightSpec *troubleshootv1beta2.Pr
 		KubernetesRestConfig:   restConfig,
 	}
 
-	logger.Debug("preflight collect phase")
+	logger.Info("preflight collect phase")
```
banjoh, line 129: Moving this to info cause they show progress when performing an installation, which is critical functionality IMO to always log

pkg/preflight/preflight.go
```diff
@@ -160,13 +160,27 @@ func Run(appID string, appSlug string, sequence int64, isAirgap bool, archiveDir
...
 			if err != nil {
 				logger.Error(errors.Wrap(err, "failed to run preflight checks"))
 				return
 			}
-			logger.Debug("preflight checks completed")
+
+			// Log the preflight results if there are any warnings or errors
+			// The app may not get installed so we need to see this info for debugging
+			if GetPreflightState(uploadPreflightResults) != "pass" {
+				// TODO: Are there conditions when the application gets installed?
```
cbodonnell, line 173: I don't think so here since we should only deploy if the preflight state is `"pass"` https://github.com/replicatedhq/kots/blob/main/pkg/preflight/preflight.go#L236-L245

cmd/kots/cli/install.go
```diff
@@ -444,11 +448,22 @@ func InstallCmd() *cobra.Command {
...
+					log.FinishSpinner()
 					if err := ValidatePreflightStatus(deployOptions, authSlug, apiEndpoint); err != nil {
-						log.FinishSpinnerWithError()
-						return errors.Wrap(err, "failed to validate preflight results")
+						log.Errorf("Preflight checks contain warnings or errors. The application was not deployed")
+						perr := preflightError{}
+						if errors.As(err, &perr) {
+							cmd.SilenceErrors = true
+							var s strings.Builder
+							s.WriteString("\nPreflight check results (state - title - message)")
```
cbodonnell, line 458: We have a `pkg/print` where have logic for pretty-printing various things.  I think [`pkg/print/config.go`](https://github.com/replicatedhq/kots/blob/main/pkg/print/config.go) could be a good reference since those print config validation errors, which might be similar to something like preflight warnings/errors. This logic could live there.

cmd/kots/cli/install.go
```diff
@@ -444,11 +448,22 @@ func InstallCmd() *cobra.Command {
 				switch status {
 				case storetypes.VersionPendingPreflight:
 					log.ActionWithSpinner("Waiting for preflight checks to complete")
+					log.FinishSpinner()
 					if err := ValidatePreflightStatus(deployOptions, authSlug, apiEndpoint); err != nil {
-						log.FinishSpinnerWithError()
-						return errors.Wrap(err, "failed to validate preflight results")
+						log.Errorf("Preflight checks contain warnings or errors. The application was not deployed")
+						perr := preflightError{}
+						if errors.As(err, &perr) {
```
cbodonnell, line 455: Any benefit to handling specific error types this way vs. how do it elsewhere in the codebase? For example: https://github.com/replicatedhq/kots/blob/main/pkg/handlers/snapshots.go#L202

pkg/preflight/preflight.go
```diff
@@ -194,7 +208,7 @@ func Run(appID string, appSlug string, sequence int64, isAirgap bool, archiveDir
 			// preflight reporting
 			if isDeployed {
 				if err := reporting.WaitAndReportPreflightChecks(appID, sequence, false, false); err != nil {
-					logger.Debugf("failed to send preflights data to replicated app: %v", err)
+					logger.Errorf("failed to send preflights data to replicated app: %v", err)
```
cbodonnell, line 211: i'm not sure this message is always accurate anymore given the work to support airgap reporting.  since we're exposing it as a `error` log with this change, how about we change the message to something more generic like:
```
failed to report preflight check results
```

cmd/kots/cli/install.go
```diff
@@ -426,6 +429,7 @@ func InstallCmd() *cobra.Command {
 				case err := <-errChan:
 					if err != nil {
 						log.Error(err)
+						// TODO: Why is this a negative exit code?
```
cbodonnell, line 432: Good question... I think this end up resulting in a 255 exit code:
```
$ echo $?
255
```

cmd/kots/cli/install.go
```diff
@@ -444,11 +448,22 @@ func InstallCmd() *cobra.Command {
...
 				case storetypes.VersionPendingPreflight:
 					log.ActionWithSpinner("Waiting for preflight checks to complete")
+					log.FinishSpinner()
 					if err := ValidatePreflightStatus(deployOptions, authSlug, apiEndpoint); err != nil {
-						log.FinishSpinnerWithError()
-						return errors.Wrap(err, "failed to validate preflight results")
+						log.Errorf("Preflight checks contain warnings or errors. The application was not deployed")
+						perr := preflightError{}
+						if errors.As(err, &perr) {
+							cmd.SilenceErrors = true
```
cbodonnell, line 456: Do we need to silence errors if we're always going to call `os.Exit(1)` at the end of this code path?

cmd/kots/cli/install.go
banjoh: > Any benefit to handling specific error types this way vs. how do it elsewhere in the codebase? For example: https://github.com/replicatedhq/kots/blob/main/pkg/handlers/snapshots.go#L202

They both achieve the same result I think. `errors.Cause()` is from [`github.com/pkg/errors`](https://github.com/pkg/errors#retrieving-the-cause-of-an-error) library while `errors.As()` is part of [go standard library](https://pkg.go.dev/errors)

pkg/preflight/preflight.go
banjoh: > i'm not sure this message is always accurate anymore given the work to support airgap reporting. since we're exposing it as a `error` log with this change, how about we change the message to something more generic like:
> 
> ```
> failed to report preflight check results
> ```

I've seen a number of these debug logs all over the code. I think I'll revert this change for now and leave it as is. Later on we can align the logs.

cmd/kots/cli/install.go
banjoh: > Do we need to silence errors if we're always going to call `os.Exit(1)` at the end of this code path?

Perhaps not. Removing it

cmd/kots/cli/install.go
```diff
@@ -445,10 +450,15 @@ func InstallCmd() *cobra.Command {
 				case storetypes.VersionPendingPreflight:
 					log.ActionWithSpinner("Waiting for preflight checks to complete")
 					if err := ValidatePreflightStatus(deployOptions, authSlug, apiEndpoint); err != nil {
-						log.FinishSpinnerWithError()
-						return errors.Wrap(err, "failed to validate preflight results")
+						log.FinishSpinner()
```
cbodonnell, line 453: to keep the old behavior where it will show a red X instead of a green checkmark if preflights do not pass
```suggestion
						log.FinishSpinnerWithError()
```

cmd/kots/cli/install.go
```diff
@@ -445,10 +450,15 @@ func InstallCmd() *cobra.Command {
...
+						log.Errorf("Preflight checks contain warnings or errors. The application was not deployed")
+						perr := preflightError{}
+						if errors.As(err, &perr) {
+							print.PreflightErrors(log, perr.Results)
+							os.Exit(1)
+						}
+						return err
 					}
-					log.FinishSpinner()
```
cbodonnell, line 451: We should leave this as we need to stop the spinner in the case where preflights pass too.  Otherwise we are left with an indefinite spinner:
```
  • Deploying Admin Console
    • Creating namespace ✓  
    • Waiting for datastore to be ready ✓  
  • Waiting for Admin Console to be ready ✓  
  • Waiting for installation to complete ✓  
  • Waiting for preflight checks to complete ⠇
  • Press Ctrl+C to exit
  • Go to http://localhost:8800 to access the Admin Console

  • Waiting for preflight checks to complete ⠏
```

banjoh:
Some tests are failing. Looking into them now.

cmd/kots/cli/install.go
sgalsaleh: if they achieve the same thing, let's just stick to `errors.Cause()` for consistency?

pkg/preflight/preflight.go
```diff
@@ -216,6 +229,21 @@ func Run(appID string, appSlug string, sequence int64, isAirgap bool, archiveDir
 	return nil
 }
 
+func preflightState(p troubleshootpreflight.UploadPreflightResult) string {
```
sgalsaleh, line 232: why not use the existing `GetPreflightState` function`?

pkg/print/config.go
```diff
@@ -26,3 +27,26 @@ func ConfigValidationErrors(log *logger.CLILogger, groupValidationErrors []confi
...
+func PreflightErrors(log *logger.CLILogger, results []*preflight.UploadPreflightResult) {
+	var s strings.Builder
+	s.WriteString("\nPreflight check results (state - title - message)")
+	for _, result := range results {
+		s.WriteString(fmt.Sprintf("\n%s - %q - %q", preflightState(result), result.Title, result.Message))
+	}
+	log.Info(s.String())
+}
+
+func preflightState(r *preflight.UploadPreflightResult) string {
```
sgalsaleh, line 40: ditto

pkg/print/config.go
```diff
@@ -26,3 +27,26 @@ func ConfigValidationErrors(log *logger.CLILogger, groupValidationErrors []confi
 	log.FinishSpinnerWithError()
 	log.Errorf(sb.String())
 }
+
+func PreflightErrors(log *logger.CLILogger, results []*preflight.UploadPreflightResult) {
```
sgalsaleh, line 31: why not following the existing pattern of printing? https://github.com/replicatedhq/kots/blob/bc8004b7b2a9d2a14ee1fb03af7c9aba3337ca46/pkg/print/restore.go#L25-L49

that would help align the column titles with the values.

cmd/kots/cli/install.go
sgalsaleh: let's fix this and set it to 255

cmd/kots/cli/install.go
```diff
@@ -446,7 +451,13 @@ func InstallCmd() *cobra.Command {
 					log.ActionWithSpinner("Waiting for preflight checks to complete")
 					if err := ValidatePreflightStatus(deployOptions, authSlug, apiEndpoint); err != nil {
 						log.FinishSpinnerWithError()
-						return errors.Wrap(err, "failed to validate preflight results")
+						log.Errorf("Preflight checks contain warnings or errors. The application was not deployed")
+						perr := preflightError{}
+						if errors.As(err, &perr) {
+							print.PreflightErrors(log, perr.Results)
+							cmd.SilenceErrors = true // Stop Cobra from printing the error, we print it ourselves
```
sgalsaleh, line 458: instead of silencing the errors, can we just return/print a generic error message?

pkg/preflight/preflight.go
banjoh: The state from `GetPreflightState` is of the entire list of preflights i.e if there is an error in any, the state == `error`. `preflightState` gets the state of each preflight for logging purposes.

pkg/print/config.go
banjoh: I'll DRY this out in favour of `preflights.PreflightState`

pkg/print/config.go
banjoh: > why not following the existing pattern of printing?

No particular reason. I'll try to adopt the same approach

cmd/kots/cli/install.go
banjoh: > instead of silencing the errors, can we just return/print a generic error message?

That extra line after the formatted preflight check results at the end did not feel right. This is how it ends up looking like

```sh
evans@evans-vm3:~$ ./kots install thanos-reloaded -n default --shared-password password --license-file ~/thanos_license.yaml --kubeconfig /etc/rancher/k3s/k3s.yaml
  • Deploying Admin Console
    • Creating namespace ✓
    • Waiting for datastore to be ready ✓
  • Waiting for Admin Console to be ready ✓
  • Waiting for installation to complete ✓
  • Waiting for preflight checks to complete ✓
  • Preflight checks contain warnings or errors. The application was not deployed

STATE    TITLE                          MESSAGE
PASS     Required Kubernetes Version    Your cluster meets the recommended and required versions of Kubernetes.
PASS     Container Runtime              containerd container runtime was found.
WARN     Kubernetes Distribution        Unable to determine the distribution of Kubernetes.
WARN     Total CPU Cores                The cluster should contain at least 4 cores.

Error: There are preflight check warnings for the application. The app was not deployed.
```

NOTE: Any other errors will get printed, we just silence the errors when we format the preflight checks table since we are printing our own error message.

banjoh: Here it is as an image to show the colours

image

cbodonnell (approved):

PR #4051, opened 2023-09-27
Detect and configure the replicated SDK chart

cbodonnell:

Fixes #

```release-note

```

### DIFF
--- a/pkg/upstream/write.go
+++ b/pkg/upstream/write.go
@@ -3,11 +3,11 @@ package upstream
 import (
 	"bytes"
 	"encoding/base64"
-	"io/ioutil"
 	"os"
 	"path"
 
 	"github.com/pkg/errors"
+	"github.com/replicatedhq/kots/pkg/archives"
 	"github.com/replicatedhq/kots/pkg/crypto"
 	"github.com/replicatedhq/kots/pkg/kotsutil"
 	"github.com/replicatedhq/kots/pkg/upstream/types"
@@ -39,7 +39,7 @@ func WriteUpstream(u *types.Upstream, options types.WriteOptions) error {
 	if err == nil {
 		_, err = os.Stat(path.Join(renderDir, "userdata", "installation.yaml"))
 		if err == nil {
-			c, err := ioutil.ReadFile(path.Join(renderDir, "userdata", "installation.yaml"))
+			c, err := os.ReadFile(path.Join(renderDir, "userdata", "installation.yaml"))
 			if err != nil {
 				return errors.Wrap(err, "failed to read existing installation")
 			}
@@ -100,7 +100,16 @@ func WriteUpstream(u *types.Upstream, options types.WriteOptions) error {
 			u.Files[i] = file
 		}
 
-		if err := ioutil.WriteFile(fileRenderPath, file.Content, 0644); err != nil {
+		if archives.IsTGZ(file.Content) {
+			updatedContent, err := configureChart(file.Content, u, options)
+			if err != nil {
+				return errors.Wrap(err, "failed to configure replicated sdk")
+			}
+			file.Content = updatedContent
+			u.Files[i] = file
+		}
+
+		if err := os.WriteFile(fileRenderPath, file.Content, 0644); err != nil {
 			return errors.Wrap(err, "failed to write upstream file")
 		}
 	}
@@ -131,6 +140,7 @@ func WriteUpstream(u *types.Upstream, options types.WriteOptions) error {
 			ReleaseNotes:             u.ReleaseNotes,
 			ReplicatedRegistryDomain: u.ReplicatedRegistryDomain,
 			ReplicatedProxyDomain:    u.ReplicatedProxyDomain,
+			ReplicatedChartNames:     u.ReplicatedChartNames,
 			EncryptionKey:            encryptionKey,
 		},
 	}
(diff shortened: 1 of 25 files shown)


pkg/upstream/write.go
```diff
@@ -198,3 +232,557 @@ func maybeEncryptIdentityConfig(identityConfig *kotsv1beta1.IdentityConfig) ([]b
...
+
+type ReplicatedValuesOptions struct {
+	LicenseYAML []byte
+	ClusterID   string
+	AppID       string
+}
+
+func addReplicatedValuesForSDK(doc *yaml.Node, opts ReplicatedValuesOptions, replicatedSDKSubchartName string, isReplicatedSDK bool) error {
+	replicatedValues := map[string]interface{}{
+		"license":               string(opts.LicenseYAML),
```
cbodonnell, line 530: should only need to do this in airgap

pkg/upstream/write.go
```diff
@@ -198,3 +232,557 @@ func maybeEncryptIdentityConfig(identityConfig *kotsv1beta1.IdentityConfig) ([]b
...
+		for i, n := range targetNode.Content {
+			if n.Value == "replicated" {
+				targetNode = targetNode.Content[i+1]
+				hasGlobalReplicated = true
+				break
+			}
+		}
+	}
+
+	replicatedValues := map[string]interface{}{
```
cbodonnell, line 609: should only need to do this in airgap

pkg/upstream/write.go
```diff
@@ -100,7 +115,25 @@ func WriteUpstream(u *types.Upstream, options types.WriteOptions) error {
...
-		if err := ioutil.WriteFile(fileRenderPath, file.Content, 0644); err != nil {
+		if archives.IsTGZ(file.Content) {
+			// this is a helm chart, so we need to check if it is or contains the replicated-sdk
+			reader := bytes.NewReader(file.Content)
+			replicatedSDKChartName, isReplicatedSDK, err := FindReplicatedSDKChart(reader, u.ReplicatedChartNames)
+			if err != nil {
+				return errors.Wrap(err, "failed to find replicated-sdk subchart")
+			}
+
+			if replicatedSDKChartName != "" {
```
sgalsaleh, line 126: why only if it's a subchart?

pkg/upstream/write.go
```diff
@@ -198,3 +232,557 @@ func maybeEncryptIdentityConfig(identityConfig *kotsv1beta1.IdentityConfig) ([]b
 
 	return b.Bytes(), nil
 }
+
+// FindReplicatedSDKChart will look for a chart with the name "replicated" or "replicated-sdk" in the archive
```
sgalsaleh, line 236: should update this comment to reflect that it uses `replicatedChartNames` and not hardcoded values
```suggestion
// FindReplicatedSDKChart will look for a chart with the name "replicated" or "replicated-sdk" in the archive
```

pkg/upstream/write.go
```diff
@@ -198,3 +232,557 @@ func maybeEncryptIdentityConfig(identityConfig *kotsv1beta1.IdentityConfig) ([]b
...
+					ChartName string `json:"name" yaml:"name"`
+				}{}
+				if err := yaml.Unmarshal(content, &chartInfo); err != nil {
+					return "", false, errors.Wrapf(err, "failed to unmarshal %s", header.Name)
+				}
+
+				if replicatedChartNamesMap[chartInfo.ChartName] {
+					// check if the sdk is the parent chart or a subchart based on the path
+					replicatedSDKChartName := chartInfo.ChartName
+					isReplicatedSDK := !strings.Contains(filepath.Dir(header.Name), string(os.PathSeparator))
```
sgalsaleh, line 278: wouldn't there be a top-level directory before the filename? e.g. `replicated/Chart.yaml`.

pkg/upstream/write.go
```diff
@@ -198,3 +232,557 @@ func maybeEncryptIdentityConfig(identityConfig *kotsv1beta1.IdentityConfig) ([]b
...
+			content, err := io.ReadAll(tarReader)
+			if err != nil {
+				finalErr = errors.Wrap(err, "failed to read file")
+				return
+			}
+
+			if filepath.Base(header.Name) == "values.yaml" {
+				// only get the values.yaml from the top level chart
+				p := filepath.Dir(header.Name)
+				if !strings.Contains(p, string(os.PathSeparator)) {
```
sgalsaleh, line 371: ditto

pkg/upstream/write.go
```diff
@@ -198,3 +232,557 @@ func maybeEncryptIdentityConfig(identityConfig *kotsv1beta1.IdentityConfig) ([]b
...
+	if hasGlobalReplicated || hasGlobal {
+		targetNode.Content = kotsutil.MergeYAMLNodes(targetNode.Content, additionalNode.Content[0].Content)
+	} else {
+		targetNode.Content = append(targetNode.Content, additionalNode.Content[0].Content...)
+	}
+
+	return nil
+}
+
+func convertLicenseFieldsToYamlNodes(licenseFields map[string]LicenseField) ([]*yaml.Node, error) {
```
sgalsaleh, line 650: can't you do a `yaml.Marshal` then `yaml.Unmarshal` into nodes?

pkg/upstream/write.go
cbodonnell: this should have a value if it's found at all, regardless of whether it's a parent or subchart. the `isReplicatedSDK` variable (now renamed to `isSubchart`) is what makes the distinction

pkg/upstream/write.go
sgalsaleh: agreed.

sgalsaleh (approved):

PR #4062, opened 2023-10-09
configure sdk metrics endpoint

cbodonnell:

This PR add logic to configure the Replicated SDK `additionalMetricsEndpoint` for deployments managed by KOTS.

https://app.shortcut.com/replicated/story/88175/add-a-helm-value-to-the-sdk-to-accept-an-endpoint-where-it-retrieves-additional-metrics-to-be-reported

```release-note

```

### DIFF
--- a/pkg/upstream/helm.go
+++ b/pkg/upstream/helm.go
@@ -17,6 +17,7 @@ import (
 	"github.com/replicatedhq/kots/pkg/buildversion"
 	"github.com/replicatedhq/kots/pkg/kotsutil"
 	"github.com/replicatedhq/kots/pkg/upstream/types"
+	"github.com/replicatedhq/kots/pkg/util"
 	"gopkg.in/yaml.v3"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/upstream/helm.go
```diff
@@ -268,10 +269,11 @@ func addReplicatedValues(doc *yaml.Node, replicatedChartName string, isSubchart
...
 	replicatedValues := map[string]interface{}{
-		"replicatedID": options.KotsadmID,
-		"appID":        options.AppID,
-		"userAgent":    buildversion.GetUserAgent(),
-		"isAirgap":     options.IsAirgap,
+		"replicatedID":              options.KotsadmID,
+		"appID":                     options.AppID,
+		"userAgent":                 buildversion.GetUserAgent(),
+		"isAirgap":                  options.IsAirgap,
+		"additionalMetricsEndpoint": fmt.Sprintf("%s/api/v1/app/metrics", fmt.Sprintf("http://kotsadm.%s.svc.cluster.local:3000", util.PodNamespace)),
```
cbodonnell, line 276: Anything else we should do here to support flows that don't have a cluster context? Looking here for example: https://github.com/replicatedhq/kots/blob/main/pkg/supportbundle/spec.go#L359-L368

sgalsaleh: The `additionalMetricsEndpoint` will only be used within the SDK's API, and currently, that can only run as a service inside the cluster. I don't think there are use cases without a cluster context when things are managed by KOTS.

PR #4105, opened 2023-10-20
write instance and preflight reports to a secret

cbodonnell:

This PR modifies airgap reporting so that instance and preflight data is written to a k8s secret. Additional dynamic collectors were added to the KOTS support bundle spec to write the data in these secrets to support bundles. This data will be written to `secrets/[namespace]/kotsadm-[appSlug]-instance-report/report.json` and `secrets/[namespace]/kotsadm-[appSlug]-preflight-report/report.json` for the instance and preflight reports, respectively.

https://app.shortcut.com/replicated/story/88191/store-airgap-telemetry-for-the-sdk-in-a-kubernetes-secret

```release-note
NONE
```

NONE


pkg/reporting/preflight_airgap.go
```diff
@@ -1,29 +1,117 @@
...
+	}
+
+	if err := CreatePreflightReportEvent(r.clientset, util.PodNamespace, app.Slug, event); err != nil {
+		return errors.Wrap(err, "failed to create preflight report event")
+	}
+
+	return nil
+}
+
+func CreatePreflightReportEvent(clientset kubernetes.Interface, namespace string, appSlug string, event PreflightReportEvent) error {
```
cbodonnell, line 58: Thoughts on refactoring this function and `CreateInstanceReportEvent` so that both the instance and preflight reports use the same function? I did a smaller DRY for some of the utility functions, but thought that having this function accept a `report any` might make it somewhat less understandable.

pkg/reporting/app_airgap.go
```diff
@@ -1,28 +1,147 @@
...
+	}
+
+	return nil
+}
+
+func GetInstanceReportEvent(licenseID string, reportingInfo *types.ReportingInfo) InstanceReportEvent {
+	// not using the "cursor" packages because it doesn't provide access to the underlying int64
+	downstreamSequence, err := strconv.ParseUint(reportingInfo.Downstream.Cursor, 10, 64)
+	if err != nil {
+		logger.Debugf("failed to parse downstream cursor %q: %v", reportingInfo.Downstream.Cursor, err)
```
cbodonnell, line 56: This maintains existing behavior by not bailing if we fail to parse the downstream sequence. Any reason to change this?

pkg/reporting/util.go
```diff
@@ -87,3 +94,62 @@ func isDevEndpoint(endpoint string) bool {
...
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "Secret",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			// since this secret is created by the kotsadm deployment, we should set the owner reference
+			// so that it is deleted when the kotsadm deployment is deleted
+			OwnerReferences: []metav1.OwnerReference{
```
cbodonnell, line 109: Added this for consistency, but we don't really do this with other KOTS resources, so might not be necessary.

pkg/reporting/app_airgap.go
sgalsaleh: nope

pkg/reporting/util.go
```diff
@@ -87,3 +94,62 @@ func isDevEndpoint(endpoint string) bool {
...
 	return result
 }
+
+func AirgapReportSecret(name string, namespace string, kotsadmUID apimachinerytypes.UID, data []byte) *corev1.Secret {
+	return &corev1.Secret{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "Secret",
+		},
+		ObjectMeta: metav1.ObjectMeta{
```
sgalsaleh, line 104: What do you think of adding the necessary labels for this to be included in snapshots?

pkg/reporting/util.go
sgalsaleh: It feels wrong from an architectural standpoint, as the IDs (instance and cluster IDs) are not tied to the `kotsadm` deployment like the case with the `replicated` sdk deployment. The cluster ID is stored in the `kotsadm-confg` configmap, and the instance ID is stored in rqlite, so the `kotsadm` deployment is ephemeral in that sense, and recreating it shouldn't cause this data to be deleted.

Another caveat is that the data will be lost during a migration from a `kotsadm` deployment to statefulset.

With that being said, we should figure out how to reset the data in case any of the IDs don't match, which can happen due to several reasons (re-installing the app, kots, etc...)

pkg/reporting/preflight_airgap.go
sgalsaleh: Yeah. You can create a `Report` interface, and maybe move all of this logic to `report.go` in this `reporting` package. You can then implement functions like `GetEventLimit`, `GetSecretKey`, etc...

You can then have a single `CreateReportEvent` function that accepts this interface. I'd also suggest renaming `AirgapReportSecret` to `ReportSecret`, `DecodeAirgapReport` to `DecodeAReport`, `EncodeAirgapReport` to `EncodeReport`.

pkg/reporting/util.go
sgalsaleh: Don't forget to enforce type checking for that new interface with something like:
```go
var _ Report = &InstanceReport{}
var _ Report = &PreflightReport{}
```

pkg/reporting/util.go
cbodonnell: I think we should do it since rqlite data would have been restored

pkg/reporting/util.go
cbodonnell: > With that being said, we should figure out how to reset the data in case any of the IDs don't match, which can happen due to several reasons (re-installing the app, kots, etc...)

Might be a good use-case for another top-level key in the secret? Could store IDs there and ensure they match before appending data. Which IDs do we need to validate? License, Cluster, and Instance?

pkg/reporting/util.go
sgalsaleh: I agree.

pkg/reporting/util.go
sgalsaleh: Good idea, but we probably don't need top-level fields for that. How about just reading the IDs from the first event?

pkg/reporting/report.go
```diff
@@ -0,0 +1,251 @@
...
+	compressedData, err := util.GzipData(data)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to gzip report")
+	}
+	encodedData := base64.StdEncoding.EncodeToString(compressedData)
+
+	return []byte(encodedData), nil
+}
+
+func DecodeReport(encodedData []byte, existingReport *Report, reportType string) error {
```
sgalsaleh, line 219: nit pick: i would suggest having this function return a decoded report instead of modifying `existingReport`.

pkg/reporting/report_test.go
```diff
@@ -0,0 +1,274 @@
...
+					},
+				),
+				namespace: "default",
+				appSlug:   "test-app-slug",
+				event:     testEvent,
+			},
+			wantNumEvents: 1,
+		},
+		// {
+		// 	name: "secret exists with max number of events",
```
sgalsaleh, line 251: why is this commented out?

pkg/reporting/report.go
```diff
@@ -0,0 +1,251 @@
...
+
+func (p *PreflightReportEvent) GetReportEventLimit() int {
+	return ReportEventLimit
+}
+
+func (p *PreflightReportEvent) GetReportMtx() *sync.Mutex {
+	return &preflightReportMtx
+}
+
+func CreateReportEvent(clientset kubernetes.Interface, namespace string, appSlug string, event ReportEvent) error {
```
sgalsaleh, line 135: It was a bit confusing to me at first glance why the `ReportEvent` is the interface rather than the `Report`. What do you think of making `Report` be the interface, renaming this function to `UpsertReport`, and the caller can decide how many events to send (in our case, it'll be a single event for now)?

pkg/reporting/report.go
```diff
@@ -0,0 +1,251 @@
...
+
+func (i *InstanceReportEvent) GetReportEventLimit() int {
+	return ReportEventLimit
+}
+
+func (i *InstanceReportEvent) GetReportMtx() *sync.Mutex {
+	return &instanceReportMtx
+}
+
+type PreflightReport struct {
```
sgalsaleh, line 97: i would also suggest moving preflight/instance specific types to its own files. e.g. `preflight_report.go` and `instance_report.go` along with the implementations of the interface functions.

pkg/reporting/report_test.go
cbodonnell: good catch

pkg/reporting/util.go
cbodonnell: Removed owner references, but leaving secrets as-is

pkg/reporting/report_test.go
```diff
@@ -0,0 +1,266 @@
...
+	encodedPreflightReport, err := EncodeReport(testPrelightReport)
+	req.NoError(err)
+
+	decodedPreflightReport, err := DecodeReport(encodedPreflightReport, testPrelightReport.GetType())
+	req.NoError(err)
+
+	req.Equal(testPrelightReport, decodedPreflightReport)
+}
+
+func Test_CreateReportEvent(t *testing.T) {
```
sgalsaleh, line 88:
```suggestion
func Test_AppendReport(t *testing.T) {
```

pkg/reporting/report_test.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+type CreateReportEventTestArgs struct {
+	clientset kubernetes.Interface
+	namespace string
+	appSlug   string
+	report    Report
+}
+
+func createTestsForEvent(t *testing.T, testReport Report) []CreateReportEventTest {
```
sgalsaleh, line 181:
```suggestion
func createTestsForReport(t *testing.T, testReport Report) []CreateReportEventTest {
```

pkg/reporting/report_test.go
```diff
@@ -0,0 +1,266 @@
...
+			req.NotNil(secret.Data[tt.args.report.GetSecretKey()])
+
+			report, err := DecodeReport(secret.Data[tt.args.report.GetSecretKey()], tt.args.report.GetType())
+			req.NoError(err)
+			req.Equal(tt.args.report, report)
+		})
+	}
+}
+
+type CreateReportEventTest struct {
```
sgalsaleh, line 167:
```suggestion
type AppendReportTest struct {
```

pkg/reporting/report_test.go
```diff
@@ -0,0 +1,266 @@
...
+	name          string
+	args          CreateReportEventTestArgs
+	wantNumEvents int
+	wantErr       bool
+}
+
+type CreateReportEventTestArgs struct {
```
sgalsaleh, line 174:
```suggestion
type AppendReportTestArgs struct {
```

sgalsaleh (approved):

PR #4113, opened 2023-10-31
change troubleshoot.io to troubleshoot.sh

cbodonnell:

Changes the label added to support bundle configmaps and secrets from `troubleshoot.io/kind` to `troubleshoot.sh/kind`, which is the correct label documented here: https://troubleshoot.sh/docs/support-bundle/discover-cluster-specs/

https://app.shortcut.com/replicated/story/92280/support-bundles-generated-from-kots-ui-don-t-include-sdk-airgap-telemetry

```release-note
Fixes an issue where KOTS support bundles were not discovering specs with the `troubleshoot.sh/kind=support-bundle` label
```

NONE

### DIFF
--- a/pkg/kotsadm/types/constants.go
+++ b/pkg/kotsadm/types/constants.go
@@ -13,7 +13,7 @@ const ExcludeValue = "true"
 const BackupLabel = "kots.io/backup"
 const BackupLabelValue = "velero"
 
-const TroubleshootKey = "troubleshoot.io/kind"
+const TroubleshootKey = "troubleshoot.sh/kind"
 const TroubleshootValue = "support-bundle"
 
 const DefaultSupportBundleSpecKey = "default"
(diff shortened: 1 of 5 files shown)


pkg/kotsadm/types/constants.go
```diff
@@ -13,7 +13,7 @@ const ExcludeValue = "true"
 const BackupLabel = "kots.io/backup"
 const BackupLabelValue = "velero"
 
-const TroubleshootKey = "troubleshoot.io/kind"
+const TroubleshootKey = "troubleshoot.sh/kind"
```
cbodonnell, line 16: should we continue supporting the old label? it's not documented [here](https://troubleshoot.sh/docs/support-bundle/discover-cluster-specs/), but maybe someone started using it at some point?

sgalsaleh (approved):

PR #4120, opened 2023-11-06
Cap report secrets data at 1MB and improve tests

sgalsaleh:

Fixes an issue where the reporting data stored in secrets in air gapped installations could exceed the size of the secret (1MB).

Fixes #

```release-note
Fixes an issue where the reporting data stored in secrets in air gapped installations could exceed the size of the secret (1MB).
```


cbodonnell (approved):

AmberAlston:
I don't believe we document the airgap telemetry feature just yet given it's in alpha, but as we move that to Beta in the next few weeks and add docs, we should explicitly call out this data storage limit. Since I'm likely to be out of office when this is relevant, tagging folks who can help ensure we cover this when appropriate @paigecalvert @obfuscurity @grantmiller

PR #4126, opened 2023-11-08
Use CLI to validate minimal RBAC tests instead of testim

sgalsaleh:

Uses the CLI to validate minimal RBAC tests instead of testim to reduce the number unnecessary testim tests.

Fixes #

Image 2023-11-09 at 7 04 47 AM

```release-note
- Surfaces preflight check errors that are due to insufficient RBAC permissions to the CLI.
```

NONE


.github/workflows/build-test.yaml
```diff
@@ -697,37 +697,118 @@ jobs:
...
+          echo ${{ secrets.MINIMAL_RBAC_LICENSE }} | base64 -d > license.yaml
+          ./bin/kots \
+            install "$APP_SLUG/automated" \
+            --license-file license.yaml \
+            --no-port-forward \
+            --namespace "$APP_SLUG" \
+            --shared-password password \
+            --kotsadm-registry ttl.sh \
+            --kotsadm-namespace automated-${{ github.run_id }} \
+            --skip-preflights \
```
cbodonnell, line 747: Just double checking that we want to skip preflights. I see in the testim test that we validated that this message was shown in the UI:
> The Kubernetes RBAC policy that the Admin Console is running with does not have access to complete the Preflight Checks. It’s recommended that you run these manually before proceeding.

Anything we would want to check on the CLI to confirm similar behavior?

cbodonnell (approved):

.github/workflows/build-test.yaml
sgalsaleh: Hmm. Looks like RBAC preflights errors are not surfaced to the CLI install command. I'll work on fixing that.

cmd/kots/cli/install.go
```diff
@@ -447,14 +448,14 @@ func InstallCmd() *cobra.Command {
 				log.FinishSpinner()
 
 				switch status {
-				case storetypes.VersionPendingPreflight:
+				case storetypes.VersionPendingPreflight, storetypes.VersionPending:
```
cbodonnell, line 451: Will this run when there are no preflight checks defined?

sgalsaleh: It shouldn't, because after the `ValidateAutomatedInstall` function above, the status will either be `pending_config`, `deployed`, or `deploying` if there were no preflights defined, because the application will be automatically deployed in that case.

sgalsaleh: Hmm, we do deploy the version in a goroutine, so there might be a race condition there: https://github.com/replicatedhq/kots/blob/main/pkg/version/version.go#L83

cbodonnell (approved):

cmd/kots/cli/install.go
sgalsaleh: This should help: https://github.com/replicatedhq/kots/pull/4126/commits/6b388ee6d7dca149d1ab2a867acf978451c4de75

cbodonnell (approved):

PR #4133, opened 2023-11-14
Use roles from the embedded-cluster config

laverya:

Fixes #

```release-note

```


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/replicatedhq/kots
 
-go 1.20
+go 1.21.0
+
+toolchain go1.21.3
```
laverya, line 5: this appears to be required as part of importing embedded-cluster-operator

cbodonnell: would it make it an easier import if there was a separate module for the kinds/apis as we have for kotskinds and kurlkinds?

pkg/embeddedcluster/util.go
```diff
@@ -52,3 +57,35 @@ func ClusterID(client kubernetes.Interface) (string, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get kubebuilder client: %w", err)
+	}
+
+	var installationList embeddedclusterv1beta1.InstallationList
+	err = kbClient.List(ctx, &installationList, &kbclient.ListOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to list installations: %w", err)
+	}
+
```
cbodonnell, line 83: are there scenarios where the list returned could be empty?

laverya: There's an installation created with every embedded-cluster install, so ideally no

cbodonnell (approved):

go.mod
laverya: Might be something to consider certainly

cbodonnell (approved):

ricardomaraschini (approved):

PR #4191, opened 2023-11-30
Use Chainguard's minio, rqlite, and dex images

sgalsaleh:

Uses Chainguard's minio, rqlite, and dex images to get rid of CVEs

Fixes #

```release-note

```


cbodonnell (approved):
lgtm. one thing i have a question about:

the `build-ttl.sh` make command still does a `docker build`. any reason this should or should not use the new build method?

PR #4201, opened 2023-12-04
feat: add support for embedded cluster updates

ricardomaraschini:
TBD

Fixes #

```release-note

```


pkg/embeddedcluster/monitor.go
```diff
@@ -0,0 +1,139 @@
...
+
+// getApp returns the app deployed on top of the embedded cluster.
+func (m *monitor) getApp(ctx context.Context) (*apptypes.App, error) {
+	apps, err := m.store.ListInstalledApps()
+	if err != nil {
+		return nil, fmt.Errorf("failed to list installed apps: %w", err)
+	} else if len(apps) == 0 {
+		return nil, nil
+	}
+	return apps[0], nil
```
cbodonnell, line 52: confirming that embedded clusters only support a single application? might be good to leave a comment about this here

pkg/embeddedcluster/monitor.go
```diff
@@ -0,0 +1,139 @@
...
+		clientset: clientset,
+	}
+	go mon.start(ctx)
+	return nil
+}
+
+// monitor is a struct that groups all methods needed to monitor the embedded cluster installation.
+type monitor struct {
+	store     store.Store
+	clientset *kubernetes.Clientset
```
cbodonnell, line 41: let's use the more generic `kubernetes.Interface` type, if possible. it will make writing unit tests easier

pkg/embeddedcluster/monitor.go
```diff
@@ -0,0 +1,139 @@
...
+	return apps[0], nil
+}
+
+// maybeStartUpgrade checks if the embedded cluster is in a state that requires an upgrade. If so,
+// it starts the upgrade process. We only start an upgrade if the following conditions are met:
+// - We have an app deployed on top of the embedded cluster.
+// - The deployed app version is in ready state.
+// - The app has an embedded cluster configuration.
+// - The app embedded cluster configuration differs from the current embedded cluster config.
+func (m *monitor) maybeStartUpgrade(ctx context.Context) error {
```
cbodonnell, line 61: ideally we should find a way to unit test this logic as it's the critical path for determining whether  a cluster upgrade happens or not.

`store.Store` can be mocked, so the only place i see issues with is the `ClusterConfig` function which instantiates it's own client. maybe the cluster config could be an argument to this function and we split the logic into a `NeedsClusterUpgrade` and `StartClusterUpgrade`? this way we could unit test the important logic for `NeedsClusterUpgrade`. thoughts?

web/src/features/Dashboard/components/Dashboard.tsx
cbodonnell, line 1: the web linter is throwing some errors for this component https://github.com/replicatedhq/kots/actions/runs/7127520622/job/19407618444

migrations/tables/app_status.yaml
```diff
@@ -21,3 +21,5 @@ spec:
         type: integer
       - name: sequence
         type: integer
+      - name: embeddedcluster_state
+        type: text
```
cbodonnell, line 25: it does feel weird to have the cluster state in the app status table, but in embedded installs, the cluster is a part of the application, so i think i'm good with it. curious to get other opinions on this one

pkg/embeddedcluster/monitor.go
```diff
@@ -0,0 +1,139 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to get app version: %w", err)
+	}
+	kinds := appv.KOTSKinds
+	notDeployed := version.Status != storetypes.VersionDeployed
+	noClusterConfig := kinds == nil || kinds.EmbeddedClusterConfig == nil
+	if notDeployed || noClusterConfig {
+		return nil
+	}
+	status, err := m.store.GetAppStatus(app.ID)
```
sgalsaleh, line 86: i would move this up a bit to save a few unnecessary DB hits when the app isn't ready yet.

pkg/kotsutil/kots.go
```diff
@@ -129,6 +133,10 @@ func IsKotsKind(apiVersion string, kind string) bool {
 	if apiVersion == "kurl.sh/v1beta1" {
 		return true
 	}
+	// In addition to kotskinds, we exclude the embedded cluster configuration.
+	if apiVersion == "embeddedcluster.replicated.com/v1beta1" {
```
sgalsaleh, line 137: did we add this to the list of kots kinds in replicated.app and airgap-builder?

pkg/store/kotsstore/embedded_cluster_store.go
```diff
@@ -67,3 +69,36 @@ func (s *KOTSStore) GetEmbeddedClusterInstallCommandRoles(token string) ([]strin
 
 	return rolesArr, nil
 }
+
+func (s *KOTSStore) SetEmbeddedClusterState(state string) error {
+	db := persistence.MustGetDBSession()
+	query := "update app_status set embeddedcluster_state = ?"
```
sgalsaleh, line 75: should this be tied to a "sequence"?

pkg/store/kotsstore/embedded_cluster_store.go
```diff
@@ -67,3 +69,36 @@ func (s *KOTSStore) GetEmbeddedClusterInstallCommandRoles(token string) ([]strin
...
+	})
+	if err != nil {
+		return fmt.Errorf("failed to write: %v: %v", err, wr.Err)
+	}
+	return nil
+}
+
+func (s *KOTSStore) GetEmbeddedClusterState(appID string) (string, error) {
+	db := persistence.MustGetDBSession()
+	query := `select embeddedcluster_state from app_status where app_id = ?`
```
sgalsaleh, line 88: should this be tied to a "sequence"?

migrations/tables/app_status.yaml
sgalsaleh: I would suggest creating a new table for the embeddedcluster status as it's a different from the app status and is updated/read separately. It's even shown separately in the UI. Similar to how we have preflight results/status in a separate table even though it's technically part of the app.

pkg/embeddedcluster/monitor.go
```diff
@@ -0,0 +1,139 @@
...
+	} else if len(apps) == 0 {
+		return nil, nil
+	}
+	return apps[0], nil
+}
+
+// maybeStartUpgrade checks if the embedded cluster is in a state that requires an upgrade. If so,
+// it starts the upgrade process. We only start an upgrade if the following conditions are met:
+// - We have an app deployed on top of the embedded cluster.
+// - The deployed app version is in ready state.
```
laverya, line 58: I think we specifically do NOT want to wait for the app to be in a ready state - and in fact might want to deploy the cluster update _even if the app could not be applied to the cluster_.

This is for situations where an app version uses a resource introduced in k8s 1.X, and the cluster is currently running 1.X-1 (among other possibilities)

pkg/kotsutil/kots.go
laverya: I do not believe we have

pkg/store/kotsstore/embedded_cluster_store.go
laverya: A sequence and an app ID

pkg/embeddedcluster/monitor.go
laverya: we no longer check the app status here

migrations/tables/app_status.yaml
laverya: it's been moved to its own table now

pkg/embeddedcluster/monitor.go
laverya: this has essentially done now, since this function has faaaaaar less logic and essentially just calls 'needsClusterUpgrade' and 'startClusterUpgrade'

pkg/embeddedcluster/monitor.go
laverya: monitor type has been deleted entirely

pkg/embeddedcluster/monitor.go
laverya: this has been changed to deploy the cluster type as part of the app deploy process

multiple apps with cluster specs WILL stomp on each other though

pkg/kotsutil/kots.go
laverya: oh neat, looks like we did update it then

web/src/features/Dashboard/components/Dashboard.tsx
laverya: fixed

pkg/embeddedcluster/monitor.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	return nil
+}
+
+// watchClusterState checks the status of the installation object and updates the cluster state
+// after the cluster state has been 'installed' for 30 seconds, it will exit the loop.
+// this function is blocking and should be run in a goroutine.
+// if it is called multiple times, only one instance will run.
+func watchClusterState(ctx context.Context, store store.Store) {
+	stateMut.Lock()
```
cbodonnell, line 68: do we need to unlock this when the function finishes so it could run again, if needed?

pkg/operator/operator.go
```diff
@@ -408,9 +409,23 @@ func (o *Operator) DeployApp(appID string, sequence int64) (deployed bool, deplo
 	}
 	deployed, err = o.client.DeployApp(deployArgs)
 	if err != nil {
+		go func() {
+			logger.Info("app deploy failed, starting cluster upgrade in the background")
+			err2 := embeddedcluster.MaybeStartClusterUpgrade(context.Background(), o.k8sClientset, o.store, kotsKinds.EmbeddedClusterConfig)
+			if err2 != nil {
+				logger.Error(errors.Wrap(err2, "failed to start cluster upgrade"))
+			}
+			logger.Info("cluster upgrade started")
```
cbodonnell, line 418: would these logs show even if it wasn't an embedded cluster?

pkg/embeddedcluster/monitor.go
laverya: Whoops

pkg/operator/operator.go
laverya: Yes, I'll fix that

pkg/embeddedcluster/monitor.go
laverya: fixed

pkg/operator/operator.go
laverya: updated to put this within an isEmbeddedCluster check

cbodonnell (dismissed):

laverya (approved):

PR #4202, opened 2023-12-05
KOTS reporting to replicated-app

FourSigma:
Injects relevant header metadata when KOTS reaches out to Replicated.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/replicatedapp/api.go
+++ b/pkg/replicatedapp/api.go
@@ -60,6 +60,22 @@ func GetLatestLicenseForHelm(licenseID string) (*LicenseData, error) {
 	return licenseData, nil
 }
 
+func getLicenseById(licenseID string) (*kotsv1beta1.License, error) {
+
+	allLicenses, err := store.GetStore().GetAllAppLicenses()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get all app licenses")
+	}
+
+	for _, l := range allLicenses {
+		if l.Spec.LicenseID == licenseID {
+			return l, nil
+		}
+	}
+
+	return nil, nil
+}
+
 func getLicenseFromAPI(url string, licenseID string) (*LicenseData, error) {
 	req, err := util.NewRequest("GET", url, nil)
 	if err != nil {
@@ -67,6 +83,16 @@ func getLicenseFromAPI(url string, licenseID string) (*LicenseData, error) {
 	}
 
 	req.SetBasicAuth(licenseID, licenseID)
+
+	l, err := getLicenseById(licenseID)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get license by id")
+	}
+
+	if l != nil {
+		reportingInfo := reporting.GetReportingInfoByAppSlug(l.Spec.AppSlug)
+		reporting.InjectReportingInfoHeaders(req, reportingInfo)
+	}
 
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
--- a/pkg/replicatedapp/api.go
+++ b/pkg/replicatedapp/api.go
@@ -180,14 +180,6 @@ func getApplicationMetadataFromHost(host string, endpoint string, upstream *url.
 		return nil, errors.Wrap(err, "failed to call newrequest")
 	}
 
-	appID, err := store.GetStore().GetAppIDFromSlug(r.AppSlug)
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to get app id from slug")
-	}
-
-	reportingInfo := reporting.GetReportingInfo(appID)
-	reporting.InjectReportingInfoHeaders(getReq, reportingInfo)
-
 	getResp, err := http.DefaultClient.Do(getReq)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to execute get request")
--- a/pkg/replicatedapp/api.go
+++ b/pkg/replicatedapp/api.go
@@ -12,6 +12,7 @@ import (
 	"github.com/pkg/errors"
 	apptypes "github.com/replicatedhq/kots/pkg/app/types"
 	"github.com/replicatedhq/kots/pkg/logger"
+	"github.com/replicatedhq/kots/pkg/persistence"
 	"github.com/replicatedhq/kots/pkg/reporting"
 	"github.com/replicatedhq/kots/pkg/store"
 	"github.com/replicatedhq/kots/pkg/util"
@@ -60,14 +61,14 @@ func GetLatestLicenseForHelm(licenseID string) (*LicenseData, error) {
 	return licenseData, nil
 }
 
-func getAppIdFromLicenseId(licenseID string) (string, error) {
-	apps, err := store.GetStore().ListInstalledApps()
+func getAppIdFromLicenseId(s store.Store, licenseID string) (string, error) {
+	apps, err := s.ListInstalledApps()
 	if err != nil {
 		return "", errors.Wrap(err, "failed to get all app licenses")
 	}
 
 	for _, a := range apps {
-		l, err := store.GetStore().GetLatestLicenseForApp(a.ID)
+		l, err := s.GetLatestLicenseForApp(a.ID)
 		if err != nil {
 			return "", errors.Wrap(err, "failed to get latest license for app")
 		}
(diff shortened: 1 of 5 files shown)


pkg/replicatedapp/api.go
```diff
@@ -67,6 +84,16 @@ func getLicenseFromAPI(url string, licenseID string) (*LicenseData, error) {
 
 	req.SetBasicAuth(licenseID, licenseID)
 
+	l, err := getLicenseById(licenseID)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get license by id")
+	}
+
+	if l != nil {
+		reportingInfo := reporting.GetReportingInfoByAppSlug(l.Spec.AppSlug)
```
divolgin, line 93: The app slug is generated and won't always match the one in the license.  I think you will need to use `ListInstalledApps` and then `GetLatestLicenseForApp` and then just call `GetReportingInfo` with the right app ID

FourSigma:
@sgalsaleh or @cbodonnell Would love you thoughts on how to globally mock a `store.Store` for failing unit test.   I can't find a good pathway to dependency injection neatly.  

Basically, the test is failing because of a call`store.GetStore()` which is `nil`.

pkg/replicatedapp/api.go
```diff
@@ -149,6 +180,14 @@ func getApplicationMetadataFromHost(host string, endpoint string, upstream *url.
 		return nil, errors.Wrap(err, "failed to call newrequest")
 	}
 
+	appID, err := store.GetStore().GetAppIDFromSlug(r.AppSlug)
```
cbodonnell, line 183: It looks like `store.GetStore()` is returning `nil` here as well based on the failing e2e jobs. This function is called in the context of the KOTS CLI, which is why it does not have direct access to the store.

cbodonnell:
> @sgalsaleh or @cbodonnell Would love you thoughts on how to globally mock a `store.Store` for failing unit test. I can't find a good pathway to dependency injection neatly.
> 
> Basically, the test is failing because of a call`store.GetStore()` which is `nil`.

In the SDK, we have [this function](https://github.com/replicatedhq/replicated-sdk/blob/main/pkg/store/store_interface.go#L41-L43), which allows setting the global store. Something like that would allow you to set and `defer` unset. In general, we've run into this issue a lot wrt the store dependency and it often requires a bit of refactoring to have functions accept the `store.Store` interface so we can unit test appropriately.

FourSigma:
> > @sgalsaleh or @cbodonnell Would love you thoughts on how to globally mock a `store.Store` for failing unit test. I can't find a good pathway to dependency injection neatly.
> > Basically, the test is failing because of a call`store.GetStore()` which is `nil`.
> 
> In the SDK, we have [this function](https://github.com/replicatedhq/replicated-sdk/blob/main/pkg/store/store_interface.go#L41-L43), which allows setting the global store. Something like that would allow you to set and `defer` unset. In general, we've run into this issue a lot wrt the store dependency and it often requires a bit of refactoring to have functions accept the `store.Store` interface so we can unit test appropriately.

Fixed this issue and the `ci-test` passes. 

However, a large number of integration tests are failing with the error `error connecting to rqlite: url specified is impossibly short`.   What the remediation for this particular error case?

pkg/replicatedapp/api.go
cbodonnell: So `store.GetStore()`, might not be returning `nil` here as I had thought above, but this would get called at a time where rqlite has not yet been deployed by KOTS and is the cause of the failing tests. Looking at the CI jobs, the `kots install` fails before any of the components are deployed. My suggestion would be to remove reporting from the metadata request here since it happens before a license is even installed. This initial metadata request is unauthenticated and is just used to apply branding to the initial screens.

cbodonnell:
> > > @sgalsaleh or @cbodonnell Would love you thoughts on how to globally mock a `store.Store` for failing unit test. I can't find a good pathway to dependency injection neatly.
> > > Basically, the test is failing because of a call`store.GetStore()` which is `nil`.
> > 
> > 
> > In the SDK, we have [this function](https://github.com/replicatedhq/replicated-sdk/blob/main/pkg/store/store_interface.go#L41-L43), which allows setting the global store. Something like that would allow you to set and `defer` unset. In general, we've run into this issue a lot wrt the store dependency and it often requires a bit of refactoring to have functions accept the `store.Store` interface so we can unit test appropriately.
> 
> @cbodonnell Fixed this issue and the `ci-test` passes.
> 
> However, a large number of integration tests are failing with the error `error connecting to rqlite: url specified is impossibly short`. What the remediation for this particular error case?

@FourSigma I just left a follow-up comment [here](https://github.com/replicatedhq/kots/pull/4202#discussion_r1428448828) with my thoughts. Hopefully that is all that is causing the tests to trip up since the panic originates from the call to `getApplicationMetadataFromHost ` ([link](https://github.com/replicatedhq/kots/actions/runs/7226186589/job/19692166086?pr=4202#step:6:170)).

FourSigma:
> > > > @sgalsaleh or @cbodonnell Would love you thoughts on how to globally mock a `store.Store` for failing unit test. I can't find a good pathway to dependency injection neatly.
> > > > Basically, the test is failing because of a call`store.GetStore()` which is `nil`.
> > > 
> > > 
> > > In the SDK, we have [this function](https://github.com/replicatedhq/replicated-sdk/blob/main/pkg/store/store_interface.go#L41-L43), which allows setting the global store. Something like that would allow you to set and `defer` unset. In general, we've run into this issue a lot wrt the store dependency and it often requires a bit of refactoring to have functions accept the `store.Store` interface so we can unit test appropriately.
> > 
> > 
> > @cbodonnell Fixed this issue and the `ci-test` passes.
> > However, a large number of integration tests are failing with the error `error connecting to rqlite: url specified is impossibly short`. What the remediation for this particular error case?
> 
> @FourSigma I just left a follow-up comment [here](https://github.com/replicatedhq/kots/pull/4202#discussion_r1428448828) with my thoughts. Hopefully that is all that is causing the tests to trip up since the panic originates from the call to `getApplicationMetadataFromHost ` ([link](https://github.com/replicatedhq/kots/actions/runs/7226186589/job/19692166086?pr=4202#step:6:170)).

 Makes sense @cbodonnell . I will make the change.

FourSigma:
@cbodonnell Happy Monday.  Looks like we have some failures on `validate-kots-pull`.  Since you are bit more familiar with this pathway, do you have some suggestions to get these test passing without too much refactoring?

cbodonnell:
> @cbodonnell Happy Monday. Looks like we have some failures on `validate-kots-pull`. Since you are bit more familiar with this pathway, do you have some suggestions to get these test passing without too much refactoring?

the `kots pull` command also runs from the CLI before any resources are deployed, so the underlying failure reason is the same as what we saw for `kots install`. if we need to avoid refactoring, one possibility could be to log (maybe just at the debug level) and move on without attempting to add headers if the database call fails. this way it can still function if we do have a database context available, but will not if it's running from the CLI only. thoughts?

FourSigma:
> > @cbodonnell Happy Monday. Looks like we have some failures on `validate-kots-pull`. Since you are bit more familiar with this pathway, do you have some suggestions to get these test passing without too much refactoring?
> 
> the `kots pull` command also runs from the CLI before any resources are deployed, so the underlying failure reason is the same as what we saw for `kots install`. if we need to avoid refactoring, one possibility could be to log (maybe just at the debug level) and move on without attempting to add headers if the database call fails. this way it can still function if we do have a database context available, but will not if it's running from the CLI only. thoughts?

I would be okay for this approach esp. if it just limited to the reporting portion of KOTS.

FourSigma:
@cbodonnell Looks like there is a `validate-gitops (kind, v1.28.0)` failure. Do you have some additional context on this test?

cbodonnell (approved):

cbodonnell:
> @cbodonnell Looks like there is a `validate-gitops (kind, v1.28.0)` failure. Do you have some additional context on this test?

@FourSigma It looks like the testim branch was out-of-date since that test was added somewhat recently. I've merged in changes from the base branch and re-run the test.

FourSigma:
@cbodonnell  Looks like we are now failing on the following tests:

`build-test / validate-smoke-test (aks, 1.25, alpha) (pull_request) `
`build-test / validate-backup-and-restore (eks, 1.27, stable) (pull_request)`

cbodonnell:
> @cbodonnell Looks like we are now failing on the following tests:
> 
> `build-test / validate-smoke-test (aks, 1.25, alpha) (pull_request) ` `build-test / validate-backup-and-restore (eks, 1.27, stable) (pull_request)`

@FourSigma those both looked to be flaky failures since a re-run passed. Merged this one.

FourSigma:
> > @cbodonnell Looks like we are now failing on the following tests:
> > `build-test / validate-smoke-test (aks, 1.25, alpha) (pull_request) ` `build-test / validate-backup-and-restore (eks, 1.27, stable) (pull_request)`
> 
> @FourSigma those both looked to be flaky failures since a re-run passed. Merged this one.

Thanks @cbodonnell!

PR #4222, opened 2023-12-14
Improve updating image deps workflow

sgalsaleh:

This changes the workflow for updating image tags for 3rd party images like minio, rqlite, and dex from a bottom-top approach (PR -> main -> Dockerhub -> regression tests -> release) to a top-down approach (Dockerhub -> PR -> main -> regression tests -> release), which helps avoid making unnecessary decisions in our CI for when the images need to be pushed or overwritten.

This PR also creates a daily job that:

- Retrieves the latest versions and epochs from the [Wolfi OSS packages](https://github.com/wolfi-dev/os/tree/main) for the minio, rqlite, and dex images.
- If the version or epoch is different, a new image will be built and pushed to our Dockerhub account.
- Runs the `imagedeps` script to retrieve the latest versions for all dependencies.
- Creates a PR if there are new versions available.
- Sends a Slack notification with the link to the PR.

The `imagedeps` scripts has been updated to retrieve the latest tags for the minio, rqlite, and dex images from our Dockerhub account.

Fixes [SC-94750](https://app.shortcut.com/replicated/story/94750/improve-the-workflow-for-updating-3rd-party-image-tags-like-minio-rqlite-and-dex)

```release-note
NONE
```

NONE


.github/workflows/image-deps-updater.yaml
```diff
@@ -0,0 +1,117 @@
...
+        with:
+          go-version: '^1.20.0'
+
+      - name: Run Update Script
+        env:
+          GITHUB_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          go run ./cmd/imagedeps
+
+      - name: Create Pull Request # creates a PR if there are differences
```
cbodonnell, line 85: any reason to keep the other one that runs in the release workflow?

cbodonnell (approved):

cbodonnell (approved):

PR #4239, opened 2023-12-19
Use rendered kots kinds

sgalsaleh:

Uses rendered KOTS kinds whenever they're available. This is a step forward to support templating for all KOTS kinds fields.

Fixes:

TBD

```release-note
TBD
```

TBD


pkg/apparchive/helm-v1beta2.go
```diff
@@ -159,16 +162,13 @@ func WriteV1Beta2HelmCharts(opts WriteV1Beta2HelmChartsOptions) error {
 			return errors.Wrap(err, "failed to process online images")
 		}
 
-		upstreamDir := opts.Upstream.GetUpstreamDir(opts.WriteUpstreamOptions)
```
cbodonnell, line 162: I think [this](https://github.com/replicatedhq/kots/pull/4239/files#diff-9bd889d7bad0a647ad15e3502d8a58d94cef268bbde15e348c874b4c770cb3fbR594) is the reason why we reloaded the installation file here? Want to make sure that this would not overwrite other changes that happened along the way.

pkg/pull/pull.go
```diff
@@ -600,16 +591,14 @@ func Pull(upstreamURI string, pullOptions PullOptions) (string, error) {
 		}
 	}
 
-	installationBytes, err := ioutil.ReadFile(filepath.Join(u.GetUpstreamDir(writeUpstreamOptions), "userdata", "installation.yaml"))
+	// installation spec gets updated during the process, ensure the map has the latest version
```
cbodonnell, line 594: at this point, would the `renderedKotsKinds.Installation` be the same as what's in `upstream/userdata/installation.yaml`? just curious why we need to reload it from the archive vs marshaling what we have in `renderedKotsKinds`

pkg/pull/pull.go
```diff
@@ -310,7 +310,31 @@ func Pull(upstreamURI string, pullOptions PullOptions) (string, error) {
...
+		log.FinishSpinnerWithError()
+		return "", errors.Wrap(err, "failed to render upstream")
+	}
+
+	renderedKotsKinds, err := kotsutil.KotsKindsFromMap(renderedKotsKindsMap)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to load rendered kotskinds from map")
+	}
+
+	needsConfig, err := kotsadmconfig.NeedsConfiguration(pullOptions.AppSlug, pullOptions.AppSequence, pullOptions.AirgapRoot != "", renderedKotsKinds, registrySettings)
```
cbodonnell, line 337: I believe this happened before attempting to render base because of things like this https://app.testim.io/#/project/wpYAooUimFDgQxY73r17/branch/use-rendered-kots-kinds/test/8lupz8KBYa9SGzDO/step/mFwAJVtdRncnSncN/viewer/screenshots?result-id=QSkV5cWC3frZ0dZI&path=s2IRYvVg6O8W5IWX%3AmFwAJVtdRncnSncN

pkg/pull/pull.go
sgalsaleh: they would be the same, yes. it didn't really matter which one to use. i decided to load it from the upstream directory just to make sure it's definitely the same, and just in case someone forgets to update the `renderedKotsKinds` object in a later change.

pkg/base/replicated.go
```diff
@@ -444,7 +439,7 @@ func tryGetConfigFromFileContent(content []byte, log *logger.CLILogger) *kotsv1b
 	return nil
 }
 
-func getKotsKinds(u *upstreamtypes.Upstream) (*kotsutil.KotsKinds, error) {
+func getTemplatingKotsKinds(u *upstreamtypes.Upstream) (*kotsutil.KotsKinds, error) {
```
cbodonnell, line 442: might be good to add a comment for this indicating what "templating kots kinds" is. assuming it's the kots kinds that are needed to render template functions?

sgalsaleh: yes, and correct.

cbodonnell (approved):

cbodonnell (dismissed):

cbodonnell (approved):

PR #4340, opened 2024-01-10
Ensure that KOTS resources only get scheduled on amd64 nodes

sgalsaleh:

Ensures that KOTS resources only get scheduled on amd64 nodes.

Fixes #

```release-note
Fixes an issue where Pods associated with KOTS components could be incorrectly scheduled on a non-amd64 node.
```


start-time
cbodonnell, line 1: looks like this snuck in...

cbodonnell (approved):

PR #4344, opened 2024-01-11
improvement: set acceptAnonymousUploads=1 for initial kotsadm-tls cert

banjoh:

Have `kurl-proxy` service set `acceptAnonymousUploads=1` in the `kotsadm-tls` cert by default when we generate the default self signed certificate. This allows prompting Admin Console users to upload BYO certs on first for first time installs. When KOTS is installed as a kURL addon, the [installer script sets this flag to 1](https://github.com/replicatedhq/kots/blob/650c91516e2dd6cf7668c71da42f6cb5ba16333d/deploy/kurl/kotsadm/template/base/install.sh#L424). Doing so in this service aligns with what's expected.

Fixes # [sc-94724](https://app.shortcut.com/replicated/story/94724)

```release-note

```

#### Review notes
-  I added some TODOs of some changes to consider. They are out of scope in this PR but would like your thoughts on them since the code was written a while back.

### DIFF
--- a/kurl_proxy/cmd/main.go
+++ b/kurl_proxy/cmd/main.go
@@ -8,9 +8,8 @@ import (
 	"encoding/pem"
 	"fmt"
 	"html/template"
-	"io/ioutil"
+	"io"
 	"log"
-	"math/rand"
 	"net"
 	"net/http"
 	"net/http/httputil"
@@ -53,8 +52,6 @@ type cert struct {
 func main() {
 	log.Printf("Commit %s\n", os.Getenv("COMMIT"))
 
-	rand.Seed(time.Now().UnixNano())
-
 	upstreamOrigin := os.Getenv("UPSTREAM_ORIGIN")
 	dexUpstreamOrigin := os.Getenv("DEX_UPSTREAM_ORIGIN")
 	tlsSecretName := os.Getenv("TLS_SECRET_NAME")
@@ -86,16 +83,18 @@ func main() {
 	if err != nil {
 		log.Panic(err)
 	}
+	// TODO: Assert namespace not empty, else we get secrets from all namespaces
 	secrets := clientset.CoreV1().Secrets(namespace)
 
 	_, err = secrets.Get(context.Background(), tlsSecretName, metav1.GetOptions{})
 	if err != nil {
-		log.Print("creating tls secret")
+		log.Print("creating default tls secret")
 
 		err = generateDefaultCertSecret(secrets, namespace)
 		if err != nil {
 			log.Printf("Could not regenerate default certificate: %v", err)
 		}
+		// TODO: Why are we exiting here? It leads to a pod restart
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


kurl_proxy/cmd/main.go
```diff
@@ -53,8 +52,6 @@ type cert struct {
 func main() {
 	log.Printf("Commit %s\n", os.Getenv("COMMIT"))
 
-	rand.Seed(time.Now().UnixNano())
```
marccampbell, line 56: Why remove this?

banjoh: > Why remove this?

Accidental clean up as I was removing deprecated code. I meant to put it back.

We however do not need it. It was deprecated in go1.20. The runtime automatically generates a seed for us. Also, I do not see anywhere where we use random numbers.

kurl_proxy/cmd/main.go
```diff
@@ -621,14 +620,17 @@ func generateDefaultCertSecret(secrets corev1.SecretInterface, namespace string)
...
 			},
 		},
 		Type:       "kubernetes.io/tls",
 		Data:       make(map[string][]byte),
 		StringData: make(map[string]string),
 	}
 
+	// TODO: Why is the namespace always "default"?
+	// Requests to kotsadm.embeddded-cluster.svc.cluster for example will
+	// fail with invalid cert errors
```
cbodonnell, line 633: Not sure about the historical reasons, but kots and kurl-proxy are always deployed to the default namespace with the kurl add-on. If it needs to be other namespaces, i think we could make it configurable. Maybe even use the `namespace` variable we already have in the main function that comes from [here](https://github.com/replicatedhq/kots/blob/main/deploy/kurl/kotsadm/template/base/kurl-proxy/tmpl-deployment.yaml#L33-L36)?

kurl_proxy/cmd/main.go
```diff
@@ -86,16 +83,18 @@ func main() {
...
 	_, err = secrets.Get(context.Background(), tlsSecretName, metav1.GetOptions{})
 	if err != nil {
-		log.Print("creating tls secret")
+		log.Print("creating default tls secret")
 
 		err = generateDefaultCertSecret(secrets, namespace)
 		if err != nil {
 			log.Printf("Could not regenerate default certificate: %v", err)
 		}
+		// TODO: Why are we exiting here? It leads to a pod restart
```
cbodonnell, line 97: not sure. on the surface, it seems like we dont _need_ a restart for anything

banjoh: I'll leave that behaviour for now, but leave the TODO there so as not to introduce some unwanted behaviour change.

kurl_proxy/cmd/main.go
banjoh: I left the hostname as-is, but I have added DNS alt names to the cert in case the namespace is not default. This will keep any old assumptions valid such as in [ekco's rotation logic](https://github.com/replicatedhq/ekco/blob/f879a259ecff9e25ac7a3685561aa39e2ec3b61a/pkg/cluster/kotsadm.go#L57)

kurl_proxy/cmd/main.go
banjoh: I think we are safe to remove this unless there are objections.

I reviewed the code and do not see any reason for keeping it

cbodonnell (approved):

PR #4362, opened 2024-01-15
bug: do not fail to install if no embedded cluster installation

ricardomaraschini:

if no embedded cluster installation object has been found in the cluster we can't start an upgrade. on this case (where the installation does not exists) we shouldn't be failing but moving on instead.

during the initial cluster bootstrap the installation object may take a while to be created therefore a scenario where the kots app is installed but no installation object exists is valid.

### DIFF
--- a/pkg/embeddedcluster/monitor.go
+++ b/pkg/embeddedcluster/monitor.go
@@ -2,6 +2,7 @@ package embeddedcluster
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


pkg/embeddedcluster/monitor.go
```diff
@@ -33,6 +35,12 @@ func MaybeStartClusterUpgrade(ctx context.Context, client kubernetes.Interface,
 
 	spec := conf.Spec
 	if upgrade, err := RequiresUpgrade(ctx, spec); err != nil {
+		// if there is no installation object we can't start an upgrade. this is a valid
+		// scenario specially during cluster bootstrap. as we do not need to upgrade the
+		// cluster just after its installation we can return nil here.
```
laverya, line 40:
```suggestion
		// cluster just after its installation we can return nil here.
		// (the cluster in the first kots version will match the cluster installed during bootstrap)
```

laverya (approved):

laverya (approved):

PR #4394, opened 2024-01-23
Arm64 support for KOTS images

sgalsaleh:

Adds support for running KOTS components on Linux Arm64 based clusters / nodes.

```release-note
* Adds support for running KOTS components on Linux Arm64 based clusters / nodes.
```


pkg/operator/client/deploy.go
```diff
@@ -521,7 +518,7 @@ func (c *Client) uninstallWithHelm(v1Beta1ChartsDir, v1Beta2ChartsDir string, ko
 		}
 
 		logger.Infof("running helm with arguments %v", args)
-		cmd := exec.Command(fmt.Sprintf("helm%s", version), args...)
```
divolgin, line 524: is this because v3 is the default and you just want to use the default?

pkg/imageutil/image.go
```diff
@@ -281,3 +281,27 @@ func KustomizeImage(destRegistry registrytypes.RegistryOptions, image string) ([
 	}
 	return rewrittenImages, nil
 }
+
+func ChangeImageTag(image string, newTag string) (string, error) {
```
divolgin, line 285: Could use a comment about what this function does

divolgin (dismissed):

divolgin (approved):

pkg/operator/client/deploy.go
sgalsaleh: it's because we removed the helm 2 binary a long time ago but didn't refactor this code.

PR #4398, opened 2024-01-25
embedded cluster upgrade modal

cbodonnell:

This PR improves the KOTS UI so that when it loses connection to the API during a cluster upgrade, a separate dialog is shown indicating to the user that the cluster is being updated and that connectivity will be restored once it completes.

https://www.loom.com/share/29a2622f3cec4276aac30b5670dfc35a?sid=b0460ba2-26c5-4cf1-8ed5-dce842a7563f

Screen Shot 2024-01-25 at 11 36 01 AM

https://app.shortcut.com/replicated/story/97393/detect-cluster-update-and-show-a-better-message-when-the-admin-console-goes-down

## Steps to reproduce

1. Install a KOTS app
2. Upgrade to a newer version that includes an embedded cluster upgrade (k0s version bump)

```release-note

```

### DIFF
--- a/pkg/api/handlers/types/types.go
+++ b/pkg/api/handlers/types/types.go
@@ -91,6 +91,11 @@ type ResponseGitOps struct {
 type ResponseCluster struct {
 	ID   string `json:"id"`
 	Slug string `json:"slug"`
+	// RequiresUpgrade represents whether the embedded cluster config for the current app
+	// version is different from the currently deployed embedded cluster config
+	RequiresUpgrade bool `json:"requiresUpgrade"`
+	// State represents the current state of the most recently deployed embedded cluster config
+	State string `json:"state,omitempty"`
 }
 
 type GetPendingAppResponse struct {
--- a/web/src/components/clusters/EmbeddedClusterUpgrading.tsx
+++ b/web/src/components/clusters/EmbeddedClusterUpgrading.tsx
@@ -1,0 +1,91 @@
+import { useEffect, useReducer } from "react";
+import fetch from "../../utilities/fetchWithTimeout";
+import { Utilities } from "../../utilities/utilities";
+import Loader from "@components/shared/Loader";
+
+interface Props {
+  setTerminatedState: (terminated: boolean) => void;
+}
+
+interface State {
+  seconds: number;
+  reconnectAttempts: number;
+}
+
+const EmbeddedClusterUpgrading = (props: Props) => {
+  const [state, setState] = useReducer(
+    (currentState: State, newState: Partial<State>) => ({
+      ...currentState,
+      ...newState,
+    }),
+    {
+      seconds: 1,
+      reconnectAttempts: 1,
+    }
+  );
+
+  let countdown: (seconds: number) => void;
+  let ping: () => Promise<void>;
+
+  countdown = (seconds: number) => {
+    setState({ seconds });
+    if (seconds === 0) {
+      setState({
+        reconnectAttempts: state.reconnectAttempts + 1,
+      });
+      ping();
+    } else {
+      const nextCount = seconds - 1;
+      setTimeout(() => {
+        countdown(nextCount);
+      }, 1000);
+    }
+  };
+
+  ping = async () => {
+    const { reconnectAttempts } = state;
+    await fetch(
+      `${process.env.API_ENDPOINT}/ping`,
+      {
+        headers: {
+          "Content-Type": "application/json",
+        },
+        credentials: "include",
+      },
+      10000
+    )
+      .then(async (res) => {
+        if (res.status === 401) {
+          Utilities.logoutUser();
+          return;
+        }
+        props.setTerminatedState(false);
+      })
+      .catch(() => {
+        props.setTerminatedState(true);
+        const seconds = reconnectAttempts > 10 ? 10 : reconnectAttempts;
+        countdown(seconds);
+      });
+  };
+
+  useEffect(() => {
+    ping();
+  }, []);
+
+  return (
+    <div className="Modal-body u-textAlign--center">
+      <div className="flex u-marginTop--30 u-marginBottom--10 justifyContent--center">
+        <Loader size="60" />
+      </div>
+      <h2 className="u-fontSize--largest u-textColor--primary u-fontWeight--bold u-lineHeight--normal u-userSelect--none">
+        Cluster update in progress
+      </h2>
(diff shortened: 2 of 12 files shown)


pkg/api/handlers/types/types.go
```diff
@@ -91,6 +91,11 @@ type ResponseGitOps struct {
 type ResponseCluster struct {
 	ID   string `json:"id"`
 	Slug string `json:"slug"`
+	// RequiresUpgrade represents whether the embedded cluster config for the current app
+	// version is different from the currently deployed embedded cluster config
+	RequiresUpgrade bool `json:"requiresUpgrade"`
+	// State represents the current state of the most recently deployed embedded cluster config
+	State string `json:"state,omitempty"`
```
cbodonnell, line 98: is there a more appropriate place for these response fields? `cluster` seems like an OK spot, but i know it's not necessarily meant to be for the "embedded cluster" but rather the "kots downstream cluster"

web/src/utilities/utilities.js
```diff
@@ -636,6 +636,37 @@ export const Utilities = {
...
   },
 
+  isClusterUpgrading(state) {
+    const normalizedState = this.clusterState(state);
+    return (
+      normalizedState === "Upgrading" || normalizedState === "Upgrading addons"
+    );
+  },
+
+  showClusterUpgradeModal(apps) {
```
sgalsaleh, line 646: this name indicates an action will be taken, how about:
```suggestion
  shouldShowClusterUpgradeModal(apps) {
```

cbodonnell:
@alicenstar tagging you on this PR to see if there's anything on the front-end that you may have feedback on.

alicenstar:
@cbodonnell Would you be able to take a Loom showing the full functionality (what the connection error looks like and how the modal pops up)? I am reviewing the front end code now, but it would be easier if I could also see how it functions in the browser.

cbodonnell:
> @cbodonnell Would you be able to take a Loom showing the full functionality (what the connection error looks like and how the modal pops up)? I am reviewing the front end code now, but it would be easier if I could also see how it functions in the browser.

@alicenstar Added a loom to the PR description. The new modal is meant to be shown in place of this existing modal, which is displayed when the UI loses connectivity to the API. We instead want to show a nicer message since this downtime is expected during an embedded cluster upgrade.

Screen Shot 2024-01-25 at 2 15 26 PM

alicenstar:
@cbodonnell Thank you, that context is helpful! I'm in refinement right now so my review got a bit paused, but it should be over shortly. I do see an interesting pattern in `EmbeddedClusterUpgrading` that I'd like to examine further.

web/src/components/clusters/EmbeddedClusterUpgrading.tsx
```diff
@@ -0,0 +1,91 @@
...
+      ...newState,
+    }),
+    {
+      seconds: 1,
+      reconnectAttempts: 1,
+    }
+  );
+
+  let countdown: (seconds: number) => void;
+  let ping: () => Promise<void>;
```
alicenstar, line 28: Why are we initializing these functions like this vs `const countdown = () => { .... }`?

web/src/components/clusters/EmbeddedClusterUpgrading.tsx
cbodonnell, line 1: @alicenstar want to get your thoughts on something in here. This component is pretty much the same as https://github.com/replicatedhq/kots/blob/main/web/src/ConnectionTerminated.jsx, but I went ahead and tried to convert it to a function component, but keep the logic as-is. One issue I ran into was the circular dependency between `countdown` and `ping`, which threw a linter error in the function component only, but does also exist in the class component. Defining those variables before and then assigning got around this for the function component, but I'm not sure of there's a better practice. Thoughts?

alicenstar (approved):
Love to see the front end tests. Looks good - I had just one comment on an odd way of declaring a function in `EmbeddedClusterUpgrading` if you could take a look.

web/src/components/clusters/EmbeddedClusterUpgrading.tsx
alicenstar: Ah, taking a look now.

alicenstar: @cbodonnell We're not showing the countdown anymore in this modal, so I think the logic in this file can be greatly simplified by replacing the countdown and ping logic with something like this (this is untested, but I think we can just ping on an interval and clear the interval on unmount):
```
import fetch from "node-fetch";
import { useEffect } from "react";
import { Utilities } from "../../utilities/utilities";
import Loader from "@components/shared/Loader";

interface Props {
  setTerminatedState: (terminated: boolean) => void;
}

const EmbeddedClusterUpgrading = (props: Props) => {
  const ping = async () => {
    await fetch(
      `${process.env.API_ENDPOINT}/ping`,
      {
        headers: {
          "Content-Type": "application/json",
        },
        credentials: "include",
      })
      .then(async (res) => {
        if (res.status === 401) {
          Utilities.logoutUser();
          return;
        }
        props.setTerminatedState(false);
      })
      .catch(() => {
        props.setTerminatedState(true);
      });
  };

  useEffect(() => {
    const interval = setInterval(() => {
      ping();
    }, 10000);
    return () => clearInterval(interval);
  }, []);

  return (
    <div className="Modal-body u-textAlign--center">
      <div className="flex u-marginTop--30 u-marginBottom--10 justifyContent--center">
        <Loader size="60" />
      </div>
      <h2 className="u-fontSize--largest u-textColor--primary u-fontWeight--bold u-lineHeight--normal u-userSelect--none">
        Cluster update in progress
      </h2>
      <p className="u-fontSize--normal u-fontWeight--medium u-textColor--bodyCopy u-lineHeight--more u-marginTop--10 u-marginBottom--10 u-userSelect--none">
        The API cannot be reached because the cluster is updating. Stay on this
        page to automatically reconnect when the update is complete.
      </p>
    </div>
  );
};

export default EmbeddedClusterUpgrading;
```
Some notes:
- I am using `node-fetch`, not the fetch with timeout utility.
- We no longer need the seconds/attempts state since it looks like it was all used to calculate the countdown timer.

cbodonnell: nice, thanks for this! will give it a try and let you know...

cbodonnell: works great 👍

alicenstar (approved):
Looks great, Craig!

sgalsaleh (dismissed):

sgalsaleh (approved):

PR #4402, opened 2024-01-26
show initial preflight page for helm chart preflights

cbodonnell:

This PR fixes an issue where the initial install flow preflight page would not be displayed if the application's preflight spec was included in a Secret or ConfigMap within the Helm chart templates.

Key Changes:
* Backend - We were not parsing preflights out of the rendered templates when populating the rendered `kotsKinds` directory for a given app version. We do this on-the-fly when we run preflights, which is why the checks would still run, but not having these saved as part of the kotsKinds in the archive meant that the `preflight_spec` field in the database would not be set, us returning `hasPreflight: false` back to the frontend.
* Frontend - The application details are now re-fetched after the configuration is saved successfully. This is necessary since the `hasPreflight` field may not be set until the configuration is saved and the version is rendered. This would be the case for a Helm chart that has required configuration, but no default values.

https://app.shortcut.com/replicated/story/96941/preflight-page-not-displayed-in-kots-admin-console-install-flow-for-preflight-specs-in-helm-charts

```release-note
Fixes an issue where the initial install flow preflight page would not be displayed if the application's preflight spec was included in a Secret or ConfigMap within the Helm chart templates
```

NONE


pkg/pull/pull.go
```diff
@@ -604,6 +602,31 @@ func Pull(upstreamURI string, pullOptions PullOptions) (string, error) {
...
+		if renderedPreflight != nil {
+			renderedPreflightBytes, err := kotsutil.MarshalRuntimeObject(renderedPreflight)
+			if err != nil {
+				return "", errors.Wrap(err, "failed to marshal rendered preflight")
+			}
+			renderedKotsKindsMap["helm-preflight.yaml"] = renderedPreflightBytes
+		}
+	}
+
+	if err := kotsutil.WriteKotsKinds(renderedKotsKindsMap, u.GetKotsKindsDir(writeUpstreamOptions)); err != nil {
```
sgalsaleh, line 626: nit pick, can you move this code block to be right before this line? just want to make sure the installation object gets updated with the latest information immediately before it's written to `kotsKinds`, that way it's less prone to error if something updates the installation object but forgets to update it here.
```
// installation spec gets updated during the process, ensure the map has the latest version
	installationBytes, err := os.ReadFile(filepath.Join(u.GetUpstreamDir(writeUpstreamOptions), "userdata", "installation.yaml"))
if err != nil {
	return "", errors.Wrap(err, "failed to read installation file")
}
renderedKotsKindsMap["userdata/installation.yaml"] = []byte(installationBytes)
```

pkg/rewrite/rewrite.go
```diff
@@ -339,6 +337,31 @@ func Rewrite(rewriteOptions RewriteOptions) error {
...
+		if renderedPreflight != nil {
+			renderedPreflightBytes, err := kotsutil.MarshalRuntimeObject(renderedPreflight)
+			if err != nil {
+				return errors.Wrap(err, "failed to marshal rendered preflight")
+			}
+			renderedKotsKindsMap["helm-preflight.yaml"] = renderedPreflightBytes
+		}
+	}
+
+	if err := kotsutil.WriteKotsKinds(renderedKotsKindsMap, u.GetKotsKindsDir(writeUpstreamOptions)); err != nil {
```
sgalsaleh, line 361: ditto

sgalsaleh (approved):

PR #4437, opened 2024-02-08
embedded cluster management UI cleanup

cbodonnell:

This PR includes a bunch of smaller UI changes for embedded cluster management:
* Cluster management tab in navbar stays underlined when navigating to node details
* Change "Cluster Nodes" title to just "Nodes"
* Remove the "This page lists..." subtitle from the nodes page
* Grid search fixed for node name and roles
* Adds dividers for the grid headers
* Updates copy on add node modal and removes "Command expires:"
* On the add node modal, i there's only one role defined, conditionally remove the sentence "Select one or more roles to assign to the new node.", don't show the role, and show the join command automatically
* Remove kubelet/kube-proxy/kernel version information from node details page

**Loom** detailing changes: https://www.loom.com/share/2563b05afeb34d70811088b54da11e7a?sid=3f317152-0873-456f-83b8-ee1c31cd90ce

https://app.shortcut.com/replicated/story/97020/clean-up-on-cluster-management-page

```release-note

```

### DIFF
--- a/web/src/components/apps/EmbeddedClusterManagement.tsx
+++ b/web/src/components/apps/EmbeddedClusterManagement.tsx
@@ -467,8 +467,10 @@ const EmbeddedClusterManagement = ({
             />
           </div>
           <p className="tw-text-base tw-text-gray-600">
-            Select one or more roles to assign to the new node. Copy the join
-            command and run it on the machine you'd like to join to the cluster.
+            {(rolesData?.roles || NODE_TYPES).length > 1 &&
+              "Select one or more roles to assign to the new node. "}
+            Copy the join command and run it on the machine you'd like to join
+            to the cluster.
           </p>
           {rolesLoading && (
             <p className="tw-text-base tw-w-full tw-text-center tw-py-4 tw-text-gray-500 tw-font-semibold">
--- a/web/src/components/apps/EmbeddedClusterManagement.tsx
+++ b/web/src/components/apps/EmbeddedClusterManagement.tsx
@@ -238,6 +238,8 @@ const EmbeddedClusterManagement = ({
     if (nodeTypes.length === 1) {
       // if there's only one node type, select it by default
       setSelectedNodeTypes(nodeTypes);
+    } else {
+      setSelectedNodeTypes([]);
     }
   }, [rolesData]);
 
--- a/web/src/components/apps/EmbeddedClusterManagement.tsx
+++ b/web/src/components/apps/EmbeddedClusterManagement.tsx
@@ -1,7 +1,7 @@
 import { useQuery } from "@tanstack/react-query";
 import classNames from "classnames";
 import MaterialReactTable, { MRT_ColumnDef } from "material-react-table";
-import { ChangeEvent, useMemo, useReducer, useState } from "react";
+import { ChangeEvent, useEffect, useMemo, useReducer, useState } from "react";
 import Modal from "react-modal";
 import { Link, useParams } from "react-router-dom";
 
@@ -233,6 +233,16 @@ const EmbeddedClusterManagement = ({
   // #region node type logic
   const NODE_TYPES = ["controller"];
 
+  useEffect(() => {
+    const nodeTypes = rolesData?.roles || NODE_TYPES;
+    if (nodeTypes.length === 1) {
+      // if there's only one node type, select it by default
+      setSelectedNodeTypes(nodeTypes);
+    } else {
+      setSelectedNodeTypes([]);
+    }
+  }, [rolesData]);
+
   const determineDisabledState = () => {
     return false;
   };
@@ -250,13 +260,11 @@ const EmbeddedClusterManagement = ({
   // #endregion
 
   type NodeColumns = {
-    name: string | JSX.Element;
-    roles: JSX.Element;
+    name: string;
+    roles: string;
     status: string;
     cpu: string;
     memory: string;
-    pause: JSX.Element;
-    delete: JSX.Element;
   };
 
   const columns = useMemo<MRT_ColumnDef<NodeColumns>[]>(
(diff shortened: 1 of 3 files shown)


cbodonnell:
@alicenstar I pushed up two additional commits since recording the loom after some product feedback:
* https://github.com/replicatedhq/kots/pull/4437/commits/42339640a66fc08cc8262a36a844f9913f3a079f conditionally shows/hides some text based whether there's more than one role
* https://github.com/replicatedhq/kots/pull/4437/commits/1cb5efc1f25698f7cc9a1ac24fd052253b9b2767 removes the node version details and adjusts the styling (screenshot below)
Screen Shot 2024-02-09 at 10 04 43 AM

web/src/components/apps/EmbeddedClusterManagement.tsx
```diff
@@ -233,6 +233,16 @@ const EmbeddedClusterManagement = ({
   // #region node type logic
   const NODE_TYPES = ["controller"];
 
+  useEffect(() => {
+    const nodeTypes = rolesData?.roles || NODE_TYPES;
+    if (nodeTypes.length === 1) {
+      // if there's only one node type, select it by default
+      setSelectedNodeTypes(nodeTypes);
+    } else {
+      setSelectedNodeTypes([]);
```
alicenstar, line 242: Since `selectedNodeTypes` is already initialized to an empty array, is this needed?

web/src/components/apps/EmbeddedClusterManagement.tsx
```diff
@@ -267,11 +275,40 @@ const EmbeddedClusterManagement = ({
         enableHiding: false,
         enableColumnDragging: false,
         size: 150,
+        Cell: ({ cell }) => {
+          const value = cell.getValue<string>();
+          return (
+            <Link
+              to={slug ? `/${slug}/cluster/${value}` : `/cluster/${value}`}
```
alicenstar, line 282: Can you tell me more about these two URLs and what is different on each of those pages (what happens on that page when there's a slug in the URL vs not)?

web/src/components/apps/EmbeddedClusterManagement.tsx
```diff
@@ -294,43 +331,15 @@ const EmbeddedClusterManagement = ({
           align: "right",
         },
       },
-      // {
```
alicenstar, line 297: Nice clean up & attention to detail 😊

web/src/components/apps/EmbeddedClusterManagement.tsx
```diff
@@ -457,50 +467,51 @@ const EmbeddedClusterManagement = ({
             />
           </div>
           <p className="tw-text-base tw-text-gray-600">
-            To add a node to this cluster, select the type of node you'd like to
-            add. Once you've selected a node type, we will generate a node join
-            command for you to use in the CLI. When the node successfully joins
-            the cluster, you will see it appear in the list of nodes on this
-            page.
+            {(rolesData?.roles || NODE_TYPES).length > 1 &&
```
alicenstar, line 470: Won't `NODE_TYPES` always have a length of 1? If so, I think this should just be checking `rolesData?.roles`.

web/src/components/apps/EmbeddedClusterManagement.tsx
```diff
@@ -457,50 +467,51 @@ const EmbeddedClusterManagement = ({
...
-                  htmlFor={`${nodeType}NodeType`}
-                  className="tw-block u-cursor--pointer u-userSelect--none u-textColor--primary u-fontSize--normal u-fontWeight--medium tw-text-center"
+          {rolesLoading && (
+            <p className="tw-text-base tw-w-full tw-text-center tw-py-4 tw-text-gray-500 tw-font-semibold">
+              Loading roles...
+            </p>
+          )}
+          {!rolesData && rolesError && (
+            <p className="tw-text-base tw-w-full tw-text-center tw-py-4 tw-text-pink-500 tw-font-semibold">
+              {rolesError?.message}
```
alicenstar, line 482: I think it'd be good to add a fallback error message here if the API doesn't return one for some reason. I'm thinking of something along these lines:
`{rolesError?.message || "Unable to fetch roles"}`

web/src/components/apps/EmbeddedClusterManagement.tsx
```diff
@@ -457,50 +467,51 @@ const EmbeddedClusterManagement = ({
...
+            </p>
+          )}
+          {(rolesData?.roles || NODE_TYPES).length > 1 && (
```
alicenstar, line 485: Same comment as above where `NODE_TYPES` will never have a length longer than 1, so we should just check `rolesData?.roles`.

alicenstar (commented):
Looks great, Craig! I just had a couple small comments/questions about some logic, nothing big.

web/src/components/apps/EmbeddedClusterManagement.tsx
cbodonnell: my thought was if the `rolesData` was refetched and contained more than one role, then we would want to deselect. though it doesn't look like we are refetching this data ever, so i do think we can remove that

web/src/components/apps/EmbeddedClusterManagement.tsx
cbodonnell: yep, good call. will definitely make that more readable too

web/src/components/apps/EmbeddedClusterManagement.tsx
alicenstar: That's fair. If you think we aren't refetching this data ever, I'd say it's safe to remove, but I also understand that logic.

cbodonnell: i'll go ahead and remove it for now as it seems we just fetch it when the component loads

cbodonnell: it seems that removing this may actually be causing an issue since `rolesData` is initially `undefined`, so we fall back to `NODE_TYPES` in that case. since that only has one role, we select it by default. then once `rolesData` is fetched, we don't deselect and still have that initial role from `NODE_TYPES` selected. going to add that `else` back in

web/src/components/apps/EmbeddedClusterManagement.tsx
```diff
@@ -479,12 +480,12 @@ const EmbeddedClusterManagement = ({
           )}
           {!rolesData && rolesError && (
             <p className="tw-text-base tw-w-full tw-text-center tw-py-4 tw-text-pink-500 tw-font-semibold">
-              {rolesError?.message}
+              {rolesError?.message || "Unable to fetch roles"}
             </p>
           )}
-          {(rolesData?.roles || NODE_TYPES).length > 1 && (
+          {rolesData?.roles && rolesData?.roles.length > 1 && (
```
alicenstar, line 486: nit: you don't need the `?` in the second condition because you've already verified it exists in the first condition.

cbodonnell: ah good catch :)

fixed that in a couple of places in https://github.com/replicatedhq/kots/pull/4437/commits/24cc3e871dbe11cbfeefa1fd5de24e02106fe505

alicenstar (approved):
Amazing, great work!

PR #4447, opened 2024-02-15
Report embedded cluster version

sgalsaleh:

Includes the embedded cluster version in the reporting information.

Fixes #

```release-note
NONE
```

NONE

### DIFF
--- a/migrations/tables/instance-report.yaml
+++ b/migrations/tables/instance-report.yaml
@@ -37,6 +37,8 @@ spec:
         type: text
       - name: embedded_cluster_id
         type: text
+      - name: embedded_cluster_version
+        type: text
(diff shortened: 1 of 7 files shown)


migrations/tables/instance-report.yaml
```diff
@@ -37,6 +37,8 @@ spec:
         type: text
       - name: embedded_cluster_id
         type: text
+      - name: embedded_cluster_version
+        type: text
```
cbodonnell, line 41: i think this table is not used. won't hurt to add this, but the table is generally outdated now (some missing datapoints). should we consider removing it?

cbodonnell (approved):

PR #4461, opened 2024-02-21
add pending_cluster_management status for embedded clusters

cbodonnell:

This PR adds a new version status `pending_cluster_management` for embedded clusters to represent the state during an initial install when the user must confirm their cluster management settings. All embedded cluster installs will have the initial version set to this state prior to completing the cluster management step. Once that step is completed, the version status will either be set to `pending_config`, `pending_preflight`, or `pending`, depending on whether or not there is any app configuration or preflights defined.

This fixes an issue where an application without required configuration would automatically deploy in embedded clusters, which was not the desired behavior.

https://app.shortcut.com/replicated/story/99681/disable-autodeploy

```release-note

```

### DIFF
--- a/pkg/handlers/embedded_cluster_confirm_cluster_management.go
+++ b/pkg/handlers/embedded_cluster_confirm_cluster_management.go
@@ -1,0 +1,95 @@
+package handlers
+
+import (
+	"fmt"
+	"net/http"
+	"os"
+
+	"github.com/pkg/errors"
+	"github.com/replicatedhq/kots/pkg/kotsutil"
+	"github.com/replicatedhq/kots/pkg/logger"
+	"github.com/replicatedhq/kots/pkg/preflight"
+	"github.com/replicatedhq/kots/pkg/store"
+	storetypes "github.com/replicatedhq/kots/pkg/store/types"
+)
+
+type ConfirmEmbeddedClusterManagementResponse struct {
+	VersionStatus string `json:"versionStatus"`
+}
+
+func (h *Handler) ConfirmEmbeddedClusterManagement(w http.ResponseWriter, r *http.Request) {
+	apps, err := store.GetStore().ListInstalledApps()
+	if err != nil {
+		logger.Error(fmt.Errorf("failed to list installed apps: %w", err))
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	if len(apps) == 0 {
+		logger.Error(fmt.Errorf("no installed apps found"))
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	app := apps[0]
+
+	downstreamVersions, err := store.GetStore().FindDownstreamVersions(app.ID, true)
+	if err != nil {
+		logger.Error(fmt.Errorf("failed to find downstream versions: %w", err))
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	if len(downstreamVersions.PendingVersions) == 0 {
+		logger.Error(fmt.Errorf("no pending versions found"))
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	pendingVersion := downstreamVersions.PendingVersions[0]
+
+	if pendingVersion.Status == storetypes.VersionPendingClusterManagement {
+		archiveDir, err := os.MkdirTemp("", "kotsadm")
+		if err != nil {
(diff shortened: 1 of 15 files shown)


pkg/handlers/embedded_cluster_confirm_cluster_management.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/replicatedhq/kots/pkg/preflight"
+	"github.com/replicatedhq/kots/pkg/store"
+	storetypes "github.com/replicatedhq/kots/pkg/store/types"
+)
+
+type ConfirmEmbeddedClusterManagementResponse struct {
+	VersionStatus string `json:"versionStatus"`
+}
+
+func (h *Handler) ConfirmEmbeddedClusterManagement(w http.ResponseWriter, r *http.Request) {
```
sgalsaleh, line 20: this function should return 400 if it's not an embedded cluster

cbodonnell: @sgalsaleh this seems very reasonable. should i add this for all embedded cluster handlers?

pkg/handlers/embedded_cluster_confirm_cluster_management.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	if len(downstreamVersions.PendingVersions) == 0 {
+		logger.Error(fmt.Errorf("no pending versions found"))
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	pendingVersion := downstreamVersions.PendingVersions[0]
+
+	if pendingVersion.Status == storetypes.VersionPendingClusterManagement {
```
sgalsaleh, line 49: what do you think if we return an error if the version status is something different?

pkg/handlers/embedded_cluster_confirm_cluster_management.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+
+		kotsKinds, err := kotsutil.LoadKotsKinds(archiveDir)
+		if err != nil {
+			logger.Error(fmt.Errorf("failed to load kots kinds: %w", err))
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+
+		downstreamVersionStatus := storetypes.VersionPending
```
sgalsaleh, line 72: is there any possibility to make this a shared function/logic instead of duplicating this code?

web/src/components/apps/AppDetailPage.tsx
```diff
@@ -337,15 +337,15 @@ function AppDetailPage(props: Props) {
       const firstVersion = downstream.pendingVersions.find(
         (version: Version) => version?.sequence === 0
       );
-      if (firstVersion?.status === "unknown" && props.isEmbeddedCluster) {
+      if (
```
sgalsaleh, line 340: would this ever be unknown now?

web/src/components/apps/EmbeddedClusterManagement.tsx
```diff
@@ -360,6 +362,50 @@ const EmbeddedClusterManagement = ({
   }, [nodesData?.nodes?.toString()]);
   // #endregion
 
+  const onContinueClick = async () => {
+    const res = await fetch(
+      `${process.env.API_ENDPOINT}/embedded-cluster/confirm`,
```
sgalsaleh, line 367: this route name seems ambiguous or generic. not sure what's a better name though...

cbodonnell: `/embedded-cluster/management`?

web/src/components/apps/AppDetailPage.tsx
cbodonnell: i think we could remove that. probably left over from an early implementation?

pkg/handlers/embedded_cluster_confirm_cluster_management.go
cbodonnell: added 400 response as this endpoint should not be called if the version is not `pending_cluster_management`

pkg/handlers/embedded_cluster_confirm_cluster_management.go
cbodonnell: added for all embedded-specific handlers

sgalsaleh (approved):

PR #4469, opened 2024-02-23
hide navbar options on initial embedded install

cbodonnell:

This PR hides the navbar options for embedded clusters during the initial install flow. The main change is the introduction and usage of the `isInitialAppInstall` utility function that determines if a application is in its initial install phase based on current and pending downstream versions and their statuses.

**Before:**

Screenshot 2024-02-27 at 12 54 55 PM

**After:**

Screenshot 2024-02-27 at 12 56 05 PM

https://app.shortcut.com/replicated/story/99650/remove-application-and-cluster-management-tabs-at-the-top-when-doing-an-initial-installation

```release-note
Hides the Application and Cluster Management tabs on the navbar during the initial install flow on [Embedded Clusters (Beta)](/vendor/embedded-overview).
```

### DIFF
--- a/web/src/utilities/utilities.js
+++ b/web/src/utilities/utilities.js
@@ -626,6 +626,11 @@ export const Utilities = {
       return true;
     }
 
+    if (app.downstream.pendingVersions.length > 1) {
+      // there's more than one pending version, so it's not an initial install
+      return false;
+    }
+
     const firstPendingVersion = app.downstream.pendingVersions[0];
     if (
       firstPendingVersion.status === "pending_cluster_management" ||
--- a/web/src/utilities/utilities.test.js
+++ b/web/src/utilities/utilities.test.js
@@ -132,6 +132,22 @@ describe("Utilities", () => {
       expect(Utilities.isInitialAppInstall(app)).toBe(true);
     });
 
+    it("should return false if there is more than one pending version", () => {
+      const app = {
+        downstream: {
+          pendingVersions: [
+            {
+              status: "pending_config",
+            },
+            {
+              status: "pending_config",
+            },
+          ],
+        },
+      };
+      expect(Utilities.isInitialAppInstall(app)).toBe(false);
+    });
+
     it("should return true if first pending version has status `pending_cluster_management`, `pending_config`, `pending_preflight`, or `pending_download`", () => {
       let app = {
         downstream: {


web/src/utilities/utilities.js
```diff
@@ -611,6 +611,38 @@ export const Utilities = {
...
+      return false;
+    }
+
+    if (app.downstream?.currentVersion) {
+      // a version has already been deployed
+      return false;
+    }
+
+    if (
+      !app.downstream?.pendingVersions ||
```
sgalsaleh, line 625: this can be simplified:
```suggestion
      !app.downstream?.pendingVersions?.length
```

web/src/utilities/utilities.js
```diff
@@ -611,6 +611,38 @@ export const Utilities = {
     }
   },
 
+  isInitialAppInstall(app) {
+    if (!app) {
+      return false;
```
sgalsaleh, line 616: shouldn't this be true?

web/src/utilities/utilities.js
```diff
@@ -611,6 +611,38 @@ export const Utilities = {
...
+
+    if (
+      !app.downstream?.pendingVersions ||
+      app.downstream.pendingVersions.length === 0
+    ) {
+      // there's no version that's been deployed or is pending, so it's an initial install
+      return true;
+    }
+
+    const firstPendingVersion = app.downstream.pendingVersions[0];
```
sgalsaleh, line 632: this might be an edge case, but what if there's more than 1 pending version?

web/src/utilities/utilities.js
```diff
@@ -611,6 +611,38 @@ export const Utilities = {
...
+    if (
+      firstPendingVersion.status === "pending_cluster_management" ||
```
sgalsaleh, line 634: would it make more sense to check the `pending`, `deploying`, `deployed`, and `failed` statuses instead, and return true if the status doesn't match either of them? just in case another status gets added in the future.

sgalsaleh: e.g.: check if the version is not "deployable" yet

sgalsaleh: actually, there should be a `isDeployable` field on the version that you can check. that might be sufficient here.

sgalsaleh: this: https://github.com/replicatedhq/kots/blob/2c8ddb48f3e5f383e6658ee2e7aa373f3c07d4cf/pkg/api/downstream/types/types.go#L44

sgalsaleh: actually, that field might not be set during that workflow to save DB queries. reference: https://github.com/replicatedhq/kots/blob/2c8ddb48f3e5f383e6658ee2e7aa373f3c07d4cf/pkg/api/downstream/types/types.go#L42-L43

sgalsaleh: actually, that field might not be set during that workflow to save DB queries. reference: https://github.com/replicatedhq/kots/blob/2c8ddb48f3e5f383e6658ee2e7aa373f3c07d4cf/pkg/api/downstream/types/types.go#L42-L43

web/src/utilities/utilities.js
cbodonnell: i did remember a bug where there was once two pending versions created during initial install. it seems very rare, but do you think we should be taking the last element in the list instead of the first? should we be iterating and checking all of them?

web/src/utilities/utilities.js
cbodonnell: seems reasonable to flip this. i did not manage to reproduce this case ever since the only caller right now enforces that `app` exists before calling this function.

web/src/utilities/utilities.js
cbodonnell: good question... thinking out loud, i suppose that could have the same issue if we add a status that is _not_ part of the initial install flow since we would have to potentially add that one to the check. i'm ok with checking the statuses as they are now since worst case, we end up showing the navbar when we shouldn't instead of hiding it when it should be showing. lmk if i'm missing anything there and we need to change it

web/src/utilities/utilities.js
sgalsaleh: honestly, it's probably fine regardless. this doesn't affect any critical functionality. i'd be okay with leaving it like this for now.

sgalsaleh: or, simply return `false` if there's more than 1 pending version?

cbodonnell: i'll go with that second one for now. the end-user might probably be stuck anyway if there's >1 pending version on initial install for some reason.

sgalsaleh (approved):

PR #4488, opened 2024-03-01
fix cluster upgrade modal not showing intermittently

cbodonnell:

This PR fixes a race condition issue in embedded clusters where the cluster upgrade modal would fail to show intermittently. This issue would arise if the new installation had been applied to the cluster, but had not yet had a `state` reported back to the UI before downtime occurred. In this scenario, the UI was not able to determine whether a cluster upgrade was pending.

This PR changes the app details response to also include the number of installation objects in the cluster. In this case, we can infer that if the latest installation does not yet have a state, but is not the only installation, then it has just recently been applied and a cluster upgrade may be pending. If there is not actually an upgrade pending, downtime will not occur and the UI will receive another update with the actual state shortly thereafter. If there is an upgrade pending, the downtime may begin before the UI receives an actual state, but with these changes will show the upgrade modal.

https://app.shortcut.com/replicated/story/99951/cluster-upgrade-modal-fails-to-show-intermittently

```release-note
Fixes an issue in Embedded Cluster (Beta) where the cluster upgrade modal may fail to show during upgrades
```

### DIFF
--- a/pkg/embeddedcluster/util.go
+++ b/pkg/embeddedcluster/util.go
@@ -60,6 +60,18 @@ func RequiresUpgrade(ctx context.Context, newcfg embeddedclusterv1beta1.ConfigSp
 
 // GetCurrentInstallation returns the most recent installation object from the cluster.
 func GetCurrentInstallation(ctx context.Context) (*embeddedclusterv1beta1.Installation, error) {
+	installations, err := ListInstallations(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list installations: %w", err)
+	}
+	latest, err := GetLatestInstallation(ctx, installations)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get latest installation: %w", err)
+	}
+	return latest, nil
+}
+
+func ListInstallations(ctx context.Context) ([]embeddedclusterv1beta1.Installation, error) {
 	clientConfig, err := k8sutil.GetClusterConfig()
 	if err != nil {
 		return nil, fmt.Errorf("failed to get cluster config: %w", err)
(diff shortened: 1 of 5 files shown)


pkg/handlers/app.go
```diff
@@ -341,11 +341,21 @@ func responseAppFromApp(a *apptypes.App) (*types.ResponseApp, error) {
...
+
+			cluster.NumInstallations = len(embeddedClusterInstallations)
+
+			latestEmbeddedClusterInstallation, err := embeddedcluster.GetLatestInstallation(context.TODO(), embeddedClusterInstallations)
+			if err != nil {
+				return nil, errors.Wrap(err, "failed to get latest installation")
+			}
+
+			if latestEmbeddedClusterInstallation != nil {
+				cluster.State = string(latestEmbeddedClusterInstallation.Status.State)
```
cbodonnell, line 357: if the installation was just recently created, `.Status.State` may be empty. the goal here is to provide an additional piece of data `numInstallations` to the UI so it can make the proper determination of whether or not to show the modal in these cases if downtime occurs before it receives an actual state.

web/src/utilities/utilities.js
```diff
@@ -672,11 +672,25 @@ export const Utilities = {
...
+    }
+    const normalizedState = this.clusterState(cluster.state);
+    if (
+      normalizedState === "Upgrading" ||
+      normalizedState === "Upgrading addons"
+    ) {
+      return true;
+    }
+
+    if (cluster.numInstallations > 1 && !cluster.state) {
```
sgalsaleh, line 687: Is it possible to have no state with a single installation (i.e. numInstallations equals 1)? If not, does returning the number of installations matter, or can we simply check if there's no state?

cbodonnell: potentially during an initial install right after the first installation is created (though unlikely to impact the UI). knowing that it's not the first installation lets us know that it's an update vs install.

sgalsaleh: Will the UI ever be available/accessible when the initial installation does not have a state yet?

pkg/embeddedcluster/util.go
```diff
@@ -75,14 +87,17 @@ func GetCurrentInstallation(ctx context.Context) (*embeddedclusterv1beta1.Instal
 	if err != nil {
 		return nil, fmt.Errorf("failed to list installations: %w", err)
 	}
-	if len(installationList.Items) == 0 {
+	return installationList.Items, nil
+}
+
+func GetLatestInstallation(ctx context.Context, installations []embeddedclusterv1beta1.Installation) (*embeddedclusterv1beta1.Installation, error) {
```
sgalsaleh, line 93: IMO this function name is confusing because there's already a `GetCurrentInstallation`. Listing installations from the cluster isn't an expensive process, so I would suggest basically calling both `ListInstallations` and `GetCurrentInstallation` instead of trying to save a function call.

sgalsaleh (dismissed):

web/src/utilities/utilities.js
cbodonnell: it seems very unlikely (maybe impossible), but the goal is to guard this conditional and prevent showing UI elements related to upgrades during an initial install flow.

sgalsaleh (approved):

PR #4493, opened 2024-03-05
disable rolling back the app on embedded cluster

laverya:

Fixes #

```release-note
Prevents app rollback on embedded-cluster installations.
```

### DIFF
--- a/pkg/store/kotsstore/version_store.go
+++ b/pkg/store/kotsstore/version_store.go
@@ -71,6 +71,10 @@ func (s *KOTSStore) IsRollbackSupportedForVersion(appID string, sequence int64) 
 		return false, errors.Wrap(err, "failed to load kots app from contents")
 	}
 
+	if util.IsEmbeddedCluster() {
+		return false, nil
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/store/kotsstore/version_store.go
```diff
@@ -71,6 +71,10 @@ func (s *KOTSStore) IsRollbackSupportedForVersion(appID string, sequence int64)
 		return false, errors.Wrap(err, "failed to load kots app from contents")
 	}
 
+	if util.IsEmbeddedCluster() {
```
sgalsaleh, line 74: why not at the very beginning of the function? :)

laverya: sure, why not

laverya: actually, "because that might impact behavior more than where it is now"

maybe moving it to the beginning means that 'can't get app' errors are surfaced differently

chris-sanders:
I'll leave the code review to those more familiar, but the concept of disabling this for Embedded Cluster 👍

sgalsaleh (approved):

PR #4514, opened 2024-03-19
allow embedded cluster node labels with dns prefix

cbodonnell:

This PR modifies the embedded cluster node label sanitization so that node labels with DNS prefixes are supported.

Ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

Fixes #

```release-note
Fixes an issue in embedded clusters where forward slashes were replaced with dashes in custom role labels
```

### DIFF
--- a/pkg/embeddedcluster/roles.go
+++ b/pkg/embeddedcluster/roles.go
@@ -98,6 +98,20 @@ func getRoleNodeLabels(ctx context.Context, roles []string) ([]string, error) {
 	return getRoleLabelsImpl(config, roles), nil
 }
 
+// labelKey will clean up a string to be a valid label key.
+// ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set
+func labelKey(s string) string {
+	parts := strings.SplitN(s, "/", 2)
+	if len(parts) == 1 {
+		return labelify(parts[0])
+	}
+	prefix := ConvertToRFC1123(parts[0])
+	name := labelify(parts[1])
+	return fmt.Sprintf("%s/%s", prefix, name)
+}
+
+// labelify will clean up a string to be a valid label value.
+// ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set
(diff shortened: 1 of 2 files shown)


pkg/embeddedcluster/roles_test.go
```diff
@@ -180,6 +180,36 @@ func Test_getRoleLabelsImpl(t *testing.T) {
...
+						Name: "a",
+						Labels: map[string]string{
+							"test-label.example.com/this-is-a-more-than-63-character-label-with-a-lot-of-filler-to-ensure-that": "this is a more than 63 character value with a lot of filler to ensure that",
+						},
+					},
+				},
+			},
+			roles: []string{"a"},
+			want:  []string{"test-label.example.com/this-is-a-more-than-63-character-label-with-a-lot-of-filler-to=this-is-a-more-than-63-character-value-with-a-lot-of-filler-to"},
+		},
```
ricardomaraschini, line 212: would you mind adding a unit test with an invalid domain ? something like `this.is.not.valid-/test`.

pkg/embeddedcluster/roles.go
```diff
@@ -112,6 +126,60 @@ func labelify(s string) string {
...
+
+var dns1123IllegalStartRegex = regexp.MustCompile(`^[^0-9a-z]+`)
+var dns1123IllegalEndRegex = regexp.MustCompile(`[^0-9a-z]+$`)
+var dns1123IllegalCharsRegex = regexp.MustCompile(`[^0-9a-z-]`)
+
+func ConvertToRFC1123(value string, args ...int) string {
+	value = strings.ToLower(value)
+
+	if len(value) == 0 || IsValidRFC1123(value) {
+		return strings.ToLower(value)
```
ricardomaraschini, line 159: I think we can simply return `value` here.

cbodonnell: looks like it since we've already run `value = strings.ToLower(value)` at that point

ricardomaraschini (approved):

PR #4515, opened 2024-03-19
push embedded-cluster oci artifacts

cbodonnell:

The PR updates the `admin-console push-images` CLI command to support pushing embedded cluster artifacts from airgap bundles. These artifacts will be present in an `embedded-cluster` directory. If the directory is not present, this is a no-op.

Sample output:
```
$ ./bin/kots admin-console push-images embedded-all.airgap ttl.sh/my-app
...
Pushing artifact ttl.sh/my-app/embedded-cluster/charts.tar.gz:2TiLbGfE9YRXAPM2JCD9STfS11B-76-0.1.66
Pushing artifact ttl.sh/my-app/embedded-cluster/embedded-cluster:2TiLbGfE9YRXAPM2JCD9STfS11B-76-0.1.66
Pushing artifact ttl.sh/my-app/embedded-cluster/images-amd64.tar:2TiLbGfE9YRXAPM2JCD9STfS11B-76-0.1.66
```

https://app.shortcut.com/replicated/story/100535/expand-kubectl-kots-push-images-command-to-include-embedded-cluster-additional-files

```release-note
The KOTS CLI `admin-console push-images` command now supports pushing embedded cluster artifacts from airgap bundles
```

### DIFF
--- a/pkg/image/airgap.go
+++ b/pkg/image/airgap.go
@@ -178,8 +178,12 @@ func TagAndPushImagesFromBundle(airgapBundle string, options imagetypes.PushImag
 			return errors.Wrap(err, "failed to push images from temp registry")
 		}
 
-		artifactTag := fmt.Sprintf("%s-%s", airgap.Spec.ChannelID, airgap.Spec.UpdateCursor)
-		if err := PushEmbeddedClusterArtifacts(extractedBundle, artifactTag, options); err != nil {
+		pushEmbeddedArtifactsOpts := imagetypes.PushEmbeddedArtifactsOptions{
+			Registry:   options.Registry,
+			Tag:        fmt.Sprintf("%s-%s", airgap.Spec.ChannelID, airgap.Spec.UpdateCursor),
+			HTTPClient: orasretry.DefaultClient,
+		}
+		if err := PushEmbeddedClusterArtifacts(extractedBundle, pushEmbeddedArtifactsOpts); err != nil {
 			return errors.Wrap(err, "failed to push embedded cluster artifacts")
 		}
 
@@ -189,7 +193,6 @@ func TagAndPushImagesFromBundle(airgapBundle string, options imagetypes.PushImag
 	default:
 		return errors.Errorf("Airgap bundle format '%s' is not supported", airgap.Spec.Format)
 	}
-
 }
 
 func PushImagesFromTempRegistry(airgapRootDir string, imageList []string, options imagetypes.PushImagesOptions) error {
@@ -703,7 +706,7 @@ type OCIArtifactFile struct {
 	MediaType string
 }
 
-func PushEmbeddedClusterArtifacts(airgapBundle string, tag string, options imagetypes.PushImagesOptions) error {
+func PushEmbeddedClusterArtifacts(airgapBundle string, opts imagetypes.PushEmbeddedArtifactsOptions) error {
 	embeddedClusterDir := filepath.Join(airgapBundle, "embedded-cluster")
 	if _, err := os.Stat(embeddedClusterDir); err != nil {
 		if os.IsNotExist(err) {
@@ -736,8 +739,8 @@ func PushEmbeddedClusterArtifacts(airgapBundle string, tag string, options image
 			Path:      path,
 			MediaType: EmbeddedClusterMediaType,
 		}
-		fmt.Printf("Pushing embedded-cluster artifact %s:%s\n", filepath.Join(options.Registry.Endpoint, options.Registry.Namespace, repo), tag)
-		if err := pushOCIArtifact([]OCIArtifactFile{descriptor}, EmbeddedClusterMediaType, repo, tag, options); err != nil {
+		fmt.Printf("Pushing embedded-cluster artifact %s:%s\n", filepath.Join(opts.Registry.Endpoint, opts.Registry.Namespace, repo), opts.Tag)
+		if err := pushOCIArtifact([]OCIArtifactFile{descriptor}, EmbeddedClusterMediaType, repo, opts); err != nil {
 			return errors.Wrapf(err, "failed to push embedded-cluster artifact %s", info.Name())
 		}
 
@@ -750,7 +753,7 @@ func PushEmbeddedClusterArtifacts(airgapBundle string, tag string, options image
 	return nil
 }
 
-func pushOCIArtifact(files []OCIArtifactFile, artifactType string, repo string, tag string, options imagetypes.PushImagesOptions) error {
+func pushOCIArtifact(files []OCIArtifactFile, artifactType string, repo string, opts imagetypes.PushEmbeddedArtifactsOptions) error {
 	orasWorkspace, err := os.MkdirTemp("", "oras")
 	if err != nil {
 		return errors.Wrap(err, "failed to create temp directory")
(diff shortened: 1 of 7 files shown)


pkg/image/airgap.go
```diff
@@ -675,6 +700,114 @@ func reportWriterWithProgress(imageInfos map[string]*imagetypes.ImageInfo, repor
...
+			return nil
+		}
+
+		repo := "binary"
+		if info.Name() == "charts.tar.gz" {
+			repo = "charts"
+		} else if info.Name() == "images-amd64.tar" {
+			repo = "images"
+		}
+		repo = filepath.Join("embedded-cluster", repo)
```
cbodonnell, line 734: Would like to get thoughts on this piece. Will the charts and images artifacts have well-known names?

Separately, any thoughts on the repo names that they are pushed to?

pkg/image/airgap.go
```diff
@@ -675,6 +700,114 @@ func reportWriterWithProgress(imageInfos map[string]*imagetypes.ImageInfo, repor
...
+
+		// push each file as an oci artifact to the registry
+		descriptor := OCIArtifactFile{
+			Name:      info.Name(),
+			Path:      path,
+			MediaType: EmbeddedClusterMediaType,
+		}
+		fmt.Printf("Pushing embedded-cluster artifact %s:%s\n", filepath.Join(opts.Registry.Endpoint, opts.Registry.Namespace, repo), opts.Tag)
```
cbodonnell, line 742: Do we care to incorporate the progress indicators if possible?

pkg/image/airgap.go
cbodonnell, line 734: Would like to get thoughts on this piece. Will the charts and images artifacts have well-known names?

Separately, any thoughts on the repo names that they are pushed to? `embedded-cluster/binary`, `embedded-cluster/charts`, and `embedded-cluster/images`?

pkg/image/airgap.go
```diff
@@ -162,7 +174,20 @@ func TagAndPushImagesFromBundle(airgapBundle string, options imagetypes.PushImag
...
 			return errors.Wrap(err, "falied to unarchive airgap bundle")
 		}
-		return PushImagesFromTempRegistry(extractedBundle, airgap.Spec.SavedImages, options)
+		if err := PushImagesFromTempRegistry(extractedBundle, airgap.Spec.SavedImages, options); err != nil {
+			return errors.Wrap(err, "failed to push images from temp registry")
+		}
+
+		pushEmbeddedArtifactsOpts := imagetypes.PushEmbeddedArtifactsOptions{
+			Registry:   options.Registry,
+			Tag:        airgap.Spec.VersionLabel,
```
cbodonnell, line 183: Using the version label as the tag for these artifacts. Could alternatively use `<channelID>-<channelSequence>` if we need it to be unique as a vendor could theoretically re-use a version label.

pkg/image/airgap.go
cbodonnell: Iterate over files and push each one as an artifact with a deterministic name

pkg/image/airgap.go
cbodonnell: `<channelID>-<channelSequence>-<versionLabel>` and then sanitize and truncate?

pkg/image/airgap.go
```diff
@@ -162,7 +174,20 @@ func TagAndPushImagesFromBundle(airgapBundle string, options imagetypes.PushImag
...
+		if err := PushImagesFromTempRegistry(extractedBundle, airgap.Spec.SavedImages, options); err != nil {
+			return errors.Wrap(err, "failed to push images from temp registry")
+		}
+
+		pushEmbeddedArtifactsOpts := imagetypes.PushEmbeddedArtifactsOptions{
+			Registry:   options.Registry,
+			Tag:        imageutil.SanitizeTag(fmt.Sprintf("%s-%s-%s", airgap.Spec.ChannelID, airgap.Spec.UpdateCursor, airgap.Spec.VersionLabel)),
+			HTTPClient: orasretry.DefaultClient,
+		}
+		if err := PushEmbeddedClusterArtifacts(extractedBundle, pushEmbeddedArtifactsOpts); err != nil {
```
sgalsaleh, line 186: airgap bundle format for images doesn't actually effect how to process embedded cluster artifacts as they're outside the airgap bundle registry. how about moving this outside the `switch` statement?

pkg/image/airgap.go
```diff
@@ -675,6 +700,107 @@ func reportWriterWithProgress(imageInfos map[string]*imagetypes.ImageInfo, repor
 	return pipeWriter
 }
 
+type OCIArtifactFile struct {
+	Name      string
+	Path      string
+	MediaType string
+}
+
+func PushEmbeddedClusterArtifacts(airgapBundle string, opts imagetypes.PushEmbeddedArtifactsOptions) error {
```
sgalsaleh, line 709: Nit pick: can we make this function more generic where it takes a directory and push files in it as artifacts? Can then pass the `embedded-cluster` directory as a parameter, and rename `PushEmbeddedArtifactsOptions` to be more generic.

sgalsaleh: Something like this:
```suggestion
func PushOCIArtifacts(airgapBundle string, opts imagetypes.PushOCIArtifactsOptions) error {
```

pkg/image/types/types.go
```diff
@@ -78,3 +79,9 @@ type LayerInfo struct {
 	UploadStart time.Time
 	UploadEnd   time.Time
 }
+
+type PushEmbeddedArtifactsOptions struct {
```
cbodonnell, line 83: rename this to make it non-EC specific

pkg/image/airgap.go
```diff
@@ -675,6 +700,107 @@ func reportWriterWithProgress(imageInfos map[string]*imagetypes.ImageInfo, repor
...
+	embeddedClusterDir := filepath.Join(airgapBundle, "embedded-cluster")
+	if _, err := os.Stat(embeddedClusterDir); err != nil {
```
sgalsaleh, line 711: how about also checking if we're actually running in an embedded cluster? might be a good guardrail to have.

pkg/image/airgap.go
cbodonnell: this code path is the only one where the bundle is extracted, so it seemed to fit in there best. it looks like it would require additional refactoring to move it outside of the `switch` statement

cbodonnell: the suggested change is to have `PushEmbeddedClusterArtifacts` accept a compressed bundle archive and do it's own extraction, as necessary

pkg/image/airgap_test.go
```diff
@@ -0,0 +1,143 @@
...
+			airgapFiles: map[string][]byte{
+				"airgap.yaml":                       []byte("this-is-the-airgap-metadata"),
+				"app.tar.gz":                        []byte("this-is-the-app-archive"),
+				"images/something":                  []byte("this-is-an-image"),
+				"embedded-cluster/test-app":         []byte("this-is-the-binary"),
+				"embedded-cluster/charts.tar.gz":    []byte("this-is-the-charts-bundle"),
+				"embedded-cluster/images-amd64.tar": []byte("this-is-the-images-bundle"),
+				"embedded-cluster/some-file-TBD":    []byte("this-is-an-arbitrary-file"),
+			},
+			wantArtifacts: map[string]string{
```
sgalsaleh, line 47: would be great to test the actual path in the registry including the defined registry namespace and `embedded-cluster`. would that be difficult to add?

pkg/image/airgap.go
```diff
@@ -675,6 +702,147 @@ func reportWriterWithProgress(imageInfos map[string]*imagetypes.ImageInfo, repor
...
+	for {
+		header, err := tarReader.Next()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return errors.Wrap(err, "failed to get read archive")
+		}
+
+		if header.Typeflag == tar.TypeDir {
```
sgalsaleh, line 734:
```suggestion
		if header.Typeflag != tar.TypeReg {
```

sgalsaleh (dismissed):

pkg/image/airgap_test.go
cbodonnell: i think that would just be a small change in the test. let me see...

pkg/image/airgap.go
```diff
@@ -675,6 +702,147 @@ func reportWriterWithProgress(imageInfos map[string]*imagetypes.ImageInfo, repor
...
+		if filepath.Dir(header.Name) != "embedded-cluster" {
+			continue
+		}
+
+		if err := func() error {
+			dstFilePath := filepath.Join(tmpDir, header.Name)
+			if err := os.MkdirAll(filepath.Dir(dstFilePath), 0755); err != nil {
+				return errors.Wrap(err, "failed to create path")
+			}
+			defer os.RemoveAll(dstFilePath)
```
sgalsaleh, line 747: this might remove the original tmp directory created at the top if there's no subdirectory, but i guess the above `MkdirAll` will recreate it again? not sure if that will have unexpected behaviors.

sgalsaleh (approved):

PR #4518, opened 2024-03-22
process embedded-cluster airgap artifacts during install and upgrade

cbodonnell:

The primary change is removing the need for `TagAndPushImagesFromPath` so that `TagAndPushImagesFromBundle` is used in all cases, which handles processing embedded-cluster artifacts. There was a bit of refactoring necessary to leverage the compressed instead of extracted airgap bundle.

https://app.shortcut.com/replicated/story/101172/embed-the-kots-cli-and-use-it-to-install-the-app-not-kots-in-online-and-airgap-mode

```release-note

```

### DIFF
--- a/pkg/airgap/airgap.go
+++ b/pkg/airgap/airgap.go
@@ -9,7 +9,6 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -37,7 +36,8 @@ import (
 
 type CreateAirgapAppOpts struct {
 	PendingApp             *types.PendingApp
-	AirgapPath             string
+	AirgapBundle           string
+	AirgapRootDir          string
 	RegistryHost           string
 	RegistryNamespace      string
 	RegistryUsername       string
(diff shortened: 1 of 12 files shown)


pkg/kotsadm/main.go
```diff
@@ -165,21 +166,21 @@ func Deploy(deployOptions types.DeployOptions, log *logger.CLILogger) error {
...
 		if !deployOptions.DisableImagePush {
-			err := image.TagAndPushImagesFromPath(deployOptions.AirgapRootDir, pushOptions)
+			err := image.TagAndPushImagesFromBundle(deployOptions.AirgapBundle, pushOptions)
 			if err != nil {
 				return errors.Wrap(err, "failed to tag and push app images from path")
 			}
 		}
 
-		airgapPath = deployOptions.AirgapRootDir
+		airgapPath = deployOptions.AirgapBundle
```
sgalsaleh, line 175: how about reusing `deployOptions.AppImagesPushed = true` here and remove this `airgapPath` variable that is only used to indicate the app images were pushed? And use `deployOptions.AirgapBundle` when getting the `airgap.yaml` file.

sgalsaleh: hmm, i do see `deployOptions.DisableImagePush` so it might be confusing... nevermind

pkg/airgap/airgap.go
```diff
@@ -213,8 +211,7 @@ func CreateAppFromAirgap(opts CreateAirgapAppOpts) (finalError error) {
 		ConfigFile:          configFile,
 		IdentityConfigFile:  identityConfigFile,
 		IsAirgap:            true,
-		AirgapRoot:          archiveDir,
```
sgalsaleh, line 216: Why remove this?

pkg/kotsadm/main.go
cbodonnell: addressed this in https://github.com/replicatedhq/kots/pull/4518/commits/07e0d7bf117ec9c5ae64de657a7a67e312297423

cbodonnell: and https://github.com/replicatedhq/kots/pull/4518/commits/2be05f26e568f7c7747b6552bc9431e8344ef762 :)

sgalsaleh (approved):

PR #4524, opened 2024-03-25
trigger embedded airgap upgrade

cbodonnell:

This PR updates the installation object that is created as part of an embedded cluster upgrade to include necessary OCI artifact paths required in airgap environments. KOTS will store the pushed artifacts in its own KOTS installation object and will read from there at deploy time when creating the embedded cluster installation object.

Depends on changes to the KOTS installation schema in https://github.com/replicatedhq/kotskinds/pull/8

Fixes https://app.shortcut.com/replicated/story/101432/trigger-airgap-upgrade-from-kots

```release-note

```


pkg/embeddedcluster/util.go
```diff
@@ -126,6 +163,7 @@ func startClusterUpgrade(ctx context.Context, newcfg embeddedclusterv1beta1.Conf
 			ClusterID:                 current.Spec.ClusterID,
 			MetricsBaseURL:            current.Spec.MetricsBaseURL,
 			AirGap:                    current.Spec.AirGap,
+			Artifacts:                 artifacts,
```
cbodonnell, line 166: Sounds like we will be adding a field to the EC installation object for the metadata override, so this will likely need to be updated to account for that.

pkg/embeddedcluster/util.go
```diff
@@ -102,8 +109,38 @@ func ClusterConfig(ctx context.Context) (*embeddedclusterv1beta1.ConfigSpec, err
...
+		case imagesArtifactRegex.MatchString(artifact):
+			artifacts.Images = artifact
+		case binaryArtifactRegex.MatchString(artifact):
+			artifacts.EmbeddedClusterBinary = artifact
+		default:
+			logger.Warnf("unknown artifact in installation: %s", artifact)
+		}
+	}
+
+	if artifacts.HelmCharts == "" {
```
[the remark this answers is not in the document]
cbodonnell: sounds good. i'll remove the validation from the kots-side

pkg/image/airgap.go
```diff
@@ -121,27 +121,29 @@ func CopyAirgapImages(opts imagetypes.ProcessImageOptions, log *logger.CLILogger
 		LogForUI:       true,
 	}
 
+	result := &imagetypes.CopyAirgapImagesResult{}
 	if opts.AirgapBundle != "" {
```
sgalsaleh, line 125: can we just move this to the top of the function and return if it's not set? would be a bit cleaner.

pkg/image/online.go
```diff
@@ -45,6 +45,12 @@ type UpdateInstallationImagesOptions struct {
 	DockerHubRegistryCreds registry.Credentials
 }
 
+type UpdateInstallationAirgapArtifactsOptions struct {
```
sgalsaleh, line 48: this is long but 🤷‍♂️ 
```suggestion
type UpdateInstallationEmbeddedClusterArtifactsOptions struct {
```

pkg/image/online.go
```diff
@@ -132,6 +138,20 @@ func UpdateInstallationImages(opts UpdateInstallationImagesOptions) error {
 	return nil
 }
 
+func UpdateInstallationAirgapArtifacts(opts UpdateInstallationAirgapArtifactsOptions) error {
```
sgalsaleh, line 141:
```suggestion
func UpdateInstallationEmbeddedClusterArtifacts(opts UpdateInstallationEmbeddedClusterArtifactsOptions) error {
```

pkg/image/types/types.go
```diff
@@ -46,6 +46,10 @@ type CopyImageOptions struct {
 	ReportWriter      io.Writer
 }
 
+type CopyAirgapImagesResult struct {
+	Artifacts []string
```
sgalsaleh, line 50:
```suggestion
	EmbeddedClusterArtifacts []string
```

sgalsaleh (approved):

PR #4533, opened 2024-04-02
Feature: kots template

nvanthao:
It is currently difficult to debug templating error thrown by KOTS during install.

This PR introduce new command `kots template` with `4` options to parse template.

1. Parse raw template directly from `--data` flag.

E.g.

```
kots template --license-file license.yaml --config-values config.yaml --data '{{repl ConfigOption "hostname" | ToUpper }}'

MYHOST
```

2. Parse template interactively. New REPL session will be created to parse template

```
kots template --license-file license.yaml --config-values config.yaml --interactive
```

3. Parse template by file

```
kots template --license-file license.yaml --config-values config.yaml <path_to_template>
```

4. Parse all templates in `kotsKind` directory, this behave similar to `helm template` where all files in `kotsKind` directory will be printed to stdout. The app will be pulled from upstream.

```
kots template --license-file license.yaml --config-values config.yaml 
```

5. Parse all templates from local release.

```
kots template --license-file license.yaml --config-values config.yaml --local-path <path_to_release>
```

```release-note
New command line command `kots template`
```

Yes

### DIFF
--- a/cmd/kots/cli/template.go
+++ b/cmd/kots/cli/template.go
@@ -268,7 +268,10 @@ func pullAndRender(appSlug string, licensePath string, configPath string, localP
 	upstream := pull.RewriteUpstream(appSlug)
 	_, err = pull.Pull(upstream, pullOptions)
 
-	if err != nil && err != pull.ErrConfigNeeded {
+	if err != nil {
+		if err == pull.ErrConfigNeeded {
+			return errors.New("missing required config values to render templates")
+		}
 		return errors.Wrap(err, "failed to pull upstream")
 	}
 
--- a/cmd/kots/cli/template.go
+++ b/cmd/kots/cli/template.go
@@ -39,6 +39,7 @@ func TemplateCmd() *cobra.Command {
 			licenseFile := v.GetString("license-file")
 			configFile := v.GetString("config-values")
 			interactive := v.GetBool("interactive")
+			data := v.GetString("data")
 
 			license, err := parseLicenseFile(licenseFile)
 			if err != nil {
@@ -55,17 +56,6 @@ func TemplateCmd() *cobra.Command {
 				return errors.Wrap(err, "failed to create config context")
 			}
 
-			// when no args are provided, render all mode, similar to helm template
-			// we will utilize pull command to fetch and render manifests from upstream
-			if len(args) == 0 && !interactive {
-				err := pullAndRender(license.Spec.AppSlug, licenseFile, configFile)
-				if err != nil {
-					return errors.Wrap(err, "failed to render all templates")
-				}
-				return nil
-			}
-
-			// interactive mode
 			// TODO: support other contexts
 			builderOptions := template.BuilderOptions{
 				ExistingValues: configCtx,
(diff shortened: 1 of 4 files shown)


cmd/kots/cli/template.go
```diff
@@ -0,0 +1,339 @@
...
+		}
+		manifetsToRender[file.Name()] = string(content)
+	}
+	for k, m := range manifetsToRender {
+		fmt.Println("---")
+		fmt.Printf("# Source: %s\n", k)
+		fmt.Println(m)
+	}
+
+	// also render helm charts with helm template if any
```
cbodonnell, line 297: the `rendered` directory should already have the rendered helm templates, plain k8s manifests, and additional resources created by kustomize (pull secrets). i think we may be able to instead just iterate and print the files in that directory?

the kotskinds will not be in `rendered`, so i think we can keep what you have above this to show them to the user

pkg/pull/pull.go
```diff
@@ -358,7 +358,7 @@ func Pull(upstreamURI string, pullOptions PullOptions) (string, error) {
 		ReportWriter:     pullOptions.ReportWriter,
 	}
 
-	if needsConfig {
+	if needsConfig && pullOptions.ConfigFile == "" {
```
cbodonnell, line 361: i think we need to leave this as-is. it's causing a failure in the test for required values that don't have defaults. in that case, it will skip this, try and render the helm chart, but fail because of a missing required value https://github.com/replicatedhq/kots/actions/runs/8562639222/job/23466431010#step:8:492
```
{"level":"error","ts":"2024-04-04T23:50:58Z","msg":"failed to update app from airgap bundle: failed to update app: failed to pull: failed to render upstream: failed to render helm chart airgap-smoke-test: failed to render helm chart in upstream: failed to render with helm v3: helm v3 render failed with error: execution error at (airgap-smoke-test/templates/deployment.yaml:25:24): myNewRequiredValue is required!"}
```



is there a reason we need it for `kots template`?

nvanthao: Many thanks @cbodonnell! I initially thought that this logic does not include the config values in the check but later found out that it is. I removed the incorrect check.

cmd/kots/cli/template.go
nvanthao: Many thanks Craig! I've included `rendered` directory excluding `kotsadm-` files. However I think we should also run `helm template` on `helm/` directory for HelmChart v2. What do you think?

cbodonnell: the `rendered` directory will already contain the `helm template` output in an `all.yaml` file for each HelmChart, so i think we can just iterate over all of the files there, including the `kotsadm-` files since those will ultimately be deployed along with everything else

nvanthao: gotcha! I've removed the helm template part. Please let me know how it goes!

cmd/kots/cli/template.go
```diff
@@ -0,0 +1,323 @@
...
+				return errors.Wrap(err, "failed to render template")
+			}
+
+			log.Info(rendered)
+
+			return nil
+		},
+	}
+
+	cmd.Flags().String("license-file", "", "path to a license file to use when download a replicated app")
```
laverya, line 135:
```suggestion
	cmd.Flags().String("license-file", "", "path to a license file to use to download a replicated app")
```

cmd/kots/cli/template.go
```diff
@@ -0,0 +1,323 @@
...
+	cmd.Flags().String("config-values", "", "path to a manifest containing config values (must be apiVersion: kots.io/v1beta1, kind: ConfigValues)")
+	cmd.Flags().String("data", "", "raw template data to render")
+	cmd.Flags().Bool("interactive", false, "provides an interactive command-line console for evaluating template values")
+	cmd.Flags().String("local-path", "", "specify a local-path to pull a locally available replicated app (only supported on replicated app types currently)")
```
laverya, line 139: what happens if this flag isn't provided?

Is the app not pulled, or is it pulled to the current path?

laverya: oh I see this is to not pull but to use the app from the local path
```suggestion
	cmd.Flags().String("local-path", "", "the path to a directory containing replicated app yaml to be rendered.")
```

PR #4534, opened 2024-04-03
only push from images directory for docker archive bundles

cbodonnell:

This PR modifies the code that pushes images from the older docker-archive bundle format so that only files within the `images` directory get pushed. This fixes an issue where docker-archive bundles that contain embedded cluster artifacts would fail to get pushed to the registry with errors such as:
```
Error: failed to push images: failed to push images from docker archive bundle: failed to get images info from bundle: not enough parts in image path: "embedded-cluster/charts.tar.gz"
```

Fixes #

```release-note

```

### DIFF
--- a/pkg/image/airgap.go
+++ b/pkg/image/airgap.go
@@ -556,7 +556,6 @@ func getImageInfosFromBundle(airgapBundle string, getLayerInfo bool) (map[string
 	imageInfos := make(map[string]*imagetypes.ImageInfo)
 
 	tarReader := tar.NewReader(gzipReader)
-	foundImagesFolder := false
 	for {
 		header, err := tarReader.Next()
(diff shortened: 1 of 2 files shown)


laverya (approved):
the helm charts and binaries get pushed elsewhere still, right?

cbodonnell:
> the helm charts and binaries get pushed elsewhere still, right?

@laverya yup, embedded artifacts are pushed at the end of [this function](https://github.com/replicatedhq/kots/blob/cbo/sc-102083/handle-ec-bundle-with-docker-archive/pkg/image/airgap.go#L134-L177) and happens for any bundle format. the problematic code fixed here is part of the `PushImagesFromDockerArchiveBundle` path in that function

PR #4553, opened 2024-04-17
include the airgap registry cluster IP in the node join response

laverya:

In order for airgap nodes to pull images, they need to list the internal registry as 'yes this is OK to pull from without a valid cert', and in order to do that they need the registry address.

Fixes #

```release-note

```

### DIFF
--- a/pkg/handlers/embedded_cluster_node_join_command.go
+++ b/pkg/handlers/embedded_cluster_node_join_command.go
@@ -7,10 +7,10 @@ import (
 
 	"github.com/replicatedhq/kots/pkg/embeddedcluster"
 	"github.com/replicatedhq/kots/pkg/k8sutil"
+	"github.com/replicatedhq/kots/pkg/kotsutil"
 	"github.com/replicatedhq/kots/pkg/logger"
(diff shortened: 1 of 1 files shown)


pkg/handlers/embedded_cluster_node_join_command.go
```diff
@@ -163,6 +166,17 @@ func (h *Handler) GetEmbeddedClusterNodeJoinCommand(w http.ResponseWriter, r *ht
 		ecVersion = install.Spec.Config.Version
 	}
 
+	airgapRegistryAddress := ""
+	if install.Spec.AirGap {
+		airgapRegistrySvc, err := client.CoreV1().Services("registry").Get(r.Context(), "registry", metav1.GetOptions{})
```
sgalsaleh, line 171: You can use the `GetEmbeddedRegistryCreds` that we use elsewhere which would return the port number too I think. Example: https://github.com/replicatedhq/kots/blob/ea2056ef211a8940481dd500b5a7913a2ad1cd24/pkg/kotsadmsnapshot/backup.go#L387

sgalsaleh (approved):

PR #4564, opened 2024-04-22
fix edit config link on dashboard

cbodonnell:

This PR fixes an issue where the url for the "Edit Config" link on the dashboard was invalid and contained an `undefined` reference.

Screenshot 2024-04-22 at 12 18 43 PM

Fixes [SC-102656](https://app.shortcut.com/replicated/story/102656/edit-config-button-doesn-t-work)

```release-note
Fixes an issue where the **Edit config* link on the dashboard was not functioning as expected.
```

NONE

### DIFF
--- a/web/src/features/Dashboard/components/AppStatus.tsx
+++ b/web/src/features/Dashboard/components/AppStatus.tsx
@@ -22,7 +22,6 @@ type Props = {
   hasStatusInformers?: boolean;
   links: PropLink[];
   onViewAppStatusDetails: () => void;
-  url: string | undefined;
   embeddedClusterState: string;
 };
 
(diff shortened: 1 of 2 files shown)


web/src/features/Dashboard/components/AppStatus.tsx
```diff
@@ -148,7 +147,7 @@ export default class AppStatus extends Component<Props, State> {
               </>
             )}
             <Link
-              to={`${url}/config/${app?.downstream?.currentVersion?.sequence}`}
```
cbodonnell, line 151: I'm not aware of a scenario where this was _not_ `undefined`, so I just removed it all together. Tested this with both single and multi-app installs and both render the correct link for editing config.

sgalsaleh (approved):

PR #4582, opened 2024-04-29
feat: store embedded cluster config in a secret

ricardomaraschini:

the crd fields may change after the cluster upgrade so we can't rely on the current version when starting an upgrade. i.e. we need a way to preserve current as well as yet to be introduced fields. this pr makes kots manage the embedded cluster configuration as a secret. during an upgrade the new config is injected into a secret and the secret name is populated in the installation, the operator will read the config from the secret.


ricardomaraschini:
We haven't decided yet this is the path we want to use so I am closing this until further notice.

PR #4583, opened 2024-04-29
backup more namespaces in Embedded Cluster

laverya:

Embedded Cluster snapshots require resources outside of the `kotsadm` namespace.  This adds those namespaces to the backups created if it is an Embedded Cluster installation.

It also changes the label selector to use `replicated.app/disaster-recovery` labels instead, which application vendors will need to add to their application.

Fixes #

```release-note

```


pkg/kotsadmsnapshot/backup.go
```diff
@@ -962,6 +958,11 @@ func prepareIncludedNamespaces(namespaces []string) []string {
 		uniqueNamespaces[n] = true
 	}
 
+	if util.IsEmbeddedCluster() {
+		uniqueNamespaces["embedded-cluster"] = true
+		uniqueNamespaces["kube-system"] = true
+	}
```
laverya, line 964: I think I should refactor this to require a boolean isEC argument

pkg/kotsadmsnapshot/backup.go
```diff
@@ -962,6 +958,12 @@ func prepareIncludedNamespaces(namespaces []string) []string {
 		uniqueNamespaces[n] = true
 	}
 
+	if isEC {
+		uniqueNamespaces["embedded-cluster"] = true
+		uniqueNamespaces["kube-system"] = true
+		uniqueNamespaces["openebs"] = true
```
sgalsaleh, line 964: wouldn't `openebs` be installed along with k0s and velero in order to even initiate the restore process?

laverya: I think we install k0s + velero, then restore openebs, then restore kotsadm

sgalsaleh: how would velero be installed without openebs?

laverya: velero does not require in-cluster storage

sgalsaleh: 👍

pkg/kotsadmsnapshot/backup.go
```diff
@@ -977,18 +982,34 @@ func excludeShutdownPodsFromBackup(ctx context.Context, clientset kubernetes.Int
 		"status.phase": string(corev1.PodFailed),
 	}
 
-	podListOption := metav1.ListOptions{
-		LabelSelector: labels.SelectorFromSet(veleroBackup.Spec.LabelSelector.MatchLabels).String(),
-		FieldSelector: fields.SelectorFromSet(selectorMap).String(),
+	labelSets := []string{}
+	if veleroBackup.Spec.LabelSelector.MatchLabels != nil {
```
sgalsaleh, line 986: Would this ever be `nil` given that it's initialized here? https://github.com/replicatedhq/kots/pull/4583/files#diff-5bf9dc41509334fa70fd5f20b3c75acac8ac8efa3b3ad54d1d18aa5ee2999792R1051

sgalsaleh: maybe use `len`?

laverya: oof, will do

laverya: done

cbodonnell (approved):

PR #4616, opened 2024-05-13
fix config and view files tab underlines

cbodonnell:

This PR fixes an issue where the KOTS UI subnav would not show the selected tab for the "Config", "Troubleshoot", and "View Files" tabs. This is because previously we were only considering the last element in the path, so any routes with subpaths like `/config/<sequence>` or `/troubleshoot/analyze/<bundle-id>` would cause the tabs to show as not active.

Before `/app/craig-helm/tree/0`:
Screenshot 2024-05-13 at 4 32 28 PM

After `/app/craig-helm/tree/0`:
Screenshot 2024-05-13 at 4 30 54 PM

Fixes https://app.shortcut.com/replicated/story/102653/config-and-view-files-tab-lack-blue-underline-when-clicked-on

```release-note
Fixes an issue where the "Config" and "View Files" tabs would not display as active when clicked.
```

### DIFF
--- a/.github/actions/build-custom-image-with-apko/action.yml
+++ b/.github/actions/build-custom-image-with-apko/action.yml
@@ -49,7 +49,7 @@ runs:
 
     - uses: chainguard-images/actions/apko-publish@main
       with:
-        apko-image: cgr.dev/chainguard/apko:latest
+        apko-image: ghcr.io/wolfi-dev/apko@sha256:87b47283433066d19b71e1cb4401be8d04b7d51f4f4263fde1977c3043b9754f
         config: ${{ inputs.context }}/apko.yaml
         archs: amd64,arm64
         tag: ${{ inputs.image-name }}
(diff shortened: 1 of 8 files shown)


.github/actions/kurl-addon-kots-generate/action.yml
```diff
@@ -1,4 +1,5 @@
 name: Generate kURL Add-On
+description: Generate kURL Add-On
```
cbodonnell, line 2: actionlint started failing because these were missing

.github/actions/build-custom-image-with-apko/action.yml
```diff
@@ -49,6 +49,7 @@ runs:
 
     - uses: chainguard-images/actions/apko-publish@main
       with:
+        apko-image: ghcr.io/wolfi-dev/apko@sha256:87b47283433066d19b71e1cb4401be8d04b7d51f4f4263fde1977c3043b9754f
```
cbodonnell, line 52: resolves issue with `latest` apko image https://github.com/replicatedhq/kots/actions/runs/9069367744/job/24919025968?pr=4616#step:3:140

sgalsaleh (approved):

PR #4633, opened 2024-05-21
feat: oras using tls registry

ricardomaraschini:
start using tls when pushing embedded cluster artifacts to the internal registry

### DIFF
--- a/pkg/image/airgap.go
+++ b/pkg/image/airgap.go
@@ -837,12 +837,19 @@ func pushOCIArtifact(opts imagetypes.PushOCIArtifactOptions) error {
 		}),
 	}
 
-	_, err = oras.Copy(context.TODO(), orasFS, opts.Tag, repository, opts.Tag, oras.DefaultCopyOptions)
-	if err != nil {
-		return errors.Wrap(err, "failed to copy")
+	_, err1 := oras.Copy(context.TODO(), orasFS, opts.Tag, repository, opts.Tag, oras.DefaultCopyOptions)
+	if err1 == nil {
+		return nil
 	}
 
-	return nil
+	// try again with plain http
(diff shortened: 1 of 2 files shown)


sgalsaleh (approved):

pkg/image/airgap.go
```diff
@@ -831,14 +836,20 @@ func pushOCIArtifact(opts imagetypes.PushOCIArtifactOptions) error {
...
-	_, err = oras.Copy(context.TODO(), orasFS, opts.Tag, repository, opts.Tag, oras.DefaultCopyOptions)
-	if err != nil {
-		return errors.Wrap(err, "failed to copy")
+	_, err1 := oras.Copy(context.TODO(), orasFS, opts.Tag, repository, opts.Tag, oras.DefaultCopyOptions)
+	if err1 == nil {
+		return nil
 	}
 
-	return nil
+	// try again with plain http
```
sgalsaleh, line 845: i thought we're going to check the existence of the embedded cluster registry tls secret here too? wouldn't this pollute the registry logs with red-herrings?

cbodonnell: ah I believe I misunderstood the ask here. i can try to find a good way to get a k8s client into this function or at the very least a new parameter specifying plainHTTP true/false.

on the logs piece, i'm not seeing additional logs in docker distribution, but that may depend on the registry implementation

sgalsaleh (approved):

pkg/image/airgap.go
cbodonnell: ultimately deciding to leave as-is as TLS is the go-forward in embedded-cluster so it is very unlikely that the non-TLS path will be reached. if it does and that still fails, both errors will be returned for debugging.

PR #4634, opened 2024-05-22
Allow ignoring  errors when loading kotskinds

divolgin:

There was a change in `LoadKotsKinds` that makes this function fail when it cannot parse at least one file. This adds an option to ignore these errors so any valid kotskinds file can be loaded.

Fixes #

```release-note

```


sgalsaleh (approved):

pkg/kotsutil/kots.go
```diff
@@ -666,6 +666,16 @@ func GetKotsKindsPath(archive string) string {
 // it loads the rendered kots kinds if they exist (should always be the case for app version archives created by newer kots versions).
 // otherwise it loads the non-rendered kots kinds (app version archives created by older kots versions).
 func LoadKotsKinds(archive string) (*KotsKinds, error) {
+	return LoadKotsKindsWithOpts(archive, LoadKotsKindsOptions{Strict: true})
```
divolgin, line 669: Not wrapping here because it's a one liner and there is no ambiguity. Wrapping will just make error stack longer.

PR #4661, opened 2024-06-05
ignore failing non strict preflights if skipped

cbodonnell:

This PR fixes an issue where the user could not skip/ignore failing non-strict preflights and deploy from the CLI even if all strict preflights were passing. This impacted both the `install` and `upstream upgrade` commands when using the `--skip-preflights` flag.

https://app.shortcut.com/replicated/story/95957/cannot-skip-ignore-failing-non-strict-preflights-and-deploy-from-the-cli-even-if-all-strict-preflights-pass

```release-note
Fixes an issue where users could not skip or ignore failing non-strict preflights and deploy from the CLI even if all strict preflights were passing when using the `--skip-preflights` flag.
```

### DIFF
--- a/pkg/preflight/preflight.go
+++ b/pkg/preflight/preflight.go
@@ -45,7 +45,7 @@ const (
 	SpecDataKey = "preflight-spec"
 )
 
-func Run(appID string, appSlug string, sequence int64, isAirgap bool, archiveDir string) error {
+func Run(appID string, appSlug string, sequence int64, isAirgap bool, ignoreNonStrict bool, archiveDir string) error {
 	kotsKinds, err := kotsutil.LoadKotsKinds(archiveDir)
 	if err != nil {
 		return errors.Wrap(err, "failed to load rendered kots kinds")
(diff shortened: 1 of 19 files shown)


pkg/store/kotsstore/preflight_store_test.go
```diff
@@ -126,7 +126,7 @@ func Test_hasFailingStrictPreflights(t *testing.T) {
 			name:               "expect false, error when preflightSpec with strict:true analyzer and preflightResultSpec has a empty string",
 			preflightSpecStr:   toSqlString(t, strictTruePreflightSpec),
 			preflightResultStr: gorqlite.NullString{Valid: true, String: ""},
-			want:               true,
+			want:               false,
```
cbodonnell, line 129: This seems like it was an error in the tests. The `name` of this test indicates that we would expect `false` when the preflight results are empty. This makes sense since how could we determine if the version  `hasFailingStrictPreflights` if we have no results to check?

pkg/preflight/preflight.go
```diff
@@ -45,7 +45,7 @@ const (
 	SpecDataKey = "preflight-spec"
 )
 
-func Run(appID string, appSlug string, sequence int64, isAirgap bool, archiveDir string) error {
+func Run(appID string, appSlug string, sequence int64, isAirgap bool, ignoreNonStrict bool, archiveDir string) error {
```
cbodonnell, line 48: Added this variable to the `Run` function that can be populated by the caller based on if we're skipping preflights and if there are strict preflights.

sgalsaleh (approved):

PR #4662, opened 2024-06-05
Remove usage of embedded-cluster-config configmap

sgalsaleh:

Removes usage of embedded-cluster-config configmap to remove the requirement of restoring the embedded cluster operator before the admin console.

Fixes #

```release-note
NONE
```

NONE

### DIFF
--- a/pkg/k8sutil/clientset.go
+++ b/pkg/k8sutil/clientset.go
@@ -1,11 +1,14 @@
 package k8sutil
 
 import (
+	"context"
 	"strconv"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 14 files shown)


pkg/k8sutil/clientset.go
```diff
@@ -134,3 +138,17 @@ func GetDynamicResourceInterface(gvk *schema.GroupVersionKind, namespace string)
 
 	return dr, nil
 }
+
+func GetKubeClient(ctx context.Context) (kbclient.Client, error) {
```
cbodonnell, line 142: Are we able to use the existing `GetClientset` function we have as long as the EC scheme is registered?

cbodonnell (approved):

PR #4668, opened 2024-06-06
fix airgap version check when semver is not enabled

cbodonnell:

This PR fixes an issue where users were unable to upload airgap bundle if the version label was the same as the base version for non-semver channels. This check is intentional for semver-enabled channels, but has been changed for non-semver channels so that only the channel and sequence is considered.

https://app.shortcut.com/replicated/story/106450/unable-to-upload-airgap-bundle-if-the-version-label-is-the-same-as-the-base-version

```release-note
Fixes an issue where airgap uploads would fail with an error indicating that is was already the current version in cases where the version labels matched. This version label check is intentional for semver-enabled channels, but was not for non-semver enabled channels.
```

### DIFF
--- a/pkg/airgap/update.go
+++ b/pkg/airgap/update.go
@@ -192,7 +192,12 @@ func UpdateAppFromPath(a *apptypes.App, airgapRoot string, airgapBundlePath stri
 		return errors.Wrap(err, "failed to read after kotskinds")
 	}
 
-	if err := canInstall(beforeKotsKinds, afterKotsKinds); err != nil {
+	license, err = store.GetStore().GetLatestLicenseForApp(a.ID)
+	if err != nil {
+		return errors.Wrap(err, "failed to get latest license")
+	}
+
+	if err := canInstall(beforeKotsKinds, afterKotsKinds, license); err != nil {
 		return errors.Wrap(err, "cannot install")
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/airgap/update.go
```diff
@@ -246,11 +246,13 @@ func UpdateAppFromPath(a *apptypes.App, airgapRoot string, airgapBundlePath stri
 
 func canInstall(beforeKotsKinds *kotsutil.KotsKinds, afterKotsKinds *kotsutil.KotsKinds) error {
 	var beforeSemver, afterSemver *semver.Version
-	if v, err := semver.ParseTolerant(beforeKotsKinds.Installation.Spec.VersionLabel); err == nil {
-		beforeSemver = &v
-	}
-	if v, err := semver.ParseTolerant(afterKotsKinds.Installation.Spec.VersionLabel); err == nil {
-		afterSemver = &v
+	if afterKotsKinds.License != nil && afterKotsKinds.License.Spec.IsSemverRequired {
```
cbodonnell, line 249: Checking the `afterKotsKinds` here as it seems we would care about the _current_ state of semver on the channel vs what it was for the last sequence. Any reason this might be problematic?

sgalsaleh: should get the license stored in the app table instead of kots kinds, as that would be the most recent and update to date one. you can get that license using `store.GetStore().GetLatestLicenseForApp`

cbodonnell: pushed https://github.com/replicatedhq/kots/pull/4668/commits/d5ded03526e6bfc5d2bf90d191389733f03538e0 to get the latest license from the db

sgalsaleh (approved):

PR #4678, opened 2024-06-12
feat(ec): ability to upgrade the cluster

emosbaugh:

Related https://github.com/replicatedhq/embedded-cluster-operator/pull/218

Fixes #

```release-note

```


pkg/embeddedcluster/upgrade.go
```diff
@@ -0,0 +1,154 @@
...
+
+	embeddedclusterv1beta1 "github.com/replicatedhq/embedded-cluster-kinds/apis/v1beta1"
+	"github.com/replicatedhq/kots/pkg/k8sutil"
+	kotsv1beta1 "github.com/replicatedhq/kotskinds/apis/kots/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8syaml "sigs.k8s.io/yaml"
+)
+
+const (
+	clusterUpgradeTimeout = 5 * time.Minute
```
sgalsaleh, line 23: I don't know if we want a timeout for cluster upgrades as it's not easy to predict its duration. Can vary significantly from one app to another.

pkg/embeddedcluster/upgrade.go
```diff
@@ -0,0 +1,154 @@
...
+)
+
+// runClusterUpgrade will download the new embedded cluster operator binary and run the upgrade
+// command with the provided installation data. This is needed to get the latest
+// embeddedclusterv1beta1 API version. The upgrade command will first upgrade the embedded cluster
+// operator, wait for the CRD to be up-to-date, and then apply the installation object.
+func runClusterUpgrade(ctx context.Context, installationData []byte) error {
+	log.Println("Downloading upgrade binary...")
+
+	bin, err := downloadUpgradeBinary(ctx)
```
sgalsaleh, line 91: This needs to account for airgap and pull the operator binary from the registry.

pkg/embeddedcluster/upgrade.go
emosbaugh: something has to cancel this as it happens in the background and can go forever

sgalsaleh: Oh this is actually just for distributing the infra images and upgrading the operator chart, not the entire cluster. So it's probably fine to have a timeout.

pkg/embeddedcluster/upgrade.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	log.Printf("Upgrading cluster to version %s...", newcfg.Version)
+
+	err = runClusterUpgrade(ctx, installationData)
+	if err != nil {
+		return fmt.Errorf("run cluster upgrade: %w", err)
+	}
+
+	log.Printf("Cluster upgrade to version %s completed successfully", newcfg.Version)
```
sgalsaleh, line 74: This message is a bit inaccurate as the above function will return before the cluster upgrade completes because that will be handled async by the new operator chart. KOTS will watch the installation object status async.

pkg/embeddedcluster/upgrade.go
emosbaugh: done

pkg/embeddedcluster/upgrade.go
emosbaugh: changed

pkg/embeddedcluster/upgrade.go
```diff
@@ -0,0 +1,347 @@
...
+		return "", fmt.Errorf("new download upgrade binary request: %w", err)
+	}
+
+	log.Printf("Downloading upgrade binary from %s...", req.URL)
+
+	httpClient := &http.Client{
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			if strings.Contains(req.URL.Path, "/embedded-cluster-public-files/") {
+				// Workaround as cloudflare complains about unsupported auth method
+				req.Header.Del("Authorization")
```
sgalsaleh, line 183: it's probably good practice to always remove the auth header on redirect. idk why Go doesn't do that automatically. can probably just always delete the auth header without checking the path.

pkg/embeddedcluster/upgrade.go
```diff
@@ -0,0 +1,347 @@
...
+	err := pullArtifact(ctx, srcRepo, dstDir, opts)
+	if err != nil {
+		return fmt.Errorf("pull oci artifact: %w", err)
+	}
+
+	return nil
+}
+
+func embeddedRegistryImageAddress(registryDomain string, srcImage string) (string, error) {
+	imagePath, err := image.PathToRegistryECImage(srcImage, registryDomain)
```
sgalsaleh, line 341: the second parameter is the registry namespace, not the domain. these registry settings should be passed by the deploy function from kots.

pkg/embeddedcluster/upgrade.go
emosbaugh: im fixing this in the proxy https://github.com/replicatedhq/cloudflare-workers/pull/348/files

PR #4681, opened 2024-06-13
save config api for upgrade service

cbodonnell:

This PR adds an API endpoint to the upgrade service for saving config. The handler will trigger preflights to run in the background after successfully saving the config and rendering the base archive.

https://app.shortcut.com/replicated/story/105494/upgrade-service-api-side-of-the-config-page-in-the-deployment-wizard

```release-note

```

### DIFF
--- a/pkg/preflight/execute.go
+++ b/pkg/preflight/execute.go
@@ -1,7 +1,6 @@
 package preflight
 
 import (
-	"encoding/json"
 	"strings"
 	"sync"
 	"time"
@@ -19,7 +18,7 @@ import (
 
 // Execute will Execute the preflights using spec in preflightSpec.
 // This spec should be rendered, no template functions remaining
-func Execute(preflightSpec *troubleshootv1beta2.Preflight, ignorePermissionErrors bool, setProgress, setResults func(b []byte) error) (*types.PreflightResults, error) {
+func Execute(preflightSpec *troubleshootv1beta2.Preflight, ignorePermissionErrors bool, setProgress func(progress map[string]interface{}) error, setResults func(results *types.PreflightResults) error) (*types.PreflightResults, error) {
 	progressChan := make(chan interface{}, 0) // non-zero buffer will result in missed messages
 	defer close(progressChan)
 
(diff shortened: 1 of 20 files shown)


pkg/upgradeservice/preflight/preflight.go
```diff
@@ -0,0 +1,199 @@
...
+	Progress map[string]interface{}  `json:"progress"`
+	Results  *types.PreflightResults `json:"results"`
+}
+
+var PreflightDataFilepath string
+
+func init() {
+	tmpDir, err := os.MkdirTemp("", "preflights")
+	if err != nil {
+		panic(errors.Wrap(err, "failed to create temp dir"))
```
sgalsaleh, line 36:
```suggestion
		panic(errors.Wrap(err, "failed to create preflights data dir"))
```

pkg/preflight/preflight.go
```diff
@@ -141,8 +142,23 @@ func Run(appID string, appSlug string, sequence int64, isAirgap bool, archiveDir
 		var preflightErr error
 		defer func() {
 			if preflightErr != nil {
-				err := setPreflightResult(appID, sequence, &types.PreflightResults{}, preflightErr)
+				preflightResults := &types.PreflightResults{
```
sgalsaleh, line 145: how about combining this with the `setResults` function below? it can also be defined outside of entire function to make it smaller. same for setProgress.

pkg/preflight/execute.go
```diff
@@ -18,7 +18,7 @@ import (
 
 // Execute will Execute the preflights using spec in preflightSpec.
 // This spec should be rendered, no template functions remaining
-func Execute(preflightSpec *troubleshootv1beta2.Preflight, ignorePermissionErrors bool, setProgress func(progress map[string]interface{}) error, setResults func(results *types.PreflightResults) error) (*types.PreflightResults, error) {
+func Execute(preflightSpec *troubleshootv1beta2.Preflight, ignorePermissionErrors bool, setProgress func(progress map[string]interface{}) error, setResults func(results *types.PreflightResults, runError error) error) (*types.PreflightResults, error) {
```
sgalsaleh, line 21: why is `runError` param needed here?

pkg/upgradeservice/preflight/preflight.go
```diff
@@ -144,12 +134,21 @@ func Run(app *apptypes.App, archiveDir string, sequence int64, registrySettings
...
 
-func setPreflightResults(results *types.PreflightResults) error {
+func setPreflightResults(results *types.PreflightResults, runError error) error {
 	preflightData, err := getPreflightData()
 	if err != nil {
 		return errors.Wrap(err, "failed to get preflight data")
 	}
 	preflightData.Results = results
+	if runError != nil {
+		if preflightData.Results.Errors == nil {
```
sgalsaleh, line 144: sorry to be nit picky, but to be consistent with the other implementation, let's update `results` before updating `preflightData.Results`. would probably look cleaner.

sgalsaleh (approved):

PR #4693, opened 2024-06-17
upgrade service preflights api

cbodonnell:

This PR adds two endpoints `POST /preflight/run` and `GET /preflight/result` to support running preflights and fetching preflight results, respectively.

https://app.shortcut.com/replicated/story/105495/upgrade-service-api-side-of-the-preflights-page-in-the-deployment-wizard

```release-note

```


pkg/upgradeservice/preflight/preflight.go
```diff
@@ -142,31 +147,61 @@ func Run(app *apptypes.App, archiveDir string, sequence int64, registrySettings
...
+	// convert to troubleshoot type so we can use the existing function
+	uploadResults := &troubleshootpreflight.UploadPreflightResults{}
+	uploadResults.Results = results.Results
+	for _, e := range results.Errors {
+		uploadResults.Errors = append(uploadResults.Errors, &troubleshootpreflight.UploadPreflightError{
+			Error: e.Error,
+		})
+	}
+	return troubleshootpreflight.HasStrictAnalyzersFailed(uploadResults)
+}
```
cbodonnell, line 186: This felt weird to write, but there are two types that are nearly identical [`PreflightResults`](https://github.com/replicatedhq/kots/blob/2002ad802ee9cabb2595d5b5297f4425219956b8/pkg/preflight/types/types.go#L23-L26) and [`UploadPreflightResults`](https://github.com/replicatedhq/troubleshoot/blob/a9c5baca1464d9e8d87d2817aa540985114530b4/pkg/preflight/types.go#L18-L21), the latter of which is from troubleshoot. I wanted to keep using the same check for failing strict preflights, but if there's a better way to do this, I'm all ears.

pkg/upgradeservice/preflight/preflight.go
```diff
@@ -24,8 +25,8 @@ import (
 )
 
 type PreflightData struct {
-	Progress map[string]interface{}  `json:"progress"`
-	Results  *types.PreflightResults `json:"results"`
+	Progress string                 `json:"progress,omitempty"`
+	Result   *types.PreflightResult `json:"result"`
```
cbodonnell, line 29: Modified this type to match more closely with what we store in the database

pkg/upgradeservice/preflight/preflight.go
```diff
@@ -195,3 +230,10 @@ func setPreflightData(preflightData *PreflightData) error {
 	}
 	return nil
 }
+
+func ResetPreflightData() error {
+	if err := os.RemoveAll(PreflightDataFilepath); err != nil {
+		return errors.Wrap(err, "failed to remove preflight data")
```
sgalsaleh, line 236: maybe don't fail if it doesn't exist?

pkg/upgradeservice/preflight/preflight.go
sgalsaleh: it's fine for now.

pkg/upgradeservice/preflight/preflight.go
```diff
@@ -142,31 +147,61 @@ func Run(app *apptypes.App, archiveDir string, sequence int64, registrySettings
...
+	preflightData.Result = &types.PreflightResult{
+		Result:                     string(resultsBytes),
+		CreatedAt:                  &createdAt,
+		AppSlug:                    appSlug,
+		ClusterSlug:                "this-cluster",
+		Skipped:                    false,
+		HasFailingStrictPreflights: hasFailingStrictPreflights(results),
+	}
+	// clear the progress once the results are set
+	preflightData.Progress = ""
```
sgalsaleh, line 169: nit pick: can this be inside the struct above?

pkg/upgradeservice/preflight/preflight.go
cbodonnell: I believe [`os.RemoveAll`](https://pkg.go.dev/os#RemoveAll) will do this. Just switched it in https://github.com/replicatedhq/kots/pull/4693/commits/dbd8cd02a9acf4a9e7dd2828d9dd38030895647a for this same reason.

pkg/upgradeservice/preflight/preflight.go
cbodonnell: made a change in https://github.com/replicatedhq/kots/pull/4693/commits/7a4d7bb34a02a9b1e766f9607cd4febd0fbfeb9e to address this one

sgalsaleh (approved):

PR #4701, opened 2024-06-20
cli airgap uploads

cbodonnell:

This PR adds a hidden KOTS CLI command `airgap-upload` as well as a corresponding API endpoint to support airgap bundle uploads from the CLI. This supports work for the new KOTS upgrade flow using the upgrade service.

https://app.shortcut.com/replicated/story/105572/review-airgap-updates-via-ec-cli-uses-old-kots

```release-note

```


pkg/handlers/update.go
```diff
@@ -420,3 +430,93 @@ func (h *Handler) GetAdminConsoleUpdateStatus(w http.ResponseWriter, r *http.Req
...
+			return
+		}
+
+		if part.FormName() != "application.airgap" {
+			continue
+		}
+
+		foundAirgapBundle = true
+
+		destDir := filepath.Join(updatechecker.AvailableUpdatesPath, uuid.New().String())
```
cbodonnell, line 485: As it is now, you could theoretically have the same bundle uploaded multiple times and KOTS will not error. We could potentially address this by making the destination path deterministic based on channel ID and sequence or by KOTS storing a map of the currently available versions.

pkg/updatechecker/updatechecker.go
```diff
@@ -774,3 +791,47 @@ func isUpdateDeployable(updateCursor string, updates []upstreamtypes.Update) (bo
...
+		if info.IsDir() {
+			return nil
+		}
+
+		airgapMetadata, err := kotsutil.FindAirgapMetaInBundle(path)
+		if err != nil {
+			return errors.Wrap(err, "failed to find airgap metadata")
+		}
+
+		if airgapMetadata.Spec.AppSlug != app.Slug {
```
sgalsaleh, line 810: should check the channel id as well

pkg/updatechecker/updatechecker.go
```diff
@@ -39,6 +43,15 @@ var jobs = make(map[string]*cron.Cron)
 var mtx sync.Mutex
 var store = storepkg.GetStore()
 
+var AvailableUpdatesPath = filepath.Join(os.TempDir(), "available-updates")
```
sgalsaleh, line 46:
```suggestion
var AvailableUpdatesDir = filepath.Join(os.TempDir(), "available-updates")
```

pkg/archives/archives.go
```diff
@@ -179,3 +179,39 @@ func IsTGZ(b []byte) bool {
 	_, err = tarReader.Next()
 	return err == nil
 }
+
+func CreateFilteredAirgapBundle(airgapBundle string, filesToInclude []string) (io.Reader, error) {
+	buf := bytes.NewBuffer(nil)
```
sgalsaleh, line 184: can this write to a file instead of memory? binaries can be large if we add more in the future.

pkg/archives/archives.go
```diff
@@ -179,3 +179,39 @@ func IsTGZ(b []byte) bool {
...
+	gw := gzip.NewWriter(buf)
+	tw := tar.NewWriter(gw)
+
+	for _, path := range filesToInclude {
+		contents, err := GetFileFromAirgap(path, airgapBundle)
```
sgalsaleh, line 189: an optimization here would be to iterate over the bundle only once and write desired files to disk directly rather than loading each file in memory and having to iterate over the bundle files for each file.

cmd/kots/cli/airgap-upload.go
```diff
@@ -0,0 +1,160 @@
...
+			appSlug := args[0]
+			log := logger.NewCLILogger(cmd.OutOrStdout())
+
+			filesToInclude := []string{
+				"airgap.yaml",
+				"app.tar.gz",
+			}
+
+			if util.IsEmbeddedCluster() {
+				filesToInclude = append(filesToInclude, "embedded-cluster/artifacts/kots")
```
cbodonnell, line 56: TODO: figure out what the actual name of the KOTS binary will be

pkg/updatechecker/updatechecker.go
```diff
@@ -752,19 +767,23 @@ func GetAvailableUpdates(kotsStore storepkg.Store, app *apptypes.App, license *k
...
 		}
 		if updates[i].IsRequired {
-			requiredUpdates = append(requiredUpdates, updates[i])
+			requiredUpdates = append(requiredUpdates, updates[i].VersionLabel)
 		}
 	}
+	return canDeployVersion(requiredUpdates)
+}
+
+func canDeployVersion(requiredUpdates []string) (bool, string) {
```
sgalsaleh, line 782: how about renaming this function and use it only if there are `requiredUpdates`?
```suggestion
func getNonDeployableCause(requiredUpdates []string) (bool, string) {
```
that way it's just about building the actual message.

cbodonnell: great idea!

pkg/updatechecker/updatechecker.go
```diff
@@ -774,3 +793,47 @@ func isUpdateDeployable(updateCursor string, updates []upstreamtypes.Update) (bo
...
+		airgapMetadata, err := kotsutil.FindAirgapMetaInBundle(path)
+		if err != nil {
+			return errors.Wrap(err, "failed to find airgap metadata")
+		}
+
+		if airgapMetadata.Spec.AppSlug != license.Spec.AppSlug && airgapMetadata.Spec.ChannelID != license.Spec.ChannelID {
+			return nil
+		}
+
+		missingPrereqs, err := airgap.GetMissingRequiredVersions(app, airgapMetadata)
```
sgalsaleh, line 816: this should just use the same `isUpdateDeployable` function above. `GetMissingRequiredVersions` is not applicable here as we don't check against deployment history anymore.

pkg/updatechecker/updatechecker.go
```diff
@@ -752,25 +767,76 @@ func GetAvailableUpdates(kotsStore storepkg.Store, app *apptypes.App, license *k
...
+		if info.IsDir() {
+			return nil
+		}
+
+		airgapMetadata, err := kotsutil.FindAirgapMetaInBundle(path)
+		if err != nil {
+			return errors.Wrap(err, "failed to find airgap metadata")
+		}
+
+		if airgapMetadata.Spec.AppSlug != license.Spec.AppSlug && airgapMetadata.Spec.ChannelID != license.Spec.ChannelID {
```
sgalsaleh, line 816: would probably be better to split them up
```suggestion
		if airgapMetadata.Spec.AppSlug != license.Spec.AppSlug || airgapMetadata.Spec.ChannelID != license.Spec.ChannelID {
```

pkg/handlers/update.go
```diff
@@ -239,7 +235,27 @@ func (h *Handler) GetAvailableUpdates(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	latestLicense, _, err := license.Sync(app, "", false)
+	latestLicense, err := store.GetLatestLicenseForApp(app.ID)
```
sgalsaleh, line 238: looks like this is only used for airgap? can probably be moved inside the if statement below

PR #4720, opened 2024-06-26
feat(ec): ability to upgrade the cluster

emosbaugh:

Original PR https://github.com/replicatedhq/kots/pull/4678

Related https://github.com/replicatedhq/embedded-cluster-operator/pull/218

Fixes #

```release-note

```


pkg/embeddedcluster/oras_test.go
```diff
@@ -0,0 +1,38 @@
...
+	tests := []struct {
+		name    string
+		args    args
+		wantErr bool
+	}{
+		{
+			name: "test1",
+			args: args{
+				ctx:     context.Background(),
+				srcRepo: "ttl.sh/ethan/embedded-cluster-operator-bin:24h",
```
sgalsaleh, line 23: is this what we want to use for testing?

pkg/embeddedcluster/upgrade.go
```diff
@@ -0,0 +1,380 @@
...
+		}
+		args = append(args, "--local-artifact-mirror-image", localArtifactMirrorImage)
+	}
+	args = append(args, "--installation", "-")
+
+	cmd := exec.CommandContext(ctx, bin, args...)
+	cmd.Stdin = strings.NewReader(string(installationData))
+	pr, pw := io.Pipe()
+	defer pw.Close()
+	cmd.Stdout = pw
```
sgalsaleh, line 158: why not just 
```suggestion
	cmd.Stdout = os.Stdout
```

pkg/embeddedcluster/upgrade.go
```diff
@@ -0,0 +1,380 @@
...
+// operator, wait for the CRD to be up-to-date, and then apply the installation object.
+func runClusterUpgrade(
+	ctx context.Context, k8sClient kubernetes.Interface,
+	in *embeddedclusterv1beta1.Installation, registrySettings registrytypes.RegistrySettings,
+	license kotsv1beta1.License, versionLabel string,
+) error {
+	var bin string
+
+	if in.Spec.AirGap {
+		image := in.Spec.Artifacts.AdditionalArtifacts["operator"]
```
sgalsaleh, line 109: `image` is a bit confusing. how about `location`?
```suggestion
		location := in.Spec.Artifacts.AdditionalArtifacts["operator"]
```

pkg/embeddedcluster/upgrade.go
```diff
@@ -0,0 +1,380 @@
...
+		if image == "" {
+			return fmt.Errorf("missing operator image in airgap artifacts")
```
sgalsaleh, line 111:
```suggestion
			return fmt.Errorf("missing operator binary in airgap artifacts")
```

pkg/embeddedcluster/oras_test.go
emosbaugh: the test is prefixed with an x_ so it does not run. do you want me to remove it?

pkg/embeddedcluster/upgrade.go
emosbaugh: this is equivalent. why would i change it?

pkg/embeddedcluster/oras_test.go
sgalsaleh: if it does not run, then we should probably just remove it yeah.

pkg/embeddedcluster/upgrade.go
sgalsaleh: i meant for simplicity without having the below goroutine entirely.

sgalsaleh: something like this https://github.com/replicatedhq/kots/blob/kots-upgrader/pkg/upgradeservice/process.go#L95-L96

sgalsaleh: I guess your way has the benefit of distinguishing it's the binary's output

sgalsaleh (approved):

pkg/embeddedcluster/upgrade.go
```diff
@@ -0,0 +1,380 @@
...
+	kotsv1beta1 "github.com/replicatedhq/kotskinds/apis/kots/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"oras.land/oras-go/v2/registry/remote/auth"
+	"oras.land/oras-go/v2/registry/remote/credentials"
+	k8syaml "sigs.k8s.io/yaml"
+)
+
+const (
+	clusterUpgradeTimeout = 5 * time.Minute
```
sgalsaleh, line 35: Is this timeout sufficient to upgrade the operator chart and run the LAM jobs specially on a large number of nodes?

emosbaugh: ive increased it to 15 minutes

pkg/embeddedcluster/upgrade.go
```diff
@@ -0,0 +1,380 @@
...
+
+// startClusterUpgrade will create a new installation with the provided config.
+func startClusterUpgrade(
+	ctx context.Context, newcfg embeddedclusterv1beta1.ConfigSpec,
+	artifacts *embeddedclusterv1beta1.ArtifactsLocation,
+	registrySettings registrytypes.RegistrySettings,
+	license kotsv1beta1.License, versionLabel string,
+) error {
+	// TODO(upgrade): put a lock here to prevent multiple upgrades at the same time
+
```
laverya, line 46: Global mutex? Should be easy, no?

laverya: DB lock would be better but then you have issues with failures

pkg/embeddedcluster/upgrade.go
```diff
@@ -0,0 +1,380 @@
...
+			ClusterID:                 current.Spec.ClusterID,
+			MetricsBaseURL:            current.Spec.MetricsBaseURL,
+			HighAvailability:          current.Spec.HighAvailability,
+			AirGap:                    current.Spec.AirGap,
+			Artifacts:                 artifacts,
+			Config:                    &newcfg,
+			EndUserK0sConfigOverrides: current.Spec.EndUserK0sConfigOverrides,
+			BinaryName:                current.Spec.BinaryName,
+			LicenseInfo:               &embeddedclusterv1beta1.LicenseInfo{IsDisasterRecoverySupported: license.Spec.IsDisasterRecoverySupported},
+		},
```
laverya, line 82: At some point we should start doing spec = current.spec and then just replacing the artifacts and configs

laverya (approved):

sgalsaleh (approved):

PR #4723, opened 2024-06-27
disable update checker in EC installations

laverya:

This disables the automated update checker in EC installations to prevent automated updates.

It will not disable automated updates if they were enabled before embedded cluster was updated to this version.

Fixes #

```release-note

```


laverya:
this might not be the best place to change it, might be better to do so farther inside the update check

web/src/Root.tsx
```diff
@@ -264,6 +264,8 @@ const Root = () => {
           return;
         }
 
+        console.log("received data", JSON.stringify(data));
```
sgalsaleh, line 267: cleanup

sgalsaleh (approved):

PR #4742, opened 2024-07-05
add preferred channel slug

FourSigma:

Fixes #

```release-note

```

### DIFF
--- a/cmd/kots/cli/install.go
+++ b/cmd/kots/cli/install.go
@@ -163,6 +163,11 @@ func InstallCmd() *cobra.Command {
 
 			upstream := pull.RewriteUpstream(args[0])
 
+			preferredChannelSlug := ""
+			if len(args) > 1 {
+				preferredChannelSlug = args[1]
+			}
+
 			namespace := v.GetString("namespace")
 
 			if namespace == "" {
@@ -252,6 +257,7 @@ func InstallCmd() *cobra.Command {
 
 			deployOptions := kotsadmtypes.DeployOptions{
 				Namespace:              namespace,
+				PreferredChannelSlug:   preferredChannelSlug,
 				Context:                v.GetString("context"),
 				SharedPassword:         sharedPassword,
 				ApplicationMetadata:    applicationMetadata.Manifest,
--- a/pkg/kotsadm/objects/configmaps_objects.go
+++ b/pkg/kotsadm/objects/configmaps_objects.go
@@ -12,6 +12,7 @@ import (
 
 func KotsadmConfigMap(deployOptions types.DeployOptions) *corev1.ConfigMap {
 	data := map[string]string{
+		"preferred-channel-slug":    fmt.Sprintf("%v", deployOptions.PreferredChannelSlug),
 		"kots-install-id":           fmt.Sprintf("%v", deployOptions.InstallID),
 		"initial-app-images-pushed": fmt.Sprintf("%v", deployOptions.AppImagesPushed),
 		"skip-preflights":           fmt.Sprintf("%v", deployOptions.SkipPreflights),
--- a/pkg/kotsadm/types/deployoptions.go
+++ b/pkg/kotsadm/types/deployoptions.go
@@ -10,6 +10,7 @@ import (
 
 type DeployOptions struct {
 	Namespace              string
+	PreferredChannelSlug   string
 	Context                string
 	SharedPassword         string
 	SharedPasswordBcrypt   string


cmd/kots/cli/install.go
```diff
@@ -163,6 +163,11 @@ func InstallCmd() *cobra.Command {
 
 			upstream := pull.RewriteUpstream(args[0])
 
+			preferredChannelSlug := ""
+			if len(args) > 1 {
+				preferredChannelSlug = args[1]
```
divolgin, line 168: This should be in the same arg as app slug, as shown in the docs: https://docs.replicated.com/reference/kots-cli-install#examples

```
kubectl kots install kots-sentry/stable
```

sgalsaleh: +1

pkg/replicatedapp/api.go
```diff
@@ -40,8 +44,37 @@ type LicenseData struct {
 	License      *kotsv1beta1.License
 }
 
+func GetPreferredChannelSlug() (string, error) {
```
sgalsaleh, line 47: I would suggest using the existing `GetInstallationParams` function: https://github.com/replicatedhq/kots/blob/2dae2bf18b640b2360f85e669228804f5d582a8e/pkg/kotsutil/kots.go#L1144

sgalsaleh: Example usage: https://github.com/replicatedhq/kots/blob/2dae2bf18b640b2360f85e669228804f5d582a8e/pkg/automation/automation.go#L238-L241

sgalsaleh: Also, if you don't mind, can you just have that `GetInstallationParams` always use `kotsadmtypes.KotsadmConfigMap` and not have to accept it as a parameter?

pkg/replicatedapp/api.go
```diff
@@ -40,8 +44,37 @@ type LicenseData struct {
...
+
 func GetLatestLicense(license *kotsv1beta1.License) (*LicenseData, error) {
+	channelSlug, err := GetPreferredChannelSlug()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get preferred channel slug")
+	}
+
 	url := fmt.Sprintf("%s/license/%s", license.Spec.Endpoint, license.Spec.AppSlug)
+	if channelSlug != "" {
+		url = fmt.Sprintf("%s/%s", url, channelSlug)
```
sgalsaleh, line 76: just to confirm, is the idea that this decision will be recorded by replicated.app and this channel will be used automatically by replicated.app for subsequent requests?

cmd/kots/cli/install.go
divolgin: yes, and it's optional. should default to "stable"

pandemicsyn: ah cool 👍 will send along one more update to use the default.

pkg/replicatedapp/api.go
pandemicsyn: fyi - went ahead updated `RemoveAppVersionLabelFromInstallationParams` too (but easy to undo if I shouldn't have).

sgalsaleh (approved):

divolgin (dismissed):

pkg/replicatedapp/api.go
pandemicsyn: @sgalsaleh i think the answer is yes. At least for requests to this endpoint, I'm not sure how this plays out in other areas. But hoping @marccampbell or maybe @divolgin can confirm. 

Over on the pending replicated.app pr from Siva [he mentioned](https://github.com/replicatedhq/vandoor/pull/5813/files/f26385692619093bd8f051825242d42784963cf6#r1667019824):

> We now check for the existence of a row in the license_instance table. If it exists, we use the channelId from that table to generate the license in L245. This prevents the downstream from switching channels once initialized and recorded on the license_instance table.

And does look like it actually does that. If there was a licenseInstance entry it will use that channel id instead:

```javascript
license = await stores.licenseStore.getLicenseFromIdAndChannelSlugId(licenseInstance.licenseId, licenseInstance.downstreamChannelId);
```

divolgin: Hard to comment on code that has been removed 😄 :
- [Handler](https://github.com/replicatedhq/vandoor/pull/5813/files/f26385692619093bd8f051825242d42784963cf6#diff-2665a6a6cd084fcdf907fb35a2e73b113fd435d3fa642c20f7cefa3155b93df2R191) still has channel slug in it while client does not?: `@Get("/:appSlug/:channelSlug")`
- Instance data is recoded asynchronously and may be delayed.
- When channel on a license changes, instance can make any number of requests before reporting new channel
    - There is also a distinction between what channel license is assigned to and what channel is currently deployed
- This should be tested e2e with a channel other than "stable" with a brand new instance

sgalsaleh (approved):

pandemicsyn:
closing in favor of (wip) #4767

PR #4756, opened 2024-07-11
feat: use an upgrade service deployment mechanism for embedded clusters

sgalsaleh:
#### What this PR does:

This PR changes the deployment mechanism for embedded clusters so that on upgrades, the version of KOTS / embedded cluster included in the release is used instead of the currently deployed version of KOTS / embedded cluster to fetch, render, and deploy the release.

It also changes the versioning mechanism so it is more similar to how Helm works where a new version/revision is created only when a change is actually deployed. When you deploy, you take all your inputs (manifests, config values, charts, registry settings, etc.), put them together, and deploy the new revision. There is no pre-rendering of versions that never get deployed, and you don’t have to make multiple changes to change multiple inputs.

**NOTE:** This new versioning / deployment mechanism is currently for embedded clusters only, and the plan is to adopt it for existing/kURL clusters later on.

#### Why we need it:

Main problem targeted by this change:

A key design goal for embedded cluster is that the infra and the app are coupled together, so the vendor tests and deploys their release and expects a particular version of their application to work on a particular version of embedded cluster and KOTS. During upgrades, however, the currently deployed version of KOTS pulls in and renders newly available versions. These newly available versions might specify a different embedded cluster version that includes a different KOTS version, which means the vendor might make use of new KOTS functionality. The old version of KOTS may be unable to render the newly available version, which can block the end-user from upgrading unless the vendor creates new releases that first upgrade KOTS and only later leverage the new functionality available in that version.

Other problems that are solved or improved by this approach include:

- Unblocks the work for adding the ability to retract a release that has already been promoted to a channel.
- Confusion around the version history page, its ordering, etc. The UX of the version history page can be very challenging beyond basic use.
- We have a lot of complex and difficult logic for the version history page to handle features like semantic versioning, required releases, pending versions, etc. We can eventually eliminate all of that code with this approach.

Fixes [SC-105486](https://app.shortcut.com/replicated/epic/105486)

```release-note

```

TBD


hack/dev/skaffold.Dockerfile
```diff
@@ -2,7 +2,6 @@ FROM kotsadm:cache AS builder
 
 ENV PROJECTPATH=/go/src/github.com/replicatedhq/kots
 WORKDIR $PROJECTPATH
-RUN mkdir -p web/dist && touch web/dist/README.md
```
laverya, line 5: are these not needed anymore? I thought this ensured the web stuff worked

migrations/tables/api_task_status.yaml
```diff
@@ -1,4 +1,3 @@
-## no longer used, must keep for migrations to complete
```
laverya, line 1: is this used now, or was this removed mistakenly?

sgalsaleh: It is now used again.

hack/dev/skaffold.Dockerfile
sgalsaleh: This is for the dev env and web is served from the kotsadm-web service not the dist directory.

pkg/embeddedcluster/upgrade.go
```diff
@@ -0,0 +1,381 @@
...
+	kotsv1beta1 "github.com/replicatedhq/kotskinds/apis/kots/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"oras.land/oras-go/v2/registry/remote/auth"
+	"oras.land/oras-go/v2/registry/remote/credentials"
+	k8syaml "sigs.k8s.io/yaml"
+)
+
+const (
+	clusterUpgradeTimeout = 15 * time.Minute
```
laverya, line 35: could we base this on the number of nodes somehow?

pkg/kotsadmupstream/upstream.go
```diff
@@ -36,8 +36,8 @@ func DownloadUpdate(appID string, update types.Update, skipPreflights bool, skip
 		go func() {
 			for {
 				select {
-				case <-time.After(time.Second):
-					if err := store.GetStore().UpdateTaskStatusTimestamp(taskID); err != nil {
+				case <-time.After(time.Second * 2):
```
laverya, line 39: I see a lot of these "one second to two seconds" changes, and that worries me
are we actually hitting issues with one second, and if so does two give us enough margin for customer environments?

sgalsaleh: No issues with one second. Just that 1 second is unnecessarily short.

pkg/embeddedcluster/upgrade.go
sgalsaleh: hmm, so this timeout is for distributing the infra artifacts and upgrading ECO. @emosbaugh thoughts?

emosbaugh: I think we should remove this and find a logical way to cancel the operation through this context

https://github.com/replicatedhq/kots/pull/4756/files/e7559fa89645292447dad8fda47473da02e44543#diff-96da550adbaf3fbec7af8856e898e960a8ec9d2be4600202dc1a19f46f7d7d3fR132

laverya (approved):
I have not adequately reviewed the TS/TSX stuff yet, but go looks ok

laverya (dismissed):

pkg/embeddedcluster/upgrade.go
sgalsaleh: removed

alicenstar (approved):
Approved from front end perspective. Didn't see anything that really jumped out at me. Love to see the organizational and TypeScript improvements! Would be nice to add unit tests in the future. But no blockers from me.

alicenstar:
Sharing the Looms that were shared with me (for posterity):

- https://www.loom.com/share/8f7b9122e5744276861207bd677eef00?sid=1e54e1e8-0698-45b4-9357-8e791f95f373
- https://www.loom.com/share/747a8c96f33242cd87aed34ef10ce797?sid=48088689-ef36-48ec-8d2c-73e6e1d4b7b4

PR #4794, opened 2024-07-30
Embedded Cluster upgrade: Rerun preflights when config has changed

miaawong:
#### What this PR does / why we need it: https://app.shortcut.com/replicated/story/108899/preflights-don-t-rerun-when-going-back-to-config-page-changing-something-and-then-proceeding-to-preflights-again
https://www.loom.com/share/712881200a9e4047946e7dd3b0919d33?sid=304dcb04-2459-4d0c-a62f-84b476dddae0

Fixes #

```release-note
Reruns preflight checks when user updates config in Embedded Cluster upgrade
```

### DIFF
--- a/web/src/components/upgrade_service/PreflightChecks.tsx
+++ b/web/src/components/upgrade_service/PreflightChecks.tsx
@@ -12,6 +12,7 @@ import { useGetPrelightResults, useRerunPreflights } from "./hooks/index";
 
 import { KotsParams } from "@types";
 import { useUpgradeServiceContext } from "./UpgradeServiceContext";
+import { isEqual } from "lodash";
 
 const PreflightCheck = ({
   setCurrentStep,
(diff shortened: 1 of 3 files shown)


web/src/components/upgrade_service/PreflightChecks.tsx
```diff
@@ -43,6 +48,13 @@ const PreflightCheck = ({
...
   useEffect(() => {
     setCurrentStep(1);
+
+    if (
+      !isEqual(prevConfig, config) &&
+      preflightCheck?.preflightResults.length > 0
+    ) {
+      rerunPreflights();
+    }
   }, []);
```
sgalsaleh, line 58: wouldn't this rerun preflights whenever the state changes? should there be a separate `useEffect` that only captures `config` and `prevConfig`?

sgalsaleh (approved):

web/src/components/upgrade_service/PreflightChecks.tsx
miaawong: @sgalsaleh With an empty dependency, it will only run once, when the page mounts.

sgalsaleh: Ah ok then

PR #4795, opened 2024-07-31
Fix error in Cluster Management page when clicking continue 

miaawong:
https://app.shortcut.com/replicated/story/107731/uncaught-in-promise-error-unable-to-confirm-cluster-managementuncaught-in-promise-error-unable-to-confirm-cluster-management
https://app.shortcut.com/replicated/story/108918/preflights-box-is-too-big-during-initial-install
https://www.loom.com/share/1efd8f1e2c094a919e05829c8e41c026

Fixes # [sc-108918] [sc-107731]
This PR 
- remove the check for pending versions in pending_cluster_management status, so now it will not fail.
- make preflight box smaller and remove decimal.
- Fix going back on preflight page will now go to config page, not cluster management.  

```release-note
Improve UI of Preflight Results page 
Fix error when clicking continue in Cluster Management 
```

### DIFF
--- a/web/src/components/apps/EmbeddedClusterManagement.tsx
+++ b/web/src/components/apps/EmbeddedClusterManagement.tsx
@@ -370,32 +370,6 @@ const EmbeddedClusterManagement = ({
   }, [nodesData?.nodes?.toString(), hasNodesChanged]);
   // #endregion
 
-  // temporary redirect logic - when user has clicked continue during initial install and went back to the cluster management page
-  useEffect(() => {
-    if (fromLicenseFlow) {
-      const appNeedsConfiguration =
-        app?.downstream?.pendingVersions?.length > 0;
-
-      if (appNeedsConfiguration) {
-        const downstream = app.downstream;
-        const firstVersion = downstream.pendingVersions.find(
-          (version: Version) => version?.sequence === 0
-        );
-        if (
-          firstVersion?.status === "pending_preflight" ||
-          firstVersion?.status === "pending"
-        ) {
-          navigate(`/${app.slug}/preflight`);
-          return;
-        }
-        if (firstVersion?.status === "pending_config") {
-          navigate(`/${app.slug}/config`);
-          return;
-        }
-      }
-    }
-  }, []);
-
   const onContinueClick = async () => {
     const res = await fetch(
       `${process.env.API_ENDPOINT}/embedded-cluster/management`,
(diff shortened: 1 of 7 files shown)


web/src/components/apps/EmbeddedClusterManagement.tsx
```diff
@@ -369,6 +370,29 @@ const EmbeddedClusterManagement = ({
...
+    const appNeedsConfiguration = app?.downstream?.pendingVersions?.length > 0;
+
+    if (appNeedsConfiguration) {
+      const downstream = app.downstream;
+      const firstVersion = downstream.pendingVersions.find(
+        (version: Version) => version?.sequence === 0
+      );
+      if (
+        firstVersion?.status === "pending_preflight" ||
+        firstVersion?.status === "pending"
```
sgalsaleh, line 384: shouldn't this redirect to the dashboard if the status is `pending`?

web/src/components/apps/EmbeddedClusterManagement.tsx
```diff
@@ -369,6 +370,29 @@ const EmbeddedClusterManagement = ({
   }, [nodesData?.nodes?.toString(), hasNodesChanged]);
   // #endregion
 
+  // temporary redirect logic - when user has clicked continue during initial install and went back to the cluster management page
+  useEffect(() => {
+    const appNeedsConfiguration = app?.downstream?.pendingVersions?.length > 0;
```
sgalsaleh, line 375: wouldn't this redirect even if it's not the initial install workflow? i can have many deployed versions and some pending versions, which would then redirect every time i visit the cluster management page later after the initial install, right?

web/src/components/apps/EmbeddedClusterManagement.tsx
miaawong: @sgalsaleh Once preflight checks are done, before clicking deploy. the status is "pending".
Screenshot 2024-08-01 at 12 14 28 PM
Redirecting to dashboard doesn't seem right the user hasn't kicked off deployment.

web/src/components/apps/EmbeddedClusterManagement.tsx
miaawong: @sgalsaleh ah you're right. Updated to have a check if it's `fromLicenseFlow`

sgalsaleh (approved):

PR #4799, opened 2024-08-02
Embedded Cluster: Improve UX for license installation process

miaawong:

Add loading page when waiting for license to install
Redirect users when trying to access /upload-license

https://app.shortcut.com/replicated/story/107723/if-i-visit-the-admin-console-with-trailing-slash-i-am-not-guided-through-the-first-time-install-wizard-thingy#activity-109532
This ended up being hard to make a loom for. The flow of this is now
In EC - we will redirect the user if they land on license upload page
they will be redirected to an intermediate screen /cluster/loading which looks like this
Screenshot 2024-08-02 at 3 17 04 PM

from there it will show the loading message and wait for the license to install then it will either

redirect to dashboard if there is an app with current version or if there's multiple pending versions. 
redirect to /cluster/management if there is a pending version

Fixes #

```release-note
Improve UX for license installation process
```

### DIFF
--- a/web/src/components/apps/AppDetailPage.tsx
+++ b/web/src/components/apps/AppDetailPage.tsx
@@ -99,8 +99,10 @@ function AppDetailPage(props: Props) {
     // navigate to first app if available
     if (appsList && appsList?.length > 0) {
       navigate(`/app/${appsList[0].slug}`, { replace: true });
-    } else if (Utilities.isLoggedIn()) {
+    } else if (Utilities.isLoggedIn() && !props.isEmbeddedCluster) {
       navigate("/upload-license", { replace: true });
+    } else if (Utilities.isLoggedIn && props.isEmbeddedCluster) {
+      navigate("/cluster/loading", { replace: true });
     } else {
       navigate("/secure-console", { replace: true });
     }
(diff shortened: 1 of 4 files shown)


web/src/components/apps/AppDetailPage.tsx
```diff
@@ -99,8 +99,10 @@ function AppDetailPage(props: Props) {
     // navigate to first app if available
     if (appsList && appsList?.length > 0) {
       navigate(`/app/${appsList[0].slug}`, { replace: true });
-    } else if (Utilities.isLoggedIn()) {
+    } else if (Utilities.isLoggedIn() && !props.isEmbeddedCluster) {
       navigate("/upload-license", { replace: true });
+    } else if (Utilities.isLoggedIn && props.isEmbeddedCluster) {
```
sgalsaleh, line 104: not sure if this affects the behavior you tested, but:
```suggestion
    } else if (Utilities.isLoggedIn() && props.isEmbeddedCluster) {
```

miaawong: @sgalsaleh fixed and tested.

sgalsaleh (approved):

PR #4823, opened 2024-08-12
fix: adjust kots install/pull fallbacks when no channel slug is requested

pandemicsyn:

Rather than defaulting to `stable` when no channel slug is specified during install's or pull's - use whatever channel is marked as the default in the license (or the first channel when only one is present, which is always the default anyway).

Additionally, air gapped install's started from pre-multi-channel licenses (where we couldn't update the license - because we are air gapped) will still resolve to use the sole channel id in the license.   

Fixes https://app.shortcut.com/replicated/story/110362/adjust-kots-install-pull-fallbacks-when-no-channel-slug-is-requested

```release-note

- Updates kots `install` & `pull` behaviors  to use the default channel of the license when no channel slug is specified
```

### DIFF
--- a/pkg/kotsutil/kots.go
+++ b/pkg/kotsutil/kots.go
@@ -1602,6 +1602,10 @@ func GetECVersionFromAirgapBundle(airgapBundle string) (string, error) {
 }
 
 func FindChannelIDInLicense(requestedSlug string, license *kotsv1beta1.License) (string, error) {
+	if requestedSlug == "" { // an install started with a legacy license would not have had a channel slug
+		return GetDefaultChannelIDFromLicense(license), nil
+	}
+
 	matchedChannelID := ""
 	if requestedSlug != "" {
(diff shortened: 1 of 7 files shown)


pkg/kotsutil/kots.go
```diff
@@ -1602,6 +1602,10 @@ func GetECVersionFromAirgapBundle(airgapBundle string) (string, error) {
 }
 
 func FindChannelIDInLicense(requestedSlug string, license *kotsv1beta1.License) (string, error) {
+	if requestedSlug == "" { // an install started with a legacy license would not have had a channel slug
```
sgalsaleh, line 1605: isn't this already handled in this function below?
```
        } else { // this is an install from before the channel slug was added to the configmap
		logger.Debug("requested channel slug not found in configmap, using top level channel id from license")
		matchedChannelID = license.Spec.ChannelID
	}
```

PR #4878, opened 2024-09-11
add PrivateCACert template function

laverya:

```release-note
Add a new template function 'PrivateCACert' that returns the name of a configmap containing additional CA certificates provided by the end user.
```

https://github.com/replicatedhq/replicated-docs/pull/2633

### DIFF
--- a/pkg/template/static_context.go
+++ b/pkg/template/static_context.go
@@ -110,7 +110,7 @@ func (ctx StaticCtx) FuncMap() template.FuncMap {
 
 	funcMap["Lookup"] = ctx.lookup
 
-	funcMap["PrivateCACert"] = ctx.additionalTLSCACert
+	funcMap["PrivateCACert"] = ctx.privateCACert
 
 	return funcMap
 }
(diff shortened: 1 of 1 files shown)


sgalsaleh (dismissed):

pkg/template/static_context.go
```diff
@@ -110,6 +110,8 @@ func (ctx StaticCtx) FuncMap() template.FuncMap {
 
 	funcMap["Lookup"] = ctx.lookup
 
+	funcMap["AdditionalTLSCACert"] = ctx.additionalTLSCACert
```
ajp-io, line 113: let's match the flag so it's more intuitive to link this all together (not a suggestion you would commit since I'm just changing this in one place)
```suggestion
	funcMap["PrivateCACert"] = ctx.additionalTLSCACert
```

pkg/template/static_context.go
```diff
@@ -676,3 +678,8 @@ func (ctx StaticCtx) lookup(apiversion string, resource string, namespace string
 	}
 	return obj
 }
+
+func (ctx StaticCtx) additionalTLSCACert() string {
```
sgalsaleh, line 682: can you please rename this function to match the template name?

sgalsaleh (approved):

PR #4879, opened 2024-09-11
Move Dev Environment to MacOS

sgalsaleh:

Changes the dev environment to run locally on MacOS instead of Github Codespaces.

Benefits:

- Much faster feedback loop
- Better caching
- Ability to integrate with and develop against the Embedded Cluster project

NONE

```release-note
NONE
```

NONE

#### Demo

https://www.loom.com/share/79da18e8cd5147efa2dc1811664b437a?sid=7600731b-3deb-40cc-ab7c-932380dd5184


Makefile
```diff
@@ -1,30 +1,13 @@
 include Makefile.build.mk
-CURRENT_USER := $(if $(GITHUB_USER),$(GITHUB_USER),$(shell id -u -n))
```
emosbaugh, line 2: any reason to remove this? will it still work in actions?

Makefile
```diff
@@ -1,30 +1,13 @@
...
-.PHONY: capture-start-time
-capture-start-time:
-	@echo $$(date +%s) > start-time
-
-.PHONY: report-metric
-report-metric:
-	@$(if ${DD_API_KEY}, $(call sendMetrics))
-	@rm start-time
+OS ?= linux
+ARCH ?= $(shell go env GOARCH)
```
emosbaugh, line 10: the go build commands do not respect these vars. is that intended?

README.md
```diff
@@ -61,22 +58,65 @@ cosign verify-blob --key sbom/key.pub --signature sbom/kots-sbom.tgz.sig sbom/ko
 
 # Development
 
-## Github Codespaces
+### Requirements
+
+- MacOS
```
emosbaugh, line 63: mac is required?

dev/dockerfiles/kotsadm/Dockerfile.local
```diff
@@ -1,32 +1,10 @@
...
 
-ENV PROJECTPATH=/go/src/github.com/replicatedhq/kots
-WORKDIR $PROJECTPATH
-COPY Makefile ./
-COPY Makefile.build.mk ./
-COPY go.mod go.sum ./
-COPY cmd ./cmd
-COPY pkg ./pkg
-COPY web/webcontent.go ./web/webcontent.go
+RUN go install github.com/go-delve/delve/cmd/dlv@v1.22.1
```
emosbaugh, line 3:
```suggestion
RUN go install github.com/go-delve/delve/cmd/dlv@v1.23.0
```

dev/manifests/kotsadm/deployment.yaml
```diff
@@ -6,28 +6,43 @@ metadata:
...
         - name: kotsadm
+          image: kotsadm-api-dev
+          imagePullPolicy: Never
+          ports:
+            - name: http
+              containerPort: 3000
+            - name: dlv
+              containerPort: 30001
+          resources:
+            limits:
```
emosbaugh, line 35: any reason to limit this in dev?

dev/scripts/up.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+set -e
```
emosbaugh, line 3: Its always best to start with `-euo pipefail` to prevent bugs down the line

README.md
sgalsaleh: yes

dev/manifests/kotsadm/deployment.yaml
sgalsaleh: yes, to mimic prod as there were times where that bit us. i guess we remove the limits on "up" so that might be sufficient?

Makefile
sgalsaleh: yes because the kots cli will typically run on your mac, not inside the container. the one inside the container gets built differently.

Makefile
sgalsaleh: this was added for github codespaces, and that'll no longer be a supported dev env.

dev/dockerfiles/kotsadm/Dockerfile.local
sgalsaleh: i think there was a compatibility issue that one of the CRE engineers had to deal with when using the latest version, so i didn't wanna upgrade it.

dev/scripts/up.sh
sgalsaleh: i had it like that, but it was failing for me for some reason. i can try it again

emosbaugh (approved):

PR #4909, opened 2024-09-19
feat(ec): support for custom admin console and lam ports

emosbaugh:

So restore and join can use the correct ports.

```release-note

```


emosbaugh:
Failing test is unrelated but needs to be fixed in a follow up pr

sgalsaleh (approved):

PR #4934, opened 2024-10-03
chore: Add check for replicated.app to support bundle specs missing the check

banjoh:

Collect information of whether connection to replicated.app succeeded or not, and analyse the results. The check already existed [here](https://github.com/replicatedhq/kots/blob/a38c002e6e59954e1f671c3f1070789a00a267cf/pkg/supportbundle/staticspecs/kurlspec.yaml#L8) but was missing in other specs

[sc-113418](https://app.shortcut.com/replicated/story/113418/check-that-kots-is-able-to-connect-to-replicated-app-license-server)

NONE

```release-note
Test connection to replicate.app and analyse results when collection support bundles in non-kURL clusters
```

NONE

### DIFF
--- a/pkg/supportbundle/defaultspec/spec.yaml
+++ b/pkg/supportbundle/defaultspec/spec.yaml
@@ -173,7 +173,7 @@ spec:
         containerPath: /tmp/last-preflight-result
         containerName: kotsadm
     - http:
-        collectorName: curl-replicated-app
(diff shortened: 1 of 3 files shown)


banjoh:
Support bundle tests are expected to fail according to https://github.com/replicatedhq/kots/pull/4935#issuecomment-2400599245.

sgalsaleh (approved):

PR #4936, opened 2024-10-04
remove clusterResources from the default SupportBundle Spec

diamonwiggins:

This PR removes `clusterResources: {}` from the default spec. Today, this will take precedent over a more specific instantiation of the `clusterResources` collector leading to a confusing experience for users.

By default if no `clusterResources` collector is specified in a `SupportBundle` spec, `troubleshoot` will add one https://github.com/replicatedhq/troubleshoot/blob/main/pkg/supportbundle/collect.go#L116-L117

[SC-112609](https://app.shortcut.com/replicated/story/112609)

```release-note

```

### DIFF
--- a/pkg/supportbundle/spec.go
+++ b/pkg/supportbundle/spec.go
@@ -430,14 +430,14 @@ func createDefaultSpec(app *apptypes.App, b *troubleshootv1beta2.SupportBundle, 
 	}
 
 	if isKurl {
-		kurlSupportBunlde, err := staticspecs.GetKurlSpec(app)
+		kurlSupportBundle, err := staticspecs.GetKurlSpec(app)
 		if err != nil {
 			logger.Errorf("Failed to load kurl support bundle spec: %v", err)
 			return nil, err
 		}
 
-		supportBundle.Spec.Collectors = append(supportBundle.Spec.Collectors, kurlSupportBunlde.Spec.Collectors...)
-		supportBundle.Spec.Analyzers = append(supportBundle.Spec.Analyzers, kurlSupportBunlde.Spec.Analyzers...)
+		supportBundle.Spec.Collectors = append(supportBundle.Spec.Collectors, kurlSupportBundle.Spec.Collectors...)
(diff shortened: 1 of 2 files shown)


diamonwiggins:
Now if a user specifies the following, it'll be respected
```
    - clusterResources:
        namespaces:
        - kube-system
 ```

**Review by sgalsaleh — APPROVED** [review]

**Comment by chris-sanders** [comment]
Commenting here to carry my shortcut forward.
We shouldn't implement this by just dropping the collection from KOTS. This puts us in a situation where we do not collect support bundles for KOTS in the use case the user has. We're basically asking the vendor to be responsible for covering our support collection needs.

We should instead have KOTS install the namespaces it _needs_ rather than all namespaces as it was doing before. We're responsible for that, we should manage collecting the right support information.

This was already approved as an acceptable approach by the customer requesting this feature. Their use case isn't that KOTS doesn't collect information, it's that KOTS shouldn't be collecting globally it should only be collecting it's own namespace(s).

File: pkg/supportbundle/spec.go [hunk 1]
```diff
@@ -727,6 +729,13 @@ func addDefaultDynamicTroubleshoot(supportBundle *troubleshootv1beta2.SupportBun
 func getDefaultDynamicCollectors(app *apptypes.App, imageName string, pullSecret *troubleshootv1beta2.ImagePullSecrets) []*troubleshootv1beta2.Collect {
 	collectors := make([]*troubleshootv1beta2.Collect, 0)
 
+	// Collect Cluster Resources for the namespace that the Admin Console is running in
```
sgalsaleh, line 732: should this also include the app namespaces? i think we already have logic to determine what those namespaces are for minimal rbac that can be reused

sgalsaleh (approved):

pkg/supportbundle/spec.go
diamonwiggins: yeah, so some additional background is that we're making this change because a customer specifically wanted to exclude certain application namespaces from being collected and instead respect their instantiation of `clusterResources` which we weren't allowing before. 

But you do raise a good point. There might be some folks relying on the default behavior and not actually specifying all needed namespaces in their spec.

PR #4942, opened 2024-10-10
fix(upgrade-service): allow no preflights and no config

JGAntunes:

```release-note

```

### DIFF
--- a/web/src/components/upgrade_service/PreflightChecks.tsx
+++ b/web/src/components/upgrade_service/PreflightChecks.tsx
@@ -8,7 +8,7 @@ import SkipPreflightsModal from "@components/shared/modals/SkipPreflightsModal";
 import PreflightsProgress from "@components/troubleshoot/PreflightsProgress";
 import "../../scss/components/PreflightCheckPage.scss";
 
-import { useGetPrelightResults, useRerunPreflights } from "./hooks/index";
+import { useGetPrelightResults, useRunPreflights } from "./hooks/index";
 
 import { KotsParams } from "@types";
 import { useUpgradeServiceContext } from "./UpgradeServiceContext";
@@ -35,10 +35,13 @@ const PreflightCheck = ({
 
   const { sequence = "0", slug } = useParams<keyof KotsParams>() as KotsParams;
 
+  const {
+    mutate: runPreflights,
+    error: runPreflightsError,
+    isSuccess: runPreflightsSuccess,
+  } = useRunPreflights({ slug, sequence });
   const { data: preflightCheck, error: getPreflightResultsError } =
-    useGetPrelightResults({ slug, sequence });
-  const { mutate: rerunPreflights, error: rerunPreflightsError } =
-    useRerunPreflights({ slug, sequence });
+    useGetPrelightResults({ slug, sequence, enabled: runPreflightsSuccess });
 
   if (!preflightCheck?.showPreflightCheckPending) {
     if (showConfirmIgnorePreflightsModal) {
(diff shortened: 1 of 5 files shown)


miaawong (approved):
just two little linter errors! Great work Joao!

laverya:
lint-web has a few failures - any reason those can't be fixed?

pkg/upgradeservice/handlers/config.go
```diff
@@ -379,13 +378,6 @@ func (h *Handler) SaveConfig(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	if err := upgradepreflight.Run(params); err != nil {
-		response.Error = "failed to run preflights"
-		logger.Error(errors.Wrap(err, response.Error))
-		JSON(w, http.StatusInternalServerError, response)
-		return
-	}
-
```
laverya, line 388: what happens when deploying the app via the CLI? Or do we just not allow that yet with the new workflow

sgalsaleh: Upgrade workflow is only supported via the UI currently.

sgalsaleh (commented):
I see a unit test for info API call, but not one for validating that the UI shows what's expected based on the values of `hasPreflight` and `isConfigurable`.

web/config/jest-setup.ts
```diff
@@ -0,0 +1,2 @@
+import "intersection-observer";
+import "@testing-library/jest-dom";
```
JGAntunes, line 2: There's a whole set of things which aren't pollyfilled by `jsdom` which is what the `@testing-library/jest-dom` is for. We also need to pollyfill the `intersection-observer` for the `AppConfig` component to render.

web/package.json
```diff
@@ -192,6 +194,14 @@
...
   ],
   "jest": {
+    "setupFilesAfterEnv": [
+      "<rootDir>/config/jest-setup.ts"
+    ],
+    "testEnvironmentOptions": {
+      "customExportConditions": [
+        ""
+      ]
+    },
```
JGAntunes, line 204: - https://mswjs.io/docs/migrations/1.x-to-2.x#cannot-find-module-mswnode-jsdom

web/src/jest-setup.ts
```diff
@@ -0,0 +1,2 @@
+import "intersection-observer";
+import "@testing-library/jest-dom";
```
JGAntunes, line 2: There's a whole set of things which aren't pollyfilled by jsdom which is what the @testing-library/jest-dom is for. We also need to pollyfill the intersection-observer for the AppConfig component to render.

JGAntunes:
Ok, after fighting a lot with `jest`, react router and its related friends I think this is ready for review. Could I get a 👀  again @sgalsaleh and @miaawong 🙏  There's a bit of noise in the tests because some of the components we're testing end up logging a lot but the tests are 🟢 if y'all want I can eventually remove the console.* calls we have spread out, or potentially silence them in some way in `jest`.

web/src/components/upgrade_service/UpgradeService.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+      const { findByTestId, getByText } = render(
+        <MemoryRouter initialEntries={[`/upgrade-service/app/${slug}`]}>
+          <Routes>
+            <Route path="/upgrade-service/*" element={<UpgradeService />} />
+          </Routes>
+        </MemoryRouter>
+      );
+
+      await findByTestId("preflight-check-area");
```
sgalsaleh, line 111: is this right? should this be available if preflights are not available?

sgalsaleh (dismissed):

web/src/components/upgrade_service/UpgradeService.test.tsx
JGAntunes: Ah nice catch, thanks! This doesn't make sense though, there's something in the tests that's leading to this behaviour because if I `it.only` this test it fails as expected, if I run the whole file it passes 🙃 need to debug this further tomorrow 😩

sgalsaleh (approved):

PR #4944, opened 2024-10-11
bug: only download bundle uri if in a connected environment

ricardomaraschini:

stop attempting to download the support bundle using the `uri` field when in disconnected (airgap) environments.


ricardomaraschini:
This doesn't solve the problem. Closing it. It seems like we need to add an extra option on `troubleshoot` instead.

PR #4945, opened 2024-10-11
bug: stop fetching support bundle uri on airgap

ricardomaraschini:

the function `ParseSupportBundleFromDoc()` always attempt to fetch the support bundle from its uri property (if present). there is another function that allows us to define if we want or not use the uri property:

```
ParseSupportBundle()
```

this new function requires a second argument called `followURI` and we need to set it to false when kots is running in an airgap environment.


pkg/supportbundle/supportbundle.go
```diff
@@ -312,7 +312,12 @@ func CreateSupportBundleAnalysis(appID string, archivePath string, bundle *types
 	if err != nil {
 		logger.Errorf("Failed to check if cluster is kurl: %v", err)
 	}
-	analyzer.Spec.Analyzers = append(analyzer.Spec.Analyzers, getDefaultAnalyzers(isKurl)...)
+
+	defaultAnalyzers, err := getDefaultAnalyzers(isKurl, foundApp.IsAirgap)
+	if err != nil {
```
emosbaugh, line 317: should you wrap and log the error here? it looks like everything else in this function does

sgalsaleh (approved):

pkg/supportbundle/spec.go
```diff
@@ -709,12 +706,17 @@ func deduplicatedAfterCollection(supportBundle *troubleshootv1beta2.SupportBundl
 	return b
 }
 
-func getDefaultAnalyzers(isKurl bool) []*troubleshootv1beta2.Analyze {
-	defaultAnalyzers := defaultspec.Get().Spec.Analyzers
+func getDefaultAnalyzers(isKurl, isAirgap bool) ([]*troubleshootv1beta2.Analyze, error) {
+	defaultSpec, err := defaultspec.Get(isAirgap)
+	if err != nil {
+		return nil, err
```
sgalsaleh, line 712: ditto (what Ethan said about wrapping / logging the error)

sgalsaleh (approved):

PR #4958, opened 2024-10-21
feat: enable remote host collector in KOTS

nvanthao:

Story: https://app.shortcut.com/replicated/story/114418/kots-to-be-updated-to-read-hostcollectors-spec-when-running-support-bundle

This PR will:
- include host collector and host analyzer in the final spec during Support Bundle generation from KOTS Admin Console
- dedup host collector and host analyzer specs in addition to current collector and analyzer spec
- always run host collector in pod

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ LVP_TAG ?= v0.6.7
 PACT_PUBLISH_CONTRACT ?= false
 
 OS ?= linux
-ARCH ?= amd64 
+ARCH ?= $(shell go env GOARCH) 
 
 .PHONY: test
 test:
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ LVP_TAG ?= v0.6.7
 PACT_PUBLISH_CONTRACT ?= false
 
 OS ?= linux
-ARCH ?= $(shell go env GOARCH) 
+ARCH ?= $(shell go env GOARCH)
 
 .PHONY: test
 test:
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ LVP_TAG ?= v0.6.7
 PACT_PUBLISH_CONTRACT ?= false
 
 OS ?= linux
-ARCH ?= $(shell go env GOARCH)
+ARCH ?= amd64 
 
 .PHONY: test
 test:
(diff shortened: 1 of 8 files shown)


DexterYan (dismissed):

DexterYan (approved):

Makefile
```diff
@@ -7,7 +7,7 @@ LVP_TAG ?= v0.6.7
 PACT_PUBLISH_CONTRACT ?= false
 
 OS ?= linux
-ARCH ?= $(shell go env GOARCH)
+ARCH ?= $(shell go env GOARCH)
```
emosbaugh, line 10: remove this space

pkg/kotsutil/troubleshoot.go
```diff
@@ -50,3 +51,26 @@ func LoadTSKindsFromPath(dir string) (*troubleshootloader.TroubleshootKinds, err
 	}
 	return tsKinds, nil
 }
+
+func Dedup[T any](objs []T) []T {
```
emosbaugh, line 55: can you move this to the supportbundle package? this is not a generic dedup package as it relies on marshalling the object and this will not work for many types

emosbaugh (changes requested):

pkg/kotsutil/troubleshoot.go
sgalsaleh: what types don't work with json.Marshal?

emosbaugh: lower case fields of structs

emosbaugh: if it were generic it would use reflection

emosbaugh (approved):

emosbaugh (changes requested):
Im not sure i understand this feature enough to approve it. I am getting more details from the troubleshoot team

emosbaugh (approved):

ajp-io:
For future reference, this is a pretty important change for users and needs a release note, but the release note section of the PR was blank. I'll add a release note to the docs though, so don't worry about it.

ajp-io:
Now I'm second guessing. Is this only part of the work to collect host level information from KOTS?

PR #4962, opened 2024-10-22
fix: persist http proxy settings and security context on upgrades

emosbaugh:

```release-note
- Fixes an issue that causes proxy settings to be removed on `kotsadm admin-console upgrade`.
- Fixes an issue that causes `--strict-security-context` to be removed on `kotsadm admin-console upgrade`.
```


sgalsaleh (approved):

emosbaugh:
The failing test seems to be from docker rate limiting. Merging through the failure

Screenshot 2024-10-22 at 2 19 07 PM

PR #4972, opened 2024-10-29
feat(ec): allow rollbacks for embedded cluster

emosbaugh:

Allows rollbacks for a given version if allowRollback is true for the latest downloaded version, and this is embedded cluster, and the embedded cluster config has changed between the version and current.

https://www.loom.com/share/fca871a6b1ca4031bcfde3ee966ce421?sid=334fa25b-40f3-4a83-affb-6257b53b2940

```release-note
Adds support for rollbacks in Embedded Cluster installations for versions where the embeddedcluster.replicated.com/v1beta1.Config has not changed between versions.
```

https://github.com/replicatedhq/replicated-docs/pull/2794

### DIFF
--- a/pkg/store/kotsstore/downstream_store.go
+++ b/pkg/store/kotsstore/downstream_store.go
@@ -426,20 +426,8 @@ func (s *KOTSStore) GetDownstreamVersions(appID string, clusterID string, downlo
 		if err := s.AddDownstreamVersionDetails(appID, clusterID, v, false); err != nil {
 			return nil, errors.Wrap(err, "failed to add details to latest downloaded version")
 		}
-		var currentECConfig, newECConfig []byte
-		if util.IsEmbeddedCluster() {
-			if result.CurrentVersion != nil {
-				currentECConfig, err = s.getRawEmbeddedClusterConfigForVersion(appID, result.CurrentVersion.Sequence)
-				if err != nil {
-					return nil, errors.Wrapf(err, "failed to get embedded cluster config for current version %d", result.CurrentVersion.Sequence)
-				}
-			}
-			newECConfig, err = s.getRawEmbeddedClusterConfigForVersion(appID, v.Sequence)
-			if err != nil {
-				return nil, errors.Wrapf(err, "failed to get embedded cluster config for version %d", v.Sequence)
-			}
-		}
-		v.IsDeployable, v.NonDeployableCause = isAppVersionDeployable(appID, v, result, license.Spec.IsSemverRequired, currentECConfig, newECConfig)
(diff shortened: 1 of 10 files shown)


pkg/store/kotsstore/downstream_store.go
```diff
@@ -423,8 +427,10 @@ func (s *KOTSStore) GetDownstreamVersions(appID string, clusterID string, downlo
 		if err := s.AddDownstreamVersionDetails(appID, clusterID, v, false); err != nil {
 			return nil, errors.Wrap(err, "failed to add details to latest downloaded version")
 		}
-		v.IsDeployable, v.NonDeployableCause = isAppVersionDeployable(v, result, license.Spec.IsSemverRequired)
-		break
```
emosbaugh, line 427: i cant understand the reason for this break here but it makes it so that the loop only iterates for the first non-pending_download version

sgalsaleh: the answer is in the comment above:
> // checking if a version is deployable requires getting all versions again.
> // check if latest version is deployable separately to avoid cycle dependencies between functions.

sgalsaleh: populating if a version is deployable or not only happens in the version history handler because it's an expensive process.

pkg/store/kotsstore/kots_store.go
```diff
@@ -34,6 +35,8 @@ type KOTSStore struct {
 }
 
 func init() {
+	store.SetStore(StoreFromEnv())
```
sgalsaleh, line 38: hmm, why the pattern change here? this makes it harder to add more implementations of the store interfaces in the future.

pkg/store/kotsstore/downstream_store.go
```diff
@@ -423,7 +426,7 @@ func (s *KOTSStore) GetDownstreamVersions(appID string, clusterID string, downlo
 		if err := s.AddDownstreamVersionDetails(appID, clusterID, v, false); err != nil {
 			return nil, errors.Wrap(err, "failed to add details to latest downloaded version")
 		}
-		v.IsDeployable, v.NonDeployableCause = isAppVersionDeployable(v, result, license.Spec.IsSemverRequired)
+		v.IsDeployable, v.NonDeployableCause = isAppVersionDeployable(appID, v, result, license.Spec.IsSemverRequired, nil, nil)
```
sgalsaleh, line 429: this means this new EC config comparison logic would not apply to the latest available version. why not keep the previous logic you had but just keep the `break`?

emosbaugh: how can you rollback to the latest available version?

sgalsaleh (approved):

sgalsaleh (approved):

sgalsaleh (approved):

PR #5002, opened 2024-11-13
chore(local-dev): add node affinity on the patches for local dev with EC

JGAntunes:

When running local dev with embedded cluster, running `make kotsadm-up-ec` will build the local dev images and load them into `node0` (by default). If we're running a multi node cluster setup it so happens that the scheduled pods might land in any of the other nodes other than `node0` (which won't have the loaded image). This affinity ensures we always schedule the pods on the correct node we set using `EC_NODE` env var.

NONE

NONE

```release-note
NONE
```

NONE


sgalsaleh:
You can use the EC_NODE env var to connect to a different node than node0:
```
export EC_NODE=node1
```
This is because you can have multiple clusters and kotsadm should not necessarily always run on node0.

JGAntunes:
> You can use the EC_NODE env var to connect to a different node than node0:

Right but even if you do, for this to work, you need to run `make kotsadm-up-ec` for each node so that `ec_build_and_load` loads the image into all the cluster nodes:
- https://github.com/replicatedhq/kots/blob/55b35f47fce43faf14bef8afb86d1fc15dcfad68/dev/scripts/up-ec.sh#L21-L22
- https://github.com/replicatedhq/kots/blob/55b35f47fce43faf14bef8afb86d1fc15dcfad68/dev/scripts/common.sh#L106-L112

Else the deployment will likely fail with pull image backoff errors if the pod lands on a node where the image isn't available.

I thought it would just be easier to pin the deployment to a single node when doing local dev as away to overcome this.

sgalsaleh:
How about we pin it to the value of `EC_NODE` then? or `node0` if `EC_NODE` is not specified? Also, need to do this for `kurl-proxy` as well.

JGAntunes:
> How about we pin it to the value of EC_NODE then? or node0 if EC_NODE is not specified? Also, need to do this for kurl-proxy as well.

Ah got it, yeah that makes sense. I can look into that 👍

JGAntunes:
Done @sgalsaleh. Let me know what you think.

Makefile
```diff
@@ -90,20 +90,24 @@ mock:
 	mockgen -source=pkg/handlers/interface.go -destination=pkg/handlers/mock/mock.go
 	mockgen -source=pkg/operator/client/client_interface.go -destination=pkg/operator/client/mock/mock.go
 
+.PHONY: install-gomplate
```
sgalsaleh, line 93: i would suggest putting this in the dev dependencies in the README instead like all the other dependencies.

- KOTS: https://github.com/replicatedhq/kots?tab=readme-ov-file#requirements
- EC: https://github.com/replicatedhq/embedded-cluster?tab=readme-ov-file#requirements

README.md
```diff
@@ -62,7 +62,8 @@ cosign verify-blob --key sbom/key.pub --signature sbom/kots-sbom.tgz.sig sbom/ko
 
 - MacOS
 - Docker Desktop with Kubernetes enabled
-- jq
+- gomplate (we'll automatically install this for you)
+- jq (we'll automatically install this for you using brew if available)
```
sgalsaleh, line 66: i would just remove these two requirements and add `homebrew`

sgalsaleh (approved):

PR #5004, opened 2024-11-13
feat(ec_join): update handler to return tcp connections required

JGAntunes:

On join, return all the TCP endpoints the joining node needs to preflight for to ensure a successful connection to the cluster

https://app.shortcut.com/replicated/story/113015/preflight-remote-access-on-join

Yes, added tests to the handler and to the newly added method

Also run some manual tests on an EC local dev deployment:
```
root@node1:/replicatedhq/embedded-cluster# curl -k -v "https://172.17.0.2:30000/api/v1/embedded-cluster/join?token=FGjr0qVT2dRHrMialHJk7zd8" | jq .
(...)
{
  "clusterID": "19750e22-03ff-44ae-8fcb-c41e70ca4db5",
  "k0sJoinCommand": "/usr/local/bin/k0s install controller --enable-worker --no-taints --labels kots.io/embedded-cluster-role=total-1,kots.io/embedded-cluster-role-0=controller-test,controller-label=controller-label-value",
  "k0sToken": "...",
  "embeddedClusterVersion": "1.19.0+k8s-1.30-11-gbb2a7ab9-jgantunes",
  "airgapRegistryAddress": "",
  "tcpConnectionsRequired": [
    "172.17.0.2:6443",
    "172.17.0.2:9443",
    "172.17.0.2:2380",
    "172.17.0.2:10250"
  ],
  "installationSpec": {
    (...)
  }
}
```

```release-note
Enable embedded cluster node joins to preflight Node to Node communications
```

NONE

### DIFF
--- a/pkg/handlers/embedded_cluster_node_join_command_test.go
+++ b/pkg/handlers/embedded_cluster_node_join_command_test.go
@@ -143,14 +143,22 @@ func TestGetEmbeddedClusterNodeJoinCommand(t *testing.T) {
 			).Build(),
 		},
 		{
-			name:              "controller and worker node IPs are returned",
+			name:              "tcp connections required are returned based on the controller role provided",
 			httpStatus:        http.StatusOK,
 			embeddedClusterID: "cluster-id",
 			validateBody: func(t *testing.T, h *testNodeJoinCommandHarness, r *GetEmbeddedClusterNodeJoinCommandResponse) {
 				req := require.New(t)
 
-				req.Equal([]string{"192.168.0.101"}, r.WorkerNodeIPs)
-				req.Equal([]string{"192.168.0.100"}, r.ControllerNodeIPs)
+				req.Equal([]string{
+					"192.168.0.100:6443",
+					"192.168.0.100:9443",
+					"192.168.0.100:2380",
+					"192.168.0.100:10250",
+					"192.168.0.101:10250",
+				}, r.TCPConnectionsRequired)
+			},
+			getRoles: func(t *testing.T, token string) ([]string, error) {
+				return []string{"controller-role"}, nil
 			},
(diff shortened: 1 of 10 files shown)


pkg/handlers/kubeclient/kubeclient.go
```diff
@@ -0,0 +1,35 @@
...
+// MockBuilder is a mock implementation of KubeClientBuilder. It returns the client that was set in the struct allowing
+// you to set a fakeClient for example.
+type MockBuilder struct {
+	Client kbclient.Client
+}
+
+// GetKubeClient returns the client that was set in the struct.
+func (b *MockBuilder) GetKubeClient(ctx context.Context) (kbclient.Client, error) {
+	return b.Client, nil
+}
```
JGAntunes, line 35: The general idea is to allow us to get a new k8s client on the fly within the handler VS the alternative which would be to have a single client instantiated when the handler is created and shared between all the endpoints. This ensures we keep functionality as is while having the ability to stub/mock the kube client.

pkg/handlers/embedded_cluster_node_join_command_test.go
```diff
@@ -0,0 +1,279 @@
...
+			httpStatus:        http.StatusInternalServerError,
+			embeddedClusterID: "cluster-id",
+			token:             "some-token",
+			getRoles: func(t *testing.T, token string) ([]string, error) {
+				require.Equal(t, "some-token", token)
+				return nil, fmt.Errorf("some error")
+			},
+		},
+		{
+			name:              "bootstrap token secret creation succeeds and it matches returned K0SToken",
```
JGAntunes, line 70: Went a bit overboard with testing other (unrelated) scenarios, but thought would be useful + it was a good opportunity for me to see how the join process works. Happy to move these to a separate PR if folks prefer.

sgalsaleh (approved):

PR #5007, opened 2024-11-15
chore: fix melange build action

ricardomaraschini:

manually pull, build, and use the latest version of the melange binary.


ricardomaraschini:
I am giving up on this approach. Let's try to do using a makefile like it is done for embedded-cluster.

PR #5019, opened 2024-11-20
add debug logging for the kotsadm startup process

laverya:

Debugging why kotsadm is not starting is quite painful. We can make this easier by:
adding a debug log after each major startup function
adding a 'this is still waiting' debug message in the db and filestore wait loops

We can also surface timeouts in the s3 connection function by using the context-aware s3 client function.

```release-note

```

### DIFF
--- a/pkg/apiserver/server.go
+++ b/pkg/apiserver/server.go
@@ -17,6 +17,7 @@ import (
 	identitymigrate "github.com/replicatedhq/kots/pkg/identity/migrate"
 	"github.com/replicatedhq/kots/pkg/informers"
 	"github.com/replicatedhq/kots/pkg/k8sutil"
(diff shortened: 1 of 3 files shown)


pkg/apiserver/server.go
```diff
@@ -145,6 +167,7 @@ func Start(params *APIServerParams) {
 			log.Println("Failed to run automated installs:", err)
 		}
 	}
+	logger.Debug("automated airgap installs are completed")
```
sgalsaleh, line 170: i would be confused if i saw this in an online install

laverya:
```suggestion
	logger.Debug("automated airgap installs are completed if requested")
```

PR #5020, opened 2024-11-21
feat(support-bundle): allow cluster resource to be overwrite by different spec or merging namespaces

DexterYan:

- KOTS will only ever collect the information from its own namespace from default support bundle spec
- now you can use other spec to overwrite combine namespaces

[sc-116165](https://app.shortcut.com/replicated/story/116165/support-bundles-generated-in-embedded-cluster-and-kurl-may-not-include-administrative-namespaces)

NONE

```release-note
NONE
```

NONE

### DIFF
--- a/pkg/supportbundle/spec.go
+++ b/pkg/supportbundle/spec.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"regexp"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


banjoh (changes requested):
I do not think merging `clusterResources` collectors logic should be here. We already have this implemented in https://github.com/replicatedhq/troubleshoot/blob/9f5f0633cf68e701e3d987aca211ea9bd2d151cc/pkg/collect/cluster_resources.go#L55

chris-sanders:
We don't think we are moving forward with this

PR #5032, opened 2024-12-06
feat(ec): add new DR backup type `ReplicatedBackup`

JGAntunes:
The Backups we show to the user in the UI now have a specific name and can contain 1 or more backups (infra and app backups). We should create a new go type ReplicatedBackup which includes a Name, an array of Backups containing each of the individual backups it holds, an ExpectedBackupCount holding the expected number of backups for this ReplicatedBackup to hold and a field to identify this type as a new Backup type.

https://app.shortcut.com/replicated/story/116339/create-a-new-replicatedbackup-type-including-a-name-property-and-an-array-of-backups

NONE we're just adding a type for now

```release-note
NONE
```

NONE

### DIFF
--- a/pkg/kotsadmsnapshot/types/types.go
+++ b/pkg/kotsadmsnapshot/types/types.go
@@ -13,6 +13,14 @@ type App struct {
 	AppIconURI string `json:"iconUri"`
 }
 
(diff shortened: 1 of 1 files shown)


JGAntunes:
Closing this one in favour of https://github.com/replicatedhq/kots/pull/5033.

PR #5037, opened 2024-12-10
feat: [sc-117029] Add keep annotation to KOTS

nvanthao:
Story details: https://app.shortcut.com/replicated/story/117029
Demo: https://www.loom.com/share/3a33f7af9a4b4df39c2ff309aaa99f48?sid=a032f5d3-16e0-4e8a-aac9-655bacc3bfff

## Changes
* implement new annotation `kots.io/keep: "true"` for KOTS to skip deleting the manifest if the annotation present in non-Helm install
* misc update in KOTS debug setup. I run `DEBUG_KOTSADM=1 make debug` to debug the code changes.

### DIFF
--- a/pkg/operator/client/delete.go
+++ b/pkg/operator/client/delete.go
@@ -82,7 +82,7 @@ func (c *Client) diffAndDeleteManifests(opts DiffAndDeleteOptions) error {
 		// if this is a keep resource, don't delete it
 		// e.g. for migration to Helm v1beta2
 		if keep, ok := o.Metadata.Annotations["kots.io/keep"]; ok && keep == "true" {
-			logger.Infof("Skipping deletion of resource %s/%s", o.Kind, o.Metadata.Name)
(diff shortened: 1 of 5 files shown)


pkg/operator/client/delete.go
```diff
@@ -79,6 +79,13 @@ func (c *Client) diffAndDeleteManifests(opts DiffAndDeleteOptions) error {
 			}
 		}
 
+		// if this is a keep resource, don't delete it
+		// e.g. for migration to Helm v1beta2
+		if keep, ok := o.Metadata.Annotations["kots.io/keep"]; ok && keep == "true" {
+			logger.Infof("skipping deletion of resource %s/%s", o.Kind, o.Metadata.Name)
```
sgalsaleh, line 85: would be nice to mention the reason in the log message as well
```suggestion
			logger.Infof("skipping deletion of resource %s/%s because kots.io/keep annotation is set", o.Kind, o.Metadata.Name)
```

sgalsaleh (approved):

PR #5055, opened 2024-12-16
Use dev migrations image for EC dev

sgalsaleh:


dev/scripts/up-ec.sh
```diff
@@ -21,6 +21,11 @@ fi
 # Build and load the image into the embedded cluster
 ec_build_and_load "$component"
 
+# Use the dev image for kotsadm migrations
+if [ "$component" == "kotsadm" ]; then
+  ec_load "kotsadm-migrations"
```
JGAntunes, line 26: Shouldn't this be `kotsadm-migrations`?

Also won't we need to make sure `kotsadm-migrations-dev` is built too as part of the `%-up-ec` make target? I believe right this is only done as part of the `dev` target?

JGAntunes (commented):
Thanks for looking into this! Left a few questions.

dev/scripts/up-ec.sh
sgalsaleh: component should be `kotsadm`. the migrations is just an init container in the kotsadm deployment. there's no separate migrations component that you can "up" into.

sgalsaleh: yes, it needs to be built ,i'll update

JGAntunes: > component should be kotsadm. 

Sorry I mistyped 🤦 what I meant was shouldn't:
```
 ec_load "kotsadm-migrations"
```
Be:
```
ec_load "kotsadm-migrations-dev"
```

But I just found out:
- https://github.com/replicatedhq/kots/blob/92a3cee3f862f989684c7d42b0d2fdc20546c6bd/dev/metadata.json#L9

So I answered my own question :D

JGAntunes (approved):

sgalsaleh:
Merging as this does not affect tests.

PR #5058, opened 2024-12-16
feat(snapshots): GetBackup details endpoint should support new dr

emosbaugh:

Adds support to backup details endpoint for new DR.

```release-note

```

### DIFF
--- a/pkg/handlers/backup.go
+++ b/pkg/handlers/backup.go
@@ -130,10 +130,9 @@ func (h *Handler) ListInstanceBackups(w http.ResponseWriter, r *http.Request) {
 }
 
 type GetBackupResponse struct {
-	BackupDetails           []snapshottypes.BackupDetail `json:"backupDetails"`
-	Deprecated_BackupDetail *snapshottypes.BackupDetail  `json:"backupDetail"`
-	Success                 bool                         `json:"success"`
-	Error                   string                       `json:"error,omitempty"`
+	BackupDetails []snapshottypes.BackupDetail `json:"backupDetails"`
+	Success       bool                         `json:"success"`
+	Error         string                       `json:"error,omitempty"`
 }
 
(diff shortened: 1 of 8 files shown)


pkg/handlers/backup.go
```diff
@@ -130,22 +130,27 @@ func (h *Handler) ListInstanceBackups(w http.ResponseWriter, r *http.Request) {
 }
 
 type GetBackupResponse struct {
-	BackupDetail *snapshottypes.BackupDetail `json:"backupDetail"`
-	Success      bool                        `json:"success"`
-	Error        string                      `json:"error,omitempty"`
+	BackupDetails           []snapshottypes.BackupDetail `json:"backupDetails"`
+	Deprecated_BackupDetail *snapshottypes.BackupDetail  `json:"backupDetail"`
```
JGAntunes, line 134: I wonder if this is better VS changing where we use this in the UI?
- https://github.com/replicatedhq/kots/blob/ea574d82fb8702318ef2c8c96831a474675237e0/web/src/components/RestoreSnapshotRow.jsx#L35
- https://github.com/replicatedhq/kots/blob/ea574d82fb8702318ef2c8c96831a474675237e0/web/src/components/snapshots/SnapshotDetails.jsx#L161

Else, should we have a store to clean this up once the UI is updated?

JGAntunes (approved):
LGTM, left a suggestion/question

pkg/handlers/backup.go
emosbaugh: Is something like this correct? 

```javascript
result.backupDetails && result.backupDetails.length > 0 ? result.backupDetails[0] : null
```

emosbaugh: Or this?

```javascript
result.backupDetails?.[0]
```

emosbaugh: updated

sgalsaleh (approved):

PR #5061, opened 2024-12-18
feat(backups): change list instance backups response payload

JGAntunes:

As per our design doc - https://docs.google.com/document/d/1JVgVSnrPWeLhS3qm3CC_MdDGDvFpM5TSABkySbHum-w/edit?tab=t.0#heading=h.ocsy1d20qe90 - we're changing the list instance backup endpoint again to better convey the 2 backup system we've designed.

https://app.shortcut.com/replicated/story/117470/change-the-listinstancebackups-endpoint-to-return-a-flat-array-structure-of-replicated-backups

Tests added

Also run some manual tests with an EC cluster to validate current functionality remains unchanged:

image

image

```release-note
NONE
```

NONE


web/src/components/snapshots/Snapshots.jsx
```diff
@@ -108,12 +108,8 @@ class Snapshots extends Component {
       }
       const response = await res.json();
 
-      const snapshots = response.backups?.flatMap(
-        (replBackup) => replBackup?.backups
-      );
-
       this.setState({
-        snapshots: snapshots?.sort((a, b) =>
+        snapshots: response.backups?.sort((a, b) =>
```
JGAntunes, line 112: This just reverts what we've done in - https://github.com/replicatedhq/kots/pull/5033/files#diff-84ad0b83d6595420183ed4f5d5f3b47be290c7a5a014f1cc4729329c2ac3dc2a

The main difference we introduced has been 2 extra fields (the backup count and expected) and a reduction of potential status phases:
- https://github.com/replicatedhq/kots/blob/05573a551f67706e5fa91fc29e271c4188cc497f/pkg/kotsadmsnapshot/types/types.go#L66-L75

I've checked in:
- https://github.com/replicatedhq/kots/blob/5fbcb34f8a5ba4a2999c2bb02e510138b115cf9b/web/src/components/snapshots/SnapshotRow.jsx#L36-L142

And it doesn't look like we've introduced anything different or non standard to what we're already using ☝️  CC @miaawong if you could confirm 👀

pkg/kotsadmsnapshot/types/types_test.go
```diff
@@ -0,0 +1,144 @@
...
+			expected: BackupStatusFailed,
+		},
+		{
+			backupStatuses: []BackupStatus{
+				BackupStatusCompleted,
+				BackupStatusCompleted,
+			},
+			expected: BackupStatusCompleted,
+		},
+	}
```
JGAntunes, line 87: This encodes the matrix states we talked about in the design doc here:
- https://docs.google.com/document/d/1JVgVSnrPWeLhS3qm3CC_MdDGDvFpM5TSABkySbHum-w/edit?tab=t.0#heading=h.qb8vv9xg5rhi

pkg/kotsadmsnapshot/types/types.go
```diff
@@ -59,29 +60,38 @@ type App struct {
...
+	Sequence        int64        `json:"sequence"` // TODO: remove with app backups
+	StartedAt       *time.Time   `json:"startedAt,omitempty"`
+	FinishedAt      *time.Time   `json:"finishedAt,omitempty"`
+	ExpiresAt       *time.Time   `json:"expiresAt,omitempty"`
+	SupportBundleID string       `json:"supportBundleId,omitempty"`
+	IncludedApps    []App        `json:"includedApps,omitempty"`
+	// number of velero backups expected to exist for the Backup to be considered done
+	ExpectedBackupCount int `json:"expectedBackupCount"`
+	// number of velero backups that actually exist
+	BackupCount int `json:"backupCount"`
```
JGAntunes, line 93: With us no longer having the `ReplicatedBackup` struct I believe these 2 fields end up being useful, specifically when we convey `Failed` states it's possible for the client to check if the counts match. They weren't part of the design doc proposal but I ended up adding them as I worked through the code.

emosbaugh (approved):

PR #5062, opened 2024-12-18
Always build and load migrations image in dev env

sgalsaleh:


sgalsaleh:
Merging as this only affects dev env.

PR #5066, opened 2024-12-20
chore(velero): bump velero client to latest 1.15.0

JGAntunes:

This PR bumps the velero client to the latest version (and the one we use in embedded cluster) v1.15.0.

The biggest change for our side is the deprecation of the generated client via:
- https://github.com/vmware-tanzu/velero/pull/8114

It's now expected that we use the `controller-runtime` client instead. There's an opportunity to simplify the clients we rely on a in a couple of places (we have methods using both `clientset` k8s `go-client` and the `controller-runtime` client) however I've opted to leave it out of scope for this set of changes to not aggravate an already large PR. 

https://app.shortcut.com/replicated/story/117774/bump-velero-to-the-latest-version-v1-15-0-on-kots

I've run some manual tests locally, creating backups and listing them, both in EC and existing cluster. Nevertheless there's a large set of code branches we're touching here so if folks could also do some smoke tests on specific paths you deem tricky I would appreciate it.

```release-note
NONE
```

NONE


pkg/handlers/handlers.go
```diff
@@ -35,7 +35,7 @@ func NewHandler() *Handler {
 func init() {
 	kotsscheme.AddToScheme(scheme.Scheme)
 	troubleshootscheme.AddToScheme(scheme.Scheme)
-	veleroscheme.AddToScheme(scheme.Scheme)
+	velerov1.AddToScheme(scheme.Scheme)
```
JGAntunes, line 38: Since all the velero object interactions now rely on:
- https://github.com/replicatedhq/kots/blob/a422a554bd22d7b1e7e0feed4f76e33d77f0fc57/pkg/k8sutil/clientset.go#L157-L171

Which has the scheme set via:
- https://github.com/replicatedhq/kots/blob/a422a554bd22d7b1e7e0feed4f76e33d77f0fc57/pkg/k8sutil/clientset.go#L36-L40

We probably don't need this anymore. Nevertheless keeping it for now just be cautious. Once we refactor our client usage across the board and drop a lof of our clientset usages we can take care of removing these.

pkg/k8sutil/clientset.go
```diff
@@ -163,27 +163,7 @@ func GetKubeClient(ctx context.Context) (kbclient.Client, error) {
...
-	cfg, err := GetClusterConfig()
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to get cluster config")
-	}
-	scheme := runtime.NewScheme()
-	velerov1.AddToScheme(scheme)
-	kcli, err := kbclient.New(cfg, kbclient.Options{
-		Scheme: scheme,
-	})
+	kcli, err := kbclient.NewWithWatch(cfg, kbclient.Options{})
```
JGAntunes, line 166: We need the watch interface for our informer:
- https://github.com/replicatedhq/kots/blob/a422a554bd22d7b1e7e0feed4f76e33d77f0fc57/pkg/informers/informers.go#L48-L50

pkg/kotsadmsnapshot/backup.go
```diff
@@ -194,12 +193,12 @@ func CreateApplicationBackup(ctx context.Context, a *apptypes.App, isScheduled b
 		logger.Error(errors.Wrap(err, "failed to exclude shutdown pods from backup"))
 	}
 
-	backup, err := veleroClient.Backups(kotsadmVeleroBackendStorageLocation.Namespace).Create(ctx, veleroBackup, metav1.CreateOptions{})
+	err = veleroClient.Create(ctx, veleroBackup)
```
JGAntunes, line 196: Namespace is already set above - https://github.com/replicatedhq/kots/blob/a422a554bd22d7b1e7e0feed4f76e33d77f0fc57/pkg/kotsadmsnapshot/backup.go#L153

pkg/kotsadmsnapshot/backup.go
```diff
@@ -275,20 +269,20 @@ func CreateInstanceBackup(ctx context.Context, cluster *downstreamtypes.Downstre
 	}
 
 	logger.Infof("Creating instance backup CR %s", veleroBackup.GenerateName)
-	backup, err := veleroClient.Backups(metadata.backupStorageLocationNamespace).Create(ctx, veleroBackup, metav1.CreateOptions{})
+	err = ctrlClient.Create(ctx, veleroBackup)
```
JGAntunes, line 272: Namespace needs to be set in `getInfrastructureInstanceBackupSpec`:
- https://github.com/replicatedhq/kots/blob/a422a554bd22d7b1e7e0feed4f76e33d77f0fc57/pkg/kotsadmsnapshot/backup.go#L435

pkg/kotsadmsnapshot/backup.go
```diff
@@ -438,6 +432,7 @@ func getInfrastructureInstanceBackupSpec(ctx context.Context, k8sClient kubernet
 	veleroBackup := &velerov1.Backup{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:         "",
+			Namespace:    metadata.backupStorageLocationNamespace,
```
JGAntunes, line 435: Namespace needs to be set given the client changes.

pkg/kotsadmsnapshot/backup.go
```diff
@@ -543,8 +538,10 @@ func getAppInstanceBackupSpec(k8sClient kubernetes.Interface, metadata instanceB
 
 		appVeleroBackup = appMeta.kotsKinds.Backup.DeepCopy()
 		restore = appMeta.kotsKinds.Restore.DeepCopy()
+		restore.Namespace = metadata.backupStorageLocationNamespace
```
JGAntunes, line 541: We also set the namespace for the restore spec that's added as annotation.

pkg/kotsadmsnapshot/backup.go
```diff
@@ -543,8 +538,10 @@ func getAppInstanceBackupSpec(k8sClient kubernetes.Interface, metadata instanceB
...
 
 		appVeleroBackup.Name = ""
+		appVeleroBackup.Namespace = metadata.backupStorageLocationNamespace
```
JGAntunes, line 544: Namespace needs to be set given the client changes.

pkg/kotsadmsnapshot/backup.go
```diff
@@ -752,13 +750,16 @@ func ListBackupsForApp(ctx context.Context, kotsadmNamespace string, appID strin
 		}
 
 		if veleroBackup.Status.StartTimestamp != nil {
-			backup.StartedAt = &veleroBackup.Status.StartTimestamp.Time
+			startedAt := veleroBackup.Status.StartTimestamp.Time.UTC()
+			backup.StartedAt = &startedAt
```
JGAntunes, line 754: Seems like timestamps are now being returned unmarshalled with `time.Local`

pkg/snapshot/store.go
```diff
@@ -792,31 +793,24 @@ func cloudCredentialsEnvVars() []corev1.EnvVar {
...
+	veleroClient, err := k8sutil.GetKubeClient(ctx)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to create velero clientset")
+		return nil, errors.Wrap(err, "failed to create velero client")
 	}
 
-	_, err = veleroClient.BackupStorageLocations(bsl.Namespace).Get(ctx, bsl.Name, metav1.GetOptions{})
-	if err == nil {
-		updated, err := veleroClient.BackupStorageLocations(bsl.Namespace).Update(ctx, bsl, metav1.UpdateOptions{})
+	if err := veleroClient.Update(ctx, bsl); err == nil {
```
JGAntunes, line 801: We can just attempt to update and fallback to creation if the object does not exist.

laverya (dismissed):

pkg/informers/informers.go
```diff
@@ -97,14 +101,8 @@ func Start() error {
 							continue
 						}
 
-						updatedBackup, err := veleroClient.Backups(backup.Namespace).Get(context.TODO(), backup.Name, metav1.GetOptions{})
```
emosbaugh, line 100: Is this tested in ci or was it tested manually? Are you sure it works without refreshing the resource? I assume it does because it is passing a pointer.

pkg/snapshot/store.go
emosbaugh: this code looks odd. i dont think it actually handles an error if it is not not found

emosbaugh (changes requested):

pkg/informers/informers.go
laverya: good question, I think this will work but I'll give it a look

laverya: The status does update in the UI, which I think is handled by this, so I would assume it's working

pkg/snapshot/store.go
laverya: oh good catch

laverya: it didn't previously handle the error due to variable masking, should work now

pkg/snapshot/store.go
```diff
@@ -798,15 +798,13 @@ func upsertBackupStorageLocation(ctx context.Context, bsl *velerov1.BackupStorag
 		return nil, errors.Wrap(err, "failed to create velero client")
 	}
 
-	if err := veleroClient.Update(ctx, bsl); err == nil {
-		if err != nil {
-			return nil, errors.Wrap(err, "failed to update backup storage location")
-		}
+	err = veleroClient.Update(ctx, bsl)
+	if err == nil {
```
emosbaugh, line 802: could you change line 814 to `return nil, errors.Wrap(err, "failed to update backup storage location")`

emosbaugh (changes requested):

emosbaugh (approved):

PR #5085, opened 2025-01-07
fix: update spec with new scripts dir when updating kotsadm

emosbaugh:

```release-note

```


pkg/kotsadm/objects/kotsadm_objects.go
```diff
@@ -163,11 +163,37 @@ func UpdateKotsadmDeployment(existingDeployment *appsv1.Deployment, desiredDeplo
...
-	existingDeployment.Spec.Template.Annotations["pre.hook.backup.velero.io/command"] = `["/scripts/backup.sh"]`
+	updateKotsadmDeploymentScriptsPath(existingDeployment)
 
 	return nil
 }
 
+func updateKotsadmDeploymentScriptsPath(existing *appsv1.Deployment) {
+	if existing.Spec.Template.Annotations != nil {
+		existing.Spec.Template.Annotations["pre.hook.backup.velero.io/command"] = `["/scripts/backup.sh"]`
+	}
```
laverya, line 174: I think we should create the map if it is nil, but this is probably fine in practice

laverya (approved):

PR #5084, opened 2025-01-07
Show app config navbar for Embedded Cluster

miaawong:
#### What this PR does / why we need it:

https://replicated.slack.com/archives/C0499EGNPNY/p1736268683199299

#### Which issue(s) this PR fixes:

[sc-118290]
https://app.shortcut.com/replicated/story/118290/investigate-lack-of-config-menu-in-ec-kots
#### Does this PR require a release note?

```release-note
Fixes an issue where the Config navbar was not visible on the Config page
```

PR #5086, opened 2025-01-07
only hide config sidebar during guided install

laverya:

```release-note

```


laverya:
https://github.com/replicatedhq/kots/pull/5084 is a better solution

PR #5112, opened 2025-01-23
chore: remove unnecessary concurrency

emosbaugh:

Unnecessary concurrency. Fix flakey test due to race condition with clients global map.

```release-note

```

### DIFF
--- a/pkg/plan/manager_test.go
+++ b/pkg/plan/manager_test.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"net/url"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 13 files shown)


pkg/plan/manager_test.go
```diff
@@ -153,23 +150,38 @@ spec:
...
+			ts := NewTestServer(t, ws)
 			defer ts.Close()
 
+			wg := sync.WaitGroup{}
+			wg.Add(len(tt.managers))
+
 			// Mock EC managers
 			for _, m := range tt.managers {
-				go newTestECManager(ts, m.nodeName, m.version)
+				m := m
```
sgalsaleh, line 173: what does this do?

PR #5192, opened 2025-03-11
feat(ec): add env vars for custom domains

emosbaugh:

Embedded cluster installs will now get the api endpoint and registry and proxy domain from the environment variables REPLICATED_API_ENDPOINT, REPLICATED_REGISTRY_DOMAIN and PROXY_REGISTRY_DOMAIN respectively. It will no longer use the license or headers from the api response.

```release-note

```


pkg/util/util.go
```diff
@@ -172,6 +166,68 @@ func EmbeddedClusterVersion() string {
...
+	}
+	return DefaultReplicatedAPIEndpoint(), nil
+}
+
+func maybePrependHTTPS(endpoint string) string {
+	if strings.HasPrefix(endpoint, "http://") || strings.HasPrefix(endpoint, "https://") {
+		return endpoint
+	}
+	return fmt.Sprintf("https://%s", endpoint)
+}
```
laverya, line 188: I'm perfectly happy to always pass http/https to the chart here

pkg/metrics/install.go
```diff
@@ -107,7 +106,7 @@ func getEndpoint(license *kotsv1beta1.License) string {
 		return license.Spec.Endpoint
```
laverya, line 106: this doesn't feel right

pkg/util/util.go
```diff
@@ -172,6 +166,68 @@ func EmbeddedClusterVersion() string {
...
+
+func DefaultReplicatedAPIEndpoint() string {
+	return "https://replicated.app"
+}
```
laverya, line 192: Why does this function exist and why is it called from outside the package? Would we not always use the env var when provided?

pkg/metrics/install.go
emosbaugh: this isnt new code

pkg/util/util.go
emosbaugh: updated

pkg/metrics/install.go
laverya: I know, but it feels like the env var should override this endpoint if present

pkg/util/util.go
```diff
@@ -169,24 +169,20 @@ func EmbeddedClusterVersion() string {
...
 		}
 		return "", errors.New("REPLICATED_API_ENDPOINT environment variable is required")
 	}
 
 	if license != nil && license.Spec.Endpoint != "" {
-		return maybePrependHTTPS(license.Spec.Endpoint), nil
+		if strings.HasPrefix(license.Spec.Endpoint, "http://") || strings.HasPrefix(license.Spec.Endpoint, "https://") {
+			return license.Spec.Endpoint, nil
+		}
+		return fmt.Sprintf("https://%s", license.Spec.Endpoint), nil
```
laverya, line 181: I don't think you need to do this - our APIserver code includes the scheme in the license format (and it's not user/vendor editable)

pkg/metrics/install.go
emosbaugh: oh i see... yes

pkg/util/util.go
emosbaugh: removed

pkg/util/util.go
emosbaugh: updated

pkg/util/util_test.go
```diff
@@ -331,3 +332,79 @@ func TestConvertToSingleDocs(t *testing.T) {
...
+			isEmbedded:  true,
+			envEndpoint: "https://replicated.app",
+			want:        "https://replicated.app",
+		},
+		{
+			name:       "embedded cluster without env endpoint",
+			license:    nil,
+			isEmbedded: true,
+			wantError:  true,
+		},
```
laverya, line 383:
```suggestion
		{
			name:       "embedded cluster without env endpoint",
			license:    nil,
			isEmbedded: true,
			wantError:  true,
		},
		{
			name:       "embedded cluster without env endpoint but with license",
			license: &kotsv1beta1.License{
				Spec: kotsv1beta1.LicenseSpec{
					Endpoint: "https://replicated.app:8443",
				},
			},
			isEmbedded: true,
			want: "https://replicated.app:8443",
		},
```

emosbaugh: that should return an error

pkg/docker/registry/registry.go
```diff
@@ -57,7 +58,22 @@ var secretAnnotations = map[string]string{
...
 
-func GetRegistryProxyInfo(license *kotsv1beta1.License, installation *kotsv1beta1.Installation, app *kotsv1beta1.Application) *RegistryProxyInfo {
+func GetRegistryProxyInfo(license *kotsv1beta1.License, installation *kotsv1beta1.Installation, app *kotsv1beta1.Application) (*RegistryProxyInfo, error) {
+	if util.IsEmbeddedCluster() {
+		registryProxyInfo, err := getEmbeddedClusterRegistryProxyInfo(license)
+		if err != nil {
+			return nil, errors.Wrap(err, "embedded cluster")
+		}
+
+		// getting the registry and proxy from the installation spec takes precedence over the application spec
```
sgalsaleh, line 68: i don't think this is true. we shouldn't fall back to anything kots does. if the env vars aren't set, we should fall back to defaults

emosbaugh: fixed

pkg/docker/registry/registry.go
```diff
@@ -113,12 +122,28 @@ func getRegistryProxyEndpointFromKotsApplication(kotsApplication *kotsv1beta1.Ap
...
+func getEmbeddedClusterRegistryProxyInfo() (*RegistryProxyInfo, error) {
+	info := &RegistryProxyInfo{
+		Upstream: util.DefaultReplicatedRegistryDomain(),
+	}
+
+	registryDomain := os.Getenv("REPLICATED_REGISTRY_DOMAIN")
+	if registryDomain == "" {
+		return nil, errors.New("REPLICATED_REGISTRY_DOMAIN environment variable is required")
 	}
+	info.Registry = registryDomain
```
sgalsaleh, line 134: should also set `info.Upstream`

emosbaugh: i think the needs to remain registry.replicated.com

pkg/template/config_context.go
```diff
@@ -320,7 +321,15 @@ func (ctx ConfigCtx) localImageName(imageRef string) string {
 	installation := &kotsv1beta1.Installation{
 		Spec: InstallationSpecFromVersionInfo(ctx.VersionInfo),
 	}
-	registryProxyInfo := registry.GetRegistryProxyInfo(ctx.license, installation, ctx.app)
+	registryProxyInfo, err := registry.GetRegistryProxyInfo(ctx.license, installation, ctx.app)
+	if err != nil {
+		// TODO: log
```
sgalsaleh, line 326: i don't get this use case, shouldn't this still be an error?

pkg/template/license_context.go
```diff
@@ -93,7 +94,15 @@ func (ctx licenseCtx) licenseDockercfg() string {
 	installation := &kotsv1beta1.Installation{
 		Spec: InstallationSpecFromVersionInfo(ctx.VersionInfo),
 	}
-	registryProxyInfo := registry.GetRegistryProxyInfo(ctx.License, installation, ctx.App)
+	registryProxyInfo, err := registry.GetRegistryProxyInfo(ctx.License, installation, ctx.App)
+	if err != nil {
+		// TODO: log
```
sgalsaleh, line 99: ditto

pkg/template/config_context.go
```diff
@@ -368,7 +377,15 @@ func (ctx ConfigCtx) localRegistryImagePullSecret() string {
 		installation := &kotsv1beta1.Installation{
 			Spec: InstallationSpecFromVersionInfo(ctx.VersionInfo),
 		}
-		registryProxyInfo := registry.GetRegistryProxyInfo(ctx.license, installation, ctx.app)
+		registryProxyInfo, err := registry.GetRegistryProxyInfo(ctx.license, installation, ctx.app)
+		if err != nil {
+			// TODO: log
```
sgalsaleh, line 382: ditto

pkg/template/config_context.go
emosbaugh: updated

pkg/template/config_context.go
emosbaugh: updated

sgalsaleh (approved):

PR #5198, opened 2025-03-12
Reduce the amount of k3s CMX versions to test to the last three major.minor versions

jtuchscherer:
This change is in the GHA job that fetches the supported k8s versions from the CMX endpoint. It adds the config option to limit the number of major minor versions supported by CMX to a configurable number of versions. It will take the most recent versions. If this value is not configured, it will continue to return all options.
I also added a simple test file for this that is currently on to be run manually, not on the pipeline.

NONE

NONE

NONE

### DIFF
--- a/.github/actions/cmx-versions/index.js
+++ b/.github/actions/cmx-versions/index.js
@@ -4,6 +4,7 @@ const semverCoerce = require('semver/functions/coerce');
 const semverMajor = require('semver/functions/major');
 const semverMinor = require('semver/functions/minor');
 const semverGt = require('semver/functions/gt');
+const semverRSort = require('semver/functions/rsort');
 
 async function getClusterVersions() {
     const url = 'https://api.replicated.com/vendor/v3/cluster/versions';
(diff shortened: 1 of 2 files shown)


jtuchscherer:
CMS versions before my change:
```json

File: .github/actions/cmx-versions/index.js [hunk 1]
```diff
@@ -122,21 +124,43 @@ function getLatestVersion(distribution) {
...
     distribution.versions.forEach((version) => {
         const parsed = semverCoerce(version);
+        // Check if majorVersions is null, undefined, empty, or includes the current version's major
         const majorMinor = `${semverMajor(parsed)}.${semverMinor(parsed)}`;
-        if (latestMinorVersions[majorMinor] === undefined) {
-            latestMinorVersions[majorMinor] = version;
-        } else {
-            const currentVersion = latestMinorVersions[majorMinor];
-            if (semverGt(parsed, semverCoerce(currentVersion))) {
+        if(majorMinorVersionFilter === null || majorMinorVersionFilter === undefined || majorMinorVersionFilter.size === 0 || majorMinorVersionFilter.includes(majorMinor)) {
```
sgalsaleh, line 151: i think this can just be (tested locally and passes tests):
```suggestion
        if(!majorMinorVersionFilter?.length || majorMinorVersionFilter.includes(majorMinor)) {
```

sgalsaleh (approved):

PR #5221, opened 2025-03-19
Wait for Velero in kurl-airgap-install regression test

sgalsaleh:

Wait for Velero in kurl-airgap-install regression test before validating snapshots functionality as joining a worker node can cause Velero to become unavailable for a bit.

```release-note
NONE
```

NONE


sgalsaleh:
Merging as this doesn't affect PR tests.

PR #5224, opened 2025-03-20
feat(customdomains): do not use support bundle spec uri for embedded cluster

emosbaugh:

In order to fully support custom domains in EC we cannot make requests to urls other than the domain. Remove the support bundle uri feature until we add support.

```release-note

```


sgalsaleh (approved):

emosbaugh:
This failure seems unrelated to the tests. Merging

PR #5226, opened 2025-03-20
adds new image scanning process 

St0rmz1:

- Adds new image scanning process 
- Scan the actual public image on dockerhub
- Use Grype as the scanner
- Use Sarif format as attachment to the repo so it gets added to the code scan results
- Enrich the Sarif with properties we use for reporting purposes
- To determine the image tag we use the release version from this repo

We scan the public image to ensure we are scanning what people are using. If we were to build the image and scan, we would actually be scanning a new image instead of the current release.

None

NONE

```release-note
NONE
```

NONE

### DIFF
--- a/.github/workflows/daily-container-scans.yml
+++ b/.github/workflows/daily-container-scans.yml
@@ -22,7 +22,7 @@ jobs:
     steps:
       - name: Get latest release
         id: get_release
-        uses: actions/github-script@v6
+        uses: actions/github-script@v7
         with:
           script: |
             const release = await github.rest.repos.getLatestRelease({
--- a/.github/workflows/daily-container-scans.yml
+++ b/.github/workflows/daily-container-scans.yml
@@ -31,28 +31,9 @@ jobs:
             });
             core.setOutput('tag_name', release.data.tag_name);
 
-  get-tag-from-image-env:
-    name: Get Tag From Image Env
-    runs-on: ubuntu-latest
-    outputs:
-      dex_tag: ${{ steps.dotenv.outputs.DEX_TAG }}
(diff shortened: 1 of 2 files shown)


.github/workflows/daily-container-scans.yml
```diff
@@ -0,0 +1,68 @@
...
+      - name: Get latest release
+        id: get_release
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const release = await github.rest.repos.getLatestRelease({
+              owner: context.repo.owner,
+              repo: context.repo.repo
+            });
+            core.setOutput('tag_name', release.data.tag_name);
```
laverya, line 32: latest _release_, or latest tag?

I'm not sure which one this does, but we also make tags for nightly builds

.github/workflows/scan-container-image.yml
```diff
@@ -0,0 +1,87 @@
+name: Scan Container Image
+
+on:
+  workflow_call:
```
laverya, line 4: should we call this workflow on merge?

St0rmz1: We can. I don't think it will hurt, but the idea for these scans are to be an async check for the current release. Basically, we need to track the state of the most recent version'ed release. Not the nightly or a new build-in-time because those don't represent the most recent versioned release on dockerhub.

.github/workflows/daily-container-scans.yml
laverya: No that makes sense if it's what you're looking for!

laverya (approved):

PR #5243, opened 2025-03-27
Fix support-bundle test flakiness

sgalsaleh:

Files in the redaction list are randomly sorted, and one of the files in the IP addresses redactions doesn't exist in the support bundle and can end up the one being selected in the test. This PR is a bandaid to address the flakiness in the meantime by selecting the second row in the redaction list where the files always exist until we figure out the underlying cause.

```release-note
NONE
```

NONE


sgalsaleh:
Not sure where this file comes from, but it's not in the bundle when I extract it:

Support Bundle Analysis support-bundle-halibut Admin Console 2025-03-27 at 11 22 34 AM

emosbaugh (approved):

emosbaugh (approved):

PR #5274, opened 2025-04-14
use EC kinds for the EC node join response, and include the app version

laverya:

This uses an imported struct from embedded-cluster for the embedded-cluster node join response, and also includes the current app version label in that response when an app is installed.

```release-note

```


pkg/handlers/embedded_cluster_node_join_command.go
```diff
@@ -178,13 +169,37 @@ func (h *Handler) GetEmbeddedClusterNodeJoinCommand(w http.ResponseWriter, r *ht
...
+		logger.Error(fmt.Errorf("failed to parse cluster id: %w", err))
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	var currentAppVersionLabel string
+	// attempt to get the current app version label from the installed app
+	installedApps, err := store.GetStore().ListInstalledApps()
+	if err == nil && len(installedApps) > 0 {
+		appVersion, err := store.GetStore().GetAppVersion(installedApps[0].ID, installedApps[0].CurrentSequence)
```
sgalsaleh, line 183: i know it's a weird name, but `CurrentSequence` here doesn't mean currently deployed, it means the latest available version for the app in the admin console / version history page.

pkg/handlers/embedded_cluster_node_join_command.go
```diff
@@ -178,13 +169,37 @@ func (h *Handler) GetEmbeddedClusterNodeJoinCommand(w http.ResponseWriter, r *ht
...
+		if err != nil {
+			logger.Error(fmt.Errorf("failed to get app version: %w", err))
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+		currentAppVersionLabel = appVersion.VersionLabel
+	} else {
+		// if there are no installed apps, we can't get the current app version label
+		logger.Info("no installed apps found")
```
sgalsaleh, line 192: i think this should fail. same if the list of installed apps != 1

pkg/handlers/embedded_cluster_node_join_command.go
laverya: oh well that would cause a problem
thanks for catching this

pkg/handlers/embedded_cluster_node_join_command.go
laverya: You can join nodes before installing an app - or at least that's my understanding of the sequencing

pkg/handlers/embedded_cluster_node_join_command.go
laverya: in my testing the version returned is the deployed version, not the latest available version

pkg/handlers/embedded_cluster_node_join_command.go
sgalsaleh: the app would be installed at this point. the meaning of `installed` here is a bit different... it's not that the app got deployed, it means that the license (and airgap bundle) have been processed.

pkg/handlers/embedded_cluster_node_join_command.go
laverya: https://github.com/replicatedhq/kots/blob/0589dd2ff354b3d2c0a688663abafc54b397476f/pkg/store/kotsstore/downstream_store.go#L142

I think the code bears this out too

sgalsaleh: that's not where app.CurrentSequence comes from. it comes from the app table. in your testing, did you have a newer non-deployed version that has a different version label?

https://github.com/replicatedhq/kots/blob/5727dfdae28d6011a0c94f368caed884f75bcaac/pkg/store/kotsstore/app_store.go#L207

pkg/handlers/embedded_cluster_node_join_command.go
laverya: Gotcha
I'd still rather fall back to "we don't include the app version in the response" instead of "you can't join nodes" though - this is an extra layer of safety check, but far from the only one

pkg/handlers/embedded_cluster_node_join_command.go
laverya: Yes, I have a newer version available: 
Screenshot 2025-04-16 at 11 51 56

But the install response includes "appver-dev-36Xpeq" not "local-test"

sgalsaleh: and rollbacks aren't supported in EC, so the app current sequence will always be the deployed sequence. i'm fine with that, just need to add a comment explaining this... not sure if we plan to support rollbacks in the future

pkg/handlers/embedded_cluster_node_join_command.go
```diff
@@ -178,13 +169,41 @@ func (h *Handler) GetEmbeddedClusterNodeJoinCommand(w http.ResponseWriter, r *ht
...
+	}
+
+	var currentAppVersionLabel string
+	// attempt to get the current app version label from the installed app
+	installedApps, err := store.GetStore().ListInstalledApps()
+	if err == nil && len(installedApps) > 0 {
+		// "CurrentSequence" is the latest available version of the app in a non-embedded cluster.
+		// However, in an embedded cluster, the "CurrentSequence" is the latest version of the app.
+		// This is because EC uses the new upgrade flow, which only downloads and renders app releases when
+		// the app version is being installed.
```
sgalsaleh, line 186:
```suggestion
		// "CurrentSequence" is the latest available version of the app in a non-embedded cluster.
		// However, in an embedded cluster, the "CurrentSequence" is also the currently deployed version of the app.
		// This is because EC uses the new upgrade flow, which only creates a new app version when
		// the app version gets deployed. And because rollbacks are not supported in embedded cluster yet.
```

sgalsaleh (approved):

PR #5280, opened 2025-04-16
add hidden 'get join-command' command for embedded-cluster

laverya:

```release-note

```

### DIFF
--- a/cmd/kots/cli/get-joincommand.go
+++ b/cmd/kots/cli/get-joincommand.go
@@ -5,16 +5,17 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
 	"github.com/replicatedhq/kots/pkg/api/handlers/types"
 	"github.com/replicatedhq/kots/pkg/auth"
 	"github.com/replicatedhq/kots/pkg/k8sutil"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
-	"io"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
-	"net/http"
-	"strings"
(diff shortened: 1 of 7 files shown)


cmd/kots/cli/get-joincommand.go
```diff
@@ -0,0 +1,172 @@
...
+			namespace, err := getNamespaceOrDefault(v.GetString("namespace"))
+			if err != nil {
+				return fmt.Errorf("failed to get namespace: %w", err)
+			}
+
+			joinCmd, err := getJoinCommandCmd(cmd.Context(), clientset, namespace)
+			if err != nil {
+				return err
+			}
+			fmt.Println(joinCmd)
```
sgalsaleh, line 49: i think we should also support a `-ojson` output as people will want to use this command mostly in ci

laverya: done

cmd/kots/cli/get-joincommand.go
```diff
@@ -46,58 +47,78 @@ func GetJoinCmd() *cobra.Command {
...
+				}
+				fmt.Println(string(b))
+				return nil
+			}
+
+			return fmt.Errorf("invalid output format: %s", format)
 		},
 	}
 
+	cmd.Flags().StringVarP(&format, "output", "o", "string", "Output format. One of: string, json")
```
sgalsaleh, line 73: sorry, but to be consistent with other commands, we should use `""` instead of `string`

laverya: okay - so `""` or `string` gets string output, `json` gets json output?

sgalsaleh: no need for `string` at all IMO

laverya: updated again

PR #5293, opened 2025-04-24
Add endpoint that returns the EC infra images from the k0s directory

sgalsaleh:

Adds endpoint that returns the EC infra images from the k0s directory.

[SC-122568](https://app.shortcut.com/replicated/story/122568)

In EC repo.

```release-note
NONE
```

NONE

### DIFF
--- a/pkg/handlers/embedded_cluster_artifacts.go
+++ b/pkg/handlers/embedded_cluster_artifacts.go
@@ -7,8 +7,6 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
-	"sort"
-	"strings"
 
 	"github.com/klauspost/pgzip"
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


pkg/handlers/embedded_cluster_artifacts.go
```diff
@@ -114,3 +113,104 @@ func (h *Handler) GetEmbeddedClusterBinary(w http.ResponseWriter, r *http.Reques
...
+
+	if len(infraImages) == 0 {
+		logger.Error(errors.New("no infrastructure image files found"))
+		w.WriteHeader(http.StatusNotFound)
+		return
+	}
+
+	// Sort by modification time, newest first
+	sort.Slice(infraImages, func(i, j int) bool {
+		return infraImages[i].ModTime().After(infraImages[j].ModTime())
```
emosbaugh, line 188: Can you give me an example of the file names you expect?

pkg/handlers/embedded_cluster_artifacts.go
```diff
@@ -114,3 +113,104 @@ func (h *Handler) GetEmbeddedClusterBinary(w http.ResponseWriter, r *http.Reques
...
+		}
+
+		// Get file info for sorting by modification time
+		fileInfo, err := entry.Info()
+		if err != nil {
+			logger.Error(errors.Wrapf(err, "failed to get file info for %s", entry.Name()))
+			continue
+		}
+
+		fmt.Println("fileInfo", fileInfo.Name(), fileInfo.ModTime())
```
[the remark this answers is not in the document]
sgalsaleh: whoops. no. removing.

pkg/handlers/embedded_cluster_artifacts.go
sgalsaleh: `images-amd64.tar` on install, and `ec-images-amd64.tar` on upgrade. we should consolidate those names.

emosbaugh: Perhaps a comment explaining that?

pkg/handlers/embedded_cluster_artifacts.go
```diff
@@ -114,3 +113,107 @@ func (h *Handler) GetEmbeddedClusterBinary(w http.ResponseWriter, r *http.Reques
...
+
+	// Filter and sort image files
+	var infraImages []os.FileInfo
+	for _, entry := range entries {
+		if entry.IsDir() {
+			continue
+		}
+
+		// Check if it's an infra images file
+		// Note: The directory can contain two types of image files:
```
emosbaugh, line 164: Why use mod time and not hardcode to the one with the ec prefix? I'm still not sure I understand the logic here

sgalsaleh: was trying to make it a bit resilient to future name changes in ec, but probably didn't have to. i can make only check for these two names for now

sgalsaleh: updated

emosbaugh (approved):

PR #5303, opened 2025-04-28
Add endpoint that returns the EC charts from the data dir

sgalsaleh:

Adds endpoint that returns the EC charts from the data dir.

[SC-122784](https://app.shortcut.com/replicated/story/122784)

Yes, in EC repo.

```release-note
NONE
```

NONE

### DIFF
--- a/pkg/handlers/embedded_cluster_artifacts.go
+++ b/pkg/handlers/embedded_cluster_artifacts.go
@@ -112,6 +112,99 @@ func (h *Handler) GetEmbeddedClusterBinary(w http.ResponseWriter, r *http.Reques
 	}
 }
 
+// GetEmbeddedClusterCharts returns the charts in the <data-dir>/charts directory as a .tgz file
+func (h *Handler) GetEmbeddedClusterCharts(w http.ResponseWriter, r *http.Request) {
+	if !util.IsEmbeddedCluster() {
+		logger.Error(errors.New("not an embedded cluster"))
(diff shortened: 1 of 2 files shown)


pkg/handlers/embedded_cluster_artifacts.go
```diff
@@ -182,3 +239,71 @@ func (h *Handler) GetEmbeddedClusterK0sImages(w http.ResponseWriter, r *http.Req
...
+
+		// Copy file data into tar writer
+		if _, err := io.Copy(tarWriter, chartFile); err != nil {
+			return errors.Wrapf(err, "failed to copy file %s to tar archive", filePath)
+		}
+
+		return nil
+	})
+
+	if err != nil {
```
emosbaugh, line 304: you should remove the file here

emosbaugh (changes requested):

pkg/handlers/embedded_cluster_artifacts.go
sgalsaleh: why? the file gets removed at the end of the handler after streaming.

sgalsaleh: oh nevermind. i see what you're saying.

sgalsaleh: updated. ty.

emosbaugh (approved):

PR #5313, opened 2025-05-07
feat(ec): do not allow upgrading more than 1 k8s minor

JGAntunes:

This is a temporary measure while we work on releasing 1.31 (see https://github.com/replicatedhq/embedded-cluster/pull/2148), to avoid having customers jump between more than 1 kubernetes minor version (e.g. moving from 1.29 to 1.31).

What we do is return an error that gets exposed in the UI when the user triggers a deploy. This is how it's currently showing up after a user presses `deploy` under the upgrade:

image

https://app.shortcut.com/replicated/story/113724/release-ec-with-support-for-1-31

Yes

```release-note
* Limit Kubernetes version upgrades in Embedded Cluster to no more than one minor version at a time.
```

Maybe?


pkg/upgradeservice/bootstrap.go
```diff
@@ -40,6 +46,41 @@ func bootstrap(params types.UpgradeServiceParams) (finalError error) {
...
+// at most one minor version ahead of the current version.
+func updateWithinKubeRange(params types.UpgradeServiceParams) error {
+	currentVersion, err := extractKubeVersion(params.CurrentECVersion)
+	if err != nil {
+		return errors.Wrap(err, "failed to extract current kube version")
+	}
+	updateVersion, err := extractKubeVersion(params.UpdateECVersion)
+	if err != nil {
+		return errors.Wrap(err, "failed to extract update kube version")
+	}
```
JGAntunes, line 59: We're erroring out if we fail to extract a valid kube version effectively halting the upgrade process altogether. So my question would be:
- Could there be EC releases out there without the `*+k8s-<version>*` version syntax?
- Should we be more cautious here and instead ignore the check altogether if we fail to extract any of the versions?

pkg/upgradeservice/bootstrap_test.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestUpdateWithinKubeRange(t *testing.T) {
+	tests := []struct {
+		name           string
+		params         types.UpgradeServiceParams
+		expectError    bool
+		expectedErrMsg string
+	}{
```
JGAntunes, line 17: I don't think we have a good way to test the `bootstrap` and `Serve` methods right now, particularly because of the large volume of side effects the methods seem to have. I've opted for some unit tests here but curious to hear your thoughts:
- https://github.com/replicatedhq/kots/blob/1153828cf4b252ccdc0d05da60c22e810e4c8a99/pkg/upgradeservice/server.go#L23-L29

pkg/upgradeservice/bootstrap.go
```diff
@@ -40,6 +46,41 @@ func bootstrap(params types.UpgradeServiceParams) (finalError error) {
...
+	if currentVersion.Major() != updateVersion.Major() {
+		return errors.Errorf("major version mismatch: current %s, update %s", currentVersion.String(), updateVersion.String())
+	}
+	if updateVersion.Minor() > currentVersion.Minor()+1 {
```
emosbaugh, line 63: do you want to prevent downgrades here as well? or is that handled somewhere else?

pkg/upgradeservice/bootstrap.go
```diff
@@ -40,6 +46,41 @@ func bootstrap(params types.UpgradeServiceParams) (finalError error) {
...
+	currentVersion, err := extractKubeVersion(params.CurrentECVersion)
+	if err != nil {
+		return errors.Wrap(err, "failed to extract current kube version")
+	}
+	updateVersion, err := extractKubeVersion(params.UpdateECVersion)
+	if err != nil {
+		return errors.Wrap(err, "failed to extract update kube version")
+	}
+	if currentVersion.Major() != updateVersion.Major() {
+		return errors.Errorf("major version mismatch: current %s, update %s", currentVersion.String(), updateVersion.String())
```
emosbaugh, line 61: the %s formatter calls the string function when implemented

https://pkg.go.dev/fmt#Stringer

```suggestion
		return errors.Errorf("major version mismatch: current %s, update %s", currentVersion, updateVersion)
```

pkg/upgradeservice/bootstrap.go
```diff
@@ -40,6 +46,41 @@ func bootstrap(params types.UpgradeServiceParams) (finalError error) {
...
+	}
+	if updateVersion.Minor() > currentVersion.Minor()+1 {
+		return errors.Errorf("cannot update more than one minor version: current %s, update %s", currentVersion.String(), updateVersion.String())
```
emosbaugh, line 64:
```suggestion
		return errors.Errorf("cannot update more than one minor version: current %s, update %s", currentVersion, updateVersion)
```

pkg/upgradeservice/bootstrap_test.go
emosbaugh: I think a unit test where you have one is the best option at this point.

pkg/upgradeservice/bootstrap.go
JGAntunes: Done via https://github.com/replicatedhq/kots/pull/5313/commits/f12427d45c8c1382cf6f2a5aae11ff3feecd0d33

emosbaugh (dismissed):

pkg/upgradeservice/bootstrap.go
```diff
@@ -22,6 +24,10 @@ import (
 )
 
 func bootstrap(params types.UpgradeServiceParams) (finalError error) {
+	if err := updateWithinKubeRange(params); err != nil {
+		return errors.Wrap(err, "kubernetes version update is not within allowed range. Please update at most one minor version at a time")
```
ajp-io, line 28:
```suggestion
		return errors.Wrap(err, "kubernetes version update is not within allowed range")
```

pkg/upgradeservice/bootstrap.go
```diff
@@ -40,6 +46,44 @@ func bootstrap(params types.UpgradeServiceParams) (finalError error) {
...
+	}
+	updateVersion, err := extractKubeVersion(params.UpdateECVersion)
+	if err != nil {
+		return errors.Wrap(err, "failed to extract update kube version")
+	}
+	if currentVersion.Major() != updateVersion.Major() {
+		return errors.Errorf("major version mismatch: current %s, update %s", currentVersion, updateVersion)
+	}
+	if currentVersion.GreaterThan(updateVersion) {
+		return errors.Errorf("cannot downgrade the version: current %s, update %s", currentVersion, updateVersion)
```
ajp-io, line 64:
```suggestion
		return errors.Errorf("cannot downgrade the kubernetes version: current %s, update %s", currentVersion, updateVersion)
```

pkg/upgradeservice/bootstrap.go
```diff
@@ -40,6 +46,44 @@ func bootstrap(params types.UpgradeServiceParams) (finalError error) {
...
+	}
+	if updateVersion.Minor() > currentVersion.Minor()+1 {
+		return errors.Errorf("cannot update more than one minor version: current %s, update %s", currentVersion, updateVersion)
```
ajp-io, line 67:
```suggestion
		return errors.Errorf("cannot update by more than one kubernetes minor version: current %s, update %s", currentVersion, updateVersion)
```

ajp-io:
mind updating the screenshot in the description with what the error now looks like?

emosbaugh (approved):

JGAntunes:
> mind updating the screenshot in the description with what the error now looks like?

Missed it sorry. Yeah sure.

JGAntunes:
@ajp-io updated the screenshot.

emosbaugh (approved):

pkg/upgradeservice/bootstrap.go
sgalsaleh: new versions should all have `k8s-<version>`. i think this looks good for now.

sgalsaleh (approved):

PR #5317, opened 2025-05-10
update upgrade task status timestamp safely

sgalsaleh:

Updates upgrade task status timestamp safely in case of panics so they don't halt the upgrade progress in the UI if rqlite temporarily lost quorum.

[SC-123136](https://app.shortcut.com/replicated/story/123136)

```release-note
NONE
```

NONE


sgalsaleh:
Because we don't want the upgrade to fail completely for this transient rqlite error.

emosbaugh (approved):

PR #5345, opened 2025-06-02
Fix goreleaser builds

sgalsaleh:

Build fails with:
```bash
link: duplicated definition of symbol dlopen, from github.com/ebitengine/purego and github.com/ebitengine/purego
```
Which is fixed in https://github.com/ebitengine/purego/releases/tag/v0.8.3

NONE

```release-note
NONE
```

NONE


sgalsaleh:
Merging as this only affects goreleaser which runs on release not on PR.

PR #5364, opened 2025-06-27
allow updating service account token

laverya:

This allows rolling the service account token in online embedded cluster installations, by adding:
1. an additional dialog button to the license page allowing the user to paste a new token
2. backend logic to not reject serviceaccount token changes when the token identity has not changed
3. frontend logic to allow bypassing the "same license ID" check if the service account IDs still match in airgap

#### Screenshots:
Screenshot 2025-06-27 at 14 39 58
Screenshot 2025-06-27 at 14 40 16
Screenshot 2025-06-27 at 14 40 41
Screenshot 2025-06-27 at 14 41 28

Currently serviceaccounts are only used with Embedded Cluster installs, so no.

```release-note
Allow rotating service account tokens.
```


pkg/util/auth.go
```diff
@@ -0,0 +1,35 @@
...
+		}
+	}
+
+	return licenseID, licenseID
+}
+
+func GetLicenseAuthHeader(licenseID string) string {
+	username, password := GetLicenseCredentials(licenseID)
+	return "Basic " + base64.StdEncoding.EncodeToString([]byte(username+":"+password))
+}
```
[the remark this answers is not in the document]
laverya: It seems like it was go formatted besides the missing newline

pkg/kotsadmlicense/serviceaccount.go
```diff
@@ -0,0 +1,54 @@
...
+
+// ValidateServiceAccountToken checks if the service account token is valid and if it matches the current license identity.
+// It returns the service account token, a boolean indicating if the secret has been updated, and an error if any.
+func ValidateServiceAccountToken(token, currentLicenseID string) (*ServiceAccountToken, bool, error) {
+	newToken, err := extractIdentityFromLicenseID(token)
+	if err != nil {
+		return nil, false, fmt.Errorf("failed to extract new token identity: %w", err)
+	}
+
+	currentToken, err := extractIdentityFromLicenseID(currentLicenseID)
```
sgalsaleh, line 22: not all license ids are in token format, no?

laverya: Correct - if your license ID is a classic license ID and not a token, then the button to update the token will not be shown

sgalsaleh: should probably check if it's a service account token similar to how the ui does it?

laverya: That's what this is doing though? Apparently I need more comments on the functions

laverya: added a comment on the 'extractIdentityFromLicenseID' function, and then renamed it to 'extractIdentityFromToken'

PR #5420, opened 2025-07-15
Add support for additional license fields in template context

0xJMart:
## Summary

This PR adds support for additional license fields in the KOTS template context, enabling KOTS configs to access more license field values through the `LicenseFieldValue` template function.

### New License Fields Added:
- **`isDisasterRecoverySupported`** - camelCase version of the existing field
- **`isKotsInstallEnabled`** - enables access to KOTS install permission
- **`isEmbeddedClusterDownloadEnabled`** - enables access to embedded cluster download permission  

### Testing
- Added comprehensive test coverage for all new license fields
- Tests cover both true and false boolean cases
- Ensures proper string conversion for template rendering

## Use Case

This enables KOTS configuration files to display license capabilities:

```yaml
- name: dr_support
  title: DR
  type: text
  value: "{{repl LicenseFieldValue \"isDisasterRecoverySupported\"}}"
  readonly: true
```

## Backward Compatibility

- All existing license field template functions remain unchanged
- New fields are additive and don't affect existing functionality
- Falls back to custom entitlements if fields aren't found

Resolves the need for accessing additional license field values in KOTS configuration templates.

Relavent shortcut:

https://app.shortcut.com/replicated/story/126762/add-support-for-additional-license-fields-in-template-context

### DIFF
--- a/pkg/template/license_context.go
+++ b/pkg/template/license_context.go
@@ -44,6 +44,8 @@ func (ctx licenseCtx) licenseFieldValue(name string) string {
 		return strconv.FormatBool(ctx.License.Spec.IsSnapshotSupported)
 	case "IsDisasterRecoverySupported":
 		return strconv.FormatBool(ctx.License.Spec.IsDisasterRecoverySupported)
+	case "isDisasterRecoverySupported":
+		return strconv.FormatBool(ctx.License.Spec.IsDisasterRecoverySupported)
 	case "isGitOpsSupported":
 		return strconv.FormatBool(ctx.License.Spec.IsGitOpsSupported)
 	case "isSupportBundleUploadSupported":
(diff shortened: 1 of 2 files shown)


pkg/template/license_context.go
```diff
@@ -44,6 +44,8 @@ func (ctx licenseCtx) licenseFieldValue(name string) string {
 		return strconv.FormatBool(ctx.License.Spec.IsSnapshotSupported)
 	case "IsDisasterRecoverySupported":
 		return strconv.FormatBool(ctx.License.Spec.IsDisasterRecoverySupported)
+	case "isDisasterRecoverySupported":
```
diamonwiggins, line 47: looks like a duplicate of what's above?

pkg/template/license_context.go
```diff
@@ -76,6 +78,10 @@ func (ctx licenseCtx) licenseFieldValue(name string) string {
 		return util.ReplicatedAppEndpoint(ctx.License)
 	case "licenseID", "licenseId":
 		return ctx.License.Spec.LicenseID
+	case "isKotsInstallEnabled":
```
diamonwiggins, line 81: I don't see this in - https://github.com/replicatedhq/kotskinds/blob/main/apis/kots/v1beta1/license_types.go#L112-L136

pkg/template/license_context.go
0xJMart: Specifically this is a normal camelCase option, leaving the other one there for backwards compatibility in case users are currently using the `Is` vs `is`

pkg/template/license_context.go
0xJMart: Oh interesting, do I go ahead and add it in there? The field is currently on our license.yamls.

PR #5472, opened 2025-08-13
feat(ec-updates): don't allow updates of >1 minor kube versions for airgap

JGAntunes:

This PR depends on:
- https://github.com/replicatedhq/kots/pull/5471

Implements that same logic but for airgap updates

https://app.shortcut.com/replicated/story/127829/update-kots-admin-console-so-version-that-are-1-minor-version-of-kubernetes-ahead-of-the-currently-deployed-version-cannot

Yes

```release-note
NONE
```

Yes (but it's EC specific).


pkg/update/required.go
```diff
@@ -119,3 +151,16 @@ func getNonDeployableCause(requiredUpdates []string) string {
...
+
+// getKubeVersionNonDeployableCause constructs a non-deployable cause message based on the kube range validation error message
+func getKubeVersionNonDeployableCause(err error) string {
+	switch {
+	case errors.Is(err, util.ErrKubeMinorRangeMismatch):
+		return "Before you can update to this version, you need to update to an earlier version that includes the required infrastructure update."
+	case errors.Is(err, util.ErrKubeVersionDowngrade):
+		return "Release includes a downgrade of the infrastructure version, which is not allowed. Cannot use release."
+	case errors.Is(err, util.ErrKubeMajorVersionUpgrade):
+		return "Release includes a major version upgrade of the infrastructure version, which is not allowed. Cannot use release."
```
ajp-io, line 163: I wonder if we even need this given that another major version of kubernetes doesn't exist. like how could this even happen right now?

pkg/update/required.go
```diff
@@ -119,3 +151,16 @@ func getNonDeployableCause(requiredUpdates []string) string {
...
+	}
+	return "Cannot validate the infrastructure version compatibility for this update. Cannot use release."
```
ajp-io, line 165:
```suggestion
	return "This version cannot be deployed because the required infrastructure compatibility could not be verified."
```

sgalsaleh:
Closing in favor of https://github.com/replicatedhq/kots/pull/5473 as there's many conflicts in this PR

PR #5525, opened 2025-09-04
fix: bump timeout for apply stdout logs validation in regression tests

diamonwiggins:

bumps timeout in the version history validation to prevent timeouts in the regression tests

```release-note

```


sgalsaleh:
Merging as the change only affects regression tests that only run on main.

PR #5590, opened 2025-09-24
chore: remove cpu resource limits from k8s objects

emosbaugh:

Removes limits for all pods

```release-note
Removes CPU resource limits for all KOTS Admin Console pods.
```


banjoh (commented):
It looks like you removed memory limits as well. Was that intentional? CPU limits are the problematic ones.

emosbaugh:
> It looks like you removed memory limits as well. Was that intentional? CPU limits are the problematic ones.

thanks. updated

PR #5601, opened 2025-09-29
fix(web): sidebar disappears on update

JGAntunes:

The sidebar set state function passed to the `AppConfig` component was being called with an incompatible object type that it wasn't expecting. This lead to the sidebar to disappear on changes in the form fields.

Fix demo:
- https://www.loom.com/share/c31092816aad4b31bcfbb1cc37299a81?sid=8f200b2b-30e4-4590-9260-1d325ac286bd

https://app.shortcut.com/replicated/story/128409/configuration-options-in-admin-console-side-nav-bar-disappear-during-first-installation

Updated the TS types to hopefully prevent this in the future.

```release-note
Fixes issue when editing the release config on Embedded Cluster where the sidebar items would disappear.
```

NONE

### DIFF
--- a/web/src/features/AppConfig/components/AppConfig.tsx
+++ b/web/src/features/AppConfig/components/AppConfig.tsx
@@ -18,7 +18,7 @@ import { Utilities } from "../../../utilities/utilities";
 import Icon from "@src/components/Icon";
 
 // Types
-import { App, KotsParams, Version } from "@types";
+import { App, KotsParams, Version, NavbarConfigGroup } from "@types";
 
 type Props = {
   location: ReturnType<typeof useLocation>;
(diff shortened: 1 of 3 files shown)


web/src/features/AppConfig/components/AppConfig.tsx
```diff
@@ -606,7 +606,7 @@ class AppConfig extends Component<Props, State> {
         const changed = this.isConfigChanged(newGroups);
         this.setState({ configGroups: newGroups, changed });
         if (this.props.isEmbeddedCluster) {
-          this.props.setNavbarConfigGroups({ newGroups, changed });
```
JGAntunes, line 609: This was the actual problem, we were passing an object instead of an array of config group.

web/src/types/index.ts
```diff
@@ -345,3 +345,22 @@ export type SnapshotSettings = {
...
   };
 };
+
+export type ConfigGroupItem = {
+  name: string;
+  title: string;
+  type: string;
+  hidden: boolean;
+  validationError: string;
+  error: string;
```
JGAntunes, line 355: Changed these 2 values to string VS boolean based on the types we had defined in `AppConfig` and what I could find in the code (I do believe these should be strings). TSC said ok so I believe we're fine.

PR #5790, opened 2026-02-26
fix: prevent deploying updates when current required version is failed to be deployed

nvanthao:

Similar to https://github.com/replicatedhq/kots/pull/5787 but for Embedded Cluster

image

Re-config required release still allows to be deployed

image

After deployed successfully, future updates can be deployed again

image

```release-note

```


banjoh (approved):
LGTM

PR #5879, opened 2026-05-01
fix(ci): copy minio from proxy registry instead of building with apko

emosbaugh:
## Summary

This PR contains four related fixes for the minio image dependency workflow:

### 1. Fix image-deps-updater workflow: copy minio from proxy registry instead of building with apko

The image-deps-updater workflow was failing because it referenced deploy/minio/apko.yaml, which no longer exists (it was renamed to apko.yaml.tmpl and is now removed entirely).

Instead of building minio with apko, this PR changes the workflow to copy minio from proxy.replicated.com/library/minio using skopeo copy --all, consistent with how rqlite and dex are already handled. The minio tag pattern in the workflow was also corrected to match RELEASE.YYYY-MM-DDTHH-MM-SSZ tags instead of semver.

Fixes failing workflow run: https://github.com/replicatedhq/kots/actions/runs/25203464852

### 2. Fix race condition in findReplicatedChart causing flaky tests

The Test_configureChart test was intermittently failing in CI.

Root cause: findReplicatedChart iterates through Chart.yaml files in tar order. Tests build tars via util.FilesToTGZ which iterates over a map[string]string — Go map iteration order is randomized. When a subchart's Chart.yaml appeared before the root Chart.yaml, the alias lookup failed because topLevelDeps was still empty.

Fix: findReplicatedChart now does a two-pass scan:
1. First pass: collect all relevant Chart.yaml files (root + direct subcharts)
2. Process the root Chart.yaml first to populate topLevelDeps
3. Then scan subcharts for a match, using the now-populated alias info

This makes the function deterministic regardless of tar entry order.

### 3. Update cmd/imagedeps for new minio RELEASE tag format

The imagedeps tool that generates pkg/image/constants.go, .image.env, and Makefile variables was using semver parsing to find the latest minio tag. Since minio now uses RELEASE.YYYY-MM-DDTHH-MM-SSZ tags (not semver), the tool was broken.

Changes:
- Updated image-spec regex to match RELEASE.* tags
- Added getLatestMinioTagFromRegistry() that normalizes tags by removing non-digits and sorts lexicographically
- Updated test data and expected outputs
- Removed the unused deploy/minio/apko.yaml.tmpl file
- Ran go run ./cmd/imagedeps to update all image constants

### 4. Use /bin/sh instead of /bin/bash in minio migration init container scripts

The upstream minio image uses busybox and does not include /bin/bash. The init container migration scripts used bash shebangs (#!/bin/bash) and bashisms like shopt -s dotglob, causing the containers to fail with 'bash: applet not found'.

Changes:
- Changed shebangs to #!/bin/sh in copy-minio-client.sh, export-minio-data.sh, and import-minio-data.sh
- Replaced shopt -s dotglob with a POSIX-compatible loop in import-minio-data.sh


sgalsaleh (approved):

emosbaugh:
[sc-137352]

emosbaugh:
[sc-137353]

PR #5888, opened 2026-05-04
Bump rqlite to 10.0.2 and unpin version constraint

emosbaugh:
This PR updates rqlite from 9.4.5 to 10.0.0 and removes the hardcoded major version pin (9.x) in the image-deps-updater workflow so future updates can pick up any rqlite major version.


emosbaugh:
## Error

> [store] 2026/05/04 20:42:12 opening store with node ID kotsadm-rqlite-0, listening on kotsadm-rqlite-0.kotsadm-rqlite-headless.qakotsregression.svc.cluster.local:4002
> [store] 2026/05/04 20:42:12 ensuring data directory exists at /rqlite/file/data
> [store] 2026/05/04 20:42:12 old v7 snapshot directory does not exist at /rqlite/file/data/snapshots, nothing to upgrade
> [store] 2026/05/04 20:42:12 old v8 snapshot directory does not exist at /rqlite/file/data/rsnapshots, nothing to upgrade
> [snapshot-store] 2026/05/04 20:42:12 store initialized using /rqlite/file/data/wsnapshots
> [snapshot-store] 2026/05/04 20:42:12 completed CRC32 check of 1 snapshot directories in 261.961µs
> [store] 2026/05/04 20:42:12 1 preexisting snapshots present
> [store] 2026/05/04 20:42:12 detected successful prior snapshot operation, skipping restore
> [store] 2026/05/04 20:42:12 CRC32 checksum mismatch during clean snapshot check - aborting

## Analysis

The rqlite v10 release notes claim upgrades are "seamless," but that refers to multi-node rolling upgrades where nodes are upgraded one-by-one while the cluster stays online. In that scenario, a new v10 node either starts fresh or joins an existing cluster and receives snapshots from the v9 leader through the cluster protocol.

But KOTS deploys rqlite as a single-node StatefulSet (-bootstrap-expect=1). There's no cluster to roll through. On upgrade:
1. The v9 pod is killed
2. A new v10 pod mounts the same PVC with existing data
3. rqlited v10 starts and successfully upgrades the v8-format snapshot store (rsnapshots) to the v10 format (wsnapshots) — this part works
4. rqlited then tries to skip restoring the SQLite database from the snapshot store by reusing the on-disk DB file directly
5. To do this safely, it reads a clean_snapshot fingerprint file that v9 wrote during its last successful snapshot. This file contains the DB file's size, mtime, and a CRC32 checksum
6. v10 computes the CRC32 using the Castagnoli polynomial, but v9 wrote the fingerprint using the IEEE polynomial (PR #2488 (https://github.com/rqlite/rqlite/pull/2488)). The values don't match
7. rqlited treats this as data corruption and aborts with CRC32 checksum mismatch during clean snapshot check - aborting

The backward-compatibility logic in FileFingerprint.Compare only handles a zero/missing CRC32 field, not a polynomial change. Since the fingerprint file exists and has a non-zero CRC32, v10 never falls back to "I don't know what this is, let me restore from the snapshot store."

This is a gap between how rqlite expects upgrades to happen and how Kubernetes StatefulSets actually work when you change the container image while the data directory persists on a volume. The upstream project likely never tested this exact pattern: a single-node deployment where the container is destroyed and recreated with a new image, leaving behind a clean_snapshot fingerprint written by the old version.

PR #5933, opened 2026-06-12
feat(reporting): generate new instance IDs upon snapshot restore

kriscoleman:
## Summary

When a vendor restores from a snapshot, the restored installation reuses the same `instance_id` as the original, so every environment restored from one backup reports as a single instance (sc-130386 has 3,000+ installations under one ID). This PR detects when a backup lands in a different environment and regenerates the reported instance ID, recording lineage so the portal can link the new instance back to the one it came from.

Shortcut: [sc-133130](https://app.shortcut.com/replicated/story/133130)

## How it works

The short version of the design:

- We store an environment fingerprint in `kotsadm_params` (key `ENVIRONMENT_FINGERPRINT`), so it travels inside every backup. The fingerprint is the `kube-system` namespace UID (canonical cluster identity, immune to IP changes) with the kotsadm pod namespace UID as a fallback for namespace-scoped RBAC.
- At startup, `reporting.Init` compares the stored fingerprint against the live cluster. No stored fingerprint means adopt and keep (that's the upgrade path, nothing regenerates on rollout). Same `kube-system` UID means keep. Different UID means regenerate. Nothing comparable means keep. We fail safe: never regenerate when uncertain.
- The app ID is a primary key with FKs everywhere, so we don't touch it. Instead the reported instance ID is decoupled into two new nullable `app` columns: `instance_id` and `instance_id_lineage` (JSON array, oldest first). NULL `instance_id` means we report the app ID, exactly today's behavior.
- On a detected restore, each installed app gets a new KSUID and the old ID is appended to the lineage. Reports gain `restored_from_instance_id` (the immediate parent) plus an `X-Replicated-RestoredFromInstanceID` header, so a chain of restores (A → B → C) is reconstructable server-side.

This covers in-place DR (same cluster, ID kept), plain IP changes (no addresses in the fingerprint, ID kept), CI environments restoring the same snapshot repeatedly (each gets its own ID, chained), and original + restored running in parallel in different clusters (distinct IDs).

## What this does NOT cover

- **Parallel byte-identical VM clones.** A VM cloned byte-for-byte carries the same `kube-system` UID, and there's no in-cluster signal that distinguishes it from the original (IP/MAC changes are explicitly not safe signals). Deduplicating those needs server-side correlation, e.g. spotting interleaved or regressing report timelines on one `instance_id`. That's out of scope here and tracked as follow-up bead ko-z28.
- **DR onto a rebuilt cluster gets a new ID.** "Same environment" is defined as "same Kubernetes cluster". If you destroy and rebuild the cluster, the restore regenerates, but lineage points at the old instance so the portal can present continuity.
- **Preflight reporting is unchanged.** The online preflight endpoint encodes the app ID in its URL path, so it keeps using the app ID.

## Changes

- `pkg/reporting/restore_detection.go`: fingerprint capture, comparison, and regeneration logic, run from `reporting.Init`
- `migrations/tables/app.yaml`: new nullable `instance_id` and `instance_id_lineage` columns
- `pkg/store/kotsstore`: `GetAppInstanceID` / `RegenerateAppInstanceIDs` and fingerprint params accessors
- `pkg/api/reporting/types`, `pkg/reporting/{app,app_airgap,instance_report,util}.go`: resolve and report `instance_id` + `restored_from_instance_id` (events and update-check header)
- Tests: fingerprint comparison matrix, regeneration + lineage chaining, store round-trips

## UAT

1. Fresh install on this branch, then check `app.instance_id` is NULL and reports carry the app ID. Restart kotsadm; nothing regenerates.
2. Take a snapshot, restore it into a different cluster. On startup kotsadm logs the regeneration, `app.instance_id` is a new KSUID, `instance_id_lineage` contains the old ID, and instance reports include `restored_from_instance_id`.
3. Restore the same snapshot again into a third cluster and confirm the lineage chains (two entries, parent is the second ID).
4. Restore into the SAME cluster (in-place DR): ID kept.
5. Upgrade an existing install to this version: no stored fingerprint, so it adopts silently and the ID is kept.


kriscoleman:
**Code review (Go) — verdict: PASS** (2 cycles, all green)

Posting for the record. The fingerprint comparison matrix implements the design rules exactly (adopt / kube-system UID / pod-namespace UID fallback / fail-safe keep), and the crash-recovery invariant holds in code: the fingerprint is persisted as the last step after the per-app regeneration loop, so a mid-loop crash re-runs detection next boot. Verified with a `gomock.InOrder` test plus an error-path case asserting the fingerprint is never persisted when an app update fails.

Startup can't be bricked: detection errors are logged and swallowed, and `reporting.Init` failure is itself non-fatal. The synchronous namespace reads are now bounded by a 10s timeout. Migration is additive (two nullable text columns, `NULL` falls back to the app id), so existing installs are unaffected.

Two concerns worth recording, both confirmed safe: the keep-refresh fingerprint merge cannot mask a genuine restore (the keep/regenerate decision is made before and independent of the merge, which only backfills empty fields for future boots), and the lineage self-heal can't lose real history on transient corruption because lineage is only ever written via `json.Marshal`, so an unparseable value means external tampering, not a code path. Degrading beats bricking detection on every boot.

kriscoleman:
**Security review — verdict: PASS** (2 cycles, no blocking findings)

Posting for the record.

RBAC: no permission widening. Cluster installs already run with a cluster-wide role, so the new kube-system namespace read is covered; namespace-scoped installs fall back to the pod-namespace UID, and if nothing is readable, detection is a clean no-op (fail-safe keep, never crash).

Information disclosure: only the immediate lineage parent is reported, and the header is omitted entirely when empty. The full lineage chain stays in the local DB. The new header and report field go only to Replicated endpoints that already receive the instance id. Namespace UIDs in the fingerprint are not secrets and never leave the cluster.

Startup DoS: a malformed fingerprint cannot crash-loop kotsadm (errors are caught and non-fatal, an unparseable stored fingerprint re-adopts rather than erroring), and the 10s timeouts actually reduce the prior unbounded-context exposure. The lineage self-heal and fingerprint merge stay within the documented DB-write threat model: an attacker who can corrupt those columns can already spoof reports wholesale, so neither change adds a capability. SQL is parameterized throughout; migration is additive with no backfill.

One thing to flag for product, not a code issue: parallel byte-identical VM clones are out of scope here (same cluster UID, same everything) and need server-side correlation. Tracked as follow-up ko-z28.

pkg/api/reporting/types/types.go
```diff
@@ -2,6 +2,7 @@ package types
 
 type ReportingInfo struct {
 	InstanceID              string         `json:"instance_id" yaml:"instance_id"`
+	RestoredFromInstanceID  string         `json:"restored_from_instance_id" yaml:"restored_from_instance_id"`
```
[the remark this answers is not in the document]
kriscoleman: Good catch, fixed in 8f07642a0. Added omitempty to both the json and yaml tags so a never-restored instance no longer serializes the empty field, and there's a test now asserting it's omitted when empty and present when set.

pkg/reporting/restore_detection.go
```diff
@@ -0,0 +1,174 @@
...
+			merged.PodNamespaceUID = stored.PodNamespaceUID
+		}
+		mergedJSON, err := json.Marshal(merged)
+		if err != nil {
+			return errors.Wrap(err, "failed to marshal merged environment fingerprint")
+		}
+		if storedJSON != string(mergedJSON) {
+			return errors.Wrap(kotsStore.SetEnvironmentFingerprint(string(mergedJSON)), "failed to refresh environment fingerprint")
+		}
+		return nil
```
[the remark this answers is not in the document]
kriscoleman: Agreed, swapped the raw-JSON-string compare for a structural comparison of the fingerprint struct in 8f07642a0. New subtest feeds a semantically-equal stored value with reordered keys and extra whitespace and asserts no write, so the guard survives future serialization changes.

pkg/reporting/restore_detection.go
```diff
@@ -0,0 +1,177 @@
...
+		merged := current
+		if merged.KubeSystemUID == "" {
+			merged.KubeSystemUID = stored.KubeSystemUID
+		}
+		if merged.PodNamespaceUID == "" {
+			merged.PodNamespaceUID = stored.PodNamespaceUID
+		}
+		// compare structurally, not byte-wise: the stored JSON may differ in key order
+		// or whitespace from what json.Marshal produces without being a real change
+		if merged != stored {
```
sgalsaleh, line 144: is this basically always true because it compares objects?

sgalsaleh (commented):
I noticed that `reporting.Init()` is called later in the bootstrap / start process. Is there any process before it that could potentially send reporting information with the old instance id before this new code updates it?

pkg/reporting/restore_detection.go
kriscoleman: Yea, I actually had to double-check my own read on this one. I don't think it's always true, but I get the instinct. If it were comparing pointers it'd basically always differ. These are both `environmentFingerprint` values though (just two strings), and Go compares structs field-by-field with `==`/`!=`, not by identity. So I believe `merged != stored` only comes back true when one of the UIDs has actually changed.

Steady state, where nothing about the environment moved, `merged` should land equal to `stored` and we skip the write. It's true only when a field genuinely drifted, e.g. kube-system still matches (so we keep the instance id) but the pod-namespace UID changed and we want to refresh what's stored.

There's a test pinning this if it's a useful sanity check: `semantically equal stored fingerprint with different serialization does not write` feeds a reordered/whitespaced-but-equal value and asserts no write, which is exactly what'd break if the compare were always-true. Happy to add an explicit identical-value case too if you'd want the extra coverage. 🤔

kriscoleman:
> I noticed that `reporting.Init()` is called later in the bootstrap / start process. Is there any process before it that could potentially send reporting information with the old instance id before this new code updates it?

🤔 that's a good point. 

I will investigate the actual kots bootstrap order and trace where Init sits relative to reporting paths, and pull this thread a bit.

kriscoleman:
Alright, traced it. Short version: I don't think anything can report with the old id, but let me lay out the path.

The regeneration runs inside `reporting.Init()` itself, and it happens *before* `Init` wires up the `reporter` (both synchronous, same function). In the startup sequence `Init()` also lands before the update checker, the watchers, the snapshot scheduler, and before the HTTP server is even built, so all the request- and timer-driven reports come up after the new id is already in place.

The operator is the one piece that starts before `Init`, and it reports through `GetReporter()`, which returns nil until `Init` assigns it. So even though it's running, it can't actually emit a report until regeneration has happened. And `GetReportingInfo` reads the instance id straight from the DB each call rather than caching it, so there's no stale copy hanging around either.

So I think we're good. The one spot I want to be sure of is the operator's `deployApp` path: if a deploy somehow fired before `Init`, `GetReporter()` is nil there. Deploys are event-driven and land after Init in practice, so it shouldn't happen, but I'd rather confirm it no-ops cleanly than lean on timing. If we want to kill the ordering dependency entirely I can move the restore check ahead of the operator start. 🤔

pkg/reporting/restore_detection.go
sgalsaleh: yeah i think you're right because the struct doesn't contain any map,slice,functions,etc...

kriscoleman:
Traced it the rest of the way. `op.Start()` kicks off `resumeDeployments` in a goroutine before `Init` runs, and that only re-deploys an app if its current version isn't already `Deployed`/`Failed` (and bails if a restore's in progress), so normally it's a no-op on startup. If an app *were* mid-deploy at startup it'd hit `deployApp` → `SubmitAppInfo` while `GetReporter()` is still nil, and since `reporter` is a nil interface there, that's a panic rather than a stale-id send. So the good news is the old id still can't leak.

The catch is it's a pre-existing nil-reporter race, not something this change introduced — `Init` has always assigned the reporter, this PR just added the restore check ahead of it. I think it's worth a tiny hardening either way: make `GetReporter()` nil-safe, or move the restore detection ahead of the operator start so the ordering isn't load-bearing. Leaning toward the latter. 🤔

sgalsaleh (approved):

PR #5940, opened 2026-06-15
fix(airgap): make push-images idempotent against tag-immutable registries

banjoh:

Makes `kubectl kots admin-console push-images` (and other airgap-push flows) idempotent against registries that enforce tag immutability — Artifactory's "Block Pushing Existing Tags", Harbor Tag
Immutability rules, JFrog Container Registry's "Disable Re-Deployment", Quay per-tag immutability, and any OCI registry behind WORM (Write Once, Read Many) storage or a 403-on-overwrite proxy.

Before this change, the second push of a release whose image tags were unchanged from the prior release failed at the first re-pushed manifest, blocking the customer's upgrade. The fix adds two layers
  of defense in `pkg/image/online.go`:

1. **Manifest precheck in `CopyImage`.** Before invoking the copy library, open the destination ref, fetch its current manifest, and compare the digest against the source. If they match, return early
— no PUT is issued, so the registry never sees an overwrite attempt to reject. A destination that is missing or unreachable falls through to the normal push path; only source-side failures bubble up.

2. **`OptimizeDestinationImageAlreadyExists: true` on `copy.Options`.** Covers per-child images if the parent manifest list differs but children are already present. The upstream library skips this
optimization for the parent manifest-list write in the multi-arch path (known [bug](https://github.com/podman-container-tools/container-libs/issues/918)), which is why the KOTS-side precheck is still required.

Both apply to every caller of `CopyImage` — embedded-cluster artifact pushes, docker-registry-format airgap bundles, docker-archive-format airgap bundles, and the online copy path — without per-caller
  plumbing.

**Screenshots**
image

image (1)

https://app.shortcut.com/replicated/story/138458

Yes — unit tests added in `pkg/image/online_test.go` covering:
- destination manifest matches source → skip (idempotent path)
- destination manifest differs from source → push
- destination tag missing → push
- destination unreachable → fall through, no error
- source unreachable → error surfaces

Manual verification against a tag-immutable registry (Harbor with Tag Immutability rule on `**`/`**`, or Artifactory with "Block Pushing Existing Tags"): first `push-images` succeeds; second
`push-images` of the same release completes successfully where it previously failed at the first unchanged tag.

```release-note
Fix `kubectl kots admin-console push-images` so re-pushing a release with unchanged image tags succeeds against registries that enforce tag immutability. KOTS now skips images whose destination manifest already matches the source instead of attempting an overwrite the registry will reject.
```

NONE

### DIFF
--- a/pkg/image/online.go
+++ b/pkg/image/online.go
@@ -306,7 +306,24 @@ func CopyImage(opts types.CopyImageOptions) error {
 		imageListSelection = copy.CopyAllImages
 	}
 
-	_, err := CopyImageWithGC(context.Background(), opts.DestRef, opts.SrcRef, &copy.Options{
+	// Idempotency precheck: if the destination tag already holds a manifest matching
+	// the source, skip the copy. Avoids re-pushing manifests to tag-immutable registries
+	// (Artifactory "Block Pushing Existing Tags", Harbor Tag Immutability, JFrog
+	// "Disable Re-Deployment") that reject overwrite even when the digest is unchanged.
+	matches, err := destinationManifestMatches(context.Background(), opts, srcCtx, destCtx)
+	if err != nil {
+		return errors.Wrap(err, "failed manifest precheck")
+	}
+	if matches {
+		destRefName := ""
+		if opts.DestRef.DockerReference() != nil {
+			destRefName = opts.DestRef.DockerReference().String()
+		}
+		logger.Infof("Skipping push: destination manifest already matches source (dest=%s)", destRefName)
+		return nil
+	}
+
+	_, err = CopyImageWithGC(context.Background(), opts.DestRef, opts.SrcRef, &copy.Options{
 		RemoveSignatures:      true,
 		SignBy:                "",
 		ReportWriter:          opts.ReportWriter,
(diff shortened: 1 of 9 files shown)


pkg/image/online.go
```diff
@@ -438,6 +466,96 @@ func getPolicyContext() (*signature.PolicyContext, error) {
...
+// so the caller proceeds with the normal push. Only source-side failures surface as
+// errors, since a missing or unreadable destination should not block the push.
+func destinationManifestMatches(ctx context.Context, opts types.CopyImageOptions, srcCtx, destCtx *containerstypes.SystemContext) (bool, error) {
+	destImg, err := opts.DestRef.NewImageSource(ctx, destCtx)
+	if err != nil {
+		// Destination does not exist or is unreachable; fall through to the normal push.
+		// Most commonly this is a 404 on first push, but it also masks transient network
+		// errors, 401/403 from misconfigured destination auth, and 429 rate limits — log
+		// at debug to leave a trail when the optimization isn't firing as expected.
+		logger.Debugf("manifest precheck: opening destination failed, falling through to push: %v", err)
```
sgalsaleh: i think this should be an error log. otherwise we won't see it in support bundles i think.

pkg/image/online.go
```diff
@@ -438,6 +466,96 @@ func getPolicyContext() (*signature.PolicyContext, error) {
...
+		return false, nil
+	}
+	defer destImg.Close()
+
+	destManifest, destMIME, err := destImg.GetManifest(ctx, nil)
+	if err != nil {
+		// Destination manifest unreadable; fall through to the normal push.
+		logger.Debugf("manifest precheck: reading destination manifest failed, falling through to push: %v", err)
```
sgalsaleh: ditto

sgalsaleh (approved):

PR #5950, opened 2026-06-25
fix: plumb skip existing images toggle to kotsadm pod

banjoh:

The "Skip Pushing Images That Already Exist" toggle was ignored when uploading an airgap bundle, so uploads to tag-immutable registries still failed. The toggle was only wired through the online registry-settings flow; on the airgap upload path the value was dropped at every layer (UI → API → pull). This threads `skipExistingImages` end-to-end through `AirgapRegistrySettings`/`UploadAirgapBundle` → `CreateAppFromAirgapRequest` → `CreateAirgapAppOpts` → `PullOptions` → `ProcessImageOptions`.

Yes — a handler test asserting the flag reaches `CreateAirgapAppOpts`. Can add if wanted.

```release-note
Fixed the "Skip Pushing Images That Already Exist" setting being ignored during airgap bundle upload, which caused failures against registries that enforce tag immutability.
```

NONE

### DIFF
--- a/dev/dockerfiles/kotsadm-migrations/Dockerfile.ttlsh
+++ b/dev/dockerfiles/kotsadm-migrations/Dockerfile.ttlsh
@@ -12,6 +12,7 @@ ENV DEBIAN_FRONTEND=noninteractive
 RUN apt-get update \
   && apt-get install -y --no-install-recommends \
     ca-certificates \
+    wget \
   && apt-get install -y --only-upgrade --no-install-recommends \
     passwd login \
   && apt-get clean \
--- a/pkg/kotsadm/objects/scripts/wait-for-rqlite.sh
+++ b/pkg/kotsadm/objects/scripts/wait-for-rqlite.sh
@@ -8,6 +8,11 @@
 timeout=300
 elapsed=0
 
+if ! command -v wget >/dev/null 2>&1; then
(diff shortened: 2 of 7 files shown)


dev/dockerfiles/kotsadm-migrations/Dockerfile.ttlsh
```diff
@@ -12,6 +12,7 @@ ENV DEBIAN_FRONTEND=noninteractive
 RUN apt-get update \
   && apt-get install -y --no-install-recommends \
     ca-certificates \
+    wget \
```
banjoh, line 15: Initialising kots in local envs was failing.

pkg/kotsadm/objects/scripts/wait-for-rqlite.sh
```diff
@@ -8,6 +8,11 @@
 timeout=300
 elapsed=0
 
+if ! command -v wget >/dev/null 2>&1; then
```
sgalsaleh, line 11: but this script is used in production. does wget exist in the production kotsadm image?

sgalsaleh: nvm it's used below 🤦‍♂️

sgalsaleh (approved):

PR #5953, opened 2026-06-30
feat(snapshots): make Velero uploader-type version-aware and block LVP on 1.17+

nvanthao:

Velero 1.17 removed the `--uploader-type=restic` install flag (kopia has been the only valid uploader since the 1.10 default). This PR makes KOTS's Velero install guidance version-aware and prevents configuring unsupported filesystem stores on newer Velero.

```
 velero version
Client:
        Version: v1.17.2
        Git commit: 7013a4097f17a97a0201c0ad3c9dd3f810d32bf6
Server:
        Version: v1.17.2

./bin/kots velero configure-hostpath --with-minio=false --hostpath /tmp/kots-backup -n foo
Error: filesystem (NFS/Host Path) snapshots via the local-volume-provider plugin are not supported on Velero 1.17+; use an object-store backend or a Velero version < 1.17
```

image

image

Yes — unit tests added (`velero_flags_test.go`, `velero_test.go`) covering the version→flags mapping and LVP support boundary; e2e CLI setup updated.

```release-note
Velero install guidance is now version-aware: KOTS recommends the correct file-system-backup uploader flags for the target Velero version (kopia on Velero 1.17+). Configuring NFS/Host Path filesystem snapshots via the local-volume-provider plugin is now blocked on Velero 1.17+, which no longer supports the plugin.
```

NONE


xavpaice:
Some early notes:
* Test environment: I have upgraded Velero from 1.16 to 1.17, with Local Volume Provider (no Minio)
* I thought `--uploader-type` is optional now?  Given that so much of the migration is auto, I'd prefer to reduce the amount of switches we ask folks to use.
* The Snapshot settings dialog still has NFS and Host Path but these should be disabled
* When I go to add a new destination and click NFS or Host Path, it links me to the local volume provider repo

New install:
* Test environment: new, include Minio, Velero 1.17
* LGTM

Installed with no Minio, 1.17, and enabled the AWS plugin.
* the snapshot settings dialog correctly limits options to those that are enabled

Stil reviewing the actual code change.

xavpaice:
# Code review: Velero 1.17 Kopia support

## Overall verdict

The install-flag / version-aware layer is solid and well tested. The main gap is that the **destination picker is not Velero-version-aware**, so users on Velero 1.17+ with Minio disabled still see LVP (NFS/Host Path) options and only learn they are unsupported when the backend rejects the request.

## What is implemented correctly

| Proposal criterion | Status | Notes |
|---|---|---|
| Version-aware install flag matrix | ✅ | `pkg/snapshot/types/velero_flags.go` correctly maps `<1.10` → `--use-restic`, `1.10–1.16` → `--use-node-agent --uploader-type=restic`, `>=1.17` → `--use-node-agent --uploader-type=kopia`. |
| CLI install instructions | ✅ | `pkg/print/velero.go` defaults to Kopia flags for unknown/new installs. |
| Web install instructions | ✅ | `web/src/components/snapshots/ConfigureSnapshots.jsx` adds `isVelero17OrNewer()` and returns the correct flags. |
| e2e install helpers | ✅ | `e2e/velero/cli.go` and `e2e/playwright/regression/shared/cli.ts` use the version-aware helper. |
| Block LVP on 1.17+ | ✅ | `cmd/kots/cli/velero.go` and `pkg/handlers/snapshots.go` reject LVP configuration with a clear error. |
| Unit tests for flag matrix | ✅ | `pkg/snapshot/types/velero_flags_test.go` and `pkg/print/velero_test.go` cover the version cases. |

## What still needs to change

### 1. Destination picker must be version-aware (`SnapshotStorageDestination.tsx`) — **blocking**

`web/src/components/snapshots/SnapshotStorageDestination.tsx` is unchanged. It still decides between filesystem-Minio and LVP based only on `snapshotSettings.isMinioDisabled`. On Velero 1.17+ the LVP links should be hidden or disabled, and HostPath/NFS should route to the S3-compatible filesystem Minio path (which already works for 1.17+).

Without this, the UI still offers NFS/HostPath via the local-volume-provider plugin to 1.17+ clusters; the user only discovers the dead end after clicking and hitting the backend error.

### 2. In-product LVP warning banner — **missing**

`web/src/components/snapshots/SnapshotSettings.tsx` does not show the informational banner described in the proposal for an LVP-backed Restic BSL detected on Velero 1.17+.

### 3. Remove unused `resticRepoBase` — **cleanup**

`cmd/kots/cli/velero.go:33` declares `resticRepoBase = "/var/velero-local-volume-provider"` but it is not referenced anywhere in the repo. Should be removed.

### 4. `pkg/snapshot/store.go` not defensive — **limitation to note**

The LVP guard is only in the HTTP handler and CLI entry points. A future caller that invokes `snapshot.ConfigureStore` directly with an LVP provider on 1.17+ would still create an invalid BSL. Moving the version check into `ConfigureStore`/`updateLvpFileSystemStore` would be safer, but this is not an immediate blocker.

### 5. Other minor items

- `e2e/velero/cli.go` pins `docker.io/velero/velero-plugin-for-aws:v1.6.1`. This may cause issues when testing against Velero 1.17/1.18; consider deriving the plugin version from the Velero version or making it configurable.
---

**Bottom line:** merge this only if the destination-picker gap is acceptable for now; otherwise add the version-aware UI routing before merging.

pkg/print/velero.go
xavpaice: this should be updated to 1.17 (or a generic latest location)

cmd/kots/cli/velero.go
xavpaice: I think we can remove this now?

xavpaice (commented):
e2e/scripts/deps.sh has a hard coded ref to Velero 1.16, we might need to consider updating that?

Couple of nits inline.

Regards matching the proposal, this seems to do the right things, and operates cleanly.

e2e/velero/cli.go
xavpaice: --plugins velero/velero-plugin-for-aws:v1.14.2

nvanthao: Done — bumped to `velero-plugin-for-aws:v1.14.2` in fe922aacb.

cmd/kots/cli/velero.go
nvanthao: done

pkg/print/velero.go
nvanthao: done

xavpaice (approved):
Tested, looks good.

PR #5980, opened 2026-07-20
feat(prune): add operator-controlled cleanup of old support bundles a…

nvanthao:
### What

Add an opt-in cleanup policy that prunes old support bundle archives and application version archives to relieve storage pressure. It is disabled by default and configured via the kotsadm-confg ConfigMap (prune-* keys).

When enabled, a pass runs immediately at startup and then daily, retaining a configurable number of the newest support bundles and app versions per app. The currently-deployed app version and any newer, not-yet-deployed versions are always protected. Deletions remove both the metadata (rqlite / Secret) and the backing object in storage, and are logged with the app ID and sequence/bundle ID.

Sample log

```
{"level":"info","ts":"2026-07-20T02:57:08Z","msg":"prune enabled: retaining 2 support bundles and 2 app versions per app"}
{"level":"info","ts":"2026-07-20T02:57:08Z","msg":"prune: deleting app version for app 3GkTaCIRWzH9mffBAH80SzKellc sequence 1"}
{"level":"info","ts":"2026-07-20T02:57:09Z","msg":"prune: deleting support bundle 3gkuqcnpccrgvydpxi2yzxsztab for app 3GkTaCIRWzH9mffBAH80SzKellc"}
{"level":"info","ts":"2026-07-20T02:57:09Z","msg":"prune: deleting support bundle 3gkun8bwzyoj657covrehq6j8xv for app 3GkTaCIRWzH9mffBAH80SzKellc"}
```

### How to enable and test

Prune reads its config only at apiserver startup, so enabling it (or changing the counts) requires restarting the `kotsadm` deployment. The immediate startup pass then runs right away instead of waiting for the daily cron.

Set `NAMESPACE` to the namespace where kotsadm is installed.

1. Enable prune and set retention counts on the `kotsadm-confg` ConfigMap (merge patch leaves other keys untouched):

   ```sh
   kubectl patch configmap kotsadm-confg -n "$NAMESPACE" --type merge -p \
     '{"data":{"prune-enabled":"true","prune-support-bundle-count":"5","prune-app-version-count":"5"}}'
   ```

2. Restart the deployment to load the new config and trigger an immediate pass:

   ```sh
   kubectl rollout restart deployment kotsadm -n "$NAMESPACE"
   kubectl rollout status deployment kotsadm -n "$NAMESPACE"
   ```

3. Watch the prune log lines:

   ```sh
   kubectl logs -n "$NAMESPACE" deployment/kotsadm | grep -i prune
   ```

To disable, set `prune-enabled` to `"false"` (or remove the key) and restart again.

```release-note

```


pkg/kotsutil/kots.go
```diff
@@ -1258,8 +1258,19 @@ type InstallationParams struct {
...
+
+	// Prune is an operator-controlled escape hatch (set directly on the kotsadm-confg
+	// ConfigMap) for cleaning up old support bundle and app version archives.
+	PruneEnabled            bool
+	PruneSupportBundleCount int // Max support bundle archives retained per app
+	PruneAppVersionCount    int // Max app version archives retained per app
 }
 
+const (
+	defaultPruneSupportBundleCount = 25
```
sgalsaleh, line 1270: i think 25 support bundles is actually too much specially when we store them in rqlite like in EC v2. perhaps 5 is a more reasonable number?

pkg/kotsutil/kots.go
```diff
@@ -1301,6 +1315,18 @@ func GetInstallationParamsWithClientset(clientset kubernetes.Interface, configMa
 	autoConfig.AppVersionLabel = kotsadmConfigMap.Data["app-version-label"]
 	autoConfig.RequestedChannelSlug = kotsadmConfigMap.Data["requested-channel-slug"]
 
+	autoConfig.PruneEnabled, _ = strconv.ParseBool(kotsadmConfigMap.Data["prune-enabled"])
+	if v, ok := kotsadmConfigMap.Data["prune-support-bundle-count"]; ok {
+		if n, err := strconv.Atoi(v); err == nil && n >= 0 {
```
sgalsaleh, line 1320: let's log a warning if we can't parse the number?

pkg/kotsutil/kots.go
```diff
@@ -1301,6 +1315,18 @@ func GetInstallationParamsWithClientset(clientset kubernetes.Interface, configMa
...
+			autoConfig.PruneSupportBundleCount = n
+		}
+	}
+	if v, ok := kotsadmConfigMap.Data["prune-app-version-count"]; ok {
+		if n, err := strconv.Atoi(v); err == nil && n >= 0 {
```
sgalsaleh, line 1327: let's log a warning if we can't parse the number?

sgalsaleh (dismissed):

pkg/kotsutil/kots.go
nvanthao: done

pkg/kotsutil/kots.go
nvanthao: done

sgalsaleh (approved):

PR #5993, opened 2026-07-28
fix(persistence): bump gorqlite to avoid panic on empty rqlite write results

kriscoleman:
## Summary

Bumps the vendored gorqlite client so a successful-but-empty write result no longer panics kotsadm, and bumps schemahero to a compatible release.

## Problem

Older gorqlite versions read the first element of the write-results slice without checking that the slice is non-empty. When rqlite answers a write with HTTP 200 and an empty `results` array, gorqlite panics:

```
panic: runtime error: index out of range [0] with length 0
github.com/rqlite/gorqlite.(*Connection).WriteOne  write.go:68
```

KOTS issues writes during startup migrations (`migrateSkippedPreflights`) and from the app-status monitor (`SetAppStatus`), so this can crash kotsadm and leave it in CrashLoopBackOff whenever rqlite returns that empty-but-successful response, for example when the embedded rqlite cluster is in a degraded state.

## Fix

- Bump `github.com/rqlite/gorqlite` to a revision whose result handling guards the empty case (it returns an empty result rather than indexing out of range).
- Bump `github.com/schemahero/schemahero` to a release compatible with that gorqlite revision.
- gorqlite's `Open` now returns a `*Connection`, so `MustGetDBSession` is adjusted to match.

## Testing

- `go build ./cmd/...` and the gorqlite-importing packages (`pkg/persistence`, `pkg/store`, `pkg/tasks`, `pkg/app`, `pkg/filestore`, `pkg/version`) compile cleanly.
- Reproduced the original panic against the previous gorqlite by driving the exact `migrateSkippedPreflights` write at an endpoint returning `200 {"results":[]}`, then confirmed the bumped gorqlite returns a result and does not panic under the same scenario.


pkg/persistence/persistence.go
```diff
@@ -26,6 +26,6 @@ func MustGetDBSession() *gorqlite.Connection {
 		fmt.Printf("error connecting to rqlite: %v\n", err)
 		panic(err)
 	}
-	db = &newDB
+	db = newDB
```
kriscoleman, line 29: the upstream gorqlite version made a breaking type-signature change, so this was required after the bump

kriscoleman:
Superseded by #5994. Moved the branch onto the main repo so CI can run (fork PRs skip the build/test jobs via the can-run-ci gate). Same commit and diff.

PR #5994, opened 2026-07-28
fix(persistence): fix gorqlite empty-write panic and move to schemahero 0.25

kriscoleman:
## Summary

This started as a one-line gorqlite bump to stop kotsadm panicking on empty rqlite writes. That gorqlite revision changed its API, which forced a schemahero bump (0.22.1 -> 0.25.1), and 0.25 reworked how schemahero reaches its DB drivers. So the PR is really two things: the gorqlite fix, and moving cleanly onto schemahero 0.25.

## The gorqlite panic

[sc-139123]

Older gorqlite reads the first element of the write-results slice without checking it's non-empty. When rqlite answers a write with HTTP 200 and an empty `results` array, gorqlite panics:

```
panic: runtime error: index out of range [0] with length 0
github.com/rqlite/gorqlite.(*Connection).WriteOne  write.go:68
```

KOTS issues writes during startup migrations (`migrateSkippedPreflights`) and from the app-status monitor (`SetAppStatus`), so this can crash kotsadm into CrashLoopBackOff whenever rqlite returns that empty-but-successful response, for example when the embedded rqlite cluster is degraded.

## Moving to schemahero 0.25

The bumped gorqlite changes `Open` to return a `*Connection`, so schemahero has to move to 0.25.1 to stay compatible. Two things change with 0.25, and they're handled separately:

- **In-process code (KOTS's own migration path).** KOTS's `pkg/persistence` imported `schemahero/pkg/database`, which in 0.25 can only reach a driver by launching a plugin subprocess. `UpdateDBSchema` now imports the driver lib directly (`plugins/{postgres,rqlite}/lib`) and drives `PlanTableSchema` -> `DeployStatements` in-process. This is the path used for the postgres schema during the postgres->rqlite migration.

- **The schemahero CLI in the init containers (rqlite schema).** The rqlite tables are still created by the `schemahero-plan`/`schemahero-apply` init containers running the schemahero binary from the `kotsadm-migrations` image. That's unchanged in structure. But 0.25's CLI loads the rqlite driver as a go-plugin binary, and if it can't find the plugin locally it **downloads it from an OCI registry**, which hangs in a no-network init container. That download is what left kotsadm stuck in `Init` and failed the e2e suite.

## The plugin fix

The `schemahero-0.25-rqlite` apk installs the plugin to `/var/lib/schemahero/plugins/schemahero-rqlite`, which is one of schemahero's default search paths. Declaring it only as a melange subpackage runtime dep didn't reliably land it in the image, so this installs it explicitly in the `kotsadm-migrations` apko image and sets `SCHEMAHERO_PLUGIN_PATH` so discovery always finds the bundled plugin and never attempts a download.

Reproduced locally with the schemahero 0.25.1 CLI + rqlite plugin: with the plugin absent, schemahero's `[plugin-downloader]` kicks in and fetches over the network; with it present at the search path, `plan` and `apply` both run against rqlite and create all 24 tables in-process, with zero download attempts.

Deployment objects (`kotsadm_objects.go`), the init containers, and the `kotsadm-migrations` image are all kept as they are; the only image change is making the bundled plugin discoverable.

## Changes

- Bump `github.com/rqlite/gorqlite` to a revision whose result handling guards the empty case.
- Bump `github.com/schemahero/schemahero` 0.22.1 -> 0.25.1 and add the `plugins/postgres` and `plugins/rqlite` modules.
- Rework `pkg/persistence` to connect the driver lib directly (in-process, no plugin subprocess) for the postgres migration path.
- Install `schemahero-0.25-rqlite` in the `kotsadm-migrations` image and set `SCHEMAHERO_PLUGIN_PATH`, so the 0.25 CLI finds the bundled rqlite plugin instead of trying to download it.
- gorqlite's `Open` now returns a `*Connection`, so `MustGetDBSession` is adjusted to match.

## Testing

- `go build ./...` and `go vet ./...` pass; `pkg/kotsadm/objects` unit tests pass.
- Reproduced the original panic against the old gorqlite by driving the exact `migrateSkippedPreflights` write at an endpoint returning `200 {"results":[]}`, then confirmed the bumped gorqlite returns a result and doesn't panic.
- Reproduced the schemahero 0.25 plugin behavior with the real 0.25.1 CLI + rqlite plugin: confirmed the download-when-missing hang, and confirmed that with the plugin on a search path, `plan`+`apply` create all 24 rqlite tables with no download.


pkg/persistence/persistence.go
```diff
@@ -26,6 +26,6 @@ func MustGetDBSession() *gorqlite.Connection {
 		fmt.Printf("error connecting to rqlite: %v\n", err)
 		panic(err)
 	}
-	db = &newDB
+	db = newDB
```
kriscoleman, line 29: the upstream gorqlite version had a breaking change with this type-signature, so this was changed due to that

kriscoleman:
Flagging a blocker on this bump before it goes further, since the fix is not a clean version bump.

**The coupling.** The empty-result guard we want only exists on gorqlite revisions that also changed `Open` to return `*Connection` (the guard and the signature change landed together upstream). Because the whole module graph resolves a single gorqlite version, bumping gorqlite for `pkg/persistence` also recompiles schemahero's in-tree rqlite driver against the new signature, which is why schemahero had to move too:

```
schemahero@v0.22.1/pkg/database/rqlite/connection.go: cannot use &db (**Connection) as *Connection
```

**The wall.** The only schemahero releases built against the new gorqlite API are v0.23.0+, and those moved the database drivers to a plugin architecture (`hashicorp/go-plugin`): postgres/rqlite are now plugin-only, launched as separate `schemahero-<driver>` binaries from `/var/lib/schemahero/plugins`, with no in-tree fallback. KOTS calls schemahero in-process (`Database{}.PlanSyncFromFile` / `ApplySync`), so v0.25 builds fine but fails at runtime, which is the CI failure here:

```
--- FAIL: TestMigrateFromPostgresToRqlite
    rqlite_migration_test.go: Failed to update postgres schema: ... postgres driver requires plugin - install schemahero-postgres plugin
```

So there is no schemahero version that is both compatible with the fixed gorqlite and usable in-process:

| schemahero | rqlite driver | gorqlite API | usable in-process by KOTS |
|---|---|---|---|
| v0.22.x (current) | in-tree | old (value `Open`) | yes, but will not compile with fixed gorqlite |
| v0.23.0+ | plugin binary | new (`*Connection`) | no, requires launching plugin binaries |

**Options I can see:**

1. Adopt the schemahero plugin model: wire a `PluginManager` at the KOTS call sites and ship `schemahero-postgres` and `schemahero-rqlite` (multi-arch, airgap-safe) in the kotsadm and migrations images. This is the forward-looking fix but touches the build/release pipeline and the airgap path.
2. Carry the guard via a Replicated-maintained gorqlite fork (old base plus the empty-result guard, keeping the value `Open`) referenced through a `go.mod replace`. Keeps schemahero untouched and is a small change, at the cost of a maintained fork. I am not in favor of this one: I would rather not take on maintaining a gorqlite fork.
3. Leave gorqlite pinned for now and guard at the call sites another way.

I do not want to force the version bump given the plugin dependency, so converting to a discussion of direction. Happy to take whichever route the maintainers prefer. For context, the underlying behavior is the panic described above: an rqlite write that returns HTTP 200 with an empty `results` array makes gorqlite index `results[0]` and crash kotsadm rather than return an error.

xavpaice:
Shortcut: [sc-139123]

banjoh:
For future travellers

PR: https://github.com/rqlite/gorqlite/pull/62
Upstream issue: https://github.com/rqlite/gorqlite/issues/41

banjoh:
We should adopt option 1 (adopting plugins architecture) in the options presented in https://github.com/replicatedhq/kots/pull/5994#issuecomment-5110415789

We already have precedent for bundling a schemahero plugin binary into an image: https://github.com/replicatedhq/kots/pull/5775 added `schemahero-rqlite` to `migrations/deploy/apko.yaml.tmpl` so the rqlite plugin ships in the `kotsadm-migrations` image instead of being downloaded at runtime. I confirmed the apk package installs straight to `/var/lib/schemahero/plugins/schemahero-rqlite`, which is schemahero's default plugin search path, so no extra wiring was needed beyond the package addition.

The postgres plugin is published in the same apk repo at the matching `0.25.1` release: `schemahero-0.25-postgres-0.25.1-r1.apk`. I downloaded it and listed its contents (an apk is just a gzipped tarball, no install needed):

```
$ tar -tzf schemahero-0.25-postgres-0.25.1-r1.apk
var/lib/schemahero/plugins/schemahero-postgres
```

Same layout as the rqlite plugin, so option 1 from the comment above is smaller than it first looked. What's needed:

1. Bump `github.com/schemahero/schemahero` in `go.mod`/`go.sum` to v0.23+ (pulls the fixed `gorqlite` transitively).
2. Wire the plugin system into `pkg/persistence/schema.go`'s `UpdateDBSchema`, since v0.23+'s `Database.GetConnection` requires a `PluginManager` and none is set today. This is `plugin.InitializePluginSystem()` plus `db.SetPluginManager(...)`, matching what schemahero's own CLI does in `root.go`.
3. Add `schemahero` and `schemahero-postgres` to `deploy/apko.yaml.tmpl` (the kotsadm image, not the migrations image, since `MigrateFromPostgresToRqlite` runs in-process inside kotsadm via `pkg/apiserver/server.go:55`). I checked and the kotsadm image ships neither today, so this is a net-new addition there, the same shape of change #5775 made for rqlite in the migrations image.
4. Make sure `integration/database/rqlite_migration_test.go` (`TestMigrateFromPostgresToRqlite`) has the plugin binary available wherever it runs in CI.

No new plugin download/bundling mechanism needed, this rides the same apk-package path #5775 already proved out.

kriscoleman:
Flagging a securebuild blocker on this PR that I could use a hand with from whoever owns the securebuild publish path (recent #6010/#6011).

**Context.** This PR fixes the gorqlite empty-write panic, but that fix is coupled to a schemahero bump: the upstream gorqlite fix landed together with a `Open() -> *Connection` signature change, and the only schemahero releases built against that API are v0.23+, which moved the postgres/rqlite DB drivers to a `hashicorp/go-plugin` architecture (separate `schemahero-<driver>` binaries, no in-tree fallback). KOTS calls schemahero in-process, so the kotsadm binary is now built against the 0.25 library that *requires* those plugin binaries to be present on disk.

**What this PR already does** (dev images): wires schemaheros `PluginManager` at the KOTS call sites and ships the plugins into the dev apko templates, airgap-safe (installed at image build time, discovered from `/var/lib/schemahero/plugins`, no runtime fetch):
- `deploy/apko.yaml.tmpl` (kotsadm): `+ schemahero-0.25-postgres`
- `migrations/deploy/apko.yaml.tmpl` (kotsadm-migrations): pinned `schemahero-0.25` + `schemahero-0.25-rqlite`

**The gap (release images).** The `securebuild/` path still pins schemahero **0.22** and ships no 0.25 plugins:
- `securebuild/package/melange.yaml` copies the 0.22 CLI into the kotsadm-migrations subpackage
- `securebuild/image/apko-kotsadm.yaml` ships no plugin

So securebuild-published (release) images would carry a kotsadm binary built against schemahero 0.25 (expecting plugins) with only the 0.22 CLI and **no 0.25 plugins present** -> runtime failure in production, and specifically an air-gap gap since theres nothing to fall back to. In other words the panic fix wont actually work in released airgap images until securebuild is updated to match.

**The ask.** Could you help me get securebuild to 0.25 + the plugins? Roughly: bump the schemahero build-dep from 0.22 -> 0.25 in `securebuild/package/melange.yaml`, and get `schemahero-0.25-postgres` + `schemahero-0.25-rqlite` into the kotsadm and kotsadm-migrations release images the same way the dev apko templates do (the apk packages `schemahero-0.25-postgres` / `schemahero-0.25-rqlite` at `0.25.1-r2` exist for both x86_64 and aarch64 and drop the plugin ELF into `/var/lib/schemahero/plugins`). Happy to pair on it or hand off whichever split makes sense — I just want the release/airgap path to match the dev images before this merges. Thanks!

securebuild/image/apko-kotsadm.yaml
```diff
@@ -17,6 +17,9 @@ contents:
     - py3.12-s3cmd
     - yarn-1.22
     - ca-certificates-bundle
+    # kotsadm runs schemahero in-process for the postgres->rqlite migration. 0.23+ ships the
+    # postgres driver as a separate go-plugin, installed to /var/lib/schemahero/plugins at build (airgap-safe).
+    - schemahero-0.25-postgres~0.25.1
```
divolgin: ~what's using this?~ nm, found the answer

securebuild/package/melange.yaml
```diff
@@ -45,7 +45,9 @@ environment:
       - yarn-1.22
       - py3.12-dateutil
       - py3.12-s3cmd
-      - schemahero-0.22
+      # schemahero 0.25 CLI (copied into the kotsadm-migrations subpackage). 0.23+ split the DB
+      # drivers into separate plugin packages, added to the apko images; this provides the CLI only.
+      - schemahero-0.25~0.25.1
```
divolgin, line 48: previous version did not pin patch. is this an intentional change?

securebuild/image/apko-kotsadm-migrations.yaml
```diff
@@ -11,6 +11,9 @@ contents:
     - curl
     - git
     - ca-certificates-bundle
+    # schemahero 0.23+ ships the rqlite driver as a separate go-plugin, not in the CLI.
+    # Installed to /var/lib/schemahero/plugins at build time (airgap-safe); pinned to go.mod.
+    - schemahero-0.25-rqlite~0.25.1
```
divolgin: this should probably be in the kotsadm-migrations package

scripts/install-schemahero-plugins.sh
```diff
@@ -0,0 +1,70 @@
...
+#
+# DEST_DIR defaults to /var/lib/schemahero/plugins (schemahero's default
+# discovery path). Point SCHEMAHERO_PLUGIN_PATH at DEST_DIR when it is not the
+# default, e.g.:
+#   scripts/install-schemahero-plugins.sh "$RUNNER_TEMP/schemahero-plugins"
+#   SCHEMAHERO_PLUGIN_PATH="$RUNNER_TEMP/schemahero-plugins" make ci-test
+#
+set -euo pipefail
+
+DEST_DIR="${1:-/var/lib/schemahero/plugins}"
```
divolgin: schemahero downloads plugins on demand. this should not be needed

migrations/deploy/apko.yaml.tmpl
```diff
@@ -14,8 +14,15 @@ contents:
...
-    - schemahero
-    - schemahero-rqlite
+    # schemahero v0.25.1 ships each DB driver as a separate go-plugin binary
+    # (installed to /var/lib/schemahero/plugins). This image runs the schemahero
+    # CLI against rqlite (schemahero-plan/apply init containers), so it needs the
+    # rqlite plugin. Pin the CLI + plugin to the exact patch (~0.25.1) that matches
+    # the schemahero library version in go.mod, so the shipped go-plugin protocol
+    # matches what CI tests. The plugin is installed into the image at build time,
+    # so nothing is downloaded at runtime (airgap-safe).
+    - schemahero-0.25~0.25.1
```
divolgin: `schemahero` and `schemahero-rqlite` will install the latest (which is 0.25.1 now).  is this pinning intentional?

deploy/apko.yaml.tmpl
```diff
@@ -19,6 +19,13 @@ contents:
     - s3cmd
     - securebuild-baselayout
     - helm~4
+    # kotsadm runs schemahero in-process for the legacy postgres->rqlite migration
+    # (pkg/persistence.UpdateDBSchema("postgres")). schemahero v0.25.1 loads the
+    # postgres driver as a separate go-plugin binary from /var/lib/schemahero/plugins,
+    # so ship the postgres plugin, pinned to the exact patch (~0.25.1) matching
+    # go.mod. It is installed at build time, so nothing is downloaded at runtime
+    # (airgap-safe).
+    - schemahero-0.25-postgres~0.25.1
```
divolgin: This makes sense, but we need to not forget to keep this version in sync with go.mod version.

scripts/install-schemahero-plugins.sh
banjoh: How about airgap?

deploy/apko.yaml.tmpl
divolgin: Alternative solution: don't call high level schemahero wrapper, but call the plugin packages directly since we know which driver we want to use

```
	return UpdateDBSchema("postgres", os.Getenv("POSTGRES_URI"), os.Getenv("POSTGRES_SCHEMA_DIR"))
```

securebuild/package/melange.yaml
kriscoleman: Yea, intentional. We pin the patch (`0.25.1`) on purpose so the shipped plugin's go-plugin protocol matches the schemahero library the kotsadm binary is built against (go.mod is at v0.25.1). If the plugin floats to a newer patch and the go-plugin handshake ever changes within 0.25.x, the plugin subprocess and the in-process client could mismatch at runtime. Pinning keeps them lockstep.

migrations/deploy/apko.yaml.tmpl
kriscoleman: Same deal as the melange one, and yea it's intentional. Pinning `schemahero` + `schemahero-rqlite` to `~0.25.1` keeps the shipped CLI and rqlite plugin on the exact patch the go.mod library is built against, so the go-plugin protocol matches. Letting them float to latest risks a handshake mismatch if a future 0.25.x bumps the protocol.

deploy/apko.yaml.tmpl
kriscoleman: Agreed, good callout. The CI install script (`scripts/install-schemahero-plugins.sh`) derives the plugin version straight from go.mod via `go list -m`, so the CI-tested plugins are always tied to the library version, and the apko pins are set to that same patch. It's still a manual bump when go.mod changes though. Happy to add a CI assert that the apko-pinned version equals the go.mod schemahero version so it can't drift silently -- want me to?

scripts/install-schemahero-plugins.sh
kriscoleman: Right that schemahero can fetch drivers on demand -- it pulls the plugin over ORAS from a registry the first time a driver is needed. The catch is that path fails in air-gapped installs: there's no outbound network to fetch it, so the migration errors at runtime. So we install the plugin into the image at build time (into `/var/lib/schemahero/plugins`, schemahero's local discovery path), which means the on-demand download never fires. This script does the same thing for the integration test -- it builds the plugin from the pinned version so the test exercises the exact "plugin present locally, no download" path the airgap images ship, instead of depending on a network fetch during CI. So it's there specifically to keep the test airgap-faithful.

securebuild/image/apko-kotsadm-migrations.yaml
kriscoleman: Good call -- moved it. The rqlite driver is a runtime requirement of the kotsadm-migrations image (that image *is* the schemahero-against-rqlite CLI), so I declared `schemahero-0.25-rqlite~0.25.1` as a `dependencies.runtime` of the kotsadm-migrations melange subpackage and dropped it from the apko image's packages list. Same end result -- apk installs it to `/var/lib/schemahero/plugins` at image assembly (airgap-safe) and the `~0.25.1` patch pin is preserved. Left the `/schemahero` root symlink in the apko image since that's an image-level entrypoint alias, same as how the kotsadm image defines `/kotsadm` and `/kots`.

scripts/install-schemahero-plugins.sh
divolgin: It should not be needed. APKO/melange specs should be installing plugins at build time.  Go code should vendor in plugin packages directly.

pkg/kotsadm/objects/kotsadm_objects.go
```diff
@@ -525,112 +518,6 @@ func KotsadmDeployment(deployOptions types.DeployOptions) (*appsv1.Deployment, e
 					ImagePullSecrets:   pullSecrets,
 					InitContainers: []corev1.Container{
 						waitForRqliteInitContainer(deployOptions),
-						{
-							Image:           GetAdminConsoleImage(deployOptions, "kotsadm-migrations"),
```
divolgin, line 529: Is this being removed because `persistence.UpdateRqliteSchema` is being called in kotsadm on start up? I think we need to keep this and run migrations in the in init containers. Removing `kotsadm-migrations` has a much bigger blast radius than this: the image itself, kurl plugin, troubleshot specs, documentation, local dev environment, etc

kriscoleman: yea I totally agree, this was not intentional, working on a different approach

pkg/persistence/schema.go
```diff
@@ -41,9 +66,41 @@ func UpdateDBSchema(driver string, uri string, schemaDir string) error {
...
+	case "rqlite":
+		return rqlitelib.Connect(uri)
+	default:
+		return nil, errors.Errorf("unsupported schemahero driver %q", driver)
+	}
+}
+
+func driverTableSchema(driver string, schema *schemasv1alpha4.TableSchema) any {
+	if schema == nil {
+		return nil
```
divolgin: Shouldn't these `return nil` be errors? if there is no schema or no valid db connection, things should fail.  Passing in an invalid driver name should also result in an error

divolgin (approved):

PR #6005, opened 2026-08-04
Updating release to use SecureBuild

divolgin:

Uses securebuild CLI to build kotsadm packages and image
- melange and apko files are updated to work in SecureBuild
- CI workflow added to use securebuild CLI to build packages and images

```release-note

```


divolgin:
implemented in https://github.com/replicatedhq/kots/pull/6021

PR #6031, opened 2026-08-19
fix(supportbundle): move support bundle sub-specs to secrets and clean up legacy configmaps

xavpaice:
## Summary

Moves the rendered support bundle sub-specs from ConfigMaps to Secrets so they are no longer collected by the default support bundle clusterResources/configMap collector.

## Changes

- `pkg/supportbundle/spec.go`: removed `createSupportBundleSpecConfigMap`; `CreateRenderedSpec` now creates the vendor, cluster-specific, and default sub-specs as Secrets using the existing `createSupportBundleSpecSecret`.
- `pkg/supportbundle/spec.go`: added per-app cleanup that deletes the three legacy ConfigMaps after the new Secrets are created.
- `pkg/supportbundle/migrate.go`: added `CleanupLegacySpecConfigMaps`, a startup migration that deletes any remaining legacy ConfigMaps matching `kotsadm-.*-supportbundle-(vendor|cluster-specific|default)`.
- `pkg/apiserver/server.go`: wired the startup migration into `Start()` so it runs automatically on every admin-console upgrade.
- `pkg/supportbundle/migrate_test.go`: added unit test for the migration.

## Related

- [sc-139178](https://app.shortcut.com/replicated/story/139178)

### DIFF
--- a/pkg/apiserver/server.go
+++ b/pkg/apiserver/server.go
@@ -91,6 +91,11 @@ func Start(params *APIServerParams) {
 		log.Println("error getting k8s clientset")
 		panic(err)
 	}
+
+	if err := supportbundle.CleanupLegacySpecConfigMaps(context.TODO(), k8sClientset, util.PodNamespace); err != nil {
+		log.Println("Failed to clean up legacy support bundle spec configmaps: ", err)
+	}
+
 	op := operator.Init(operatorClient, kotsStore, params.AutocreateClusterToken, k8sClientset)
 	if err := op.Start(); err != nil {
 		log.Println("error starting the operator")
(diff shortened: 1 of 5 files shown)


pkg/apiserver/server.go
```diff
@@ -91,6 +91,11 @@ func Start(params *APIServerParams) {
 		log.Println("error getting k8s clientset")
 		panic(err)
 	}
+
+	if err := supportbundle.CleanupLegacySpecConfigMaps(context.TODO(), k8sClientset, util.PodNamespace); err != nil {
```
kriscoleman, line 95: smart

kriscoleman (approved):
great improvement to cover a real issue

PR #6044, opened 2026-09-03
Faster image-push command

divolgin:

This makes the `kots admin-console push-images` command run faster:

1. Initial bundle scan needed to create progress page in UI is skipped when CLI is used
2. Images are extracted and pushed in parallel instead of sequential "extract" -> "push" chain of steps.
3. The final scan of the bundle that looks for Embedded Cluster images is terminated early if there are no Embedded Cluster resources in the bundle.
4. But if there are Embedded Cluster images, they are processed in parallel with application images.

Testing with a 160GB airgap bundle, timing before:

```
$ time kots admin-console push-images ./67.airgap 34.133.154.238:5000
...
real	142m48.808s
user	187m47.327s
sys	22m15.314s
```

with gzip and sequential image push

```
$ time kots admin-console push-images ./67.airgap 34.133.154.238:5000
...
real	50m10.771s
user	109m0.847s
sys	14m39.375s
```

with pgzip and podman's `ConcurrentBlobCopiesSemaphore`

```
$ time kots admin-console push-images ./67.airgap 34.133.154.238:5000
...
real	37m24.219s
user	106m45.268s
sys	15m20.171s
```

 New features:
 ```release-notes-features
 
 ```
 Bug fixes:
 ```release-notes-fixes
 
 ```
 Improvements:
 ```release-notes-improvements
 Improves the speed of the `kots admin-console push-images` command
 ```

### DIFF
--- a/pkg/image/airgap.go
+++ b/pkg/image/airgap.go
@@ -14,12 +14,12 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/distribution/reference"
 	imagespecsv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
-	"github.com/replicatedhq/kots/pkg/archives"
 	"github.com/replicatedhq/kots/pkg/archiveutil"
 	dockerarchive "github.com/replicatedhq/kots/pkg/docker/archive"
 	dockerregistry "github.com/replicatedhq/kots/pkg/docker/registry"
@@ -32,6 +32,7 @@ import (
 	"github.com/replicatedhq/kots/pkg/util"
 	kotsv1beta1 "github.com/replicatedhq/kotskinds/apis/kots/v1beta1"
 	"go.podman.io/image/v5/transports/alltransports"
+	"golang.org/x/sync/errgroup"
 	oras "oras.land/oras-go/v2"
 	orasfile "oras.land/oras-go/v2/content/file"
 	orasremote "oras.land/oras-go/v2/registry/remote"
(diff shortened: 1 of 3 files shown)


pkg/image/airgap.go
```diff
@@ -629,9 +670,23 @@ func PushImagesFromDockerArchiveBundle(airgapBundle string, options imagetypes.P
 				SkipExistingImages: options.SkipExistingImages,
 			},
 		}
-		if err := pushImage(pushImageOpts); err != nil {
-			return errors.Wrapf(err, "failed to push image %s", imagePath)
-		}
+		pushGroup.Go(func() error {
+			defer os.Remove(tmpFileName)
+			pushMutex.Lock()
+			defer pushMutex.Unlock()
```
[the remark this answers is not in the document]
divolgin: this is correct, but look at the code that was removed.  it extracted all files to temp directory. so this is an improvement and also intentional.

pkg/image/airgap.go
```diff
@@ -629,9 +670,23 @@ func PushImagesFromDockerArchiveBundle(airgapBundle string, options imagetypes.P
 				SkipExistingImages: options.SkipExistingImages,
 			},
 		}
-		if err := pushImage(pushImageOpts); err != nil {
-			return errors.Wrapf(err, "failed to push image %s", imagePath)
-		}
+		pushGroup.Go(func() error {
+			defer os.Remove(tmpFileName)
+			pushMutex.Lock()
```
sgalsaleh: wouldn't this defeat the purpose of using a pushGroup to parallalize? the mutex would make this run serially, no?

divolgin: > defeat the purpose

it does not defeat the purpose. each layer is being pushed in parallel.

the risk of pushing all images in parallel is saturating CPU and network.  parallelism is controlled automatically by podman

sgalsaleh: we have pattern elsewhere that sets a concurrency limit if we want to run N pushes in parallel without making unbounded: https://github.com/replicatedhq/kots/blob/73f89f5b80d5478e00bde5bf83d2f6749e41ea04/pkg/image/online.go#L82-L83

sgalsaleh: ah i see. this push is for a single image, but the image layers get pushed in parallel? then why do we need a goroutine and a mutex here?

pkg/image/airgap.go
```diff
@@ -183,9 +163,42 @@ func TagAndPushImagesFromBundle(airgapBundle string, options imagetypes.PushImag
 		VersionLabel: airgap.Spec.VersionLabel,
 		HTTPClient:   &http.Client{Transport: orasretry.NewTransport(transport)},
 	}
-	err = PushEmbeddedClusterArtifacts(airgapBundle, airgap.Spec.EmbeddedClusterArtifacts, pushEmbeddedArtifactsOpts)
-	if err != nil {
-		return errors.Wrap(err, "failed to push embedded cluster artifacts")
+
+	var group errgroup.Group
+	group.Go(func() error {
+		if err := PushEmbeddedClusterArtifacts(airgapBundle, airgap.Spec.EmbeddedClusterArtifacts, pushEmbeddedArtifactsOpts); err != nil {
```
sgalsaleh, line 172: note: pushing EC artifacts in parallel means both the app and EC bundles would be extracted to disk at the same time. pushing in parallel reduces time but increases disk usage. not a blocker, just something i wanted to call out.

pkg/image/airgap.go
sgalsaleh: also, what if we're pushing single-layer images? a future improvement might be to cap number of images we push based on the number of layers in them.

pkg/image/airgap.go
divolgin: indeed. though the size of EC images is bounded and independent of application.

sgalsaleh (approved):

